PR #3, opened 2016-03-03
Kern naming

jameshensman:
I've implemented GPy-style naming of the kernels

this means that if you add two kernels, you get an intuitive naming scheme for them. Example:

```
k = GPflow.kernels.RBF(1) + GPflow.kernels.White(1)
k.rbf.variance  = 1
```

the name of the kernel is given by the lower case of the class name. So RBF-> rbf, etc. Duplicate kernels are given suffixes:
```
k = GPflow.kernels.RBF(1) + GPflow.kernels.RBF(1)
k.rbf_1.variance  = 1
k.rbf_2.lengthscale = 73.
```

Product kernels work in the same way. 

Another change is that additive and product kernels no longer 'nest'. Consider

```
k1 = GPflow.kernels.RBF(1) + GPflow.kernels.White(1)
k2 = GPflow.kernels.Matern32(1) + GPflow.kernels.Linear(1)
k = k1 + k2 
```
Previously, we'd have had something like
```
k.k1.k1.lengthscale = 7 # set the lengthscale of the RBF kernel
k.k2.k2.variance = 2 # set the variance of the Linear kernel
```
Now we'll have 
```
k.rbf.lengthscale = 7 # set the lengthscale of the RBF kernel
k.linear.variance = 2 # set the variance of the Linear kernel
```


alexggmatthews:
Hi James,
I like the general idea of the change set and we seem to have reasonable test coverage.

Maybe we could pull the piece of code that does if islist unwrap else append out into a single sub routine as we have very similar code twice? 

Alex

jameshensman:
Thanks @alexggmatthews , code reused as suggested, good call.

alexggmatthews:
It is also worth noting that this pull request will break backward compatibility with the existing model front end, although the fix should be very straightforward and the new system should give a lot of benefit going forward. Maintaining strong backwards compatibility is a large development overhead that we are trying to avoid for the time being.

PR #4, opened 2016-03-07
Added a simple versioning system

jameshensman:
The version is stored in GPflow/_version.py, as a string. 

This string is then read by `__init__.py` to create `GPflow.__version__`, and also by `setup.py` for the official versioning (e.g. `python setup.py --version`). 

This slightly round-about way means that the version is only stored in one place, as recommend at http://stackoverflow.com/questions/458550/standard-way-to-embed-version-into-python-package

I also added a RELEASE.md file, where we can describe changes in future versions


jameshensman:
Oops, looks like I accidently included a commit which was a run of python-modernize. 

This changes a few things to be python 3 compatible, mostly import patterns and raise Foo(message) patterns. The tests all pass (python 2.7) and this falls in the same remit as versioning (i.e. making the package better, non-technical changes), so I'm going to leave it in.

alexggmatthews:
Hi James,
Looks good and clean in general. Thanks for your efforts.

I'm personally not sure about mixing in this modernize commit. Is the modernize thread complete? Will it now run in python 3? It seems to met that the discussion and implementation of this should be in a separate pull request. 

It isn't hard to take it out. I just did a git revert on the commit in question and the tests still pass. I'm happy to push the fix if you like. 

hth,
Alex

jameshensman:
Hi Alex,

yep, please push the fix and we'll modernize on a separate PR.

James.

Issue #7, opened 2016-03-10
Put minibatch work onto master


alexggmatthews:
We have done some work on getting minibatches working in GPflow. We should consolidate it into its simplest and cleanest form and then submit a pull request onto master.

markvdw:
Hi Alex, I'm happy to do this today.

alexggmatthews:
OK Mark great! We're trying to keep 1 issue per pull request. Then we can review it there.

markvdw:
So I have an implementation running, but before I code things together, I wanted a bit of input, in particular on:
- Which class(es) should be SVI "aware"?
- Should we add compatibility for the Tensorflow optimisers?
These two are related.

There are two ways of adding the SVI functionality:

- Inside classes that inherit from `Model`. For example, we want to add SVI to `SVGP`. In this case, we modify `SVGP.build_likelihood` to randomly subsample `self.X`, calculate the likelihood and rescale appropriately. Advantage: `Model` stays untouched.
- Inside `Model`. In this case, we make `self.X` a Tensorflow variable, which is passed into the computational graph when `self._session.run()` is called. `self.objective` could be rewritten to subsample the data if a particular flag in `Model` was set. Advantage: In the Tensorflow docs, this is the way minibatches are handled. Disadvantage: `Model` needs to be SVI aware.

Personally, I would argue for putting the SVI functionality in `Model`, for the following reasons:
- The subsampling gets done whenever the objective is run, depending on what dataset is passed into the computational graph. If the subsampling is done in `build_likelihood`, the model would need to be recompiled each time the minibatch size is changed.
- The above goes hand-in-hand with what's needed to use the Tensorflow way of handling stochastic optimisation.
- `Model` contains the optimisation procedure, which needs to be SVI aware.
- This change can be transparent to models which do not use minibatches. For models that don't support minibatches, nothing changes so long as the full dataset is passed into the computational graph. Even if the full dataset isn't passed, nothing will break (though we won't get an unbiased estimate of the objective).

An alternative would be to create another intermediate class `SVIModel` or something, from which other models which require SVI would inherit. This class would have to re-implement some stuff that `Model` does with only minor changes (which looks like code duplication to me).

[...]

PR #13, opened 2016-03-10
Svi tfopt

markvdw:
Mostly modified `models.py`.

`Model` is modified to allow optimisation with Tensorflow optimisers. `GPModel` is modified to allow for stochastic optimisation. And `SVGP` was changed to correctly rescale the likelihood terms. There's a notebook for illustration.


jameshensman:
This is being discussed in issue #7 .

PR #15, opened 2016-03-15
Svi

alexggmatthews:
Minibatch style optimization for GPflow. See also #7


alexggmatthews:
Hi Mark,
What do you think about changing "minibatch" to "minibatch_size"? It seems a bit misleading at the moment. I'm happy to do it.
Alex

markvdw:
Sure!

alexggmatthews:
OK that's done. Well done on becoming a GPflow contributor @markvdw .

PR #20, opened 2016-03-21
Sgpr

jameshensman:
The SGPR method does Sparse GP regression. The q(u) distribution is set optimally as per Titsias (2009) (see docstring for citation), and the method is limited to Gaussian likelihoods.

The method is tested in the same way as existing methods. In addition, I added test_method_equivalence.py, which ensures that our methods that should be similar are giving the same results (adapted from sanity_check.ipynb).


GPflow/sgpr.py
```diff
@@ -0,0 +1,94 @@
...
+        tmp = tf.matrix_triangular_solve(L, tf.transpose(Knm), lower=True)*tf.sqrt(1./self.likelihood.variance)
+        A_ = tf.matmul(tmp, tf.transpose(tmp))
+        A = A_ + eye(num_inducing)
+        LA = tf.cholesky(A)
+
+        tmp = tf.matrix_triangular_solve(L, tf.matmul(tf.transpose(Knm), err), lower=True)
+        b = tf.matrix_triangular_solve(LA, tmp, lower=True) * beta
+
+        #compute log marginal bound
+        bound = -0.5*tf.cast(num_data*output_dim, tf.float64)*np.log(2*np.pi)
```
alexggmatthews: The computation of the bound could do with a comment explaining how the code relates to the equations in Titsias' paper.

GPflow/sgpr.py
```diff
@@ -0,0 +1,94 @@
...
+        bound += -tf.cast(output_dim, tf.float64)*tf.reduce_sum(tf.log(tf.user_ops.get_diag(LA)))
+        bound += 0.5*tf.cast(num_data*output_dim, tf.float64)*tf.log(beta)
+        bound += -0.5*beta*tf.reduce_sum(tf.square(err))
+        bound += 0.5*tf.reduce_sum(tf.square(b))
+        bound += -0.5*(tf.reduce_sum(Kdiag)*beta - tf.reduce_sum(tf.user_ops.get_diag(A_)))
+
+        return bound
+
+    def build_predict(self, Xnew):
+        err =  self.Y - self.mean_function(self.X)
```
alexggmatthews: The code could do with some comments explaining how the computation relates to the mathematics in Titisias' paper.

alexggmatthews:
Hi James, Iooks good. I've made some simple comments in the code. Basically it would be good to improve the commenting so that readers can see the relationship between the code and Titsias' paper more clearly. 
Alex

alexggmatthews:
Hi James,
It looks like the tests now fail. Let me know when you want me to have another look.
Thanks,
Alex

jameshensman:
It's back up again.

James.

jameshensman:
Okay, I added a notebook with (extensive!) derivations of the equations. This actually took longer than writing the code! @alexggmatthews , could you take another look?

PR #27, opened 2016-03-27
Full covariance matrix in predictions

### LINKED ISSUE #25: Return off diagonals of predictive covariance
It would be useful to return predictive covariance as well as predictive variance. Clearly this is intractable for large predictive sets but still worth thinking about in my opinion for smaller ones.

jameshensman:
All the GPflow methods can now compute the full covariance matrix at predict time. The `GPmodel` class has a function `predict_f_full_cov`.

For convenience I added `predict_f_samples`, which returns samples from the corresponding distribution. 

In addition, the `conditionals.py` file has been heavily refactored. Old usage exists but raises deprecation warnings. 

Since there is new functionality and also some refactoring, I have increased the version number to 0.1.2, and changed `RELEASE.md` accordingly. 

This fixes #25 . If this is accepted, please file a new issue to remind us to remove the deprecated code in a future release.


GPflow/conditionals.py
```diff
@@ -2,246 +2,106 @@
...
-    #compute kernel stuff
-    num_data = tf.shape(X)[0]
-    Kmn = kern.K(X, Xnew)
-    Kmm = kern.K(X) + eye(num_data)*1e-6
-    Lm = tf.cholesky(Kmm)
 
-    #this is O(N M^2)
-    A = tf.matrix_triangular_solve(Lm, Kmn, lower=True)
+    #another backsubstitution in the whitened case
+    if not whiten:
```
alexggmatthews: Shouldn't the comment here say "in the unwhitened case"?

jameshensman: Yes, good spot.

GPflow/gpr.py
```diff
@@ -46,15 +46,19 @@ def build_predict(self, Xnew):
...
         Kx = self.kern.K(self.X, Xnew)
         K = self.kern.K(self.X) + eye(self.num_data) * self.likelihood.variance
         L = tf.cholesky(K)
         A = tf.matrix_triangular_solve(L, Kx, lower=True)
         V = tf.matrix_triangular_solve(L, self.Y - self.mean_function(self.X), lower=True)
         fmean = tf.matmul(tf.transpose(A), V) + self.mean_function(Xnew)
-        fvar = Kd - tf.reduce_sum(tf.square(A), reduction_indices=0)
-        return fmean, tf.tile(tf.reshape(fvar, (-1,1)), [1, self.Y.shape[1]])
+        if full_cov:
+            fvar = self.kern.K(Xnew) - tf.matmul(tf.transpose(A), A)
```
alexggmatthews: Why not use conditionals.py for predictions? Same for some other examples.

GPflow/conditionals.py
```diff
@@ -2,219 +2,106 @@
...
-    Note (and TODO):
-        At the moment, num_columns only gets used for the q_sqrt.ndim==3 case,
-        and it tells use the value of q_sqrt.shape()[2]. We need to find a way
-        to get this from the tf graph. 
-
-    See also:
-        gp_predict -- where there is no uncertainty in F
-        gaussian_gp_predict_whitened -- the same, but with whitening (centering) the F variables
-
+def conditional(Xnew, X, kern, f, num_columns, full_cov=False, q_sqrt=None, whiten=False):
```
alexggmatthews: Why do we have num_columns here? Isn't this redundant because we q_sqrt?

GPflow/gpr.py
jameshensman: So, the following methods use conditionals.py for predictions:
 - GPMC (Full covariance MCMC)
 - SGPMC (Sparse MCMC)
 - SVGP (Sparse variational GP (non-conjugate))

These methods do not:
 - GPR (Regression with Gaussian noise). 
 - SGPR (Sparse Regression)
 - VGP (Variational GP)

The reason is that for these methods, computing the posterior and then computing the corresponding conditional would introduce redundant computation.

GPflow/vgp.py
```diff
@@ -100,7 +104,10 @@ def build_predict(self, Xnew):
             A = K + tf.diag(1./tf.square(b))
             L = tf.cholesky(A)
             LiKx = tf.matrix_triangular_solve(L, tf.transpose(Kx), lower=True)
-            f_var.append( Kd - tf.reduce_sum(tf.square(LiKx),0) )
+            if full_cov:
+                f_var.append( self.kern.K(Xnew)- tf.matmul(tf.transpose(LiKx),LiKx) )
```
alexggmatthews: Why use append rather than addition in these cases?

alexggmatthews:
Hi @jameshensman 
looks good in general. I've added a few questions as line comments. 
Alex

GPflow/vgp.py
jameshensman: Here we're looping over the number of latent functions (i.e. over the columns of Y). We're going to return the variances as a NxD matrix (N test points, D functions) or as a NxNxD tensor of 'full' covariances. We want to stack them, not add them.

GPflow/conditionals.py
jameshensman: Yes, you're right, it's redundant. The num_columns integer must be the same at the number of columns of f, and the last dimension of q_sqrt. 

The reason its being passed explicitly is so that we can run a python loop over those columns: we need the integer at "python run time", not just in the graph. 

The master branch of theano now has a `tf.scan` function, which we could use for this simple loop, enabling us to remove `num_columns`. That would be a tidy addition in a future version.

PR #37, opened 2016-04-13
FITC method for GPflow

alexggmatthews:
This is not ready to merge yet. Rather it is a "help wanted" PR for @jameshensman to look at.

I working from Carl and Hannes' GPML code:

```
function [post nlZ dnlZ] = infFITC(hyp, mean, cov, lik, x, y)

% FITC approximation to the posterior Gaussian process. The function is
% equivalent to infExact with the covariance function:
%   Kt = Q + G; G = diag(g); g = diag(K-Q);  Q = Ku'*inv(Quu)*Ku;
% where Ku and Kuu are covariances w.r.t. to inducing inputs xu, snu2 = sn2/1e6
% is the noise of the inducing inputs and Quu = Kuu + snu2*eye(nu).
% We fixed the standard deviation of the inducing inputs snu to be a one per mil
% of the measurement noise's standard deviation sn.
% The implementation exploits the Woodbury matrix identity
%   inv(Kt) = inv(G) - inv(G)*V'*inv(eye(nu)+V*inv(G)*V')*V*inv(G)
% in order to be applicable to large datasets. The computational complexity
% is O(n nu^2) where n is the number of data points x and nu the number of
% inducing inputs in xu.
%
% The function takes a specified covariance function (see covFunction.m) and
% likelihood function (see likFunction.m), and is designed to be used with
% gp.m and in conjunction with covFITC and likGauss.
%
% The inducing points can be specified through 1) the 2nd covFITC parameter or
% by 2) providing a hyp.xu hyperparameters. Note that 2) has priority over 1).
% In case 2) is provided and derivatives dnlZ are requested, there will also be
% a dnlZ.xu field allowing to optimise w.r.t. to the inducing points xu. However
% the derivatives dnlZ.xu can only be computed for one of the following eight
% covariance functions: cov{Matern|PP|RQ|SE}{iso|ard}.
%
% Copyright (c) by Ed Snelson, Carl Edward Rasmussen 
%                                               and Hannes Nickisch, 2013-10-28.
%
% See also INFMETHODS.M, COVFITC.M.

if iscell(lik), likstr = lik{1}; else likstr = lik; end
if ~ischar(likstr), likstr = func2str(likstr); end
if ~strcmp(likstr,'likGauss')               % NOTE: no explicit call to likGauss
  error('Inference with inFITC only possible with Gaussian likelihood.');
end
cov1 = cov{1}; if isa(cov1, 'function_handle'), cov1 = func2str(cov1); end
if ~strcmp(cov1,'covFITC'); error('Only covFITC supported.'), end    % check cov
if isfield(hyp,'xu'), cov{3} = hyp.xu; end  % hyp.xu is provided, replace cov{3}

[diagK,Kuu,Ku] = feval(cov{:}, hyp.cov, x);         % evaluate covariance matrix
m = feval(mean{:}, hyp.mean, x);                          % evaluate mean vector
[n, D] = size(x); nu = size(Kuu,1);
cov{3} = reshape(cov{3},[],D);

sn2  = exp(2*hyp.lik);                              % noise variance of likGauss
snu2 = 1e-6*sn2;                              % hard coded inducing inputs noise
Luu  = chol(Kuu+snu2*eye(nu));                         % Kuu + snu2*I = Luu'*Luu
V  = Luu'\Ku;                                     % V = inv(Luu')*Ku => V'*V = Q
g_sn2 = diagK + sn2 - sum(V.*V,1)';          % g + sn2 = diag(K) + sn2 - diag(Q)
Lu = chol(eye(nu) + (V./repmat(g_sn2',nu,1))*V');  % Lu'*Lu=I+V*diag(1/g_sn2)*V'
r  = (y-m)./sqrt(g_sn2);
be = Lu'\(V*(r./sqrt(g_sn2)));
iKuu = solve_chol(Luu,eye(nu));                       % inv(Kuu + snu2*I) = iKuu
post.alpha = Luu\(Lu\be);                      % return the posterior parameters
post.L  = solve_chol(Lu*Luu,eye(nu)) - iKuu;                    % Sigma-inv(Kuu)
post.sW = ones(n,1)/sqrt(sn2);           % unused for FITC prediction  with gp.m

if nargout>1                                % do we want the marginal likelihood
  nlZ = sum(log(diag(Lu))) + (sum(log(g_sn2)) + n*log(2*pi) + r'*r - be'*be)/2;
  if nargout>2                                         % do we want derivatives?
    dnlZ = hyp;                                 % allocate space for derivatives
    al = r./sqrt(g_sn2) - (V'*(Lu\be))./g_sn2;      % al = (Kt+sn2*eye(n))\(y-m)
    B = iKuu*Ku; w = B*al;
    W = Lu'\(V./repmat(g_sn2',nu,1));
    for i = 1:numel(hyp.cov)
      [ddiagKi,dKuui,dKui] = feval(cov{:}, hyp.cov, x, [], i);  % eval cov deriv
      R = 2*dKui-dKuui*B; v = ddiagKi - sum(R.*B,1)';   % diag part of cov deriv
      dnlZ.cov(i) = (ddiagKi'*(1./g_sn2)+w'*(dKuui*w-2*(dKui*al))-al'*(v.*al)...
                                 - sum(W.*W,1)*v - sum(sum((R*W').*(B*W'))) )/2;
    end
    diag_dK = 1./g_sn2 - sum(W.*W,1)' - al.*al;                  % diag(dnlZ/dK)
    dnlZ.lik = sn2*sum(diag_dK);
    % since snu2 is a fixed fraction of sn2, there is a covariance-like term in
    BWt = B*W'; % the derivative as well
    dKuui = 2*snu2; R = -dKuui*B; v = -sum(R.*B,1)';    % diag part of cov deriv
    dnlZ.lik = dnlZ.lik + (w'*dKuui*w -al'*(v.*al)...
                                    - sum(W.*W,1)*v - sum(sum((R*W').*BWt)) )/2; 
    for i = 1:numel(hyp.mean)
      dnlZ.mean(i) = -feval(mean{:}, hyp.mean, x, i)'*al;
    end
    if isfield(hyp,'xu')                 % derivatives w.r.t. inducing points xu
      xu = cov{3};
      cov = cov{2};           % get the non FITC part of the covariance function
      Kpu  = cov_deriv_sq_dist(cov,hyp.cov,xu,x);           % d K(xu,x ) / d D^2
      Kpuu = cov_deriv_sq_dist(cov,hyp.cov,xu);             % d K(xu,xu) / d D^2
      if iscell(cov), covstr = cov{1}; else covstr = cov; end
      if ~ischar(covstr), covstr = func2str(covstr); end
      if numel(strfind(covstr,'iso'))>0            % characteristic length scale
        e = 2*exp(-2*hyp.cov(1));
      else
        e = 2*exp(-2*hyp.cov(1:D));
      end
      v = diag_dK-1./g_sn2;        % BdK = B * ( dnlZ/dK - diag(diag(dnlZ/dK)) )
      BdK = B.*repmat(v',nu,1) + BWt*W + (B*al)*al';
      A = Kpu.*BdK; C = Kpuu.*(BdK*B'); C = diag(sum(C,2)-sum(A,2)) - C;
      dnlZ.xu = A*x*diag(e) + C*xu*diag(e);  % bring in data and inducing points
    end
  end
end

```
(Sorry to paste apparently .m files aren't allowed as attachments). 

Currently I get numerical inversion problems when I try to run the code. 

The tests currently fail. 

Any ideas @jameshensman ?


jameshensman:
Okay, I had a look through the code. I think there was a sign error in the calculation of the mahalanobis distance. I also tried to simplify the code a little, and move it closer to pep8. 

The tests pass, aside from the tests that ensure that the full_covariance prediction matches the variance prediction: this is just because it's NotImplemented. 

I also added FITC to the SanityCheck notebook: looks great. 

I note in passing that adding a White kernel to a FITC model does not help in the same way as it does with the variational methods.

alexggmatthews:
Thank you!

alexggmatthews:
@jameshensman  Is it me or is the sign error in Carl and Hannes' expression for the NML too? 

 nlZ = sum(log(diag(Lu))) + (sum(log(g_sn2)) + n*log(2*pi) + r'*r - be'*be)/2;

jameshensman:
No, I think their code is correct. Looking back, you didn't have a sign error, but I think this introduced some broadcasting issues:
```python
tf.reduce_sum( tf.mul( tf.transpose(alpha), gamma ) ) )
```
perhaps it would all work without that transpose.

alexggmatthews:
OK yes that makes sense. There definitely shouldn't be a transpose there.

alexggmatthews:
OK I think FITC regression is finished. 

@jameshensman would you mind reviewing the full covariance predictions.

I've repeated the Snelson experiment I did in MATLAB. There are a few tweaks to the optimizer options needed to make it all work but the effect is still there. I'm going to finish it in a separate PR / issue which will reference this one.

alexggmatthews:
Also. I think we shouldn't squash this one. I'd like @jameshensman 's contribution to remain in the GitHub log.

jameshensman:
I'm not sure that's the FITC full covariance prediction. It certainly looks
like a sensible expression, but I don't think it's the expression that
you'd get from Carl's code.

If we're working with the FITC model, then far from the pseudo-points we
expect the process to look like white noise. That's not the case here. I
think the correct thing to do is to replace
```python
self.kern.K(Xnew)
```

with the equivalent FITC prior.

alexggmatthews:
Thanks for looking at it. Is it possible that what I've got is better? Would you actually want it to look like white noise? I derived it by going back to Ed and Zoubin's paper. I took the FITC posterior over points and then integrated through the GP conditional like you would in our AISTATS 2015 paper. I think this means I haven't assumed that the prediction points are independent, but why would you when you don't have to? Indeed I think Carl points out in his review paper that it wasn't clear what Zoubin and Ed intended. Perhaps I am missing the point here.

alexggmatthews:
Oh yay more test speeding up to do!

jameshensman:
Agree, it's a strange thing, but you can find plots like that in e.g. this
paper:

http://www.jmlr.org/papers/volume11/lazaro-gredilla10a/lazaro-gredilla10a.pdf

(figure 2).

What you have may well be better, and I think there may be a subtlety
between FITC and FIC. Perhaps you should mimic whatever the GPML code does?

alexggmatthews:
It smells of a straw man / misunderstanding to me. I'll ask Carl what he thinks.

markvdw:
If I remember correctly, there are two reasons why to choose the independent model, even if it seems worse. Firstly, for computational reasons. The matrix that needs to be inverted becomes as large as the number of predictions, which can be larger than M. Secondly, because if you take correlations into account in the predictions, you're using a different p(y|u) in training and test, and so the FITC model does not correspond to a well-defined Gaussan process model anymore.

jameshensman:
> the FITC model does not correspond to a well-defined Gaussan process
model anymore

This is my favourite view of FITC, that it's al alternative GP model, with
a new kernel. In this case the 'noisy' predictions make sense.

On the other hand, we're splitting hairs here, and Alex's method is also
just fine.

alexggmatthews:
I just talked to Rich Turner about it. He says do what you would do for variational. i.e integrate out q(u) using the full conditional.

Incidentally isn't GitHub useful?

@markvdw @jameshensman  I think I agree with Mark's second point but I'm not sure I think it is that important. I'd rather have q treat train and test points differently and give a better prediction. I'm not sure I agree with Mark's first point. I think prediction goes as O(nHoldOut^2) rather than O(nHoldOut^3)  (neglecting M dependence) . It is still interesting to look at correlation among a small set of hold out points and in any case we do supply a version that just gives the diagonals of the covariance and scales as O(nHoldOut).

alexggmatthews:
I sped up the tests a bit and this now passes and is ready for merge.

PR #41, opened 2016-04-13
Optimizer tolerance

jameshensman:
The tolerance of the optimizer can now be set in model.optimize()

This may break some people's existing code, if they're passing a callback and relying on the order of the arguments.


alexggmatthews:
I've merged the test savings in from the fitc branch. Therefore this should be merged (by me I guess) after @jameshensman merges the passing fitc branch.

alexggmatthews:
The tests are genuinely failing on this one rather than timing out @jameshensman .

jameshensman:
should be fixed now.

Issue #40, opened 2016-04-13
Linear kernel variance optimized to 0 with VGP model


vicow:
Hello there,

We are trying a very simple example and don't understand why the variance of a linear kernel is learnt to be 0. Here is what we do:

````
import GPflow
import numpy as np
import scipy as sp
import tensorflow as tf

# Simple training set
m = 2
n = 4
X = np.array([
[ 1.,  3.],
[ 2.,  4.],
[ 1.,  1.],
[ 2.,  1.]])
Y = np.array([1, 1, 0, 0])
Y = np.expand_dims(Y, axis=1)

# Build and train the model
kernel = GPflow.kernels.Linear(m)
probit = GPflow.likelihoods.Bernoulli()
model = GPflow.vgp.VGP(X=X, Y=Y, likelihood=probit, kern=kernel)
model.optimize()
# print model

# Predict a new data point
x_new = np.array([1.5, 5])
model.predict_y([x_new])
````

If we display the model, we see that `model.kern.linear.variance` is 0. Which leads to a predictive mean of 0.5 all the time.

Is that the expected behavior ?

**Comment by jameshensman** [comment]
Yep, that's exactly the expected behaviour. 

The reason is that the linear kernel only supports functions which pass through zero. None of the functions in your prior are able to split the training set, so the best option is to give everything probability 0.5.

You can get the behaviour you expect by doing
```python
kernel = GPflow.kernels.Linear(m) + GPflow.kernels.Bias(m)
```
which allows an offset on the functions.

lucasmaystre:
Actually, the dataset is linearly separable:

![ermau9ws5rav1wajhladalshpmizigu6kkwzqoatalshpmizigu6l0p73qkfvpuc5uaaaaaelftksuqmcc](https://cloud.githubusercontent.com/assets/278373/14522184/ade8c898-022d-11e6-9cd8-cc6f82017aeb.png)

Even if it wasn't separable, it is not intuitive to me that the "optimal" solution gives probability 0.5 to everything. Is it a consequence of taking the KL divergence from Q (the approximating distribution) instead of from P (the true posterior)?

jameshensman:
The dataset is not separable by a function that also passes through (0, 0).
All functions in the Linear kernel pass through (0, 0).

No, this is a consequence of fitting the parameters by (approximate)
maximum likelihood. There is no parameter setting that can help the model
fit, to there's a mode in the parameter likelihood at the simplest model,
i.e. with zero variance.

[...]

PR #45, opened 2016-04-17
Added constant kernel

markvdw:
A constant kernel as referenced in https://github.com/GPflow/GPflow/issues/40.


jameshensman:
Nice work @markvdw , but we already have this kernel!  I call it the Bias kernel, following Neil's naming convention.

If we want to cover both naming conventions (I'm happy to), then perhaps a cleaner solution is

```python
class Constant(Bias):
    pass
```
This has the advantage that the naming of the constant in the model will be constant, i.e.

```python
k = GPflow.kernels.Linear(1) + GPflow.kernels.Constant(1)
m = GPflow.gpr.GPR(X, Y, kern=k)
m.kern.constant.variance = 7.3  # for example
```

markvdw:
Hah! I completely missed that! I interpreted Alex's comment on a constant kernel being feasible as it not existing yet.

I would err on only keeping one naming convention and class for a particular functionality. It probably causes more confusion than it's worth to have two names for the same thing.

jameshensman:
Right, but I kind of prefer Constant to Bias!

markvdw:
Me too, as it was the name I was trying to find in `kernels.py`. I would vote to change it to constant :) But I'd leave the decision up to you and Alex.

jameshensman:
@alexggmatthews , do you have any preference for the naming of this kernel?

alexggmatthews:
No.

jameshensman:
So I fixed this by adding a Constant class, and was then confused because pushes to  my constant_kernel branch wouldn't show up here. Now that's I've pushed to constant-kernel, I think we're ready to merge.

 @alexggmatthews , could you take a quick  look when you have time?

alexggmatthews:
@jameshensman push issue not your fault I don't think. GitHub are "investigating reports of delayed notifications and delayed inbound commit processing."

jameshensman:
Oh, It was my fault. note the hyphen/underscore discrepancy! :)

PR #52, opened 2016-04-20
Shortening travis file.

alexggmatthews:


alexggmatthews:
@jameshensman I could do with your opinion on this one. We're currently failing to build because we are over time again. It is suprising this time as the exact same build was ok earlier. I think they're must be some variation in the speed in which travis gets through the script. In any case I went through the tests and looked for more savings. I found some but it meant merging some tests that might naturally be separate. Maybe have a look and see what you think? If no we need another way of making the build complete reliably.

Issue #38, opened 2016-04-13
IPython notebook showing comparison of FITC and VFE


alexggmatthews:
It would be good to have the experiment with FITC documented in an IPython notebook. 

To some extent this is a continuation of #37

jameshensman:
Done!

PR #53, opened 2016-04-22
FITC vs VFE notebook.

alexggmatthews:
Discussed in #38


jameshensman:
A couple of tiny things:

1) Could you name the data? e.g. snelson_demo_train_inputs, etc? Perhaps in a data/ directory?
2) (optional) Maybe a nice way to organize this kind of demo is to have the functions defined in a file (like your FITCvsVFE.py file) , with a `_name__=="__main__"` clause at the end to run the demo. Then import the functions into the notebook, and run the demo there too.

I really like having the notebook in the git repo so that everyone can see the results without running the code, but I also think that having the .py files is useful.

alexggmatthews:
OK @jameshensman all good points. I've done as requested.

Issue #48, opened 2016-04-18
Recommended way to serialize and de-serialize model class?


alexggmatthews:
Do we have in mind a particular way to serialize and de-serialize models so that they can be stored? Pickle doesn't currently work.

markvdw:
I did this for a while and I just stored the full parameter vector. I had to add to modify the GPflow source to ensure that if an object was created in the same way, the parameters were flattened in the same order. This involved adding a local variable inside `param.py` that counted up every time a `Parameterized` object was made. It was definitely not ideal. It was brittle in the sense that if you reconstructed the object in a different way, it may cause the parameters to be interpreted in the wrong way.

I think pickle currently doesn't work because functions are assigned to variables in the object (e.g. `._objective`), and functions aren't picklable.

I for one would like to see a lightweight pickling method that just stores the current parameter values, and not the dataset. I think this would be really useful to make scripts re-startable, or for logging past parameter values during optimisation.

alexggmatthews:
Slightly thinking out loud here. What about an xml serializer / de-serializer that uses the param class? I.e we make every param serializable and the xml tree respects the param tree?  I guess what this would be missing is how the params relate to the class. Maybe we just need to find a principled and robust way to guarantee the parameter order in the tf array. 

Thoughts? @jameshensman ?

markvdw:
That could work. Although the downside of XML is its size. A binary representation would be a lot more compact and quicker to load for large models.

I think the basic requirement is that the model object should be able to be reconstructed from the pickled parameters, without the data. In terms of the code interface, we could have `Model.save(filename)` and `Model.load(filename, X, Y)` functions. `load` would have to construct the correct Model and assign the dataset.

jameshensman:
I see three options. 

1) Pickle. 
Pros:
 - Ease of use for saving /loading models to disk

Cons:
 - possible incompatibility across GPflow versions
 - inefficient storage
 - would require some definitions of which bits of the model to pickle.

2) dump the get_free_state() vector to disk
Pros: 
 - efficient storage (don't store the data)

Cons:
 - Would require us to guarantee the order of the parameters in the model

3) other (as @markvdw suggested, or xml as @alexggmatthews suggested)
Pros:
 - efficient storage
 - compatibility across GPflow versions

Cons:
 - Would require significant coding effort to implement and maintain. 

[...]

PR #56, opened 2016-04-24
Pickle

jameshensman:
Here's some initial work on making GPflow objects picklable, see discussion in #48 . 

This allows pickling of model classes using dill. I'd like feedback on whether it works and how well. 

Even if this is not the best way to serialize model classes, it seems like a sensible option to have.


alexggmatthews:
Looks like there are currently some genuine test failures in this one.

jameshensman:
how strange, it runs just fine on my machine. Alex, could you give it a
quick go on linux?
`python testing/test_pickling.py`

alexggmatthews:
Works on my linux box. The test error on travis is complaining about reaching a maximum recursion depth. If this is genuinely the intention of dill then I'm concerned. I wouldn't use the stack to implement serialization I would use the heap. It is always possible to change one for the other.

markvdw:
Any more thoughts on this? I was looking into this a bit, and I think the biggest issue is the fact that `Param` objects have a `_parent` variable. This causes circular references, which I think cause a problem with dill...

alexggmatthews:
I think we were a bit scared of getting different answers on two Linux boxes. Also it looked like dill is using the stack to implement serialisation which I think is a bit scary. @markvdw does that match up with your looking into it?

jameshensman:
I've been having a think about json structures, and reading a little into
the different json packages that area available to python. Here are some
thoughts and discoveries:

1) it's not straightforward to put a numpy array into a json file. The json
encoder in the standard library (`import json`) won't do it.
2) We've already seen issues with the ordering of the variables in
get_free_state(). I think it would be great if we could crack that at the
same time as this.
3) serializing things to json is relatively easy (ust dump the __dict__).
It's re-loading things that's hard.

I propose we have a look at the following.
1) create a 'hash' for each object in the GPflow hierarchy. The simplest
thing would be to get the name of the object as it's printed by the __str__
method. This is unique and human readable :)
2) create a dictionary using those hashes as keys, and the _array storage
of the parameter values. This can easily be saved by numpy.savez, with
optional compression.
3) create _to_json methods for all our objects, but ignoring the actual
values of the parameters. We'll record the __class__, as well as any
pertinent properties (like whether a parameter is fixed) and all the
children of each parameterized object in a tree within json.
4) write a 'from json' method that can rebuild an 'empty' model from the
json strings above.
5) write methods that populate the empty models from the dictionary.

markvdw:
@alexggmatthews : So I haven't seen situations of different results on different machines, but I haven't really tried that yet. It's definitely true that dill uses the stack to implement serialisation. While using it, I've had to increase the maximum recursion level. Personally, this wouldn't stop me from using dill altogether, although I agree that doing the recursion manually would be better. However, I'd argue that this is not something that we should worry about. I'd rather off-load this decision onto another project like dill, and just complain about the stack being used, rather than re-implement something better from scratch :) Though, Alex, could you elaborate on your reasons for disliking the stack? I may be convinced otherwise...

@jameshensman : It sounds like it could work well, but I don't have a clear view yet of how much work it would entail to implement and maintain. E.g, how would point 4 work? It seems like this would be brittle, and like it would need to re-implement much of what's already done by pickle...

I'm curious to see exactly what goes wrong with pickling, and whether this can be fixed, while keeping the nice separation between GPflow model code and code for loading/storing objects...

jameshensman:
Thanks for the input @markvdw .

re pickle: I'm a bit scared of increasing the recursion depth! 

Well, 1 - 2 were easy, I implemented them in PR #92 

3 and 4 might turn out to be easy, we'd have to see how we can extend json's encoder/decoder objects.

alexggmatthews:
Hi @markvdw , 
Thanks for your reply.
I went back and repeated the experiment and I get a function depth exceeded issue as well. I'm not sure why this sometimes pops up and sometimes doesn't tbh. It seems to do it more often than not as far as I can tell.

I guess the question about the stack relates to whether we trust a given library. In the case of dill it is not part of the python core libraries and so the quality of the code is perhaps more in question. 

I agree with you in that my ideal solution to this problem is to out source as much serialization code as possible to a decent library. If you get it wrong serialization can be a total pain to maintain.

 @jameshensman is proposing #92  which does involve some of our own code but he has cleverly managed to get it pretty short. I'm having a look at it as we speak and I'm keen to get a solution to this decided and in soon.

Alex

alexggmatthews:
This Dill issue looks at least relevant to the discussion. I'm not sure it is exactly what we are seeing but it seems by looking at the discussion that dealing with circular references in general is a thorn in the side of all python serialization code

https://github.com/uqfoundation/dill/issues/58

markvdw:
Yeah, it's 4 that I'm mostly worried about. This may be lack of Python knowledge, but currently the way I'm envisaging, would require creating a new object through the constructor. Then we'd need to pass the correct arguments to the constructor, and fix up other properties which aren't accessible through the constructor afterwards. But, I should look into json's encoder and decoder options, haven't looked at these yet. I hope it turns out easy :)

Does anyone know how pickle does this? It seems to bypass the constructor altogether and just sets elements in `__dict__`...

I'm also uneasy about changing the recursion depth. The solution that I'd be least unhappy with, would be to change it only for the duration of the pickling process.

Oh one final thought, regarding the different ordering of parameters from `get_free_state()`. One thought that I had was that if the object creation code was the same, the ordering should be the same. To get this behaviour, I added a global variable in `param.py`, which just counted up whenever a `Param` object was made. It's a unique id, which is also has the right ordering.

alexggmatthews:
OK. @markvdw @jameshensman I'm going slightly cross eyed looking into this today so I'm going to have a break. I spent some time looking into the various pickle related options. With regards to the TensorFlow aspect it might be relevant to consider how they've done it in skflow:

https://github.com/tensorflow/skflow/commit/afb324dfe8f4aa5e350de90116b6106242dc8164

which in turn uses 

tf.train.Saver

It seems to be accepted buy those guys that pickle doesn't work well with TensorFlow graphs: 

https://github.com/tensorflow/skflow/issues/37 .

markvdw:
So my current opinion is that we don't really need to worry about saving Tensorflow graphs, since we recompile them all the time any way. What we really want is to store the model structure, as defined in the object hierarchy, as json/pickle do, while stripping out all the tensorflow variables, which can be recreated.

alexggmatthews:
@markvdw I'm leaning that way too - it seems pretty tricky to serialize the computational graph. Do you agree that makes it less likely we can use pickle or dill with minor tweaks?

markvdw:
Well, not the way I see it. I had a play around with this over the weekend, and I just excluded any tensorflow variables from being serialised in `__getstate__()`. Then on recovering, I'd just initialise new tensorflow variables in `__setstate__()`.

I think I got pretty close to something working. I think the biggest issue was the cyclic references. But I didn't exactly pin down where the pickling failed. The way to do this would be to start with a really simple model, and build it up with sub-classes slowly. An added complication is that the results aren't consistent between runs, like you observed as well. I'm pretty sure this is related to the fact that `.sorted_params` doesn't have a consistent order. This influences the order of the pickling procedure too. But, like I said, that can be fixed.

alexggmatthews:
@markvdw . That sounds like it might work to me. @jameshensman  what do you think?

In terms of debugging the pickler you could try defining a class in the following way

```
import dill
import pickle

class MyPickler (pickle.Pickler):
    def save(self, obj):
        print 'pickling object', obj, 'of type', type(obj)
        pickle.Pickler.save(self, obj)

pickler = MyPickler(open('foo','w'))
pickler.dump(bar)
```

Adapt as required.  Using pickle rather than cPickle will also give better strack traces during debug. Maybe you've already tried something like that?

jameshensman:
Encouraged by your experiments, @markvdw I made PR #96 , which I think does the correct thing with both cyclic references and tf_objects. Comments welcome.

jameshensman:
+1 for advocating pickle over cPickle for better traces.

jameshensman:
I'm closing this PR in favour of #96 . 

We might want to come back to some of this discussion before merging #96 though

Issue #57, opened 2016-04-25
Combine mean functions


PabloLeon:
What would the best way be to combine mean functions? Ideally I would like to add addition and multiplication to the MeanFunction class, but I keep running into problems with the Parameter specifications (and I feel that your setup would allow a more elegant solution that implementing multiplicative/additive mean function classes).

jameshensman:
Hi @PabloLeon 

I agree it would be nice to have addition and multiplication of the mean function classes. Maybe you could share what you've done already as a pull request and we can have a look?

Another nice thing would be to give the mean_function classes the ability to slice their inputs, like the kernels can.

alexggmatthews:
@PabloLeon @jameshensman . What is the status of this? Can we close  it?

jameshensman:
I think it's all up and running, right @PabloLeon ?

PabloLeon:
yea basic functionality should be there!

alexggmatthews:
Thanks for your efforts.

PR #58, opened 2016-04-26
Mean function Combination

PabloLeon:
This is not functional (yet) but only serves as reference to #57 .

I introduce addition and multiplication for mean functions and provide two Polynomial mean functions to handle higher order combinations (linear * linear and combinations of polynomial with some other function).
@jameshensman I need a way to access the  parameters of the constituent mean functions (e.g. function1.A, function1.b etc.). I tried wrapping them into Parameterlists and running reduce(tf.add/tf.matmul) on them, but without any luck. Any other feedback would be highly appreciated!

### DIFF
--- a/GPflow/mean_functions.py
+++ b/GPflow/mean_functions.py
@@ -1,6 +1,6 @@
 import tensorflow as tf
 import numpy as np
-from GPflow.param import Param, Parameterized, ParamList
+from GPflow.param import Param, Parameterized
 
 class MeanFunction(Parameterized):
     """
@@ -15,22 +15,15 @@ class MeanFunction(Parameterized):
         raise NotImplementedError, "Implement the __call__ method for this mean function"
     
     def __add__(self, other):
-        raise NotImplementedError, "Implement the __add__ method for this mean function"
-        
+        return Additive(self,other)   
 
     def __mul__(self, other):
-        raise NotImplementedError, "Implement the __mul__ method for this mean function"
+        return Product(self, other)
         
 
 class Zero(MeanFunction):
     def __call__(self, X):
         return tf.zeros(tf.pack([tf.shape(X)[0], 1]), dtype='float64')
-        
-    def __add__(self, other):
-        return other
-    def __mul__(self, other):
-        return self   
-
 
 class Linear(MeanFunction):
     """
@@ -47,30 +40,7 @@ class Linear(MeanFunction):
         self.b = Param(b)
     def __call__(self, X):
         return tf.matmul(X, self.A) + self.b
-        
-    def __add__(self, other):  
-        if isinstance(other, Zero):
-            return self
-        elif isinstance(other, Constant):
-            return Linear(other.A, other.b + self.c)
-        elif isinstance(other, Linear):
-            return Linear(self.A + other.A, self.b + other.b)
-        elif isinstance(other, PolyAdd):
-            return PolyAdd(self, other)
-        elif isinstance(other, PolyProd):
-            return PolyAdd(self, other)
-            
-    def __mul__(self, other):
-        if isinstance(other, Zero):
-            return other
-        elif isinstance(other, Constant):
-            return Linear(other.A, other.b * self.c)
-        elif isinstance(other, Linear):
-            return PolyProd(self, other)
-        elif isinstance(other, PolyAdd):
-            return PolyProd(self, other)
-        elif isinstance(other, PolyProd):
-            return PolyProd(self, other)
+
 
 class Constant(MeanFunction):
     """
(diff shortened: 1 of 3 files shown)


jameshensman:
Hi @PabloLeon 

Thanks for this!  

What an interesting approach. What do you see as the problem with having `add` and `mul` mean functions? I can see that your method ends up being more parsimonious in many cases, but I'd hope that tensorflow would sort that out under the hood anyway.

PabloLeon:
True, that would be a simpler solution. I did it this way since most combinations amount to the three basic  (provided) mean functions and this way I maintain information about the class of mean function (Linear, Zero, Constant in cases where the combination does not change the type) which for my needs can be useful.

jameshensman:
Okay, that sounds interesting. I think you're going to struggle with the logic in trying to implement your system with the parameter framework, since you're trying to initialize new mean functions which expect to have their own parameters, using existing constituent parameters. 

My suggestion is to use the simpler approach that I've outlined, and then use the same logic as you've applied in the class structure 'post-hoc', in order to get a simplified representation of the function. I can't say exactly how that would work without more detail from you, but for example I could envisage something like this:

```python
def is_linear(mf):
    """
    Takes a mean_function and returs a boolean indicating
    whether the structure is overall linear.
    """
    if isinstance(mf, [mean_functions.Linear, mean_functions.Bias]):
        return True
    elif isinstance(mf, [mean_functions.Add, mean_functions.Prod]):
        return is_linear(mf.first_part) and is_linear(mf.second_part)
    else:
        return False
```

and I could envisage a similar scheme to work out the 'effective' parameters of the linear function. Does that help?

PabloLeon:
I've updated the code as suggested and the combination seem to work nicely (I will check this for some synthetic data now). 
@jameshensman I will update the mean functions with type information and also change the first_part/second_part to something more informative (first_prod/first_add or similar). Do you think it would be worthwhile to update the naming for combinatorial mean functions as for combinations of kernels? Also, is there a reason why the topmost composition of kernels is not added to the naming (e.g. I have the product of a kernel with the sum of two kernels, the names only reflect the add class)?

GPflow/mean_functions.py
```diff
@@ -81,61 +51,20 @@ def __init__(self, c=np.zeros(1)):
...
-    def __add__(self, other):
-        if isinstance(other, Zero):
-            return self
-        else: return PolyAdd(self, other)        
-    def __mul__(self, other):
-        if isinstance(other, Zero):
-            return other
-        else: return PolyProd(self, other)
\ No newline at end of file
+        return tf.matmul(self.first_part(X), self.second_part(X))
```
jameshensman: Do you mean matmul here, or simply mul?

jameshensman:
Hey @PabloLeon , nice work!

Yeah, I think it would be really nice to have the same naming scheme for the mean functions as the kernels. 

If I understand your question correctly, the answer is that the topmost level is just called 'kern'. For example, let's make such a kernel:
```python
k1 = GPflow.kernels.RBF(1) * GPflow.kernels.Linear(1)
k2 = GPflow.kernels.Matern32(1)
k = k1 + k2
```
now `k` is an add instance. Naturally you can do `k.prod.rbf.variance=1`. When you add k to a model, it gets called `kern`... so you can do `m.kern.prod.rbf.variance=1`. `m.kern.add` would be tautological, and `m.add` doesn't seem descriptive enough. 

I'd love to accept this PR if 1) you implement the same naming scheme as for the kernels (good idea) and 2) you write tests to cover the added code. Thanks!

GPflow/mean_functions.py
PabloLeon: Yes it should be tf.mul, will change that!

PabloLeon:
I've changed the names to aguend, addend, multiplicant and multiplier for the individual components. This does not follow the kernel naming scheme, but seemed more intuitive considering that the mean functions are composed of two components. In addition that allows to visualize the structure of the mean functions nicely. 
E.g. for complex compositions like in:
```
comp1 = mean_functions.Linear(1) + (mean_functions.Constant(9) * mean_functions.Constant(12) * mean_functions.Constant(100))
comp2 = (mean_functions.Linear(1) + mean_functions.Constant(9)) * (mean_functions.Constant(12) * mean_functions.Constant(100))
```

This is not ideal, as it does not follow the kernel naming scheme, but wanted to leave it for discussion - I am happy to change it to add_1, add_2, prod_1, prod_2 @jameshensman what is your preference here? 

@jameshensman For the tests - I thought about the following tests: 
- a + b --> additive, a * b --> product
- (a+b) * c  = ac + bc; a + b -b = a, a + b * c = bc + a != (a+b) * c
for a constant kernel and some linear dataset for all mean function types (Linear, Zero, Constant, Additive, Product). Anything I should add to the tests (not quite sure how to test, since the interactions of kernels/mean_functions/data can get quite complex...)?

PabloLeon:
@jameshensman  I've updated the tests and changed the naming to add_1/2 prod_1/2. I still  have errors when I try to replicate test_basic_output_shape in TestMeanFuncs for the compositional mean functions.
This is the error message: 

> InvalidArgumentError:  Expected begin[0] == 0 (got 0) and size[0] == 0 (got 2) when input.dim_size(0) == 0  [[Node: Slice_3316 = Slice[Index=DT_INT32, T=DT_DOUBLE, device="/job:localhost/replica:0/task:0/cpu:0"](Slice_3315, Slice_3316/begin, Slice_3316/size)]]

Do we need to provide a _slice method for mean functions (as in the kernel class)? I am a bit confused why we need make_tf_array() in the TestMeanFuncs setUp, not calling make_tf_array() on the compositions of mean functions does not result in errors.

PR #63, opened 2016-05-04
Added `evalK` function for easy evaluations of kernels.

markvdw:
Added a simple function to evaluate kernels. Usage:

```
import numpy as np
import GPflow

kern = GPflow.kernels.RBF(1)
X = np.linspace(0, 10, 11)[:, None]
print(kern.evalK(X))
```

I'd like to add a test for this. Any ideas on where would fit?

### DIFF
--- a/GPflow/kernels.py
+++ b/GPflow/kernels.py
@@ -38,6 +38,17 @@ class Kern(Parameterized):
     def __mul__(self, other):
         return Prod([self, other])
 
+    def evalK(self, *args):
+        tf_vars = [tf.placeholder('float64', name='evalK_%i' % i) for i, x in enumerate(args)]
+        feed_dict = dict(zip(tf_vars, args))
+        fs = tf.Variable(self.get_free_state(), name='evalK_free_state')
+        self.make_tf_array(fs)
+        with tf.Session() as sess:
+            sess.run(tf.initialize_all_variables())
+            with self.tf_mode():
+                K = self.K(*args).eval(feed_dict=feed_dict)
+        return K
+
 
 class Static(Kern):
     """


alexggmatthews:
@jameshensman this one seems to have fallen by the wayside. What do you think?

jameshensman:
I like the idea in principle, but I would rather it was done via the
autoflow mechanism. This kind of things take a surprising amount of
maintainance: see the recent kerfuffle for example.

Problem is, autoflow works with the model class, not the kernel class.
Perhaps we could have an AutoFlow function in the GPModel class?

```python
def GPmodel(...):
    # <snip>
    @AutoFlow()
    def compute_kernel_matrix(self):
        return self.kern.K(self.X)
```
?

An alternative would be to make the autoflow mechanism universal, so it
didn't have to borrow from the model's `_free_state` vector or `_session`,
but had its own.

markvdw:
I'll see what I can do for the universal autoflow mechanism.

jameshensman:
@markvdw how's this going? 

As you have been experiencing, the autoflow mechanism is really nice for saving a lots of code and adding friendly features, but it's a bit of a pain to maintain. 

Perhaps we could reconsider just adding a `compute_kernel_matrix` function to the GPmodel class (then it will appear in `GPR`, `GPMC`, `VGP` etc.

markvdw:
I haven't looked at this yet. Perhaps I can take a look this weekend. I think that the nicest solution would be to get AutoFlow working with `Kernel` objects, like you mentioned. This just requires some thought and digging through the code.

I'd like to at least figure out how difficult this would be before resorting to a `compute_kernel_matrix` function in `GPmodel`. It's an easy solution, but it doesn't really make total sense, since kernel matrices exist and can be evaluated independently from models...

jameshensman:
Okay, sounds good.

I agree that it would be nicer to have the autoflow function available
anywhere. Perhaps what's needed is to have each autoflow create its own
session, _free_vars object (and perhaps its own `tf.graph` too).

jameshensman:
I had a go at this on the branch 'universal_autoflow'. I think it makes things quite a bit tidier, actually. Comments welcome.

PR #65, opened 2016-05-19
Added cholesky decomposition with jitter

### LINKED ISSUE #64: Sample from model
What would be the preferred way to sample from a GP with mean function/compositional kernel? I tired the various sample/predict and also the evalK() from #63 but I can't make it work.

PabloLeon:
Heavily based on the GPy implementation. Should fix #64.

### DIFF
--- a/GPflow/gpr.py
+++ b/GPflow/gpr.py
@@ -5,6 +5,7 @@ from .densities import multivariate_normal
 from .mean_functions import Zero
 import likelihoods
 from tf_hacks import eye
+import jitter_cholesky
 
 class GPR(GPModel):
     def __init__(self, X, Y, kern, mean_function=Zero()):
@@ -29,7 +30,7 @@ class GPR(GPModel):
 
         """
         K = self.kern.K(self.X) + eye(self.num_data) * self.likelihood.variance
-        L = tf.cholesky(K)
+        L = jitter_cholesky(K)
         m = self.mean_function(self.X)
 
         return multivariate_normal(self.Y, m, L)
@@ -47,7 +48,7 @@ class GPR(GPModel):
         """
         Kx = self.kern.K(self.X, Xnew)
         K = self.kern.K(self.X) + eye(self.num_data) * self.likelihood.variance
-        L = tf.cholesky(K)
+        L = jitter_cholesky.jitchol(K)
         A = tf.matrix_triangular_solve(L, Kx, lower=True)
         V = tf.matrix_triangular_solve(L, self.Y - self.mean_function(self.X), lower=True)
         fmean = tf.matmul(tf.transpose(A), V) + self.mean_function(Xnew)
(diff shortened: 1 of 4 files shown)


jameshensman:
Hi @PabloLeon 

Thanks for this! 

Unfortunately, I think there's an issue: tf.cholesky will never raise an error, because that only constructs the graph to compute the Cholesky. Numerical errors that occur during the Cholesky decomposition will occur only once the graph is executed.

PabloLeon:
@jameshensman yes you are right.
I could delete the except block and simply increase the jitter with num_tries?

mathDR:
If the number of tries exceed a predefined amount, you could replace the decomposition with the closest non-singular matrix. Probably should warn the user if you do this though. Let me know if you are interested and I could do a PR.

PabloLeon:
@mathDR sounds good to me!

mathDR:
After thinking about this for a while, I think the idea of adding jitter is the way to go.  (Instead of replacing with the closest non-singular matrix -- in fact, there is no *closest* as any nonzero perturbation will lead to a nonsingular matrix).  

The issue will be with the condition number of the original covariance matrix:  if it is high, then there may be numerical problems.  Maybe the answer is to report the condition number of the matrix and give a warning that the original covariance is ill conditioned?

jameshensman:
Adding jitter certainly work well for GPy. This issue here is that you'd
have to implement the detection of the CN *inside* the tensorflow kernel.

PabloLeon:
@jameshensman any advice on how to detect it? Alternatively, can we just inform the user about the added jitter at graph execution and increase the number of attempted iterations?

PabloLeon:
Any news on this? Is the approach in #72 the way to go/ shall I close this?

alexggmatthews:
Long term I'm not sure it has been conclusively resolved yet. It has been proving tricking to get a clean solution so far. @jameshensman perhaps we should have another think about this thread?

In the shorter term #74 makes it possible to add white noise that is constrained not to grow too large which may well help in many situations put is not a perfect solution.

PabloLeon:
@alexggmatthews Thanks! My models are created automatically, i.e. I have very little control and can't add white noise as needed but would need to catch the singularity issues automatically (by adding jitter/adding white noise kernels). 
So either I need to catch the error after the graph is executed (not sure how to do this automatically) or "predict" the singularity before tf creates the graph? Any advice on how/where to best integrate this functionality (sorry still struggling with tensorflow etc.)?

jameshensman:
Hey @PabloLeon 

How is numerical stability in GPflow for you these days, has it improved?

There was a small fix that went in recently that prevented parameters becoming numerically zero, and this fixed problems for several people. If your use case is also fixed, perhaps we can close this?

PabloLeon:
@jameshensman Yes it looks better now! Thanks for the fixes! We can close/delete this PR.

jameshensman:
Great! I'm pleased it's working well. If there are still issues, I guess they will be separate from those already fixed, so please do open up a new bug report describing them if so.

PR #71, opened 2016-05-27
Periodic kernel

alexisboukouvalas:
Added support for periodic kernel as described in Equation (47) of 
    D.J.C.MacKay. Introduction to Gaussian processes. In C.M.Bishop, editor, Neural Networks and Machine Learning, pages 133--165. Springer, 1998.
Added single unit test test_kerns/TestPeriodic
All tests pass.

In future work, I hope to add support for Bayesian GPLVM to use this kernel.


alexggmatthews:
Hi @alexisboukouvalas hope you all good with you. This seems to me like the sort of thing we should have. 

I'd be happy to help with the merging. 

I remember James Lloyd was at one point playing with cos(w(x-x')) as a kernel. Do we know what the relevant merits of the two variants are? Maybe we ought to have both. 

@jameshensman do you have an opinion?

markvdw:
Doesn't the cosine kernel only produce exact cosines? While the periodic produces a sqexp constrained to repeat over some interval.

alexggmatthews:
Thanks @markvdw , @alexisboukouvalas sorry for the red herring.

jameshensman:
And we already have the Cosine kernel :)

alexisboukouvalas:
Thank you all for your feedback.
Thanks @jameshensman I will have a look later on at implementing the subspace-projection kernel.
I will also try the Psi functons for BGPLVM for the periodic kernel.

Have a good weekend everybody. Let me know if you would like me to change anything for this pull request.

jameshensman:
Just looking at this now

PR #72, opened 2016-05-28
Make model objective calculation robust to singular matrix errors

alexisboukouvalas:
This is some code James wrote a couple weeks back as we were hitting 
'LLT decomposition was not successful.' during model optimisation.

I have modified the code to be more specific, that is catch only the exceptions raised by Cholesky and MatrixTriangularSolve with error code 3 that is is always generated when the invalid argument: LLT decomposition error is thrown.

I would appreciate ideas and feedback if you think this approach is worthwhile or there is a better way to make the objective calculation more robust to numerical errors.


jameshensman:
Alexis, if you merge the master branch into your feature-branch the tests should run a lot (10x) faster, which will help us look at this.

alexisboukouvalas:
thanks James - I have done the merge.

alexggmatthews:
Thank you for the pull request. There was some related discussion on #65 . I note that whereas the original idea on that PR would not have worked as intended this one will.

@alexisboukouvalas @jameshensman  I think this issue is pretty important and merits some thought and discussion. 

I understand putting inf for the objective . Why put zeros for the gradient of the objective, please? 

In terms of how this compares to automatically adding jitter I guess we are suggesting that people effectively use the white kernel to add jitter instead?

Perhaps we should consider suggesting the user use a white kernel in the warning? I guess it would be nice to be able to constrain the white kernel to have small noise but it isn't clear to me how to do that.

alexisboukouvalas:
Thanks @alexggmatthews 
I run into this issues when using GPflow on another project with James and the kernel we use there is quite sensitive to numerical errors. We were hitting these issues during kernel parameter optimisation and this change helped the optimisation get around it. What we were doing up to that point is trying random restarts or setting the maximum number of iterations to a smaller number, both of which are rather inelegant.

Note that we were using a white kernel + our kernel but the optimiser can still go to parameter areas where the White kernel variance is small enough to be inconsequential. We could a prior on that to help things.
I like the idea of modifying the error message to offer some suggestions such as inserting a White noise kernel, perhaps with a prior on the variance? Also in GPy there is the ability to hard constrain the values of hyper parameters within a specified interval, I guess this is similar to putting a uniform prior but it is reported as a constraint.
@jameshensman let us know what you think.

jameshensman:
to help with this, I just made a PR which adds a logistic transformation, #74 . @alexisboukouvalas , could you have a look and see if that helps in this case?

I'm not against accepting this PR, a similar strategy works really well for GPy, but I'm definitely in favour of keeping things simple where possible.

alexisboukouvalas:
Thanks @jameshensman, the change #74 helps fix things as we can constrain the hyperparameters. As such it is also very general. Could we add a FAQ section somewhere that then includes:
1. If I hit LLT decomposition failed errors what can I do?
-> Try adding a white kernel with a small variance or a logistic transorm to stop the optimizer from considering numerically unstable parameter values.

Shall I close this PR then? I don't think we need it as James's solution is more general and elegant.
The only disadvantage is that is it less automatic and requires user intervention.

alexggmatthews:
Hi @alexisboukouvalas @jameshensman I've got a big talk tomorrow so I'll be going to ground until Friday now. We can talk about this then, Thanks, Alex

jameshensman:
Don't close this yet, it's still interesting, just needs a little more discussion.

alexisboukouvalas:
We now have a nice script to reproduce this issue see #78 
In the master branch the code currently fails with
'InvalidArgumentError: Input matrix is not invertible.
	 [[Node: MatrixTriangularSolve = MatrixTriangularSolve[T=DT_DOUBLE, adjoint=false, lower=true, _device="/job:localhost/replica:0/task:0/cpu:0"](Cholesky, sub_1)]]
'

Using the robustobjective branch we get
'
Warning: Matrix decomposition failed due to to singular matrix, setting objective to inf.
At iterate   10    f=  5.60719D+02    |proj g|=  6.99262D+01
'

alexggmatthews:
OK so @jameshensman and I have been thinking about this aspect of the numerical stability thread. I can see the sense of trying setting the objective to Inf but I'm not so sure about setting the gradient to zero since this corresponds to convergence. This could be particularly bad for those optimization algorithms which ignore the objective and use only the gradient.

I might just throw some ideas into the mix. I'm not saying these are necessarily the solution

1) What would happen if the gradient returned was Nan? I guess it would depend on how the given optimizer handles that.

2) Although it is messy we could do something like jitchol by having a tensorflow variable that controls the (relative?) jitter for all calls to cholesky made in GPflow. Then when we caught the error in the optimizer we could increase the  jitter parameter and recall. It feels suboptimal to do this in such a global way.

alexisboukouvalas:
Thanks @alexggmatthews , @jameshensman.
I have modified the code to set the gradient to nan and this still works with the L-BFGS-B optimizer.

I have added a small script based #78 that I am building towards a full test.
I also tried the other scipy optimizers that require gradient information.
CG works, TNC does not hit the error, SLSQP gets stuck in infinite loop and BFGS converges but then the prediction produces nan (possibly because of hyperparameter values).

Note that the approached coded up here will not work when using the tensorflow optimizer as the ObjectiveWrapper is only used in optimize_np. A different approach would be needed for the tensorflow optimizer. 

Perhaps an approach like (2) would be needed then. We would catch the error outside tensoflow/scipy and then increase the jitter terms everywhere. This may be a bit slower than (1) but is more likely to work for all optimizers. This is the approach discussed in #65. 
Also related to this is #80 as increase the jitter would be better done using a tf.placeholder rather than a tf.constant which necessitates rebuilding the graph.

alexggmatthews:
Hmm @alexisboukouvalas thanks for the experiments. @jameshensman  thoughts?

alexisboukouvalas:
I will close this given the recent changes with the controllable jitter via settings.

PR #82, opened 2016-06-05
Fill constants

jameshensman:
Some simple changes to use `tf.fill` were appropriate, instead of creating an array ones ones and then multiplying by a constant.

### DIFF
--- a/GPflow/kernels.py
+++ b/GPflow/kernels.py
@@ -1,7 +1,6 @@
 from functools import reduce
 
 import tensorflow as tf
-from .tf_hacks import eye
 import numpy as np
 from .param import Param, Parameterized
 from . import transforms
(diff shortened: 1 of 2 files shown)


alexggmatthews:
Hi @jameshensman, looks good. For future reference why is tf.fill preferred please?

jameshensman:
Just because the graph will be simpler, gradient computation should be
(fractionally) faster. Take for example the Kdiag function in the
stationary kernel class. We had

`return X[:,0] * 0 + self.variance`

Which has a slice, multiply and add op. Now we have

`return tf.fill(tf.pack([tf.shape(X)[0]]), tf.squeeze(self.variance))`

which has a shape, squeeze and fill op.

The second is a more 'correct' way of doing things, to me.

J.

Issue #85, opened 2016-06-07
Autoflow fails when making predictions with different models


markvdw:
Another failure case of Autoflow:

```
import numpy as np
import GPflow
import tensorflow as tf

X = rnd.rand(10, 1)
Y = rnd.rand(10, 1)
mv = GPflow.sgpr.SGPR(X.copy(), Y.copy(), GPflow.kernels.RBF(1), Z=np.random.rand(4, 1)*np.max(X))
mf = GPflow.sgpr.GPRFITC(X.copy(), Y.copy(), GPflow.kernels.RBF(1), Z=np.random.rand(4, 1)*np.max(X))

mv.predict_y([[3]])
mf.predict_y([[3]])
mv.predict_y([[3]])
```

The cause: When running several models in parallel, AutoFlow shares its `self.tf_args` variable between models. So after `mf.predict_y` has been called, the AutoFlow object will now contain the tensorflow arguments for mf. These are the wrong placeholders when `mv.predict_y` is called again. The root of problems here, is that the decorator is called when the `GPModel` class is created, so the decorator is shared between all objects that inherit from it.

A quick fix would be to remove the caching mechanism. A different fix would be to store `tf_args` in `instance` instead of `self`.

I think the latter solution is preferable, so I'll submit a pull request for it.

markvdw:
Pull request: https://github.com/GPflow/GPflow/pull/86

jameshensman:
Thanks again for fixing this @markvdw .

PR #86, opened 2016-06-07
Fixed AutoFlow sharing `self.tf_args` between different instances

markvdw:
Issue discussed here https://github.com/GPflow/GPflow/issues/85.


alexggmatthews:
@jameshensman do you mind having a look at this one, please? I think you have done more AutoFlow than me. Thanks, A

jameshensman:
Yeah, this is clearly the right thing to do, nice work. 

Line 207 `in model.py` deletes the autoflow graphs when the model is flagged to be recompiled. Could you do the same for the tf_args, please?

jameshensman:
Oh, and please add the failure-case you specified in #85 as a test in `testing/test_autoflow.py`. I'm happy to help if it's not clear how it should work.

markvdw:
If the graph is deleted then `tf_args` is automatically recreated, so it shouldn't make a difference. I instinctively feel like it's best to rely purely on the existence of the graph to indicate whether AutoFlow is stale, rather than deleting every variable we need to store. This can be a hard thing to keep consistent.

I've added a test which passes with the changes but fails on master.

jameshensman:
okay, that's fine with me. Just merging in master and adding a docstring to your test, then I'll accept this PR if the build/coverage tests pass.

Good test btw, nice and succinct.

Issue #81, opened 2016-06-03
Autoflow fails after resizing of variables


markvdw:
Hi all,

I'm worried Autoflow still isn't working as expected. I'm on the latest commit. Below is an experiment where I add an extra inducing point. I think the bound calculation is good, but the predictions are different if I copy the parameters over to a new model.

Haven't had the chance to investigate thoroughly yet, but I thought I'd share.

[edit] More concise piece of code plus a workaround. The problem is that after adding the new inducing point, the prediction `py2` is the same as the prediction `py1`. Only after deleting the prediction graph, do I get a correct prediction.

```
import numpy as np
import matplotlib.pyplot as plt
import GPflow
import datasets

Y, X, _, _ = datasets.load_data('snelson1d')
m = GPflow.sgpr.SGPR(X, Y, GPflow.kernels.RBF(1), Z=np.random.rand(2, 1)*np.max(X))

py1, pyv1 = m.predict_y([[3]])

Z = m.Z._array.copy()
m.Z._array = np.vstack((Z, [3]))  # Add an extra inducing point
m._compile()                               # Compile to ensure correct parameter mapping

py2, pyv2 = m.predict_y([[3]])

delattr(m, '_predict_y_graph')
py3, pyv3 = m.predict_y([[3]])

print py1
print py2
print py3
```

markvdw:
Ok, so I've looked into this issue a bit more, and I think I've found the root cause. I've edited the code above to give a more concise example.

The problem is that when I change the size of `Z`, `predict_y` still only "sees" the old size of `Z`. I confirmed this by adding a print op in `build_predict`. The problem is that after `Z` has changed size, `predict_y` still lifts the old indices out of `_free_vars`, since it hasn't been re-run after the `_compile()`.

I admit this is a bit of an edge case, but it would be fixed by re-creating the graph at every call. This slows down a call by ~200ms in some simple tests for me (from just a few ms).

Any ideas on how to fix this? Or whether to even bother? I've got a workaround now, so I'm fairly happy...

jameshensman:
I'm not sure this is a bug: the convention is that since `Param._array` starts with an underscore, you shouldn't really be accessing it. In other languages this would be a private variable. 

The correct way to assign to a `Param` is to use assignment (!), i.e.
```python
m = GPflow.sgpr.SGPR(X, Y, Z, ...)
new_Z = np.random.randn(10, 1)
m.Z = new_Z
```

[...]

PR #87, opened 2016-06-07
adding value property to Param class

jameshensman:
As discussed in #81


markvdw:
Looks good James, did you just want me to sign off on the merge? Or did you also want me to look into the issue of setting `_needs_recompile` on a new `Param` object assignment?

Currently, I don't think assigning a new `Param` object will trigger a recompile. What would be a good way to do this? I can only think of the solution of writing a getter and setter for each attribute, which may be a bit verbose...

jameshensman:
Well, now that you mention it, perhaps this would be a good opportunity to fix that. Working on it now!

jameshensman:
Okay, I added a functionality so that when you replace a Param with a new Param, 

1) The old parameter gets unlinked from the tree (i.e. it can no longer reference a 'parent')
2) The new parameter gets linked to the tree (this happened anyway)
3) The model gets told to recompile (if there is a model to recompile, of course)

alexggmatthews:
@jameshensman @markvdw What will be the behaviour of the functionality with arrays. Do we want to make .value give a copy? Something to think about.

jameshensman:
I think we do, otherwise if the user does something interesting to the result, it will change the model. Example (using curent code):

```python
m = GPflow.model.Model()
m.p = GPflow.param.Param(np.arange(3))
# m.optimize, whatever

result = m.p.value
print m

# model.p transform:(none) prior:None
# [ 0.  1.  2.]

result += 4
print m
# model.p transform:(none) prior:None
# [ 4.  5.  6.]
```

markvdw:
I really like the fix to setting the `_needs_recompile` flag. Essentially this relies on the fact that `Model` inherits from `Param` right? Hadn't looked at this code yet, but I think it's really elegant.

Ok, so the current situation as I see it:
- Can read parameter values using `Param.value`
- Can set the values of the parameter using `.param_name = some_value` (does not trigger re-compilation)
- Can re-set the whole parameter by assigning a new `Param` object (triggers re-compilation)

Looks ideal to me! Just needs fixing of the typo `FailIf` in the test. Shouldn't that be `failIf`?

Also, `failIf` seems to be deprecated. It should be `assertFalse` [https://docs.python.org/2.7/library/unittest.html].

markvdw:
Ok I uploaded a small fix. I couldn't merge into master though due to some conflicts. Could someone else look at this? I don't want to erase anybody else's changes.

jameshensman:
Taking a look now.

jameshensman:
Turns out git doesn't like merging things if you rename methods. All fixed now.

PR #89, opened 2016-06-08
a simple function to work out where a parameter is on the free_state

jameshensman:
This possibly needs more work, but for lots of reasons we'd like to be able to work out where on the _free_state array any Parameter lives. For example, if we have sample from HMC, we might want to work out which column is the (transformed) lengthscale.

Comments and suggestions welcome.


markvdw:
Hi James, this is just what I need for logging parameter values during optimisation and sampling. So I'm in favour of getting this in soon. Is there much more you want to add to this, or are you happy for this to be merged if the tests pass and after I've reviewed the changes?

jameshensman:
Hi @markvdw, 

One more feature in the pipeline: turning a trace into a dictionary. Have a look at the changes I just pushed to the regression notebook: https://github.com/GPflow/GPflow/blob/45bcb9c34a5760f0d7cef8321fc06f9b939f7c5b/notebooks/regression.ipynb

Just working on improving the testing. I'll let you know when it's ready, then you can review.

jameshensman:
Okay @markvdw , I think this is ready. Feel free to add more tests if you think they're needed.

markvdw:
Ok, I've tested `get_samples_dict()`. So as I see it, it returns a dict of arrays with the variable over time. So currently there's no easy way to select a parameter setting for a particular time, and set the model's parameter to this value using `set_parameter_dict`, right?

I just want to suggest an alternative, which is to use a pandas dataframe. This behaves like a numpy array, only columns can have names.

So as an example:
```
import pandas as pd
samp_dict = model.get_samples_dict(...)
for k in samp_dict:
    # Small hack to make `.from_dict` work... Just for illustration, but only works for parameters of single value
    samp_dict[k] = samp_dict[k].T
df = pd.DataFrame.from_dict(samp_dict)
print(df)
print(df.iloc[-1, :])
print(df.iloc[-1, :][`model.kern.lengthscales`])
```

I think this is a really and flexible syntax. The disadvantage is another dependency on pandas. However, pandas is widely used and mature, so I don't see this as a problem. Many projects use pandas DataFrames as their default way of handling large quantities of data.

Thoughts?

markvdw:
Should be done now, if you want to check it out!

markvdw:
Ok I fixed a failed test in `test_hmc.py` and a small bug. I manually changed the setup code in codeship to install pandas. I'm not sure what else needs to be changed to add this as a dependency?

But otherwise, it's ready to be merged!

jameshensman:
Hey @markvdw 

I made just a couple of adjustments: 
 - get_sample_dict became get_samples_df (since it returns a dataframe, not a dict)
 - set_parameter_dict no longer 'pops', resulting in an empty dict

I hope those are okay by you. I also re-ran the notebook to cope with those adjustments. I really like the pandas format for the posterior trace. 

Since we've both contributed to this, we should get @alexggmatthews to have a read before merging.

alexggmatthews:
Hi @markvdw, @jameshensman thank you for your contribution. The code looks good. 

@jameshensman . I have only two questions.

1) Are you satisfied that pandas is sufficiently stable to make it a dependency for GPflow? 
2) Are you satisfied that pandas has functionality that is genuinely necessary and relevant to this case on top of the basic python data structures? 

I've already discussed this with @markvdw in person and know he is happy and I'm anticipating you will be too but I wanted to confirm.

jameshensman:
1) yes, 100%. pandas is now considered 'core', next to numpy/scipy. 
2) 90%. It's not completely necessary, but it is the most elegant solution, farming out a lot of code to the pandas project. Otherwise we'd have to do something more complex in gplfow: using pandas keeps our codebase down.

alexggmatthews:
OK @jameshensman @markvdw looks good merging.

PR #91, opened 2016-06-09
Bayesian GPLVM

alexisboukouvalas:
Initial code for Bayesian GPLVM. 
Very minimal testing thus far - code basically doesn't crash.

Ideas for improvements and tests eagerly sought.
Most of the code written by @jameshensman.

Added notebook/GPLVM.ipynb - optimize() hits numerical problem (matrix non-invertible) and result is the same as initialisation. So things not quite working yet.


alexggmatthews:
@jameshensman @alexisboukouvalas  I don't think the coverage test is working quite right here. The overall test coverage has reduced from master, but because it went down then slightly up on this branch codeship thinks it is fine. I guess we still need to get test coverage of some of the new code.

alexggmatthews:
Hi @alexisboukouvalas @jameshensman do we actually need the changes to  GPflow/sgpr.py ? I can't see what they do. 

What about prediction? I notice this isn't currently implemented. It might be quite fiddly in the case of  a GPLVM I guess. 

@jameshensman maybe we should think about that coverage issue I mentioned above.

alexggmatthews:
Also I notice we have some numerical issues in the notebook? I guess this was part of the motivation for #72 .

alexisboukouvalas:
Thanks @alexggmatthews  for the comments. This definitely needs more work. 
@jameshensman  could you have a look at the basic Bayesian GPLVM code and let me know if you spot anything? I will try to fix it in the next few days but would appreciate the help.

I then plan to add prediction and priors. The latter only affect the KL term and are useful in GPs with noisy inputs scenarios.

SandySpitfire:
@jameshensman, I am Sanket, one of Marc Marc Deisenroth's student at Imperial. Marc suggested me to add gplvm code to GPflow. May I know what is the current state of the gplvm @alexisboukouvalas ?  The upstream/gplvm shows test cases missing, is everything else done (apart from missing test cases) or we need more work on that ?

I need to build dynamical model on top of GPLVM, i.e. use model from Damianou2011 NIPS paper.

alexggmatthews:
@jameshensman @alexisboukouvalas did you have time to reply to @sanket-kamthe ?

jameshensman:
Hi @sanket-kamthe , 

You're right, the GPLVM is pretty much done, but some more tests are needed to make sure the coverage is at the level we'd like. It looks like it works well though, yes. 

As for extensions, well that's a special case of a deep GP. If you're interested in helping implement that, we may have to have a bit of a get-together to discuss directions, as there are others working on related things.

sanket-kamthe:
Hi @jameshensman and @alexggmatthews ,

Thank you for the update. I am very much interested in extending it to the dynamical models.

ilm:
Is there any update on the deep GP implementation ? We were planning to investigate deep GPs more in the near future, including the implementation of one. Nothing concrete yet, we are still reading up on the literature (and, hence, also welcome any pointers ).

Sorry for posting this here, is there any mailing list where these things are discussed ?

jameshensman:
No progress on deep GPs to my knowledge, @ilm, but you're welcome to contribute if you have time. 

@sanket-kamthe  is interested in this direction, but I don't know if he's got anywhere. 

There's no user mailing list at the moment. we should probably do something very 21st century and use gitter or slack or something.

alexisboukouvalas:
Hi @jameshensman 
I have added a number of tests to increase the test coverage in the gplvm branch.

There is still one problem in *test_kernel_expectations.TestPsi2D_GH*, when using composite kernel one component of which requires quadrature, it fails - uncomment line # FAILS to see it.
When using RBF/Linear kernel for which we have analytic expression for the psi statistics, there is no issue with the composite kernels. Any help appreciated.

Status of code and proposed next steps
===
   1. increase numpoints=20 in one_dimensional_psi_stats psi stats for kernel_expectations tests to match up.  This will slow down the code however.
   1. Would like to write tests for psi0 and psi2 calculations.
   1. I will also try to write some correctness tests - for example 
   1. BayesianGPLVM with small fixed variances arriving at similar solution to GPLVM.
   1. Create documentation and example notebooks - should I wait for sphinx branch to be merged?
   1. I would like to add fixed kernel. Can I use coregion kernel?

Let me know what else you think is needed for this PR. Let me know when you are back in Manchester.

markvdw:
If I read the code correctly, the phi statistics do not allow for a full covariance matrix over q(x), right? The GPSSM does need this, and I think it would be good to have a unified codebase. I have a basic implementation of what I think would be the start of a good solution here https://gist.github.com/markvdw/3079a018fef5c89048453b5b5bf0653b. This also includes the additional required expectation of <x Kzx>.

What it lacks that your code has is:
- Support for inactive dimensions
- Combination of Linear and RBF kernels

My tests currently use Monte Carlo, but my biggest struggle is to find a good test for <x Kxz> since MC has too high variance.

I feel like the kernel expectation code in GPflow should reflect the most general case, which should be applicable to both the GPLVM and GPSSM. Additionally, the ability to have any covariance matrix over q(X) is an advantage as well. And finally, I think that the kernel expectations should be housed inside a kernel object, whether it be the definition of e.g. the RBF itself, or a separate kernel object that supports kernel expectations in more limited cases. E.g. GPflow could have a file `ekernels.py`, with all kernels that support kernel expectations.

So in conclusion, I think there's value in combining our two approaches, and that it would be easier if we worked together :)

jameshensman:
Wow, this looks great @alexisboukouvalas ! Great work getting the tests passing etc.

I think the GPLVM part of this PR is ready to go. 

I also think that @markvdw 's solution to the kernel expectations is very neat, and we should adopt it. I also approve of Marks' naming convention! Much better than psi0, psi1, psi2, which is rather cryptic. 

As for non-diagonal covariances, we should support both, and detect which one we want based on the dimension of the matrix, like we do for the conditionals: 3d tensors means we have an array of posdef matrices (or an array of lower triangular matrices I assume, Mark?) and a 2d tensor represents an array of diagonal matrices. 

much of the code inside kernel expectations in this PR could be moved into
1) the base kernel class (with 1D quadrature, we can do any kernel)
2) the rbf and linear classes, for their special cases
3) the additive/product kernel classes. Most of the 'logic' code lives there too, I think. 
all using the structure in Mark's gist.

My proposal is: let's  create a new branch from this one. That way, we have a fully passing set of tests etc. In that branch, create the kernel structures and move the code out of kernel_expectations.py. Since this PR has a full set of passing tests, we (you) can develop against them.

Meanwhile on this branch, I can have a look at fixing that failing test (I think I know what's wrong).

alexisboukouvalas:
thank you both.
@jameshensman also please have a look at the hack I did for the active dimensions. By disabling the slicing when no active dimensions are specified, I am worried it may have some performance implications for the whole of GPflow. Also this change has made some code in kernels.py redundant as it's still checking both cases. 

@markvdw it would be great to work with you on the more general case. I suggest we follow @jameshensman  suggestion and create a new branch starting from this one. I am quite interested in applying these latent variable models on computational biology problems and would like to better understand what we can get from the more general case (q(x) not diagonal).
Note that in the gplvm branch we have already added informative priors in the KL term - this is quite useful when we have some prior information on the approximate location of the latent points. The latter is the case in some biology experiments for example where we are reconstructing time from coarse observations.

alexisboukouvalas:
Closing this as the Bayesian GPLVM will now be part of the #195

PR #92, opened 2016-06-13
adding methods to convert a model to a dict

jameshensman:
this is partially in response to #56 and #48


alexggmatthews:
I'm personally not sure we should merge this one until we have resolved the conversation on #56 as this PR implements a partial solution to the requirements there. I'm open to the opinions of others though.

jameshensman:
Agree, closing this. Opening another PR with fuller solution.

PR #96, opened 2016-06-13
serializing GPflow objects

jameshensman:
This branch is perhaps mis-named, it does two things, both addressing #48 

First, the ability to get all the parameters in a models in a dictionary.

Second, the ability to pickle GPflow objects. 

This is very much up for discussion, and we may want to separate the two features. I think they're both useful though, and both address the same open question.

See also #56 

```python
import numpy as np
import GPflow
import pickle
X, Y = np.random.randn(2, 10, 1)
m = GPflow.sgpr.SGPR(X, Y, GPflow.kernels.Matern32(1), Z=X)
s1 = pickle.dumps(m)
m._compile()
s2 = pickle.dumps(m)
m1 = pickle.loads(s1)
m2 = pickle.loads(s2)
print m.compute_log_likelihood()
print m1.compute_log_likelihood()
print m2.compute_log_likelihood()
```


alexggmatthews:
@jameshensman Looks like progress. There are some tests that will need fixing. Also I think we should wait for @markvdw to feed back on this one too as he has had a significant impact on it.

jameshensman:
Tests now pass, on my laptop at least. Still needed: some tests for pickling and unpickling models, perhaps based on my previous comment.

alexggmatthews:
@jameshensman Could you elaborate on what you mean about needing more tests, please? I wasn't quite sure what you meant about your previous comment.

alexggmatthews:
Just a thought but maybe @markvdw should contribute the remaining code if he wants to? Seems reasonable given that this implements some of his ideas that his name should be on the merge.

jameshensman:
Yep, this is absolutely an implementation of @markvdw 's idea, it was just easier for me to implement because I knew where all the circular dependencies were coming from and where all the tf objects were. 

For future reference, the circular dependencies are mostly around the `_parent` attribute. All of our objects are in a tree structure, in that the lengthscales are children of the kernel, which is a child of the model, etc. But they each have a `_parent` reference to the node above too, so that they can notify when things change (like a prior getting set, causing a recompile to be required). 

To make pickle work, I made `__(s|g)etstate__` methods that threw out the `_parent` references at pickle time, and reinstated them at unpickle time. 

Additionally, some attributes of the model are tensorflow elements (`m._minusF`). These just get removed at pickle time, and the `needs_recompile` flag is set so that they get reinstated through that mechanism.

Also, the `_session` and `_free_vars` attributes are discarded/reinstated by pickling/unpickling.

alexggmatthews:
@jameshensman that's very helpful. Interesting it turned out that between us we knew how to do it already. A good example of team work if ever I saw one :) @markvdw

alexggmatthews:
Looks like this one has stopped working after a merge from master.

alexggmatthews:
It looks like a merge of #100 also fixed the failures mentioned above.

alexggmatthews:
@jameshensman are you sure we need to_dict as well as pickle? I'm mindful of feature creep and the time cost of development.

jameshensman:
Happy for you to strip out the to_dict functionality into a separate PR so
we can discuss it there.

alexggmatthews:
@jameshensman OK I will do. 

@markvdw did you want to be involved with this PR as discussed above? Fine if not but we wanted to give you the option. I'm keen to get pickling in so will push on this afternoon unless I hear otherwise.

markvdw:
Woah, sorry for the silence, let me download and get back to you.

alexggmatthews:
OK I'm holding on :)

As far as I can see if you do want to get involved (other than helping work out how to do it!) we need to:

1) Decide if you are happy with the basic idea for pickling on here.
2) Remove the "to dictionary stuff" from this PR and move it to a separate PR.
3) Add some additional tests of the pickling on this PR.
4) Merge this PR.

I'm happy to do this if you don't want to but conditioned on 1) it is fairly mechanical and would mean that you contribute code to the PR that you already contributed to intellectually.

markvdw:
Ok, I think the idea is great. Thanks @jameshensman for the implementation, I had similar `__(g/s)etstates__` implemented, but I was hesitant about diving into the `_parent` part.

I'll get the remaining bookkeeping done, thanks for holding off. Great teamwork :)

markvdw:
The only thing I'm still worried about is that `id(param_obj)` may change after pickling. I'll see if I can elicit this kind of behaviour in a test. If I can't, I'll merge. Any thoughts?

jameshensman:
Great point, and that's why I think we'll need the to_dict method (although I'm happy to discuss that separately).

From run-to-run, the free_state variable change in order. so after pickling, and unpickling a model, a copy of the free_state is not valid. so this will not work

```python
m = GPflow.gpr.GPR(X, Y, kernel=GPflow.kernels.RBF(1))
m.optimize()
opt_state = m.get_free_state()
m2 = GPflow.gpr.GPR(X, Y, kernel=GPflow.kernels.RBF(1))
m2.set_state(opt_state) # Nope!
```

but a copy of the dictionary _is_ valid in these circumstances:
```python
m = GPflow.gpr.GPR(X, Y, kernel=GPflow.kernels.RBF(1))
m.optimize()
opt_dict = m.to_dict()
m2 = GPflow.gpr.GPR(X, Y, kernel=GPflow.kernels.RBF(1))
m2.from_dict(opt_dict) # Yes!
```
Similarly with pickled/unpickled models.

markvdw:
Ok @jameshensman, the branch `to_dict` is ready to be pull-requested as soon as this is merged!

alexggmatthews:
Great stuff. The tests look to be taking a fair bit longer than usual so we might want to check if there is something slow in the new ones.

markvdw:
Yeah I noticed that. They're super quick on my machine though. Didn't this happen in the past as well? What caused it?

alexggmatthews:
In the past it was because TensorFlow keeps the grad def in memory even after you would expect the lifetime of the corresponding python object to be over, a bit like a memory leak.
@jameshensman remind me how we fixed it? I can't remember exactly off the top of my head.

alexggmatthews:
@markvdw sorry we probably have to fix this before merging I know its annoying. now I remember,

Try calling:

`tf.reset_default_graph()`

at the beginning of setUp in the unit test. You can look at other tests for examples.

markvdw:
@jameshensman How long did the tests take when you ran them? They take 2 mins to run on my machine with `python -m unittest discover`. It's hard to tell which tests it gets stuck on, but it's definitely early on...

jameshensman:
just having a look now...

jameshensman:
Ran 143 tests in 92.989s

jameshensman:
```
nosetests --nologcapture --with-timer .

testing.test_method_equivalence.TestEquivalence.test_all: 15.1593s
testing.test_mean_functions.TestModelsWithMeanFuncs.test_basic_mean_function: 9.6680s
testing.test_methods.TestMethods.test_all: 7.1100s
testing.test_variational.VariationalUnivariateTest.test_build_likelihood: 4.8716s
testing.test_pickle.TestPickleSVGP.test: 4.3229s
testing.test_hmc.SampleGaussianTest.test_rng: 3.5089s
<snip>
```

jameshensman:
To get the coverage back up, move the long_name property to the to_dict branch, it's tested by the code that's been moved over there.

alexggmatthews:
@markvdw @jameshensman Can we try cPickle instead of pickle? The former should be faster where the latter is better for debugging.

alexggmatthews:
@jameshensman @markvdw 

I've done a time of the build. There may be some measurement error due to Codeship speed flucutuation.

Without this PR: 19:40 sec
With this PR: 21min 3 sec

So with the changes this PR costs us 1 minute and 40 seconds.  We should keep an eye on build times in general but I think this is acceptable so I'm going to merge given that the PR has already been held up.

markvdw:
Ok, so the serialising works really well now, but there's still an issue with storing things like a history of old parameter values, that the order in `get_free_state()` changes. The only time when the changing order is ever a problem, is when we want to reload parameter values after recreating a model. And I now think that `to_dict` is actually the perfect way to solve this.

So @jameshensman are you happy if I set up the `to_dict` branch for a merge?

PR #97, opened 2016-06-13
Universal autoflow

jameshensman:
This PR moves the autoflow functionality from the model class to the param class

This was movitated by the desire to compute the kernel matrix without constructing a model.

At the same time, the autoflow mechansim has been simplified: maintainance will hopefully be easier.

### DIFF
--- a/GPflow/kernels.py
+++ b/GPflow/kernels.py
@@ -2,7 +2,7 @@ from functools import reduce
 
 import tensorflow as tf
 import numpy as np
-from .param import Param, Parameterized
+from .param import Param, Parameterized, AutoFlow
 from . import transforms
 
 
@@ -42,6 +42,10 @@ class Kern(Parameterized):
 
     def __mul__(self, other):
         return Prod([self, other])
+
+    @AutoFlow((tf.float64, [None, None]), (tf.float64, [None, None]))
+    def compute_K(self, X, Z):
+        return self.K(X, Z)
 
 
 class Static(Kern):
--- a/GPflow/model.py
+++ b/GPflow/model.py
@@ -1,11 +1,10 @@
 from __future__ import print_function
-from .param import Parameterized
+from .param import Parameterized, AutoFlow
 from scipy.optimize import minimize, OptimizeResult
 import numpy as np
 import tensorflow as tf
 from . import hmc
 import sys
-from functools import wraps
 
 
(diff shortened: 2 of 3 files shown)


jameshensman:
This is an alternative to #63

alexggmatthews:
@jameshensman Looks good to me. Thank you.

ialong:
Hi guys, 
Sorry to bother you again, I've been trying to wrap my head around why the following code does not work:

```
kern = GPflow.ekernels.RBF(Ee, ARD=True)

eval_K = GPflow.param.AutoFlow((tf.float64,))(lambda kernel, input: kernel.K(input))
k = eval_K(kern, z)

eval_exKxz = GPflow.param.AutoFlow((tf.float64,),(tf.float64,),(tf.float64,))(lambda kernel, z, x, xx: kernel.exKxz(z, x, xx))
exkxz = eval_exKxz(kern, z, xmu, xcov)  
```

I'm constructing two AutoFlow instances with the appropriate number of arguments (the second calls a kernel expectation function) and yet the second value ```exkxz``` ends up being the same as the first (```k```). Do I need to start two independent tensorflow sessions?

@markvdw any thoughts?

Alessandro

mrksr:
The problem you are running into is essentially the difference between the wrong and the "right" solution I described in #383. The two autoflowed functions behave the same because you have a collision in the way AutoFlow does caching and therefore `eval_exKxz` just reuses the graph of `eval_K`. The reason is that the caching is done based on the name of the autoflowed function, both of which are lambdas in your case which are called `__main__.<lambda>` (or something similar) and therefore collide.

This would not happen if
1) You use different kernels so the caching domains are different
2) You rename your functions so there is no collision

This should work:
```python
kern = GPflow.ekernels.RBF(Ee, ARD=True)

def _eval_K(kernel, input):
    return kernel.K(input)
eval_K = GPflow.param.AutoFlow((tf.float64,))(_eval_K)
k = eval_K(kern, z)

def _eval_exKxz(kernel, z, x, xx):
    return kernel.exKxz(z, x, xx)
eval_exKxz = GPflow.param.AutoFlow((tf.float64,),(tf.float64,),(tf.float64,))(_eval_exKxz)
exkxz = eval_exKxz(kern, z, xmu, xcov)  
```

ialong:
@mrksr That's great, thank you very much!!

Issue #94, opened 2016-06-13
Fixing a set of parameters e.g in kernel


alexggmatthews:
Unless I've missed a thread we don't currently have the ability to fix a set of parameters all at once.

For instance it would be nice to say

`model.kern.fixed=True`

which would fix all the hyper parameters in the kernel, rather than going through fixing each one in turn which is what we currently have to do.

jameshensman:
That would be a nice feature, and easily implemented.

jameshensman:
Another thing: if we do this for fixed, we should perhaps do it for all transforms. And maybe priors too.

Also, I wonder if we should keep the `fixed` attribute on the kernel? Calling `m.kern.fixed = True` would set the fixed state of the variance and lengthscale, but the kernel actually doesn't have a fixed attribute, so presumably we'd not set one? Would it be odd to do

```python
m.kern.fixed = True
hasattr(m.kern, 'fixed')
>>> False
```

alexggmatthews:
Fixed in #98

PR #98, opened 2016-06-14
adding fixed property to parameterized

jameshensman:
As per Alex's request, #94 

You can now do something like
```python
m = GPflow.gpr.GPR(X, Y, kern=GPflow.kernels.Matern32(1))
m.fixed  # is False
m.kern.lengthscales.fixed = True  # works as before
m.kern.fixed = True
m.kern.lengthscales.fixed  # still True
m.kern.variance.fixed # now True
m.likelihood.fixed # still False
```

### DIFF
--- a/GPflow/param.py
+++ b/GPflow/param.py
@@ -331,6 +331,15 @@ class Parameterized(Parentable):
                   and key is not '_parent']
         return sorted(params, key=id)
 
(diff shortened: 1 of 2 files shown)


alexggmatthews:
Looks good. Thank you.

PR #100, opened 2016-06-14
initializing all tf variables inside AF session

### LINKED ISSUE #99: compute_log_likelihood for SVGP class results in unitialized variable error.
The following code gives an error.

```
import numpy as np
import GPflow

rng = np.random.RandomState(1)
X = rng.randn( 10, 1)
Y = rng.randn( 10, 1 ) 
Z = rng.randn( 3,1 )
model = GPflow.svgp.SVGP( X=X, Y=Y, kern=GPflow.kernels.RBF(1) ,likelihood=GPflow.likelihoods.Gaussian() , Z=Z )
model.compute_log_likelihood()
```

> FailedPreconditionError: Attempting to use uninitialized value tfY

It seems this may be down to recent changes?

jameshensman:
fixes #99

### DIFF
--- a/GPflow/param.py
+++ b/GPflow/param.py
@@ -297,6 +297,7 @@ class AutoFlow:
                 with instance.tf_mode():
                     storage['tf_result'] = tf_method(instance, *storage['tf_args'])
                 storage['session'] = tf.Session()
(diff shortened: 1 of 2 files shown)


alexggmatthews:
@jameshensman Thank you looks good.

Issue #80, opened 2016-06-03
Treating fixed values as placeholder rather than constants


alexisboukouvalas:
This is something that has been bugging me with GPflow, namely how fixed parameters are treated.

if you have a param.fixed = True then during 
_compile-> make_tf_array
it is added in the graph as a tf.constant (see Param:make_tf_array)
However if we want to do somethings like
m.kern.param.fixed=True
m.kern.param = 2
m.optimize() -> this will trigger a compile that adds a constant
if we now do
m.kern.param = 3 this will also trigger a compile during the next optimize (see Param.setattr)
m.optimize()

However tensorflow offers another mechanism that will save some computation, namely placeholders.
As we have the free state, we could have a fixed state.
When _compile gets called it creates tf.Variables for the former and tf.PlaceHolders for the latter.

Then in our above scenario, recompilation would not be needed.
Note that this would apply to transforms and priors as well (for example we want to return the optimization with different limits on the logistic transform).

This will be only important when cost of recompilation is high, that is in complex models with large graphs.
Also note that by avoiding recompilation we do not pollute the default graph space in tensorflow with  redundant graphs.

Perhaps this is not that important at this point but wanted to bring it to your attention for comments.

alexisboukouvalas:
@jameshensman, @alexggmatthews  let me know if I should close. 
My thinking here is why we are not use the tensorflow placeholders in GPflow but perhaps as compile time is not such a big issue, we can forget about it. This is also related to the serialisation work #56

alexisboukouvalas:
Okay I have been thinking about this a bit more, and I will try the following experiment in the near future:
GPflow, construct fixed_state in additiont to free_state. 
Then pass in fixed_state to graph via placeholders (feed_dict).

Compare performance to current approach on tests as well as scripts that update fixed state variables, for e.g. for i in range(1000)
	m.kern.bias = i
	m.predict()

Note this is scenario may be important when doing Bayesian optimisation to learn some of the model parameters and the others are learned by gradient optimisation.
This is exactly the case for the branching model I am developing with @jameshensman 

Let me know if you think this is not a sensible or worthwhile approach.

jameshensman:
I think this is a sensible plan, happy to accept it if the solution turns out to be clean.

jameshensman:
We're working on a fix for this over in #105 with @fujiisoup .

alexisboukouvalas:
Fixed in #105.

PR #108, opened 2016-06-24
Placeholder

jameshensman:
This PR is to merge @fujiisoup 's Placeholder suggestiongs into master. Discussion and edits can take place in this PR.


jameshensman:
Okay @fujiisoup I had a look at the changes and added some of my own. I think this is a good direction, but there are a few things to address before we can merge. 

1. storing the data in _data_dict is going to be annoying if people have previously relied on m.X, m.Y for whatever reason. Suggest using the _tf_mode switch to use the tf representation at the appropriate time. Not sure exactly how to do this: a new class for holding data could be one way, but might be overkill. 

2. Some of the code clashes with the methods in svgp.py which supplies the data in minibatches. We should make sure the two mechanisms are one-and-the-same, and share code where possible. I think this might be an opportunity to make the code clearer and more general. We should discuss with @markvdw .

3. How long does it take to gather the placeholders and the arrays, to put in the feed_dict? If this is quick, no problem. But we are doing this inside the _objective function, so time spent here happens at every step of the optimizer (or sampler). 

4. Obviously the tests must pass, and we'll need to write some new ones that target the functionality that we've added.

jameshensman:
Oh, one more thing. 

5. We've broken the ability to pickle models, I suspect through the data_dict (which can't be pickled). We'll need to remove the data_dict at pickle time, and reinstate it afterwards. (see also point 1.) Perhaps a Data object would be sensible after all.

fujiisoup:
Hi @jameshensman. 
Thank you for your review.

First of all, please let me make sure whether the code you have modified works well in your environment.
In my environment (Ubuntu 14.04 64bit, Python 2.7), the code fails at the 17th code block in [notebook](https://github.com/GPflow/GPflow/blob/Placeholder/notebooks/regression_with_update_data.ipynb).
I guess tf.placeholder object is recreated and becomes different from that in feed_dic at some point related to fix-parameters.
I struggled about this issue today resulting in the ugly code I pushed...

I also thank you for the comments 1~5.
1. 5, I will consider about the solution to satisfy the backward compatibility.

3., I did not consider the gathering time. 
It might be significant for the large problem. I don't think of anything...
The use of tensorflow optimization might be solution?

4, I will see [testing dirs](https://github.com/GPflow/GPflow/tree/Placeholder/testing) after the directions were fixed.

alexggmatthews:
Hi everyone, 

@fujisoup Thanks for your interest in GPflow and your contribution.

One observation from a brief review. Looking at the pull request there seems to be quite a lot of notebook code (the pull request as a whole has 6700 new lines). @jameshensman maybe we want to streamline that a bit or move it into python test code? I'm sure it is possible to fix.

jameshensman:
@alexggmatthews yep we can do that

jameshensman:
@fujiisoup can we have a think about whether we want to be able to change the shape of the data for VGP (and GPMC also)? Perhaps this should just raise an error, since the model has to be recompiled anyway. For the other models it's fine, but I'm not sure we should be re-creating model parameters when the data are changed - this seems like a big change to the model. What do you think? 

Anyone else got any thoughts, @GPflow/contributors ?

jameshensman:
I did some tests to have a look at the gathering time. For a SVGP with a RBF kernel, the time is 0.085ms. For a more complex kernel, it's going to take longer to traverse through the structure and gather the parameters. I tried this:
```python
ks = GPflow.kernels
k = ks.RBF(1) + ks.Linear(1) + ks.Matern32(1) * ks.Linear(1) + ks.Bias(1) + ks.White(1)
m = GPflow.svgp.SVGP(X, Y, k, Z=X[::2].copy(), likelihood=GPflow.likelihoods.StudentT())
```
and the time to gather the feed_dict was 0.207ms. To on something that iterates for 500 iterations, that's an extra 10 seconds, not insignificant.

jameshensman:
> I guess tf.placeholder object is recreated and becomes different from that in feed_dic at some point related to fix-parameters.
I struggled about this issue today..

I found that this was due to the placeholder be recreated (as you suggested) when calling an AutoFlow function (like predict, for example). 

In the PR I've proposed to merge into this one, #110 , I fixed this issue by only creating the placeholder when the Parameter becomes fixed. 

Hope that's helpful.

fujiisoup:
Thank you @jameshensman for much improvement.

I merged that [PR](https://github.com/GPflow/GPflow/pull/110).
(Is it a correct way for PR? Actually I am not yet accustomed a GitHub culture... sorry)

I added some test code in (https://github.com/fujiisoup/GPflow/tree/Placeholder) 
but found it fails when a parameter is fixed and assigned a value to param.value before the first compilation.
After any compilation, it succeeds.
(I tried to commit it to this PR but have no access ...)

Tomorrow, I will review your code more carefully and 
try to find any solutions for the above issue and the below 1)

> The advantages are:
> 1) I like the syntax m.X = X_new, instead of m.update_data(X_new, Y_new).
> 2) The minibatching in SVGP is now much simpler, and incorporated into the same mechanism
> 3) It's easier to add new models that want to do minibatching (Deep GPs, perhaps?)
> 4) We can get at the data values in the same way as parameters through m.X.value
> 5) All the tests pass

> The disadvantages are:
> 1) A new class. I'm wary of class bloat.
> 2) Some new tests are needed, and documentation needs improving before this can be merged onto master (this PR is to merge into Placeholder, for the moment).

jameshensman:
> Is it a correct way for PR?

Yes, that PR was meant for you to merge into this one once you'd had a look and made sure I wasn't doing anything silly. Once we're happy with what we have on *this* PR, we'll get @alexggmatthews and/or @markvdw to have a look and see if they agree that it should be merged on to master. Nice work so far :)

> I added some test code in ...

Now you can add the test code straight into this PR: we've given you permission to write to this branch. To do this

1. make a fresh clone of this repo `git clone git@github.com:GPflow/GPflow.git`
2. get this PR (branch) `git checkout Placeholder`
3. make changes and push `git push origin Placeholder`
4. Your changes will show up here, and the tests and coverage will run automatically. 

> ... but found it fails when a parameter is fixed and assigned a value...

Great! That's exactly the point of writing tests. Together we'll work out a clean solution so that it works for that case, and any other case that you can think of that might break the current implementation.

fujiisoup:
Thanks @jameshensman for the information.

I reviewed your changes and found my previous comment
> but found it fails when a parameter is fixed and assigned a value to param.value before the first compilation.
was wrong.
Your code works more than perfectly but my test_code was buggy.

I also prepared some documentation in codes.
Please review the changes.

I appreciate your help.

jameshensman:
Thanks @fujiisoup, those are some good additions! I'll have a go through
them today.

James.

jameshensman:
> and the time to gather the feed_dict was 0.207ms. So on something that iterates for 500 iterations, that's an extra 10 seconds, not insignificant.

Oh dear, I must have been tired writing that. 0.2ms means that for something that iterates 5000 iterations, it takes an extra 1 second. Not a problem, in my view.

jameshensman:
A few days ago I identified 5 potential issues with this PR. They are now all fixed.

> 1.storing the data in _data_dict is going to be annoying...

I think the DataHolder class is a clean addition, no more data_dict.

> 2. Some of the code clashes with the methods in svgp.py 

I think our solution is perhaps cleaner than the one currently on master: no extra code is needed in the svgp class (the DataHolder class manages the minibatches)

>3. How long does it take to gather the placeholders and the arrays,

 Not a problem in my opinion, see above comment

4. Obviously the tests must pass, and we'll need to write some new ones that target the functionality that we've added.

Done. 

> (5). We've broken the ability to pickle models

Fixed.

There's one thing we need to discuss, which is: what is the desired behaviour when changing the shape of the data?

fujiisoup:
Thank you @jameshensman.

> (3). How long does it take to gather the placeholders and the arrays,
> Not a problem in my opinion, see above comment

I Agree.

> what is the desired behaviour when changing the shape of the data?

I think the current implementation is better than to raise an exception, 
because users may expect models are just reusable.
I think I don't understand an advantage of the another implementation that raises an exception for different shape data.

Another thing.
I'm wondering about the backward compatibility.
In the previous implementation, the data used in the model can be read through
used_x = m.X
but in the current code, it should be like
used_x = m.X.value

I modified to maintain this backward compatibility. See #111.
But the code becomes a bit less beautiful.

jameshensman:
> I think the current implementation is better than to raise an exception
Well, one problem with the current implementation is that all the models
recompile, when actually four of them do not *need* to recompile (GPR,
SVGP, SGPR, SGPMC). In those models we could just set the data, and the
model would work fine.

So my proposal is that for those four models, we just set the data, no
recompile needed. For the two models that can't do that (VGP, GPMC), we
either: 1) recompile. or 2) raise an exception that tells the user to make
a new model. What do you think?

fujiisoup:
Thank you @jameshensman for the information.

I did not understand it well.
I still think that it is more user-friendly to make recompilation for these two models. 
Expert users may understand why these two models need recompilation, while beginners can use GPflow more naturally.

jameshensman:
Okay @alexggmatthews , I think this one is ready to roll. @fujiisoup , thanks so much for your help: do you have any further comments? @alexisboukouvalas might have some comments too, since this covers some of his use-cases, including #80. 

There are a lot of changes. Most notably, a unified mechanism for feeding things to `self._session.run()` that are not parameters. This is called `get_feed_dict`, and it plays well with the current structure. currently it fetches data (through the dataHolder class) and fixed parameters. 

The consequences are:
 -- data can be changed safely! `m.X = X_new`
 -- recompilation avoided if the data changes (where possible)
 -- fixed parameters can be changed (this was a bug, #80)
 -- minibatching is simpler (less code, extendable to other objects easily. I hope @markvdw approves)

I propose incrementing the version to 0.2.0 I've added notes to the RELEASE.md file.

fujiisoup:
Thank you @jameshensman and all for taking care of my request.
I am happy to hear that.

One thing.
A [notebook](https://github.com/GPflow/GPflow/blob/Placeholder/notebooks/regression_with_updated_data.ipynb) I have pushed is not so helpful for beginners.
It might be good to introduce main functionalties in the notebook for users
(I was so helped by these noteebooks).

alexggmatthews:
@jameshensman @fujiisoup OK I will have a look.

alexisboukouvalas:
This is a great improvement! I will try these changes with my use cases (branching code) as @jameshensman  suggests. 
Any tests done on impact on performance? I suspect that in scenarios where fixed values are changed frequently (e.g. data or parameters) this change will improve things significantly.

I also like @fujiisoup suggestion for a beginners notebook where we show all the different ways of updating the parameters/data and what does and does not trigger a recompile.

alexggmatthews:
OK @jameshensman @fujiisoup . Thank you both for your efforts this looks good. @fujiisoup congratulations on becoming a GPflow contributor your name is now in the acknowledgements.

PR #109, opened 2016-06-24
Multiclass

alexggmatthews:
Not to be merged yet.


alexggmatthews:
@jameshensman I've tried to get the likelihood test suite working on the multiclass likelihoods but haven't succeeded so far. I have two problems at the moment:

1) Calls to

` l.predict_mean_and_var(self.F, self.F * 0) `

seem to be returning all nans. 

2) Calls to 

`Likelihood.predict_density(l, self.Fmu, self.Fvar, y)`

Seem to be getting broadcasting errors.

Any ideas? Thanks very much.

jameshensman:
1. This was because of a sqrt function being called on `self.F * 0`. I added come clipping to try to help with this. 

2. Short answer: test not applicable. Long answer: This test was designed to check analytical implementations of the density against numerical ones. For example, in the Gaussian likelihood, we know the formula for the predictive density. The base class has the quadrature code in, so I compared the performance of the base class (`Likelihood.predict_density(l...`) to the analytical code (`l.predict_density(...`). The broadcasting error happens because the base-class assumes that the likelihoods is univariate, i.e. multiple columns of Y and F (and F_var) are treated independently, which is not the case for the multi-class likelihood. My solution is to omit this likelihood from that test.

alexggmatthews:
@jameshensman Did you push changes corresponding to your comment above? For instance in the case of 2) I can't see any changes to the likelihood test. 

Also the tests seem to fail in a very similar way to how they failed before.

jameshensman:
Sorry @alexggmatthews I didn't make any changes to the tests. 

For 1., the call that you mentioned no longer gives nan, but sadly the tests still fail.

alexggmatthews:
@jameshensman OK thanks for quick reply I've removed case 2) . I had a look into case 1)

There do currently seem to be some large differences. In some cases we have a relative difference of more than 10% .

I suggest this is a property maybe of the way we have had to implement it in TensorFlow?

alexggmatthews:
Actually @jameshensman having done some more looking into it I don't think this is entirely a numerical problem.

I removed some of the clipping which improves the numerical stability of some of the tests. I hope this is OK?

The remaining tests look like they are failing from some sort of indexing error in predict_mean_and_var. If I call the test_mean function in the failing case I get:

> In [1]: mu1
> Out[1]: 
> array([[ 0.99900001,  0.001     ],
>        [ 0.001     ,  0.99900001],
>        [ 0.99900001,  0.001     ],
>        [ 0.99900001,  0.001     ],
>        [ 0.001     ,  0.99900001],
>        [ 0.001     ,  0.99900001],
>        [ 0.99900001,  0.001     ],
>        [ 0.99900001,  0.001     ],
>        [ 0.99900001,  0.001     ],
>        [ 0.99900001,  0.001     ]], dtype=float32)
> 
> In [2]: mu2
> Out[2]: 
> array([[ 0.999,  0.001],
>        [ 0.001,  0.999],
>        [ 0.001,  0.001],
>        [ 0.001,  0.001],
>        [ 0.001,  0.001],
>        [ 0.001,  0.999],
>        [ 0.999,  0.001],
>        [ 0.999,  0.001],
>        [ 0.001,  0.001],
>        [ 0.001,  0.001]])

In some cases the output of predict_mean_and_var does not add up to 1! It is a similar case for the other failing tests. I don't think the result can be explained by numerical problems.

alexggmatthews:
@jameshensman  OK so it turned out to be numerical after all :)

alexggmatthews:
OK I'm making some progress now but still working on some more tests.

alexggmatthews:
@jameshensman I think this one is good to go.

@alexisboukouvalas since James and I have both written code for this would you like to act as the reviewer? If you are happy with the content could you please merge it.

jameshensman:
Nice work @alexggmatthews. 

Could you tell us (me) a little about the changes you made in order to make the numerics work, please?

alexggmatthews:
Hi @jameshensman I backtracked a bit so only ended up with one change . There was a division by the standard deviation that we needed to clip in the way you clipped another part. Ive tidied the test and rerun the notebook.

GPflow/likelihoods.py
```diff
@@ -321,3 +327,110 @@ def conditional_mean(self, F):
...
+        oh_on = tf.cast(tf.one_hot(tf.reshape(Y, (-1,)), self.num_classes, 1., 0.), tf.float64)
+        mu_selected = tf.reduce_sum(oh_on * mu, 1)
+        var_selected = tf.reduce_sum(oh_on * var, 1)
+
+        # generate Gauss Hermite grid
+        X = tf.reshape(mu_selected, (-1, 1)) + gh_x * tf.reshape(tf.sqrt(tf.clip_by_value(2. * var_selected, 1e-10, np.inf)), (-1, 1))
+
+        # compute the CDF of the Gaussian between the latent functions and the grid (includeing the selected function)
+        dist = (tf.expand_dims(X, 1) - tf.expand_dims(mu, 2)) / tf.expand_dims(tf.sqrt(tf.clip_by_value(var, 1e-10, np.inf)), 2)
+        cdfs = 0.5 * (1.0 + tf.erf(dist/np.sqrt(2.0)))
```
alexggmatthews: @jameshensman I clipped in the computation of dist.

alexisboukouvalas:
Hi @alexggmatthews, I will happily review and merge this. I will do so this afternoon.

alexisboukouvalas:
@alexggmatthews, @jameshensman  just one thing to check:
In MultiClass(Likelihood) the class comments says:
'We use the softmax and RobustMax inverse-link functions; [..] '
Should this  be 
'' A likelihood that can do multi-way classification.  Currently the only valid choice 
of inverse-link function (invlink) is RobustMax.''

As there is no softmax option per se - just RobustMax.

alexggmatthews:
Hi @alexisboukouvalas . Thanks for looking at this. Yes good spot I took out softmax because we only had a half implementation and I think RobustMax makes more sense algorithmically given that it requires only a one dimensional quadrature. I've made the correction you so suggest so I think we are good to go.

jameshensman:
We should have the softmax too, it's nice if you're doing MCMC. RobustMax
won't MCMC because it's not differentiable. Softmax can wait for a future
PR though.

James.

alexggmatthews:
@jameshensman To clarify I guess you mean full MCMC? We know RobustMax is OK with variational MCMC .

jameshensman:
Yes.

alexggmatthews:
OK thanks.

PR #110, opened 2016-06-25
Data object

jameshensman:
Here is a series of commits that implement an object `DataHolder` that, er, holds the data. 

The advantages are:
1) I like the syntax `m.X = X_new`, instead of `m.update_data(X_new, Y_new)`.
2) The minibatching in SVGP is now much simpler, and incorporated into the same mechanism
3) It's easier to add new models that want to do minibatching (Deep GPs, perhaps?)
4) We can get at the data values in the same way as parameters through `m.X.value`
5) All the tests pass

The disadvantages are:
1) A new class. I'm wary of class bloat.
2) Some new tests are needed, and documentation needs improving before this can be merged onto master (this PR is to merge into Placeholder, for the moment).


jameshensman:
@fujiisoup I'd welcome your thoughts on this approach.

PR #111, opened 2016-06-28
Backward compatibility to access the data

fujiisoup:
To maintain the backward compatibility, it may be preferable that the data in a model can be read through
used_x = m.X
used_y = m.Y
rather than
used_x = m.X.value
used_y = m.Y.value

But the code became a bit ugly.
How do you think?


jameshensman:
If you don't mind, @fujiisoup , I'm going to close this without merging it. 

I agree that backwards-compatibility would be nice, but I think we should prioritize cleaner code. 

We'll increment the version number and let people know about the changes to m.X and m.Y in RELEASE.md. I think the syntax m.X.value is fine as it's the same as all the parameters.

fujiisoup:
OK @jameshensman. 

I agree that clearer code is better.
And as for me this change is not a problem.
Thank you for your review.

PR #112, opened 2016-06-28
`to_dict`

markvdw:
I think @jameshensman was right to add this after all :)

When trying to keep a history of old parameter values, we hit the problem that the parameter ordering is different after unpickling or recreating a model. There's no easy way round this, but it's also a pretty brittle thing to rely on anyway. This branch gives an alternative: just store the dict. I think it shouldn't be much bigger storage wise (I may test this) and it removes any ambiguity.


alexggmatthews:
@jameshensman @markvdw OK that sounds like progress. I will have a look.

alexggmatthews:
OK @jameshensman @markvdw . I've had a look and I think this will in general be a useful addition to the project. I think the code is good.

I have two more general questions:

1) We currently have the ability to get and set using a dict . What are we recommending in terms of actually saving state to a file? Obviously there are a number of ways you could doing it using the dict functionality. Is the intention to leave that to user choice? I personally can see the merits of that actually. 

2) Do we want to provide pickle functionality as well? Aren't they two ways of solving very similar problems? Do we want to support both solutions? Maybe we have to now for backwards compatibility.

PabloLeon:
In case that this is of help - I was just having a look at the GPy page and they discuss storage/ consistency across python versions: 

> As pickle is inconsistent across python versions and heavily dependent on class structure, it behaves inconsistent across versions. Pickling as meant to serialize models within the same environment, and not to store models on disk to be used later on.

jameshensman:
Hi @alexggmatthews 

Thanks @PabloLeon , that's correct. 

We shouldn't really be recommending pickle as a way to store models, for several reasons:

1. Small changes to GPflow will make it impossible to resurrect objects from saved pickle files.
2. Pickle is not secure: who knows what objects are going to be in something we unpickle?
3. pickle is not (very) human readable.

Pickle _is_ needed though, for example to pass a GPflow instance over a network, in order to do some parallel computing -- I know that ipcluster uses pickle to send objects around, for example. And as the GPy guys wrote, it probably okay to pickle something inside a single session. 

I think we should be recommending the dictionary format for storing optimized model results. There's a handy `np.savez` for writing such dictionaries to disk. Also, it's a good way to store a series of model states as @markvdw has suggested. Perhaps in the future we can make the sampler return a list of state-dictionaries? That would be handy.

Longer term, perhaps we should think about how to write the models to disk too? What I would recommend at the moment is to separate-out the code that constructs a model, and store that, but it's not completely satisfactory. Writing and maintaining our own serializer sounds like a pain, but perhaps this library could help? http://camel.readthedocs.io/en/latest/camel.html. See also https://eev.ee/blog/2015/10/15/dont-use-pickle-use-camel/.

alexggmatthews:
OK thanks @jameshensman @markvdw and @PabloLeon . I'm satisfied that the PR is useful in its current form  and that we should maintain support for Pickle for the reasons given.  I'm therefore merging.

PR #113, opened 2016-06-28
Kern slicing

jameshensman:
Kernels with active dimensions now select columns of data using
tensorflow's gather, rather that a packed series of scalar indexes.

old: `tf.pack([X[:,i] for in in active_dims])`
new: `tf.transpose(tf.gather(tf.transpose(X), active_dims))`

the latter is preferred because the graph is much smaller, especially
for a large number of active dimensions.

In addition, the kernel tests have been tidied and extended to cover
more of kernels.py.

### DIFF
--- a/GPflow/kernels.py
+++ b/GPflow/kernels.py
@@ -22,7 +22,7 @@ class Kern(Parameterized):
         if active_dims is None:
             self.active_dims = slice(input_dim)
         else:
-            self.active_dims = active_dims
+            self.active_dims = tf.constant(np.array(active_dims, dtype=np.int32), tf.int32)
 
     def _slice(self, X, X2):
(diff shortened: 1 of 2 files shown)


markvdw:
Happy to merge this, just one question I've been wondering for a while. What is the intended use case for this functionality? Why don't we just slice from the dataset before assigning the datasets to the model? Is there a use case where we want to change the dimensions that are used on the fly?

jameshensman:
So that different kernels (or parts of a kernel) can work on different parts of the data. For example, 

```python
ks = GPflow.kernels
k = ks.RBF(1, active_dims=[0]) + ks.RBF(1, active_dims=[1])
```

markvdw:
Ah great!

PR #118, opened 2016-07-06
Convenient way to get and restore all the data in a model

fujiisoup:
Hi all.

Related to [PR#108](https://github.com/GPflow/GPflow/pull/108).

I added a few methods to acquire and update all the data in a model,
`get_data_dict()` and `set_data_dict()`
which are similar to `get_parameter_dict()` and `set_parameter_dict(),` respectively.

I'm using GPflow as a part of a larger graph, reusing a single model with different sets of data, and try to infer some global hyperparameters which are common for all the data sets.
(I also use DataHolder in kernel and likelihood.)
So, I was looking for a simple way to replace the data in a model. 

I don't think these methods are always useful,
but if these method does not bother the code cleanness, please consider to marge this PR.

p.s. I also fixed minor bugs in the overloaded version of _compile()


fujiisoup:
Oh.

I found the codes I pushed do not work for the case when some DataHolders are defined in the child classes other than GPflow.model class.
Test was insufficient.

I close this PR.
The present best way to replace data maybe simply `model.X=X` or `model.likelihood.Z=Z`.
Sorry for all the fuss.

PR #125, opened 2016-07-09
Bug-fix in VGP related to mean_function

fujiisoup:
Hi all.

I guess there is a bug in VGP related to mean_function.
The following in VGP.build_likelihood() may be wrong.
`f_mean = tf.matmul(K, self.q_alpha) + self.mean_function(self.X)`
`KL = 0.5 * (A_logdet + trAi - self.num_data * self.num_latent+ tf.reduce_sum(f_mean*self.q_alpha))`
Especially -> `f_mean*self.q_alpha`

I think this term should be like
`(f_mean - mean_function) K^{-1} (f_mean - mean_function)`
but the current implementation is equivalent to
`(f_mean) K^{-1} (f_mean - mean_function)`
This makes `VGP.optimize()` diverge.
 
I also modified test_method_equivalence.py to include mean_function.

### DIFF
--- a/GPflow/vgp.py
+++ b/GPflow/vgp.py
@@ -46,7 +46,7 @@ class VGP(GPModel):
         self.q_lambda = Param(np.ones((self.num_data, self.num_latent)),
                               transforms.positive)
 
-    def _compile(self):
+    def _compile(self, optimizer=None):
         """
         Before calling the standard compile function, check to see if the size
         of the data has changed and add variational parameters appropriately.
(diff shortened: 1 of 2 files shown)


jameshensman:
Hi @fujiisoup ,

Great spot! You're completely right, the VGP was incorrect. It worked with a zero mean-function, which is why we hadn't spotted  the bug. 

I note that you've modified the tests which show how models are equivalent to include a mean function: this covers the case of the bug. Good job. 

To merge this into GPflow, it needs to be on a branch in this repo. So I'm going to merge this onto GPflow/VGPmod, and open another PR from there. This is only because the tests and coverage will only run from a branch in our own repo.

jameshensman:
closing in place of #126

fujiisoup:
Thank you @jameshensman for your review.
I'm happy to hear the PR was merged :)

jameshensman:
Thank you @fujiisoup for your contribution!

PR #129, opened 2016-07-11
Data tensor

alexggmatthews:
Not to be merged yet.

@jameshensman I've begun the changes we discussed in a new PR. 

At the moment I've left the base class DataHolder in param.py because the dependencies would have been very fiddly to pull it out into a separate file. I've made it into an abstract base class using the core library abc.

How would you suggest I do the pickle code for the new TensorData class? 

Obviously I will need to change the model classes to allow us to pass in a TensorData class instead if we want to.

This is a good chance for me to learn this bit of the code better. Does it look right so far apart from that?


jameshensman:
> How would you suggest I do the pickle code for the new TensorData class? 
That's going to be tricky, since pickling must remove the _tf_array, which isn't under our control. Suggest you pop it in __getstate__, and leave a note on __setstate__ that if needs to be reinstated. 

Perhaps put in a sensible error message if the user tries to acces _tf_array and we don't have one?

code structure looks good, though it might be worth pulling out all the dataholder classes into a single file eventually.

alexggmatthews:
@jameshensman Re setting the tensor post unpickling- I guess this new class would therefore need a set_data method that takes a tensor rather than a numpy array. It would need to then call recompile otherwise the objective function would not reference the correct tensor. Could you explain a bit more about the _kill_autoflow lines that I have commented below please?

GPflow/data_holders.py
```diff
@@ -0,0 +1,105 @@
...
+            self._array[...] = array  # just accept the new values
+        else:
+            if self.on_shape_change == 'raise':
+                raise ValueError("The shape of this data must not change. \
+                                  (perhaps make the model again from scratch?)")
+            elif self.on_shape_change == 'recompile':
+                self._array = array.copy()
+                self.highest_parent._needs_recompile = True
+                if hasattr(self.highest_parent, '_kill_autoflow'):
+                    self.highest_parent._kill_autoflow()
```
alexggmatthews: @ Could you explain a bit more about what is going on here, please?

jameshensman: @alexggmatthews , this function gets called when the user assigns a new array to the object. If the new array is the same shape as the old array, nothing happens, we just accept the new data. If the shape has changed, that might be bad news for the model, which may have parameters whose shape matches that of the data (VGP, GPMC). In this case, we can either 1) raise an exception,  2) recompile the model or 3) do nothing

1) isnt used as far as I remember
2) is used for VGP and GPMC, which have special compile functions to cope
3) is used for SGPMC, SVGP, GPR, SGPR, which can all cope with the shape of the data being changed on-the-fly. 

edit: perhaps you were referring to the kill_autoflow switch? This removes any autoflow objects from the model, perhaps a little conservative. It's slightly messy, we should probably do that every time the _needs recompile flaga is set, automatically at the Parameterizsed or model level.

jameshensman: @alexggmatthews I tried to tidy up this behaviour in #130

alexggmatthews:
@jameshensman thanks for the PR I've merged it. We'll also decide how we want to call models now that we have different data containers. We could make it that X and Y could be numpy arrays, DictData objects or TensorData objects and triage them in the constructor. What do you think?

jameshensman:
Hi @alexggmatthews , 

Yes, triaging the data in the constructor is a good plan. At the moment, there's a little bit of triage in the `model.GPmodel` constructor. 

I have tried to keep the `model.Model` class much more general, so that it doesn't have to have any particular data associated with it. The constructor of the `GPmodel`, though, is fair game.

alexggmatthews:
I'm handing this PR over to @jameshensman since I'm now going to ground to write my thesis.

jameshensman:
@markvdw I'm pretty happy with this, Alex's work is nice. Do you want to have a quick look before we merge it?

Note that I've added a version increment to 0.2.1

alexggmatthews:
@awav is it likely that this PR will be superseded by your work?

alexggmatthews:
It is also rather old now.

jameshensman:
I agree that this is superseded by @awavs work. I’ll close this: any objections, please reopen.

Issue #132, opened 2016-07-14
A pre-computed kernel


alexggmatthews:
In my previous work I've encountered the case where you want to pre-compute a gram matrix (which has fixed hypers) and load it into some GP software without implementing all the code you used in that software. The solution I came to was to store the gram matrix and make the data inputs integer indeces into that matrix. then K(i,j) = K_{i,j} . Do we think this is of general interest?  Is the integer index trick the best way to do it?

jameshensman:
In GPy, there's a 'Fixed' kernel, which just returns a designated matrix (times a scalar parameter). 

The problem is, what to do at predict time? Presumably you just want to ignore this kernel when predicting? 

I think GPy's fixed kernel is confusing for users.

alexggmatthews:
I guess you would upload those elements of the matrix too :) . Consider a case I've come across where the data inputs are _actually_ a large set of graphs. We may not want to have to deal with that sort of complex data structure in GPflow. I can always implement it myself but its just a question of whether we think it is core and more of use to others than a hindrance.

jameshensman:
Ah. Now I have had a think about this sort of thing. What I would suggest, is that the 'X' in GPflow becomes some integer values, and the kernel stores a big lookup table. so the kernel would look like

```python
class LookupKern(GPflow.kernels.Kern):
    def __init__(self, K):
        self._K = K
    def K(X, X2):
            return self._K[X, X2]

big_K = np.load(...)
k = LookUpKern(big_K)

X = np.random.randint(0, big_K.shape[0], (10, 1))
Y = np.foo
```

Then, if you wanted to combine this kernel with other kernels, you'd use separate columns of X, and the active_dims feature of the kernels.

In this case, I think it's core, and a definite upgrade on GPy.kern.Fixed.

alexisboukouvalas:
Just some feedback which may be of use - models such as the PANAMA model use fixed kernels so this enhancement would be useful when trying to implement such models.
[PANAMA](http://journals.plos.org/ploscompbiol/article?id=10.1371/journal.pcbi.1002330) is a GPLVM model that includes known confounding factors

alexggmatthews:
OK. I'm now thesis writing so I will open this one up if someone fancies it.

jameshensman:
Note: This is already pretty-much implemented in the Coregion kernel. The difference in that in the Coregion kernel the positive-definite matrix can be optimized (sampled), whereas this proposal is for a fixed matrix.

Issue #79, opened 2016-06-02
Multi-task/outputs GPs


alexisboukouvalas:
Hi team,

I am considering implementing in GPflow the following paper as a start to multi-output GPs:
https://papers.nips.cc/paper/3189-multi-task-gaussian-process-prediction.pdf

It's a kronecker product kernel where the between-output correlation is estimated using maximum likelihood.
I had implemented this a while back in Matlab and with autodifferentiation it should be a breeze to implement in GPflow.

Do you think this is worth doing or there are much better approaches out there we should implement instead? I like the kronecker approach because it's so simple and interpretable.

jameshensman:
Hi Alexis,

That paper is exactly a linear model of coregionalization. Essentially it's
a product kernel between k(x, x') and what GPy calls a 'Coregionalization'
kernel. If the tasks are all aligned (i.e. for each input we observe all
tasks), the the kronecker structure can be used for inference.

So, you'd need two separate features to make this work.

1) a coregionalization kernel.
2) a kronecker-structured model object

I would love to see an implementation of 1.

For 2, the structure should be able to handle any separable kernel. I think
it would be great to have a think about what's possible for both GPR and
for MCMC. Here are some useful references:

http://papers.nips.cc/paper/5089-it-is-all-in-the-noise-efficient-multi-task-gaussian-process-inference-with-structured-residuals.pdf
http://jmlr.org/proceedings/papers/v37/flaxman15.pdf

For completeness, I'd also want a variational version, calling @c0g

James.

alexisboukouvalas:
Thanks James,  I will go over the papers and we can discuss the best code design.

PR #138, opened 2016-07-24
Coregionalization

fujiisoup:
Not to be merged yet.

I am preparing coregionalized models in GPflow [#79](https://github.com/GPflow/GPflow/issues/79).

Though the implementation is still in progress, I mostly fixed my direction.
Now I have the following concerns, and
I would be great if anyone could review the codes (and test-codes) and give me a comment.

I added the codes in a sub-directory 'coregionalized' and tried to keep (most of) the other parts unchanged.
(I guess coregionalization is not a main stream of GPflow)

However, I would like to make some changes for my implementation.

+ AutoFlow with DataHolder (in [param.py](https://github.com/GPflow/GPflow/blob/Coregionalization/GPflow/param.py))

The current implementation of `AutoFlow` uses `tf.placeholder`s and simply pass it to the wrapped method.
But if it handles `DataHolder` (and its child classes) rather than placeholder, 
it will be very convenient for my implementation.

I want to change the usage of AutoFlow as follows

`@AutoFlow((tf.float64, [None, None]))` -> `@AutoFlow(DataDict)`

and want to make a child class that inherits from AutoFlow.

It is the largest change I want to make in the core part of GPflow.

Please let me know if this change is acceptable.

+ DataHolderList that handles list of data_holders are added in [data_holders.py](https://github.com/GPflow/GPflow/blob/Coregionalization/GPflow/data_holders.py)

I added a class that has similar functionalities to those of ParamList.
I guess it will be convenient for other purposes, so I added it in data_holders.py rather than in coregionalized folder.

+ Lookup parameters and kernels.

Related to #132.
I added lookup kernel in constructing Kronecker's multiple in [coregionalized/lookup_kernels](https://github.com/GPflow/GPflow/blob/Coregionalization/GPflow/coregionalized/lookup_kernels.py) and [coregionalized/lookup_params](https://github.com/GPflow/GPflow/blob/Coregionalization/GPflow/coregionalized/lookup_params.py).

If this functionalities can be used for another purpose, it may be better to move these modules to the parent directory.


jameshensman:
Hi @fujiisoup 

Great! I love coregionalization models. 

There's a lot to look at here. Perhaps, to make it easier to review, we could split out the features into several pull requests. 

1) DataholderList
this seems like a sensible and useful feature. I look forward to including it. 

2) AutoFlow with DataHolder
This sounds like it might be a big change! If it's useful, and doesn't impact other users, we can include it. Otherwise, perhaps I can help you work out another solution. 

3) Lookup parameters and kernels.
I'm not too clear on what this does, but I think it's strongly related to the coregionalization?

4) the coregionalization methods themselves. 

I have some notes on Coregionalization models for you.  In GPy, we can up with a really neat solution for Coregionalization models. The trick contains two things:
1) A coregion kernel, which specifies the kernel between outputs (see #139 )
2) Use an extra column of the *inputs* to the kernel to specify the outputs. 

This takes a moment to get used to, let me give you an example. If we have three output functions, observed at three points 1.2, 2.3, 3.4, the X matrix looks like

```python
np.array([
[1.2, 0],
[2.3, 0],
[3.4, 0],
[1.2, 1],
[2.3, 1],
[3.4, 1],
[1.2, 2],
[2.3, 2],
[3.4, 2]])
```

And the Y observations are stacked up in a single vector. You can specify a Linear model of Coregionalization with this kernel:
```python
k = GPflow.kernels.Matern32(1, active_dims=[0]) * GPflow.kernels.Coregion(1, output_dim=3, rank=2, active_dims=[1])
```
Now you can use this kernel and that data inside *any* GPflow method, doing variational approximations or MCMC or whatever you need. 

Now, an additional thing that happens with coregionalization models is a Kronecker-structured matrix. This gives some significant computational benefits, but requires you to see all the outputs every time. Kronecker methods are not restricted to Coregionalized methods, though, so we should have a separate implementation for general Kronecker things. Note that the method I described above does not require kronecker structure.

fujiisoup:
Hi @jameshensman 

Thank you for the quick review and information.
I would be great if you guys help me for the development.

I guess it is better to note here what I want to realize in this implementation.
I want to construct a very flexible framework for the coregionalization, not only with the coregion kernel but also (for examples)

+ Different mean functions among output tasks

+ Different variations (or even different likelihood forms) among the output task

+ Different scales and offset of the input channel
(e.g. the tasks (X1, Y1) and (X2, Y2) are similar if some transform to the input 
`X1 -> c * X1 + delta` is introduced, with unknown coefficients. Note, this future is not implemented yet and left for the future extension.)


> 4) the coregionalization methods themselves.

I first thought of the extra column in the Data vector, which realize the coregion kernel.
But I guess it will be difficult to use different liekelihoods for each output task, in the current implementation.
(The use of different mean function may be possible if active_dims feature is introduced in mean_functions)

Therefore, I came up with an idea to define a new DataHolder object that bundles the value of X and output task label (`coregionalized.labeld_data.LabeledData`), like

`X.data  = [1.2, 2.3, 3.4, 1.2, 2.3, 3.4, 1.2, 2.3, 3.4]`
`X.label = [  0,   0,   0,   1,   1,   1,   2,   2,   2]`

and also for the the value of Y,

`Y.data  = [2.1, 3.2, 4.3, 2.1, 3.2, 4.3, 2.1, 3.2, 4.3]`
`Y.label = [  0,   0,   0,   1,   1,   1,   2,   2,   2]`

then X and Y are passed to kernel, mean_function, and likelihood.
In the kernel, with X.data the kernel values are calculated and with X.label the coregion kernel values are looked up from the parameter table.
(The `LookUpKernel` is just indroduced to handle this operation, and `coregionalized.kernels.Kern` is almost the same to `kernels.Coregion` kernel)

In the mean_function and likelihood, the input and data vector is split into multiple vectors according to the label, e.g.
`X` -> `[[1.2, 2.3, 3.4], [1.2, 2.3, 3.4], [1.2, 2.3, 3.4]]`
then passed to respective mean_functions and likelihoods and then the results are recombined (and sorted to the original order).
(`coregionalized.mean_functions.MeanFunction`) and (`coregionalized.likelihoods.Likelihood`)


> 2) AutoFlow with DataHolder 

But this idea gives a trouble in the prediction phase, 
because we have to pass data and label pair to the kernel, mean_function and likelihood, but AutoFlow only accepts a simple tf.placeholder.

That's why I want to change the AutoFlow functionalities.

(`DataHolderList` is also introduced (mainly) to customize AutoFlow)


> This gives some significant computational benefits

I might not understand the theory very well, but is it the case if X is common in multiple tasks?
If it is, I might consider to implement both the cases.


> Perhaps, to make it easier to review, we could split out the features into several pull requests.

OK. I will prepare small pull requests.

But because my implementation is largely based on the above idea, 
any advice may avoid unnecessary changes to the codes.

Actually I saw the GPy usage and found that it uses different likelihood and mean_functions among tasks.
I'm wondering if there is a super-smart solution also for GPflow.

alexisboukouvalas:
This I believe also addresses issue #79. This is a great addition and I can help with merging or as an extra set of eyes for reviewing.

fujiisoup:
@alexisboukouvalas  Thank you for your comment.
I would be very great if I could extend GPflow's functionalities.

Additions to the above my comment.
I found my current implementation is not very computationally effective because it execute the sort-operation every calculation time of kernel, mean function, and likelihood.
It may be better to make a sort only one time at the `data_holder.set_data()` method before the calculation.

My `coregionalized.kernels.Linear` is also less effective than #139.

But anyway, I'd like to fix the coding direction at first and then will make necessary improvement.

fujiisoup:
An alternative to achieve coregionalization with much less impact to the GPflow core might be 

to make all the `Likelihood` methods include `X` in their arguments,
like
`Likelihood.logp(self, F, Y)` -> `Likelihood.logp(self, F, X, Y)`
but do nothing for the normal GP models.
In the likelihood classes for  the coregionalized model (should be implemented separately), the task index contained in the extra dimension of `X` (as suggested in the above comment) is used to adopt different variance parameters among tasks.

Which ones are preferred for the GPflow concept?
If you guys have any idea, please let me know.

jameshensman:
Hi @fujiisoup 

> with much less impact to the GPflow core 

I like that idea !

In GPy, we did this by adding "Y_metadata" as an argument to the likelihoods. It was a bit annoying because it touches all the likelihood codes. 

An alternative would be to put the index in an extra column of Y *as well* as an extra column of X. Then a special likelihood could strip out the extra column and pass the correct parts of the remaining columns to the relevant likelihoods. I think I could prototype that pretty quickly.

fujiisoup:
Hi @jameshensman 

Thank you for your comment.
> to put the index in an extra column of Y as well as an extra column of X
sounds great.

It will make almost no impact on the other part!! Cool!

fujiisoup:
Hi @jameshensman 

Thank you very much for your help!
It is exactly what I want. 
I was so surprised that such a simple code realizes the coregionalization...

I will commit some appropriate test codes for this modification.

PR #139, opened 2016-07-24
adding coregion kernel

jameshensman:


markvdw:
Hi James, I see that I've been asked to review this, but I don't really understand the code or how it would be used. My understanding of coregionalisation (but I've never used it) in GPs is that you define some GP and multiply it by a matrix `W` to give correlated outputs. The covariance between outputs at the same inputs is then `B = W*W'`. So does the covariance between the vectorised outputs become `K kron B`? How would this kernel be used in that way? I don't see any reference to another kernel for the underlying GP?

jameshensman:
Hi Mark, 

Ah, sorry. I guess we need some better documentation. Your intuition is right though. 

Yes, in a coregionalization model (so-called intrinsic coregionalization) you have K GPs that you transform by a DxK matrix W to give D correlated GPs. Then you can add extra independent GPs with the same covariance to each. 

Now, the full covariance matrix is K \kron B, with B = WW^T + diag(kappa), which is where this kernel comes in. 

So if you want to build such a kernel, you just have to multiply a Coregion kernel by another. So:

```python
k_coreg = GPflow.kernels.Coregion(1, rank=1, num_outputs=2, active_dims=[0])
k_space = GPflow.kernels.Matern32(1, active_dims=[1])
k = k_coreg * k_space
```

and your data matrix will have two columns, one specifying the output
```
X = [[0, 0.1],
        [0, 0.2],
        [0, 0.3],
        [1, 0.1],
        [1, 0.2],
        [1, 0.3]]
```
There is a notebook (to be included in the doc) which demonstrates. 

You can now build multiple-output GPs using any of the GPflow GP methods. There's no accounting for the kronecker structure, that will have to be a separate model class (which will of course use this kernel).

markvdw:
Ah right, I get it. So the kronecker structure is implicit, and the full covariance is just stored as a large `output_dim x N` matrix, and also inverted as such.

I'll merge this.

PR #140, opened 2016-07-25
Making the documentation more official

jameshensman:
The notebooks that serve as our documentation are not the best format for many users, especially those new to python/github/etc. 

Sphinx is a neat tool for building documentation that has a built-in support for notebooks! with a little wrangling, I managed to get sphinx to build the content of our notebooks. That means we can have the content in html, or pdf or whatever form we need. 

to build, you need sphinx and some plugins. 

```
pip install sphinx
pip install numpydoc
pip install nbsphinx
```
numpydoc lets us include docstrings in the numpy format (which I think is best, and we're all used to since we're such heavy numpy users) and nbsphinx is a plugin that automatically parses ipynb files. 

Then, in the doc directory, either ```make html``` or ```make latex```

The documentation isn't finished yet, but this PR makes it much easier to hack on the docs since there's a build system in place and lovely pdfs can be made.


jameshensman:
Okay, this is up, here: http://gpflow.readthedocs.io/en/latest/

jameshensman:
Okay @GPflow/contributors this documentation is not finished, but I think it's ready to be released into the wild. 

In order to get the documentation to build on readthedocs.org, I had to make the whole project build on there, since building the docs imports GPflow necessarily. This means that there are some tweaks to the install, like ensuring that we include the compiled binaries. 

In general I think the docstringing of the code has been really good, but in place I think we'll want to tweak it for compilation into sphinx. In particular, there are some docstrings on constructors that should perhaps be on their classes. 

The notebooks have been moved to GPflow/doc/source/notebooks. This is to facilitate including them in the sphinx build. 

You can check out the current results at gpflow.readthedocs.io

alexggmatthews:
@jameshensman Zoubin was very impressed with the webpage. Thank you for this PR.

PR #143, opened 2016-07-27
DataHolderList

fujiisoup:
Hi.

Related #138 
I added DataHolderList to data_holders.py, which handles multiple DataHolders as a list.
append, __iter__, __getitem__ methods are prepared just like a normal list.
ScalarData is a simple DataHolder to handle a scalar, not np.array.

This PR is a part of #138, and meant for the easier review.


fujiisoup:
Closed because code direction was changed.

PR #144, opened 2016-07-27
Auto data flow

fujiisoup:
Also related to #138 

This is PR for the review of my AutoFlow modification.

I want to use a DataHolder in AutoFlow rather than a simple placeholder, and make it possible to change the behavior of the AutoFlow in its child class.

The main purpose of this modification is to make it possible to pass an object other than tf.placeholder to the predictive method such as `model.predict_f()`.
In my coregionalized model implementation, I want to pass an object that include a coordinate vector, task index as well as some methods to sort, split and merge, in a child class of AutoFlow ([LabeledAutoFlow](https://github.com/GPflow/GPflow/blob/32590b7ea7963651de493ea5c5cd6e147ed9764c/GPflow/coregionalized/param.py) in #138).

I would be very great if anyone review this codes and comment if this change is acceptable or not.
If not, I will consider alternatives.

### DIFF
--- a/GPflow/kernels.py
+++ b/GPflow/kernels.py
@@ -3,6 +3,7 @@ from functools import reduce
 import tensorflow as tf
 import numpy as np
 from .param import Param, Parameterized, AutoFlow
+from .data_holders import DictData
 from . import transforms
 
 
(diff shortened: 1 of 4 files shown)


fujiisoup:
Closed because code direction was changed in #138 .

PR #145, opened 2016-07-27
Multi likelihood

jameshensman:
A likelihood for @fujiisoup, so that we can make models with different likelihoods for different bits of the data. 

I learned a lot about tensorflow's dynamic_partition and dynamic_stitch, very useful for splitting up data by indices. 

There are no tests yet, and I suspect this may actually break tests, but it's working in principle. See the new notebook, coreg_demo.


fujiisoup:
Hi @jameshensman 

I really appreciate this addition.

I added some tests for the coregionalized model.
Everything works fine, except for predict_y (see the commented-out part in test_coregion.py).

It seems to fail because session.run() does not handle a tuple of lists of tf.variables.

Do you have any idea to hack this?

jameshensman:
Oh, great work with the tests @fujiisoup !

Yes, we need to decide what to do when we predict with such a likelihood. 

For predict_density, it should be clear: the user should pass in a Y with the corresponding values in the last column. 

For predict_mean_and var, we have some options:
  - return the mean and variance of the first likelihood (nope)
  - return the mean and variance of a specified likelihood. nicer, but I don't think there's a clear way to specify. I don't really want extra arguments into predict_y, because it breaks all the similarity with other methods. 
 - return for all the likelihoods, by horizontal concatenation. I think this is the neatest option, but it will have to be well documented.

jameshensman:
@fujiisoup I pushed a commit that returns the concatenated predictions for the likelihoods. Could you write a test for that code, please? Then I think this code can be reviewed for merge.

fujiisoup:
@jameshensman  Thank you for your addition.

I added the test codes.
I thought of another option, to add inheriting GPmodels for this purpose that overload `predict_y()`, but it may decrease the expandability of the codes.

I also think the option
> return for all the likelihoods, by horizontal concatenation.
is nicest.

fujiisoup:
@jameshensman 

I have some questions.

+ Can I add `MultiMeanFunction` that is a mean_function version of `MultiLikelihood` ?
The constructor will be like `mean = MultiMeanFunction(input_dim, active_dim, list_of_means)`.

+ How do you think if `MultiLikelihood` takes `input_dim` and `active_dim` as arguments of constructor? 
Does it make the usage complex?
I guess it will make it similar to `kernels.coregion`.

jameshensman:
Hi @fujiisoup , 

1. Yes, good plan. In a separate pull request, please. You should talk to @PabloLeon , who has some experience of dealing with the mean functions. 

2. The other likelihoods don't take input_dim or active dim at the moment. What would these arguments do?

fujiisoup:
Thank you @jameshensman.

1. I will make another PR for multi-mean and talk to @PabloLeon .

2. I feel it is unnatural that `switched_likelihood` assumes the label is in the last dimension of `Y` but `kernels.coregion` does not...  The introduction of the additional argument may just increase the code complexity as you pointed out.

Do anyone have any idea about the test fail by codeship?
I seem not to enable to access the build detail by codeship.

jameshensman:
@alexggmatthews (assigned)

This PR adds a new likelihood "switched likelihood" which allows different datapoints to be conditioned differently on the GP. This work by adding an extra column to the `Y` matrix, which contains integers which index the likelihood to be used. 

A SwitchedMeanFunction class works similarly. 

Tests are provided for both classes, and in writing the tests I found a bug in VGP.py, where multiple concurrent Y columns failed. This is easily fixed and is now covered by the new tests. 

It looks lik @fujiisoup has been using this successfully for a while on his own fork -- any comments, Keisuke?

fujiisoup:
@jameshensman 

Thank you for completing this PR.
I was tied up with another project and I could not use MultiLikelihood very often (sorry).

I found the notebook in this PR (`coreg_demo`) fails in the second optimization.
This may not be the issue for this PR, but I guess it is due to the recent update 
in GPflow particularly around default_graph (sorry, I didn't follow them well).

alexggmatthews:
@jameshensman I'm having a look at this.

As i understand we use integers to index the different likelihoods. I'm concerned this might be a bit brittle. What happens if we add another likelihood for instance? If feels a bit like the software engineering considerations in C++ or another low level language around enums in that sense.

alexggmatthews:
Scratch that last. I missed the constructor list defining the ordering. That is much nicer.

GPflow/vgp.py
```diff
@@ -105,15 +105,15 @@ def build_likelihood(self):
             tf.expand_dims(tf.transpose(self.q_lambda), 2) * K
         L = tf.cholesky(A)
         Li = tf.matrix_triangular_solve(L, I)
-        tmp = Li / tf.transpose(self.q_lambda)
+        tmp = Li / tf.expand_dims(tf.transpose(self.q_lambda), 1)
```
alexggmatthews: @jameshensman Why do we need the expand_dims here, please?

PR #146, opened 2016-07-27
adding a simple Uniform prior

jameshensman:


jameshensman:
@markvdw could you have a look at this?

markvdw:
Should `self.log_height = np.log(upper - lower)` not be `self.log_height = -np.log(upper - lower)`? Since we need to normalise the distribution? Now the height is larger if the range is larger...?

jameshensman:
yes! great catch. fixed.

Issue #49, opened 2016-04-18
`q_sqrt` over-parametrisation


markvdw:
I'm quite keen to get rid of the `q_sqrt` overparameterisation in `SVGP`. Currently a full square matrix is stored, even though only the bottom half is used. This has caused me some annoyances when optimising large amounts of inducing points. It slows down BFGS unnecessarily and makes the sizes of my parameter history files twice as large as they need to be.

I'm willing to put in some work to fix this, any tips on how to proceed? I could contribute to your current Tensorflow branch, or alternatively, the main Tensorflow branch. I think this is tangentially mentioned in https://github.com/GPflow/GPflow/issues/10.

jameshensman:
Hi Mark,

I'd like that too. It's actually not as easy as I'd like though: you need a
wrapper for

(d|s|c)tpttf

and

(d|s|c) trttf

if that looks alien to you, the first letter is double|single|complex type,
and tpttf converts a 'packed' triangular matrix into a dense matrix (with
the zeros in ) and trttf converts back again.

I'm not sure if those are available in Eigen, but they're definitely
available in BLAS. The learning curve for getting a new tensorflow op up
and running seems steep to me, but Alex will give you some pointers I'm
sure.

alexggmatthews:
See also this issue on the main tf repo
https://github.com/tensorflow/tensorflow/issues/1825

markvdw:
I've started a bespoke TensorFlow op in #150. I'd appreciate some pointers.

markvdw:
Also, why is it that `q_sqrt` stores the matrices in the outer dimensions (i.e. the most "left"), rather than the inner dimensions? E.g., to access the first q_sqrt matrix, `q_sqrt[:, :, 0]` is used, right?

Assuming a row-first memory layout, storing the matrices in the inner dimension should be faster, right? Since otherwise the strides between consecutive elements become large.

jameshensman:
> why is it that q_sqrt stores the matrices in the outer dimensions

Well, I guess my (perhaps flawed) reasoning was that the data-matrix Y
usually appears in columns, so we should keep that Fortran-style ordering
for these too.

When we make predictions of the variance,

mu, var = m.predict_f(Xnew)

we want the means and variances of the different outputs in columns, right?
var.shape should match mu.shape.

when we do predict_full_cov(Xnew)

where should we put the extra dims? mu is N x D, should var be N x N x D or
D x N x N?

[...]

PR #150, opened 2016-07-29
Fix covariance overparameterisation

markvdw:
NB: This is nowhere near ready to commit yet. I just want some comments (from @alexggmatthews in particular).

This is a start to work to fix https://github.com/GPflow/GPflow/issues/49. This needs a new TensorFlow op to turn a vector into a triangular matrix. On the plus side, it's possible to do this without having to maintain a special fork of TensorFlow. With the TensorFlow binaries installed, custom ops can be compiled in place and then imported as a standard Python module.

The op is in `/GPflow/tfops/vec_to_tri.cc`, and needs to be compiled with the command:
Linux:
`g++ -std=c++11 -shared zero_out.cc -o zero_out.so -fPIC -I $(python -c 'import tensorflow as tf; print(tf.sysconfig.get_include())')`
MacOS:
`g++ -undefined dynamic_lookup -std=c++11 -shared zero_out.cc -o zero_out.so -fPIC -I $(python -c 'import tensorflow as tf; print(tf.sysconfig.get_include())')`

There's a small example script calling the custom op in `/op.py`.

I've basically just adapted the TensorFlow custom op example, and need some pointers on how to proceed. @alexggmatthews, what resources did you use to find out the style used in TensorFlow and how to manipulate the Tensor objects etc? Right now I'm modifying individual elements in a very roundabout way, by first flattening them.

Next steps are to properly template the ops to ensure compatibility with more data types, adding gradients, tests etc...


jameshensman:
This is really super work Mark, I'm looking forward to having this in GPflow. 

It would be great to get the code compiled by `setup.py`, so that users don't have to do anything manually to get the thing working. I can help with this if you need.

markvdw:
Ok, some new developments. I've got `SVGP` working with flattened `q_sqrt`. To do this, I had to modify `conditional()`, by adding another case to the if statement. In `kullback_leiblers.py` I modified the KL functions to assume packed `q_sqrt`s, since `SVGP` is the only place they're used, as far as I could see.

Questions / comments now:
- Should the kullback leiblers be changed to handle both cases of `q_sqrt`, as in `conditional()`? This shouldn't be hard to do. I've used the convention that if a flattened q_sqrt is passed, it's stored as an 1xNxD matrix.
- How to modify `setup.py` to do the compilation of `matpackops.so`? I could work my way through the docs, but if anyone (@jameshensman?) has this experience on-hand, that may help.
- Adding tests for the new case of `conditional()`
- I've also been thinking that this whole process may be best thought of as a transform. So we access `q_sqrt` as a matrix, but in the free state it's a packed vector. This seems like the best way to deal with it, actually. This way it's not necessary to have the extra cases in the kullback leiblers etc.

I'd appreciate input!

jameshensman:
> I've also been thinking that this whole process may be best thought of as a transform

Oh, that's a nice idea. I really like that, it would make everything so simple. The correct matrices would display when you did print(m), and the correct values would be present in m.q_sqrt.value. Really nice. It would also mean no (or few) changes to the conditionals/Kullback-Leiblers.

You may have noticed that there's a big TODO marker in param.py that says "what if the transform changes the size?" (line 202). You'll need a simple mechanism for working out how many free-parameters you'll need. I suggest:

1) in transforms.Transform, define
```python
...
    def free_state_size(self, output_shape):
        return np.prod(output_shape)
```

2) in param.py, near line 202, replace
```python
x_free = free_array[:self.size]
<snip>
return self.size
```

with
```python
size = self.transform.free_state_size(self.shape)
x_free = free_array[:size]
<snip>
return size
```

and then in your shiny new `transforms.lower_triangular` transform, define free_state_size in the appropriate way ((N^2 -N)/2?)

does that make sense?

markvdw:
Makes sense, I'll get on it tomorrow morning. Any pointers on `setup.py`?

jameshensman:
> Any pointers on setup.py?

I just had a peruse of the docs, it's not as straightforward as I had hoped. 

If you write a vanilla python extension, then setup.py is really helpful. In this case, I think the best bet might be a 'DIY' solution, which manually executes the compiler commands using something like 
```python
import os
string = "gcc -I /usr/lib/python tfops/mat2vec.cc -o tfops/mat2vec.o"
os.system(string)
```
or similar. Sorry not to be more helpful :(

markvdw:
It doesn't look like codeship currently runs `setup.py`? I could manually add the compilation instruction to the startup code, but wouldn't it be better to rely on `setup.py` to do everything? This would function as an additional test that everything installs correctly on a clean machine.

jameshensman:
Yep, I guess codeship should really run python setup.py install.
James.

markvdw:
Ok, I think it's pretty much there!

- The conversion is now done by a transform.
- The fix for having transforms of different sizes was exactly as @jameshensman said, only `set_state()` had to be touched as well.
- `setup.py` now compiles the C++ code and this is done by codeship as well.

All this required a few fixes to some quick things I did earlier (like not batching `tri_to_vec`). This is all fixed now.

markvdw:
I'm a little confused why codecov is complaining about those few lines, I specifically wrote a test for them...?

jameshensman:
let me take a quick look.

jameshensman:
oh dear, it just segfaults on my macbook.

Following a rabbit hole of compiler options, etc...

markvdw:
So it compiles and runs fine on my Linux machine. I've had a "just segfaulting" issue with Theano a while back as well, and it was an issue with temporary files that were compiled against out-of-date libraries.

jameshensman:
So this works just great on my macbook, now that I've updated everything. I had an old version of g++ installed, which was interfering with the build. On a clean homebrew install of gcc and family, it works great.

One final thing to look at is why some bits of the code don't appear to be covered, even though you wrote a test for them @markvdw . I'll look into that now.

GPflow/transforms.py
```diff
@@ -170,4 +174,82 @@ def __setstate__(self, d):
...
+        """
+        N = int((y.size / self.num_matrices)**0.5)
+        y = np.reshape(y, (N, N, self.num_matrices))
+        return y[np.tril_indices(len(y), 0)].T.flatten()
+
+    def tf_forward(self, x):
+        return tf.transpose(tfh.vec_to_tri(tf.reshape(x, (self.num_matrices, -1))), [1, 2, 0])
+
+    def tf_log_jacobian(self, x):
+        return tf.zeros((1,), tf.float64) - np.inf
```
jameshensman: Why is the log Jacobian inf?
I think the mapping is just a reshape, so the jacobian is 1, so the log jacobian is 0.

markvdw:
I was thinking about this too. But I think it's not really a reshape, since we're throwing away parameters. So technically the jacobian won't be square, so there's no determinant.

Alternatively, we could consider a non-square determinant. But in this case, there are certain vales in the triangular matrix representation that the vector representation can not change. This will add zero rows in the jacobian, making the determinant zero.

This was my logic.

jameshensman:
I'm not sure we're throwing parameters away. I think those zeros in the top-right on the matrix are not parameters, they're just placeholders for emptiness. They'll never be not-zero. 

I think the log jacobian is 0 still!

GPflow/tf_hacks.py
```diff
@@ -19,8 +19,28 @@
...
+vec_to_tri = _custom_op_module.vec_to_tri
+tri_to_vec = _custom_op_module.tri_to_vec
+
+
+@tf.python.framework.ops.RegisterGradient("VecToTri")
+def _vec_to_tri_grad(op, grad):
+    return [tri_to_vec(grad)]
+
+
+# TODO: Finish registering the shape. Was unsure how to handle incomplete shape information in the input.
```
jameshensman: Is this done now? Can I help?

markvdw:
I can definitely see your point, and I think putting the log jacobian to zero may be the way to go. If we any distribution over the full matrix representation will always only have density over the lower triangular components, then we're good and the jacobian is 1. I guess this would be the use case anyway.

I was thinking about the case where we want to transform a distribution over the full matrix to the compressed representation. If there's a distribution over all elements of the matrix, then the compressed representation can't "get to" all the mass of the original matrix.

jameshensman:
Correct, but it can get to all of the mass of matrices in the posdef cone.

markvdw:
It passes! It looks like the explicit call to the function, instead of passing the function to `assertRaises` ensures that the coverage is taken into account, right?

jameshensman:
Yeah, I'm not sure what was wrong, but this seems to work okay. I think the problem might have been that you had two conditions, and the first one failed both times. Not certain though. 

I added some RegisterShape code, could you have a quick look? If you approve that, I think we can merge this one.

jameshensman:
Eek, test failed. Improved my shape registration code and tried again...

markvdw:
Shape registration looks correct by eyeballing it. I couldn't see a test for it though, but I guess it must get run by tensorflow under the hood somewhere?

jameshensman:
Yeah, it gets run under the hood, every time you make the op, it now checks the shape.

Something is up with codeship, it's not reporting, but the tests pass and the coverage is up (according to codecov.io, coveralls is garbage), so I'll merge this as soon as codeship comes back up.

I added your Copyright to the cpp files, and added some simple docstrings.

This is some really super work, Mark. It sets up how to add extra ops to GPflow, and solves a long-standing tricky issue.

markvdw:
Glad you like it. It's definitely nice to be able to add ops on demand now. I'm keen to get some input from the tensorflow guys, since I'm sure there's a lot that could be improved with style etc, but it will do for now.

PR #151, opened 2016-07-29
make kernels with active_dims pickleable

### LINKED ISSUE #149: pickle and active_dims
Hi All,

Thanks for implementing the getters and setters for pickle, makes saving experiments much easier! However I noticed a small bug, if you have active_dims set in the kernel it stops pickle from working since the active dimensions are stored as tf.constants.

example code to reproduce bug:

```python
X = np.random.randn(100).reshape((10,10))
y = np.random.randn(10)

k = GPflow.kernels.RBF(len(X[0]), active_dims=[0])
m = GPflow.gpr.GPR(X, y[:, np.newaxis], k)
```

Best,
Matt

jameshensman:
fixes #149

### DIFF
--- a/GPflow/kernels.py
+++ b/GPflow/kernels.py
@@ -1,3 +1,4 @@
+from __future__ import print_function
 from functools import reduce
 
 import tensorflow as tf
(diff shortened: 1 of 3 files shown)


markvdw:
Reviewed and merged.

Issue #147, opened 2016-07-28
Optimisation hacks


markvdw:
To me it seems like the optimisation hacks (i.e. the renaming of optimiser options for BFGS) are doing more harm than good. From what I can see from the scipy documentation ([1], [2], [3]), all the options are the same for all optimisers, it's GPflow that has inconsistent names :)

Is this something that should be fixed?

[1] http://docs.scipy.org/doc/scipy/reference/optimize.minimize-cg.html#optimize-minimize-cg
[2] http://docs.scipy.org/doc/scipy/reference/optimize.minimize-bfgs.html#optimize-minimize-bfgs
[3] http://docs.scipy.org/doc/scipy/reference/optimize.minimize-lbfgsb.html#optimize-minimize-lbfgsb

jameshensman:
Yes! 100% should be fixed. 

I'm not sure why I did that now. I guess I might have been working around an inconsisteny bug in previous versions of scipy?

markvdw:
Possibly? I'll submit a PR.

PR #153, opened 2016-08-04
Fixed optimiser options to be consistent with scipy

markvdw:
Small updates to code, keeping it up-to-date with scipy. Fixes https://github.com/GPflow/GPflow/issues/147.

I know it's slightly annoying to have changes in code interface, but it fixes inconsistencies between optimisers, and I think we should adhere to scipy's conventions.


jameshensman:
Really nice. 

Suggestion: put a VisibleDeprecation warning in to help users switch. So when people pass in the old arguments ('max_iters' and 'display'), they get automatically converted to 'maxiter' 'and 'disp', but with a warning. Then in the next+1 version, we can remove the warning.

Since this changes the api, suggest also a minor version bump in _version.py

markvdw:
Fixed.

markvdw:
@jameshensman, I'm not exactly sure about how the coverage setup works, and before I start a string of commits to figure out the correct command, do you know how to exclude a block of code from the code coverage? I don't want to write a test for the deprecation warning, and the `#pragma nocover` that I saw elsewhere in the code doesn't seem to be doing the trick.

jameshensman:
hi Mark, 

the `#pragma: no cover` *does* work, but not when I typed `#pragma: no-cover`, which is a bug.

in param.py, I correctly typed `#pragma: no cover` at line 762: this doesn't get covered. But at line s 255 and 396, I incorrectly added the hyphen. 

You can find other things that we excuse from coverage in `GPflow/.coveragerc`

PR #154, opened 2016-08-07
add Apache license headers (fix #135)

### LINKED ISSUE #135: Add Apache Licence header to each file
GPflow is released under the Apache licence. The header of each file should reflect this with the standard form.

davharris:
I've added headers to everything in the `GPflow` directory.  The headers now say `Copyright 2016 $AUTHORS`, where `$AUTHORS` is pulled from the git blame file.

For example, `GPflow/vgp.py` credits `James Hensman, Valentine Svensson, alexggmatthews, fujiisoup`. I didn't worry about adding grammatically correct conjunctions and comma placement, or about translating git names to legal names, although that should be straightforward if desired.

Let me know if you'd like any tweaks to the headers or if you'd like them added to other files as well. Hope this is useful.

### DIFF
--- a/GPflow/__init__.py
+++ b/GPflow/__init__.py
@@ -1,3 +1,17 @@
+# Copyright 2016 James Hensman
+# 
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+# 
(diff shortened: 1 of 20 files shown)


davharris:
Ack.  Pushed the wrong version.  Will push the correct version in a moment...

PR #159, opened 2016-08-10
use batch ops in KLs and conditionals

jameshensman:
This PR attempts to remove python-side loops form GPflow. 

The problem with python-side loops is that they 'unroll' the graph, which makes for potentially very large graphs. tensorflow has recently (0.9) provided some really nice ops for doing things in batches, including

 - tf.batch_matmul
 - tf.batch_matrix_triangular_solve
 - tf.batch_matrix_diag_part

which let us remove looping code from GPflow. 

The main reason for loops is where there are multiple columns of the data matrix Y. Following the GPy convention, these are treated as independent outputs. All of our loops so far have been to accommodate this. Everything still works as before, but the loops are handled by tensorflow's batching ops.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -55,3 +55,12 @@ docs/_build/
 
 # PyBuilder
 target/
+
+# Emacs backups
+*~
+
+# Pycharm IDE directory
+.idea
+
+# IPython Notebooks
+*/.ipynb_checkpoints
--- a/GPflow/__init__.py
+++ b/GPflow/__init__.py
@@ -1,3 +1,18 @@
+# Copyright 2016 alexggmatthews, James Hensman
+# 
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+# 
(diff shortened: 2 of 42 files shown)


markvdw:
Great, I've been wondering about this for a while!

One question, is it better to do `tf.matmul(A, A, transpose_a=True)` rather than just transposing `A`?

jameshensman:
I guess it is, because tf will be able to reason with the arguments a bit better. So if the matmul op sees that it has the same input twice, and 'transpose_a' is set, it can use a more efficient algorithm underneath. If you pass in (A, tranpose(A)), it has to work harder to infer how to simplify the graph.

jameshensman:
PS. it's very annoying that in matmul the transpose bools are called `transpose_a|b` and in batch_matmul they are called `adj_x|y`. This is almost as bad as `columns|numcolumns|num.columns` in R, which is my pet R peeve.

jameshensman:
Okay, this is ready to roll @markvdw , if you want to have a read though?

Note that coverage is decreasing because I'm removing (simplifying) lines of code, rather than adding new, untested code.

PR #165, opened 2016-08-15
Fixed custom op test failing

### LINKED ISSUE #163: tests fail with tensorflow 10.0rc0
One test fails with the new tensorflow pre-release, it's related to the custom-op test:
```python
ImportError: No module named 'tensorflow.python.kernel_tests'
```

In addition, there sees to be an issue with one of the notebooks ( #161 ), I'm looking into this.

markvdw:
Fixes #163. It looks like a module changed name in the new TensorFlow version or something. The documentation hasn't been updated yet either.

I don't think this ever broke anything in the regular code, since it's just regarding the gradient checks. However, this change may break the test in TensorFlow v0.9, though I didn't downgrade again to check this. Is this acceptable? Given the speed of development, I think it's reasonable to expect GPflow users to keep their TensorFlow version updated.


alexggmatthews:
@jameshensman what do you think of @markvdw 's question on backwards compatibility? I'm personally happy to break backwards compatibility with TensorFlow but we should probably change the version and setup.py .

jameshensman:
Yep I have no problem requiring the latest and greatest tensorflow release. 

We'll need to change

- setup.py (to require 0.10.0)
- RELEASE.md
- _version.py
codeship

jameshensman:
perhaps tag a new release on github also? I just tagged 0.2.1.

This should make it easier for people building on our stuff to upgrade at their own pace.

alexggmatthews:
@jameshensman did you want the new version to be 0.3.0 or 0.2.2?

jameshensman:
Perhaps, given the tri2vec ops, we should call this 0.3.0

alexggmatthews:
@jameshensman version number incremented. It looks like someone has already changed the tf version on codeship.

markvdw:
I've already updated the codeship build rules, so I reckon this is ready to go once the tests pass.

jameshensman:
Perhaps I could get my 'sphinx' branch in before 0.3.0? Minor changes only, but significant;y improved documentation. It also has a commit on there to improve setup.py, so that includes @markvdw 's binary block in an install.

markvdw:
@jameshensman: What does the adding the `.so` as an install improve?

Related: One thing that's bothering me, is that I think that if TensorFlow is updated, `setup.py` needs to be run again to recompile the extra ops. Is there a way to do this automatically? Not a huge issue, but it may confuse someone that GPflow stops working after an update of TensorFlow!

jameshensman:
If you do `python setup.py install` right now on a clean install, it will compile the `.so` files, but won't copy them over with the python source, i.e. install will fail. My fixes will make sure python bundles the `.so`s with the `.py`s. 

Yep, you're going to have to run `python setup.py develop` after updating tensorflow. I thin kthis is pretty normal for a project tracking a fast-paced project.

alexggmatthews:
@jameshensman I think the sphinx branch now needs this one as it is failing. To break the circular dependency I suggest we merge this one with the necessary two lines from setup.py that you allude to. I've committed them just now. 

I personally think that it might be best for sphinx to wait for the next version now. If not this branch will have to wait for me to review the sphinx branch. I'm not particularly fussed about waiting except we currently have #163 which is a bug that needs fixing.

PR #167, opened 2016-08-18
adding extra args to model.sample to pass through to hmc

jameshensman:


alexggmatthews:
Looks good thank you.

PR #169, opened 2016-08-18
Multiclass slicing

jameshensman:
Alex, 

This PR adds a tensorflow op and its gradient. CPU only for now, I'm afraid I hope it's still useful. 

The op is called "remove_row_elements" (and "remove row_elements_grad"). The op takes a matrix (float) and a vector (int) and returns a matrix(float) and a vector(float). 

Perhaps this is easiest explained by the equivalent python. 

``` python
def remove_row_elements(X, index):
    Y = np.empty((X.shape[0], X.shape[1]-1))
    Z = np.empty(X.shape[0])
    for i, (ind, x) in enumerate(zip(index, X)):
        Y[i] = np.delete(x, ind)
        Z[i] = x[ind]
```

I hope this makes sense in the context of the quadrature needed in the multi-class likelihood. It removes the need to use the 'blanking-out' method (via one_hot) that we were using before. I think this should increase the numerical stability, it will certainly make the graph smaller. 

It may be the case that to make a really stable multiclass-quadrature method, you'll need this op and also to re-adjust the clipping/squashing that we were doing (I have left these).


alexggmatthews:
@jameshensman thanks for the PR . As you suggest we need to consider the effect of this on GPU performance. One option would be to include a "switch" to use it only in certain circumstances but I worry this could get messy. I suggest we hold off merging whilst we discuss #170 and #171

jameshensman:
Hi Alex, 

Yep, I agree that there could be an issue with GPU performance. All our custom ops are just reshapes at the moment, so will very quick I'm sure, but there could be some GPU/CPU switching overhead. We should profile it, though, before investing time in writing GPU ops!

alexggmatthews:
@jameshensman could you say a bit more about why you think the original op could be unstable, please? Inefficient I can see...

jameshensman:
Only a suspicion, but here goes.

When we use the one-hot method, we compute some quantities that are related
to parameters of interest that we then discard. We discard them by
multiplying by zero. In the backprop computation, that multiplication gets
differentiated, and should of course result in zero gradient. But I worry
that we've expanded the graph unnecessarily, and in parts of the graph
reduction, some operations may change order: it's not really clear.

The main evidence to support this hunch is that our cython code for this
computation works great, (and works in the way that this PR enables), but
the multiply-by-one-hot method seems to less stable.

 It'd be good to compare the two in a systematic way, any ideas?

jameshensman:
closing: inactivity and the desire to have a 100% python package.

Issue #116, opened 2016-07-01
Tensorboard support


alexisboukouvalas:
1. Implement extensive scoping in the code. This will make the job of looking at the computational graph in tensorboard much easier. For example each kernel could define its own scope. We would need to agree on a systematic way of how to add scoping to the code. 

2. Logging statements. To debug issues I have extensively been using the print commmand, for example to print the first 10 entries of a  tensor and its shape:
```lik = tf.Print(lik,[pointToUpdate, tf.shape(lik), lik],message='point - lik=',name='likeddebug',summarize=10) ``` In my code, I add Print nodes under a debug flag. Perhaps something similar would be useful for GPflow.

3. Use the summary statistics to log information. I have been playing with histogram_summary but tensorflow offers storing scalars and images as well. To keep a log of the value of a tensor simply do:
```tf.histogram_summary('probvector', tf.squeeze(probVector))```
Then collect all the summaries and write to file
```
tf.merge_all_summaries()
tf.train.SummaryWriter("logs",  self._session.graph)
```

alexggmatthews:
@alexisboukouvalas just had another look at this one. We now have #183 merged which deals with 1). Do you think we want 2) and 3) to be in the code all the time?

alexisboukouvalas:
Thanks @alexggmatthews , I  like the new scoping support.
It is worth thinking about more support for debugging including the work you've done on identifying performance problems. 
However the most important change was to include scoping to make the tensorboard graphs readable.
So I am happy to close this.

alexggmatthews:
Ok thanks @alexisboukouvalas , closing now.

icouckuy:
> However I did not get anything particularly nice and useful out of this and would welcome ideas for better support.
At least tracking the optimization progress seems useful to me (likelihood and hyperparameters). I was also thinking about keeping the distributions and histograms for the MCMC methods (like Model.sample()).

Of course adding compute speed and memory metadata is very useful for debugging and profiling (slows the code down a lot), though generating chrome trace files might be better for that.

I'll add a pull request for my branch.

PR #174, opened 2016-08-22
Tracer and wrapped session for GPflow

alexggmatthews:
Not to be merged yet.

This PR currently allows one to switch on a profiling mode that stores a tracer file for calls to tf.session.run . It can be set to store a tracer file for each such call or only the last one. 

To do this there are two key concepts:

1) TensorFlow itself has the ability to do this either for both CPU and GPU processes. They key commands are 

```
run_options = tf.RunOptions(trace_level=tf.RunOptions.FULL_TRACE)
run_metadata = tf.RunMetadata()
sess.run(res, options=run_options, run_metadata=run_metadata)
tl = timeline.Timeline(run_metadata.step_stats)
ctf = tl.generate_chrome_trace_format()
with open('timeline.json', 'w') as f:
   f.write(ctf)

```

2) TensorFlow (oddly) does not have the ability to do this every time with some constant run_options. Therefore I have wrapped all calls to session in GPflow. We can then use either a standard tf.Session or a special new class that has the same interface as the normal session but is running and storing the traces each time. I have done it this way to minimize the impact on the rest of the code. 

I can imagine we might want other interfaces for the profiler which is why I'm doing the PR a bit early. 

I still need to write unit tests and update the existing ones. 

An example script is provided in notebooks/example_profiler.py


alexggmatthews:
@jameshensman @markvdw  it would be good to decide what we want to do with the interface on this one so we can get it in. I suggest that we do something a bit different from what we have currently on this PR.

How's about having three config flags, one for training profiling, one for test profiling and one for the output directory stub. The code would automatically output a file for each call to sess.run in each case so we end up with a directory containing integer labelled files e.g output_directory/profiling/training/timeline_1.json . This directory get big quickly so we would have to warn the user when the software is run in this way. 

Thoughts?

markvdw:
Hmm, if we are to have automatic speed testing, we need to be able to switch profiling on and off from within the code, right? The config currently doesn't allow this.

Otherwise, I think this would be a nice way to switch profiling on and off. It would make it possible to profile code without any changes in the code.

jameshensman:
Why would we need to switch it on in the code? We could just have a local
gpflowrc that turns the profiling on?

markvdw:
Well, I was thinking that for a testing script, it would be nice to have something like:

```
config.profiling.doit = True
for benchmark in benchmarks:
    config.profiling.output = benchmark.output_dir
    benchmark.run()
```

And it would be nice if that could be run out of the box, directly after a clone. What ideas did you have for the workflow?

jameshensman:
@alexggmatthews , what's the status of this PR? We now have the ability to set things in the gpflowrc file to specify if we want the json files to be dumped. I would love to have this ability for one of my projects.

alexggmatthews:
@jameshensman I agree it would be good to get tracing in and I'm happy to finish this. It has been a while so I'm reloading this PR in to my mental cache! 

The config file helps. If I remember rightly, I don't think we fully pinned down what we wanted from the interface. Do we want it to produce a .json file for every training tf.run call? This could get large pretty quickly. What about test time? What do we want to do with the file names of the json files?

jameshensman:
I think we should just dump the last run. I don't see that dumping multiple runs is very helpful. 

Here's my preferred option for the filenames:
1) prescribe a file prefix in gpflowrc
2) add a string giving the name of the function being called to the string, plus '.json'

the name of the function should be accessible to AutoFlow via __name__. For the main function, use 'objective'. So we might end up with 

```
my_prefix_objective.json
my_prefix_predict_f.json
my_prefix_predict_y.json
```
, for example.

icouckuy:
Is there a plan to include tensorboard support. I see there is an option in gpflowrc but it is not used in the code. I added some simple code myself to track the likelihood evolution for example, useful for long training times like mnist. Maybe it already exist in some branch (there are too many branches imo)?

Of course adding metadata to the graph (compute speed, memory) should be optional, but I think tracking some summary statistics is cheap enough.

jameshensman:
Hi @icouckuy 

As far as I know, Alex's plan is only to extract the json file that describes the runtime of the graph. A separate commit for tensorboard support would be welcome, but without much experience of tensorboard I don't know what would be needed.  Perhaps @alexisboukouvalas can comment?

alexisboukouvalas:
Similar to what @icouckuy has described I did write some code to track summary scalar statistics and display them in tensor board. However I did not get anything particularly nice and useful out of this and would welcome ideas for better support. I am happy to help with this, @icouckuy perhaps you can make a PR from your changes to get the ball rolling?

alexggmatthews:
tensorboard was discussed in #116 so perhaps best to switch discussion over there. Feel free to reopen.

markvdw:
Hi Alex, I'm interested in reviving this. Have you had any thoughts in the meantime?

markvdw:
I fixed a few merge conflicts. @alexggmatthews, what in particular were you not happy about, regarding this code? I think with a bit of cleaning it can go it.

Regarding the code interface that @jameshensman suggested, this could be implemented as follows:
- Depart from the abstract base class idea
- `getSession()` accepts all the arguments to `tf.Session()` plus some extra arguments like the function name.
- `getSession()` actually constructs the object, rather than returning the class. If we're profiling, the extra arguments for function name, output file etc, get passed to the constructor, otherwise a normal `tf.Session()` is created.

This requires `SessionFactory` to be aware of the object it will be returning, which the abstract base class setup avoids. I don't think this is a huge problem though, since we only have two Session-like classes.

markvdw:
Ok, here's my suggestion for some edits to finalise this. It's pretty much the same idea, only with the above suggestions added. All profiling can be switched on and off through `settings`, either by modifying `gpflowrc` or by `GPflow.settings.profiler.dump_timeline = True`.

GPflow/gpflowrc
```diff
@@ -15,3 +15,6 @@ ekern_quadrature = warn
 [profiling]
 dump_timeline = False
 dump_tensorboard = False
+output_file_name = timeline
```
alexggmatthews: @markvdw Would this filename only be for the profiler? Maybe profiler_file_name etc would be less ambiguous.

GPflow/model.py
```diff
@@ -147,6 +152,7 @@ def _compile(self, optimizer=None):
         sys.stdout.flush()
 
         self._feed_dict_keys = self.get_feed_dict_keys()
+
```
alexggmatthews: Not a big deal but try to avoid random whitespace. Can make it hard to see changes. No need to revert.

doc/source/notebooks/simple_regression.py
```diff
@@ -38,12 +38,13 @@ def getSamples( m ):
     samples = m.sample(100, epsilon = 0.1)
     return samples
 
-def runExperiments(plotting=True,outputGraphs=False):
+def runExperiments(sampling=True,outputGraphs=False):
```
alexggmatthews: @markvdw These changes look unrelated?

alexggmatthews: Not to worry I see you're using it for the demo.

alexggmatthews:
@markvdw I've added some comments and obviously we'll need unit tests. Generally looks fine to me. Given that we've both worked on it we'll probably need @jameshensman to review it as well.

GPflow/gpflowrc
markvdw: Is this not implied by it being under the profiling section? The property is accessed using `GPflow.settings.profiling.output_file_name`.

doc/source/notebooks/simple_regression.py
markvdw: Not sure how this crept in actually, I think this is the result of a merge rather than a conscious decision on my part?

markvdw:
Great. For unit tests, it's hard to check whether the profiling is actually being done. The best I can come up with is to check whether the correct files are being created.

GPflow/session.py
```diff
@@ -0,0 +1,77 @@
+import os
+import warnings
+import tensorflow as tf
+from tensorflow.python.client import timeline
+from ._settings import settings
+
+
+class TracerSession(tf.Session):
+    def __init__(self, *args, **kwargs):
```
jameshensman: would this not be better as 

```python
def __init__(self, output_file_name, output_directory, each_time, **kwargs)
```

GPflow/session.py
```diff
@@ -0,0 +1,77 @@
...
+        self.outputFileName = kwargs.pop("output_file_name")
```
jameshensman: snake_case please.

jameshensman:
I like this design, seems sensible and unobtrusive to me. I like that by default it just uses a standard session. 

I agree that it's hard to think of thorough tests. I suggest running a TracerSession, making sure that the file is written to disk, making sure that the file is not empty. Then making sure that running model without a tracer session does not do those things (i.e. making sure that the default still works).

GPflow/session.py
markvdw: This won't allow `*args` to be passed to `tf.Session` I believe. Not a huge problem.

jameshensman:
Correct, but we are already mandating some kwargs, so feels tidier this
way.

markvdw:
Right, I think this can be merged now. I cheated by adding 3 lines of `# pragma: no cover` on a check for whether there are any conflicting run options. We don't actually do this anywhere in the code, so I thought it was justified.

PR #176, opened 2016-08-22
LowerTriangularTransform now also takes matrix, not just array of matrices

markvdw:
A small fix to the `LowerTriangular` transform, so that it now also takes MxM matrices, rather than NxMxM arrays of matrices. This is sometimes required and just makes it a bit nicer to work with.

Additionally, in case of any questions about `likelihoods.py`, my IDE automatically enforces PEP8 style, which caused all the changes. I decided to add them anyway, but feel free to remove if you think it causes clutter.

### DIFF
--- a/GPflow/likelihoods.py
+++ b/GPflow/likelihoods.py
@@ -18,8 +18,8 @@ import tensorflow as tf
 import numpy as np
 from .param import Parameterized, Param
 from . import transforms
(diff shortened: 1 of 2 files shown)


jameshensman:
Looks great. Could you add a little docstring to explain what squeeze does? It's obvious to me now, but it might not be in a few months.

jameshensman:
Great stuff. Merging.

PR #182, opened 2016-08-25
Remove default use of the LowerTriangular transform, for speed reasons

markvdw:
Need to add a test before merge.


markvdw:
Actually, @alexggmatthews, what kind of test did you mention earlier? The LowerTriangular transform is automatically tested by the batch transform tester. It's also run by a specific tester that checks whether it raises the correct errors. Did you think it was important just to see if it works in combination with `SVGP`?

alexggmatthews:
@markvdw  I think that should do in terms of test coverage.

alexggmatthews:
@markvdw  thanks for your PR.

Issue #181, opened 2016-08-25
GPflow has become considerably slower on GPU with some recent changes.


alexggmatthews:
I have raised this as a bug because as you will see the effect is large.

The main changes I can see are:

1) The addition of user_ops for VecToTri and TriToVec
2) The use of batch matrix triangular solve.

The git hashes I'm comparing are:

new with changes 23b47f590806f0da3cb71916d68674f8bee4d875
old without changes 9c2faf62d04f8ec2ab35396553b5115d92d3c84b

I'm running MNIST training with a minibatch size of 1000.

GPU operation is 70%  slower between these two hashes.
CPU operation is about the same. Did we observe a speed up when we merged these?

The new profiling code #174 allows us to give a detailed breakdown of the slow down. 
[user_ops_comparison.tar.gz](https://github.com/GPflow/GPflow/files/436776/user_ops_comparison.tar.gz)

The can be viewed in chrome using the GUI found at chrome://tracing .

We find that the batch matrix triangular solve does not parallelize as well as the triangular_solve which is arguably a flaw in TensorFlow itself.

We find that the VecToTri and TriToVec ops are currently very slow relatively considering they just reshape something.

**What should we do about this?**

I would argue we should find out exactly which recent changes are causing this and revert them as soon as possible. The new code looks really great and requires considerably skill but this is showing us that it isn't ready for master yet IMHO. The new user ops also make installation more difficult. We want our users at any moment to have the best available experience and we have to be dedicated to this if we are going to achieve our ambitions for the software. 

Longer term I argue that we should move the code to pull requests and expedite #174 #171 and #170

@markvdw @jameshensman thoughts?

markvdw:
It is surprising that `VecToTri` and `TriToVec` are so slow, and this needs investigating. Is this an orthogonal problem to the GPU ops (#171)? I.e. do the custom ops slow down the CPU execution significantly as well?

I agree that #170 is important and would catch these issues.

jameshensman:
HI Alex,

could you point me to the code that's running  to generat these traces
please?

James.

alexggmatthews:
Hi @markvdw no it is particularly a problem with GPU it seems. Whilst the CPU time of the VecToTri and TriToVec ops is significant in both cases it only seems to be a bottleneck on GPU for this example. The overall CPU run time is about the same as can be seen on the attached .json traces.

alexggmatthews:
@jameshensman that will need about 40min or so.

[...]

PR #183, opened 2016-08-29
Scoping

jameshensman:
Added some basic scoping functionality. 

1. I made a decorator in scoping.py. 
2. That decorator is applied in kullbackleiblers.py and conditionals.py
3. That decorator is automatically applied in any parameterized class where 'scoped_keys' has been defined.

### DIFF
--- a/GPflow/conditionals.py
+++ b/GPflow/conditionals.py
@@ -15,8 +15,10 @@
 
 from .tf_hacks import eye
 import tensorflow as tf
+from .scoping import NameScoped
 
 
+@NameScoped("conditional")
 def conditional(Xnew, X, kern, f, full_cov=False, q_sqrt=None, whiten=False):
     """
     Given F, representing the GP at the points X, produce the mean and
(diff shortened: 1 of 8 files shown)


alexggmatthews:
@jameshensman why do we have two different code coverage tests at the moment? In this case we are currently waiting for one of them to be reported.

jameshensman:
@alexggmatthews I guess codecov had a bit of an outage there. 

We have codecov and coveralls, but we don't depend on coveralls (it's not required to merge a commit). I'm happy to remove coveralls, which I don't rate, but I've left it there in case others are using it. 

This is the first time I've seen codecov have an outage, or be anything less that perfect.

alexggmatthews:
@jameshensman Looks good. Let's get some test coverage to check that the names are working as expected.

@markvdw I'm pretty happy with this. As you were involved in discussing this on Thursday thought I would give you an opportunity to have a quick look. We're keen to use this to fix #181 so if I don't hear back from you 1pm today I will merge (subject to tests).

markvdw:
I think this is a very elegant solution.

alexggmatthews:
@jameshensman thank you for the PR - merging now.

PR #185, opened 2016-08-31
Float types

jameshensman:
This PR adds support for float32 through settings.py

Surprisingly, most of GPflow still works with float32. I do anticipate that most users are going to want float64 though, because of the numerical stability. 

The only two tests to fail are:

- `test_method_equivalence` where optimization fails because the Cholesky decomposition fails
- `test_methods` where a lengthscale-gradient does not seem to be the same across two similar models (presumably this is a numerical thing)

If someone want to look a these, they'd be welcome, but the immediate question is how much speed this gains us on the GPU. 

Before this could be merged, we'd need a better method of controlling float64/32 through a config file -- I think there are lots of things that would be good in a config file, which I'll look at in a separate PR.


alexggmatthews:
@jameshensman Thanks for the PR. I think the Cholesky failure could be tricky to fix. Are you looking into the GPU speed up?

jameshensman:
I don't have time to look today, would love to see the timeline-traces using float32 on GPU.

Yeah, chasing down the source of the numerical instability is going to be really hard. We might have to force 64bit precision during matrix decomposition or something. 

There's a possibility though that the choice of l-bfgs-b as a default optimizer is compounding the issue. Perhaps a less 'aggressive' optimizer could cope with some numerical instabilities.

alexggmatthews:
@jameshensman your latest commit fixed my mnist bug without needing to merge your other branch. As it didn't come out in the unit tests perhaps we should add a regression test (i.e a test that checks the bug is fixed) for this?

alexggmatthews:
@jameshensman running mnist experiments on this branch:
https://github.com/GPflow/GPflow/tree/mnist_datatypes

to see how we do with reduced numerical precision. I had to remove the "nasty merge" as it was causing a lot of test failures. I'll let you know how it goes.

alexggmatthews:
@jameshensman looks like we're close on this one, no?

jameshensman:
yep. We need to make a decision on what to test: do we want the tests to pass in float32?

alexggmatthews:
@jameshensman I have to confess a bias on that one. mnist is seriously fast with the hybrid branch I linked and ends up with the same error of 97.94% . Great work :) 

Thoughts:
Maybe if we don't pass tests we give a warning? 
How far off passing on float32 do you think we are?

jameshensman:
(errors=4, failures=2), 189 tests. Some of those are easily fixed, others
are harder.

jameshensman:
now (errors=1, failures=2)

One of those failures, in test_data_holder, is a bug and can be fixed easily. Will look at it tomorrow.

The other failure is due to numerical instability: a lengthscale gradient does not match across two equivalent models (hardly surprising, I think we compute and differentiate K^{-1} * K * K^{-1} * y, which I imagine is silly in float32).

The error is in test_method_equivalence, where we optimize a series of models and make sure they do the same thing. numerical badness leads to a linear algebra error (LLT not valid). I've played with the code, and this can be fixed by adding a White kernel to each model: each model optimizes the white noise variance to zero, but the extra-posdefness helps during optimization. FITC, though,  needs special treatment, suggest a fixed white kernel. 

For the latter two, I'm not sure if I *want* to fix them. I'd rather just mark them as 'known to fail for float32'. Agree?

jameshensman:
In addition, we should add one extra tests to cover the code in DataHolder.py. This will let the codecov requirement pass too.

markvdw:
Hi James, just out of curiosity, where is Kinv K Kinv y computed? Is this an artefact of the comparison of models, or something that is actually done intentionally?

alexggmatthews:
@jameshensman I guess we wouldn't test float32 on continuous integration anyway because we would then back to the ballooning test times? 

Once we have that data holder test so that CI passes, I'm personally inclined towards merging and adding a ticket for any outstanding issues. We would then treat float64 as the supported data type and any change as a user decision that they need to check. 
 
I'd also be interested in the answer to @markvdw 's question about the inverses.

jameshensman:
Okay, I'll fix up that last test today.

The numerical instability is in SGPMC when the data points and the inducing points are the same. I guess it's actually

Kfu Kuu^{-1} v.

Of course, tensorflow doesn't know Kfu==Kuu, we just happen to feed a copy of the same data for X and Z. It copes pretty well, only one of the lengthscale gradients is off, and of course it's fine in float64.

markvdw:
@jameshensman What may make the problem worse, is that jitter is added to Kuu, but not Kfu, which should be equal to Kuu. It would be interesting to see if removing jitter (or adding jitter to Kfu if X happens to equal Z) solves this problem.

PR #191, opened 2016-09-01
Config

jameshensman:
Here's a proposal for how a config might work. 

Three things of note. 

1) I used the official python config style, and the built-in configparser. No extra dependencies. Should work in python 2 and 3 (I tried it on 3). 

2) I search for the config file in three places:
 - the current directory
 - the user's home directory
 - the GPflow directory

This means that we can distribute demos with gpflowrc tailored to the demo (in the same dir); users can set their own preferences if they wish; new users don't need to do anything.

3) I used the namedtuple class to make accessing the settings easy. You can type
```python
jitter = GPflow.settings.numerics.jitter_level
```

...but you can't assign! Sorry @markvdw.

Discuss.


jameshensman:
Any thoughts on this method of config, @markvdw?

alexggmatthews:
@jameshensman if you get that short unit test in I'm happy to merge this one also.

jameshensman:
working on it.

jameshensman:
okay @alexggmatthews , ready to go.

markvdw:
Currently, no flags are actually defined and used in the code right? Can I add a flag for verbosity of compilation and optimisation? This has been annoying me, and it would also be nice to have an example usage before merging, I think...

alexggmatthews:
@markvdw keen to get this one in as its blocking some other PRs. I'll add that as an issue. The unit test shows the functionality works although not a practical example.

alexggmatthews:
See #201

PR #195, opened 2016-09-07
Kernel expectations

markvdw:
Ok, so again not ready to be merged, but a start to what we're discussing in #91. There are lots of things up for discussion and TODOs to merge the existing psi calculations in here. I'll try to summarise.

Notes on the commit here:
- I have implemented `exKxz` which is the expectation `<x_t K_{x_{t-1}, Z}>_q_{x_{t-1:t}}`. This a necessary piece for GPSSMs, which I'm interested in. ~~It should collapse to `eKxz` for delta q(x_t).~~ (don't think this is true anymore).
- The function `exKxz` computes the above expectation for an entire time series. So give it N+1 means (`Xmu` N+1 x D) and N+1 covariances and cross-covariances (`Xcov` 2xN+1xDxD). However, in order for this to work the cross covariance in `Xcov[1, n-1, :, :]` needs to be consistent with the covariance in `Xcov[0, n, :, :]. This is ensured by the`TriDiagonalBlockRep`transform that I have added in`etransforms.py`.
- ~~I haven't implemented `eKxz` yet.~~ This is now done.
- ~~I have Monte Carlo tests for all the expectations. However, the one for `exKxz` has high variance. I would appreciate help here.~~

I added `TriDiagonalBlockRep` along with the tests mostly because I wanted to upload something quickly. I feel like it may be a little outside of scope for GPflow, and it's not quite finished. Essentially it's a way to freely parameterise a covariance matrix with a Markovian structure (i.e. the precision is block tri diagonal). If this is something that could be useful / isn't minded in GPflow, I'll leave it in. Otherwise I'll rework `test_ekern.py` to not need it.

Overall things to do and decide:
- ~~Should we keep kernels with expectations separate from the original kernels? I.e. should the functionality in `ekernels.py' be merged with`kernels.py`?~~ No, keep them separate.
- ~~Add quadrature for 1D kernels in the kernel base class.~~
- ~~Add expectations to Linear kernel.~~
- ~~Sort out sum and product kernels.~~
- ~~Add a check that allows for diagonal q(X) as well.~~
- ~~Make tests more reliable. Perhaps we can do some low dimensional quadrature instead of relying on monte carlo? Currently variance causes things to fail sometimes.~~


alexisboukouvalas:
Thanks @markvdw 
I will start taking ideas and code from #91 that should be helpful following yours and @jameshensman suggestions.

alexisboukouvalas:
@markvdw this is still very much in transition. I have put in the code kern_kernel_expectations.py but this is temporary and I use it to test the eKernels code. I have also started on using quadrature for testing. 

In addition to the list you have given above which I am still working at, could you implement the eKxz method for the RBF kernel unless you see an easy way to get there from the output of the exKxz method?
It would be nice to implement all the kernel expectation methods for the linear kernel as well. Will you do that or should I?

markvdw:
Sure, I can implement eKxz, and the relevant methods for the linear kernel. Will get back to you tomorrow on that to keep things moving!

markvdw:
@alexisboukouvalas Sorry for the wait, but I've added the kernel expectations for `Linear`, and `eKxz` for `RBF`.

markvdw:
Right, I made some more progress today.
- `exKxz` is now implemented in `Linear`
- Multidimensional quadrature is implemented for the kernel expectations in the `Kern` base class. This may not be practical, but technically *all* kernels now support kernel expectations in any situation :)
- Multidimensional quadrature is now used for testing the exact implementations in the tests, rather than the more unreliable Monte Carlo.

markvdw:
@alexisboukouvalas: I just merged master into this branch, so the GPLVM stuff should now work with tf0.11. I also made the necessary changes to our code for compatibility with tf0.11.

I also fixed the issue of the code not being compatible with slices for active dimensions. I had to fix two things:
- Bits of your code relying on `kern.active_dims`
- Bits in my code using `kern._slice` for getting the correct active dimensions out of covariance matrices. This is now done by `kern._slice_cov`.
This was one of the things we discussed a few weeks ago.

Things left to do on this branch:
- [ ] Get a Linear+RBF kernel working
- [ ] Get sum kernels working with the remaining terms done using numerical quadrature.
- [ ] Ability for diagonal q(X)
- [ ] Switch GPLVM code to use `ekernels` module.

markvdw:
@alexisboukouvalas: I'm working on getting `Add` to work with the kernel expectations for disjoint dimensions, but having some trouble. For eKzxKxz (phi2), your code simply add's the phi2's from the different kernels. Is this really correct? I think it should be more like:

<KxzKzx> = <(Kxz1 + Kxz2) (Kzx1 + Kzx2)> = <Kxz1Kzx1> + <Kzx2Kxz2> + <Kzx1Kxz2> + <Kzx2Kxz1>

alexisboukouvalas:
@markvdw I think this is right. It also reveals a whole in our testing - could you use the the multivariate quadrature code to do a test?

markvdw:
Yes! Working on it.

markvdw:
@alexisboukouvalas I've committed the latest version. I've spent quite a lot of time on this, but I can't get `Add` to work with `eKzxKxz`, when the kernels are on separate dimensions. I think I'm really close, the issue is that the numerical integration does not correspond to the analytical version. Could you have a look at this?

alexisboukouvalas:
many thanks @markvdw I will take a look tomorrow.

markvdw:
Fixed the issue in `eKzxKxz` in `ekernels.Add`.
- The decomposition when `on_seperate_dimensions` is True only works with diagonal q(X) (d'oh)
- I added a bit of quadrature code to fix this when q(X) is not diagonal. It computes the covariance of <Kzx1Kxz2> for q(X). The rest of the terms are computed using the exact `eK...` functions. This should allow `Add` to be used in general, since only a _tiny_ bit of the integral is done using quadrature! I expect this to have very little bias.

alexisboukouvalas:
That's great @markvdw ! I have been a bit swamped with things but it looks like we are basically there.
Is there anything else to do before switching the GPLVM code to use the new kernel expectations? I am happy to do that.

markvdw:
I'm swapping over the GPLVM code at the moment!

jameshensman:
Q: what happens if a user tries to build a GPLVM with 'regular' kernels
instead of eKernels? Do we 'translate' for them? Raise an error?

J.

markvdw:
Well, currently it would silently use quadrature for the kernel expectations. As we discussed, we could add a visible warning whenever a kernel expectation from `kernels.py` is called?

jameshensman:
Ah yes, I recall. Sounds like a good plan. Perhaps the visible warning can
point users in the direction of GPflow.eKernels.*

alexisboukouvalas:
Agreed - this seems like the most sensible option! Many thanks @markvdw for working so hard on this.
I will try to get some functional tests for GPLVM that go beyond code coverage.

markvdw:
Right, new update. I removed old psi statistics code and the GPLVM now uses the new kernel expectations. I had to add a new transform that constrains to diagonal matrices, since the new kernel expectations code can't handle this properly yet, while the GPLVM expects this.

There's one change that I would like @jameshensman in particular to look at: https://github.com/GPflow/GPflow/pull/195/commits/a2dd78468a0cb14db8f130d44e631c68d1863b45#diff-da2a47880fa50096cd5c579c2f45c874R247

jameshensman:
Increased performance for diagonal q(X)  can be considered an additional feature, I think. Better to focus on getting the last few remaining bits in, and the structure of the code in place. 

I love the fact that you used a transform for the diagonal matrices. Really shows how nice the framework is :)

alexisboukouvalas:
@markvdw why don't we merge ekernels with kernels? Is there a reason to use quadrature when using an RBF kernel? It makes testing easy but introduces an extra non-intuitive step for the user.

markvdw:
We were discussing this too. The reasons were mainly the following:
- It clutters up `kernels.py` too much. If the kernel expectations were added there, each kernel would become very long, which kind of removes a lot of the elegance and clarity. Most kernels now require only a few lines of a definition of `Kern.K()`.
- A warning being raised by the kernel expectations in `kernels.py` highlights the issue to the user.
- It makes testing easier.
- Also the quadrature is desirable for kernels where closed form expressions haven't been defined yet.

Have you been running into problems with this (apart from the test that I forgot to switch over :))?

alexisboukouvalas:
The warning is very helpful and made it clear immediately why it was taking long to do GPLVM:)
It's just a usability question. Note that in GPy this is done automatically as you would expect - i.e. it does quadrature only when an analytic solution is not implemented. 
For the first point, we could redo the code as in Gpy - have a kernels directory where each file is a kernel.

markvdw:
I'm hesitant. I think the current solution is actually quite elegant, the only issue being that it's possible to inadvertently use quadrature, when it isn't desired. However, this is addressed with the warning. I think the multi-file solution with a kernels directory is quite messy for very little gain...

But I'm open to change my mind. @jameshensman, thoughts?

jameshensman:
I like the current design. Let's not over-complicate this PR with a code reshuffle: we can do that later if we need to.

alexisboukouvalas:
@markvdw could we add support for the Prod kernel for non-overlapping active dimensions? Not strictly needed but we had it in the old gplvm branch.

markvdw:
@alexisboukouvalas: This is now done, but it only supports diagonal q(X) and kernels on separate dimensions, like the old code.

alexisboukouvalas:
Hi @markvdw , some issues to consider mostly to do with performance and memory usage:

1. Could we add RBF+Linear with overlapping dimensions? Analytic solution available in old code but how could it fit in new kernel framework is not clear. See [old code](https://github.com/alexisboukouvalas/GPLVMTry/blob/master/kernel_expectations.py) - build_psi_stats_rbf_plus_linear.

1. I have added a GPLVM notebook - GPflow/doc/source/notebooks/GPLVM but it fails with
	ValueError: Cannot create a tensor proto whose content is larger than 2GB.
	Dataset is 100 X 12 with an additive RBF+Linear kernel. 

1. I When running a performance comparison - (see https://github.com/alexisboukouvalas/GPLVMTry perfcompare.py)
The times I get are
Time GPy=10.1152 , OldPsi=19.3988, kernexp=42.059 seconds.
I get the following warning with the kernel expectations:
	Warning:  /Users/mqbssaby/anaconda/lib/python3.5/site-packages/tensorflow/python/ops/gradients.py:90: UserWarning:Converting sparse IndexedSlices to a dense Tensor of unknown shape. This may consume a large amount of memory.
Also note that the likelihood is not same between old and new code.
Will this be fixed if we do the optimisation you mentioned on doing only the diagonal?

1. For testing would a setting to disable quadrature be helpful? i.e. throw error instead of warning

markvdw:
Thanks for spotting these issues. It's so hard to do this myself since I don't go through all these use cases!

1. I definitely think it's important to get an RBF+Linear for overlapping dimensions working. Currently, I think the best way to do this would be to have a separate `SumLinearRBF` class. I have an idea on how to do this neater within the `Sum` class, but it only really makes sense if we start implementing kernel expectations for many more kernels.
2. The performance and memory issues are probably caused by quadrature being used, and non-diagonal q(X)'s. If q(X) is constrained to be diagonal, no quadrature should be needed for sum kernels.
3. The difference in likelihood does worry me though, as the code uses q(X) which is limited to a diagonal matrix, even though it is represented as a full matrix. This should give the same result as previously.
4. We can definitely do this. It's probably best to add an option to gpflowrc to control this.

I see two ways forward:
- We could continue to fix up these issues and only merge once the gplvm is 'feature complete'
- Or we could find a smaller set of features which are currently already supported, and fix these issues with PRs later.

Most issues currently are caused by use cases involving sum kernels. I think a viable basic GPLVM (simply RBF kernel) is already supported by the code and could currently already be merged. @alexisboukouvalas, what do you think?

alexisboukouvalas:
I can see the advantage of submitting the code as is now and working on improvements. 
For one thing, hopefully more people will start using it and we will get a better idea of what is needed.
Also other people could use the kernel expectations for other purposes.

I have fixed up the GPLVM notebook so it's working and readable.

@markvdw  could you have a quick look at [this code](https://github.com/alexisboukouvalas/GPLVMTry/blob/master/checkLikelihood.py)? It's comparing the likelihood at model construction and performance of the old psi code and the kernel expectations code. 
Why do the likelihoods not match? I have done my best to match the two models exactly but I still get
Likelihood at start OldPsi=-1595.10, kernexp=-1541.64
The new code is also consistently slower (about 2x) and I get this warning:
[...]tensorflow/python/ops/gradients.py:90: UserWarning:Converting sparse IndexedSlices to a dense Tensor of unknown shape. This may consume a large amount of memory.
If you do not see a problem here, I am happy to merge - I guess the main thing would be to have all the code reviewed.

markvdw:
I haven't looked into the UserWarning yet, but I suspect it could be due to `active_dims` being passed as a list. I'm not sure though.

Looking into the difference in marginal likelihoods now. Already found a few issues.

markvdw:
I haven't looked into the UserWarning yet, but I suspect it could be due to `active_dims` being passed as a list. I'm not sure though.

Looking into the difference in marginal likelihoods now. Already found a few issues.

markvdw:
@alexisboukouvalas: I made the PR https://github.com/alexisboukouvalas/GPLVMTry/pull/1 with a few changes to `checkLikelihood.py`.

There's also a small fix to `GPflow.gplvm.BayesianGPLVM`. `X_var` wasn't being set properly. The issue is fixed now.

alexisboukouvalas:
thanks @markvdw for taking a look. I confirm the script is working great now. Do you suggest  we put `checkLikelihoods.py` as a test? It would necessitate maintaining the old psi-stats code so I have stayed away from that option.
When using a slice operation we don't get the warning
`ke = GPflow.ekernels.RBF(1, ARD=True, active_dims=[0])`  - warning as list is used
`ke = GPflow.ekernels.RBF(1, ARD=True, active_dims=slice(0))` - no warning when slice is used.

markvdw:
Yeah I've noticed this warning while running the tests too. I think we'll have to understand why this happens, also since it probably relates to why the old notebook using `Add` is taking up so much memory.

I think keeping it in as a test would be too much code. The bug didn't influence the correctness of the gplvm code, just the initialisation. When printing `m.X_var.value`, the correct values were displayed, they were just different from the ones we wanted to initialise to.

markvdw:
@alexisboukouvalas: I restored your old notebook that uses the `Add` kernel. In order to get it to work, quadrature had to be switched off by setting `num_gauss_hermite_points = 0`. In the future, once diagonal q(X) is fully supported, this won't be necessary any more, as we should be able to detect when quadrature is necessary.

I also added a setting to raise an error if quadrature is used.

markvdw:
@alexisboukouvalas: The cross term for Linear + RBF with overlapping dimensions has been added. Currently it only works when all dimensions are used :) It's probably not difficult to modify it so it works in general, I just haven't thought it through too much. Patches/suggestions etc welcome!

@jameshensman: I was hoping you could have some input on this. The test currently failing is `TestAddCrossCalcs` in `test_ekerns.py`. I want it to compare the calculation of cross terms needed in `Add` to quadrature. But I'm having trouble getting GPflow to do the calculation correctly. I'm getting that some parameters turn up as `None` when during `tf_mode`. Any suggestions?

alexisboukouvalas:
thanks @markvdw - I will update the tests and notebook.

markvdw:
@jameshensman @alexisboukouvalas From my point of view, this is ready to commit!
- I'm very confident about the kernel expectation quadrature code. It's been untouched for weeks now, and we test everything against it.
- We have ways of switching this off (setting `num_gauss_hermite_points` to zero for an individual kernel, or setting `settings.numerics.quadrature` to error.
- I'm very confident about the kernel expectations we implemented for `ekernels.RBF` and `ekernels.Linear`.
- I'm quite happy with `ekernels.Add`. There are a few features still missing, but in principle it will do closed form expectations when it can, and otherwise fall back on quadrature for the cross kernel expectations. I only calculate the covariance of these, and this really a tiny adjustment. I'm pleased at how this worked out. I haven't used `ekernels.Add` extensively in my private code though, so I can't really vouch for its correctness above the unit tests.
- `ekernels.Prod` only supports limited cases now, for diagonal q(X) and `on_separate_dimensions`.

Overall, I think the kernel expectations code is ready to be merged. Most of it is not just unit tested, but also mature and I use it in my personal code. It also does not interfere with other GPflow code.

As for the GPLVM code, I've run a few simple examples, but @alexisboukouvalas has more practical experience with it!

@jameshensman, what do you think? Ready to merge?

jameshensman:
Oh! I just sent you both an email asking if this was ready -- sounds like it is :)

I'll have a peruse through and make comments if I think things need clarifying, but I already approve of the overall architecture of what you're doing.

J.

GPflow/gpflowrc
```diff
@@ -9,6 +9,8 @@ int_type = int32
 
 [numerics]
 jitter_level = 1e-6
+# quadrature can be set to: allow, warn, error
+quadrature = warn
```
jameshensman: Should this be `ekern_quadrature`, instead of just `quadrature`? We use quadrature as standard for many likelihoods.

GPflow/gplvm.py
```diff
@@ -0,0 +1,187 @@
...
+        :param M: number of inducing points
+        :param Z: matrix of inducing points, size M (inducing points) x Q (latent dimensions). By default
+        random permutation of X_mean.
+        :param X_prior_mean: prior mean used in KL term of bound. By default 0. Same size as X_mean.
+        :param X_prior_var: pripor variance used in KL term of bound. By default 1.
+        """
+        GPModel.__init__(self, X_mean, Y, kern, likelihood=likelihoods.Gaussian(), mean_function=Zero())
+        del self.X  # in GPLVM this is a Param
+        self.X_mean = Param(X_mean)
+        # diag_transform = transforms.DiagMatrix(X_var.shape[1])
```
jameshensman: What's going on here? It looks like the BGPLVM at one point supported diagonalmatrixTransform, but no longer does. It would be great if we could support both diagonal and lower-triangular matrix transforms.

GPflow/kernels.py
```diff
@@ -23,9 +25,36 @@
...
+
+
+def mvhermgauss(means, covs, H, D):
+    """
+    Return the evaluation locations, and weights for several multivariate Hermite-Gauss quadrature runs.
+    :param means: NxD
+    :param covs: NxDxD
+    :param H: Number of Gauss-Hermite evaluation points.
+    :param D: Number of input dimensions. Needs to be known at call-time.
+    :return: eval_locations (H**DxNxD), weights (H**D)
```
jameshensman: Should that be ((H**D)*N) x D)?

GPflow/kernels.py
```diff
@@ -51,21 +80,45 @@ def __init__(self, input_dim, active_dims=None):
...
                 X2 = tf.transpose(tf.gather(tf.transpose(X2), self.active_dims))
-            return X, X2
+        with tf.control_dependencies([
+            tf.assert_equal(tf.shape(X)[1], tf.constant(self.input_dim, dtype=settings.dtypes.int_type))
+        ]):
+            X = tf.identity(X)
+
+        return X, X2
+
+    def _slice_cov(self, cov):
```
jameshensman: Would be great to have docstrings for both slice and slice_cov (though I realize _slice is not part of this PR.)

GPflow/kernels.py
```diff
@@ -85,6 +138,107 @@ def compute_K_symm(self, X):
...
+        self._check_quadrature()
+        Xmu, _ = self._slice(Xmu, None)
+        Xcov = self._slice_cov(Xcov)
+        X, wn = mvhermgauss(Xmu, Xcov, self.num_gauss_hermite_points, self.input_dim)  # (H**DxNxD, H**D)
+        Kdiag = tf.reshape(self.Kdiag(X, presliced=True),
+                           (self.num_gauss_hermite_points ** self.input_dim, tf.shape(Xmu)[0]))
+        eKdiag = tf.reduce_sum(Kdiag * wn[:, None], 0)
+        return eKdiag  # N
+
+    def eKxz(self, Z, Xmu, Xcov):
```
jameshensman: Docstring please.

GPflow/kernels.py
```diff
@@ -85,6 +138,107 @@ def compute_K_symm(self, X):
...
+        Xcov = self._slice_cov(Xcov)
+        N = tf.shape(Xmu)[0]
+        M = tf.shape(Z)[0]
+        HpowD = self.num_gauss_hermite_points ** self.input_dim
+        X, wn = mvhermgauss(Xmu, Xcov, self.num_gauss_hermite_points, self.input_dim)  # (H**DxNxD, H**D)
+        Kxz = tf.reshape(self.K(tf.reshape(X, (-1, self.input_dim)), Z, presliced=True), (HpowD, N, M))
+        eKxz = tf.reduce_sum(Kxz * wn[:, None, None], 0)
+        return eKxz
+
+    def exKxz(self, Z, Xmu, Xcov):
```
jameshensman: Improve docstring please: specify what this function computes.

GPflow/kernels.py
```diff
@@ -85,6 +138,107 @@ def compute_K_symm(self, X):
...
+        fXcovb = tf.concat(2, (tf.transpose(Xcov[1, :-1, :, :], (0, 2, 1)), Xcov[0, 1:, :, :]))
+        fXcov = tf.concat(1, (fXcovt, fXcovb))  # Confirmed correct
+        X, wn = mvhermgauss(fXmu, fXcov, self.num_gauss_hermite_points, D * 2)  # (H**2DxNx2D, H**2D)
+        Kxz = tf.reshape(self.K(X[:, :D], Z), (Hpow2D, N, M))
+        exKxz = tf.reduce_sum(
+            tf.expand_dims(tf.reshape(X[:, D:], (Hpow2D, N, D)), 2) * tf.expand_dims(Kxz, 3) * wn[:, None, None, None],
+            0)
+        return exKxz
+
+    def eKzxKxz(self, Z, Xmu, Xcov):
```
jameshensman: Docstring this function please.

GPflow/kernels.py
```diff
@@ -105,7 +259,7 @@ class White(Static):
     The White kernel
     """
 
-    def K(self, X, X2=None):
+    def K(self, X, X2=None, presliced=False):
```
jameshensman: What's the behaviour of the `presliced` flag, and why is it necessary?

GPflow/kernels.py
```diff
@@ -464,18 +630,32 @@ def __init__(self, kern_list):
         names = make_kernel_names(self.kern_list)
         [setattr(self, name, k) for name, k in zip(names, self.kern_list)]
 
+    @property
+    def on_separate_dimensions(self):
```
jameshensman: please be specific about what this function does.

GPflow/transforms.py
```diff
@@ -164,6 +165,46 @@ def __str__(self):
...
+        return tf.matrix_diag(tf.reshape(self._positive_transform.tf_forward(x), (-1, self.dim)))
+
+    def tf_log_jacobian(self, x):
+        return tf.zeros((1,), float_type) + self._positive_transform.tf_log_jacobian(x)
+
+    def __str__(self):
+        return 'DiagMatrix'
+
+    def free_state_size(self, variable_shape):
+        return variable_shape[0] * variable_shape[1]
```
jameshensman: This seems like a pretty vital part of the transform: perhaps a test is needed?

alexisboukouvalas:
I have been using the GPLVM aspect of things and the code is working well. I have also attempted to add some rudimentary documentation in the form of a simple notebook that shows how one can use the additive kernel with GPLVM. I am not aware of any outstanding issues in terms of correctness.
Happy to proceed with the review and merge!

GPflow/transforms.py
```diff
@@ -236,12 +277,12 @@ def tf_forward(self, x):
         return tf.squeeze(fwd) if self.squeeze else fwd
 
     def tf_log_jacobian(self, x):
```
jameshensman: I think it's best not to hard-code `tf.float64`, rather use `settings.dtypes.float_type`

GPflow/ekernels.py
```diff
@@ -0,0 +1,307 @@
...
+import warnings
+import tensorflow as tf
+from . import kernels
+from .tf_wraps import eye
+from ._settings import settings
+
+int_type = settings.dtypes.int_type
+float_type = settings.dtypes.float_type
+
+# TODO: Efficiency gain from using diagonal q(X).
```
markvdw: Sure, do you mind keeping the TODO's in the code as well? Or would you rather have them removed?

GPflow/gpflowrc
markvdw: Done.

GPflow/gplvm.py
markvdw: Previously, the kernel expectations required rank 3 tensors, so I had to use a transform to constrain the NxD matrix of variances to a NxDxD tensor of diagonal matrices.

I've since reworked the kernel expectations to accept rank 2 tensors, and assume that this implies a diagonal q(X). This is a future-proof method for doing this, as we will want to use this representation to speed up the computation of the kernexps when q(X) is diagonal.

GPflow/kernels.py
markvdw: Fixed.

GPflow/kernels.py
markvdw: Fixed.

GPflow/kernels.py
markvdw: Done.

GPflow/kernels.py
markvdw: When computing kernel expectations, it is nicer to pass in a pre-sliced `X`. This way, the quadrature can create the input points on a lower dimension, rather than compute points on many more dimensions, only for some to be discarded by the kernel function.

An alternative would be to create the quadrature grid, and then to reverse slice it to fill up the sliced away dimensions with dummy data (e.g. zeros). I thought this would be impractical to do with TensorFlow manipulations.

GPflow/ekernels.py
jameshensman: I think we should just have a github issue. TODOs in the code make it seem incomplete, which I don't think fairly reflects the status of the code.

GPflow/transforms.py
markvdw: Done.

GPflow/ekernels.py
markvdw: Ok, will do.

markvdw:
Does this mean we're ready to go? Finally?! :D

markvdw:
What an epic pull request.

icouckuy:
Haha, congratulations. Yes, I did receive a lot of e-mails regarding this PR ;)

alexggmatthews:
@markvdw @alexisboukouvalas @jameshensman great to see this one go in. Thanks very much for your sustained efforts on this.

alexisboukouvalas:
It's been a pleasure helping on this! I look forward to help with the follow up issues ( #274 )

@jameshensman when/how are the docs updated?
In https://gpflow.readthedocs.io/en/latest/intro.html is still says GPLVM is not there:)

PR #196, opened 2016-09-08
Correction to link.

alexggmatthews:
Use tf 0.10 rather than the release candidate. Minor updates to readme.


alexggmatthews:
@jameshensman  this one is ready to go.

jameshensman:
Do we _require_ tf10? or will the rc do?

alexggmatthews:
We don't require tf10 but I'm not sure if the release candidate version is officially supported. It isn't linked from the tf webpage for instance.

alexggmatthews:
@jameshensman I just noticed intro.rst needed updating too. Going forward we probably don't want to have duplicate text in two files- I'll add an issue.

alexggmatthews:
See #197 .

PR #200, opened 2016-09-09
Mnist demo.

alexggmatthews:


alexggmatthews:
@jameshensman let me know what you think.

alexggmatthews:
Happy to notebook it if you think that is better, though we should consider what it would add here.

jameshensman:
I think this is fine as a python file, but it would be more discoverable if it were a notebook: we could include it in the documentation that way. 

Would be nice if
 - it worked with python 3
 - it downloaded the 'official' mnist data from Yann's site, instead of our pickle (which won't load in python 3 either. Because pickle is a naughty format.)

alexggmatthews:
Hmm I'm not personally sure about requiring Python 3 compatability for PRs. Is that our policy now?

alexggmatthews:
Also the notebook will take a few days to generate.

jameshensman:
I don't think it's a policy, but the rest of the code has py3
compatibility. I'm about ready to advocate for a switch to py3.5.

We don't have to run the notebook, just indicate how to reproduce the
result.

alexggmatthews:
Asking for Python 3 compatibility before merging PRs seems like policy! 

I'll add an issue to discuss.

I'm happy to switch this script over to a notebook.

jameshensman:
Fair comment :)

alexggmatthews:
See #203

icouckuy:
Why don't we use the mnist (and other data sets) code provided with tensorflow?

from tensorflow.examples.tutorials.mnist import input_data
mnist = input_data.read_data_sets("MNIST_data/", one_hot=False)

alexggmatthews:
@icouckuy good idea I will have a look.

jameshensman:
This looks great.

1) Could we have it a bit more visible -- perhaps a link from one of the documentation notebooks?

2) typo: "#achieves 97.94% error." you mean accuracy :)

AtheMathmo:
I've been trying to get a GP classifier working on MNIST myself without any luck. I saw this PR and tried running it myself but I cannot successfully train the model - my accuracy stays close to 10% .

My own attempts to reproduce the Sparse VI GP experiments from [this paper](https://arxiv.org/pdf/1706.07258.pdf) have also fallen short:  (Figure 5, Top Left).

Could somebody verify for me that this PR runs correctly for them and achieves ~98% accuracy? Otherwise, is it worth me opening a new issue for my own code to resolve the problems there? (From what I can tell my model parameters hardly change during optimization - even with different constraints, priors, and optimization algorithms/learning rates).

hughsalimbeni:
Are you using a GPU and tf1.1? If so, have a look at #415

alexggmatthews:
@AtheMathmo I haven't run this PR for a while but aside from any software issues yes it should work so if it doesn't we should get to the bottom of it. The original code was in GPy so this in fact replicates the experiments from our original paper. I'm afraid I don't know about the software for the paper you referenced. Let me know how it goes.

AtheMathmo:
I just tried this with tensorflow 1.2 instead of 1.1 and it worked as expected. Looks like I was just running into an odd (previously reported) bug.

Thanks!

alexggmatthews:
@athemathmo strange - I got a similar result using py 36 tf 1.2 and float32. Suggests something subtle has changed since I last ran it some months ago, which could be in either GPflow or tf itself. I'll keep you posted. The experiments take a couple of hours each. Worth noting this is still only a PR though :) 
The result was originally produced in different code and looking at Daniel's paper that you linked looks to have been repeated by others. I'm still therefore confident this is an issue with this implementation rather than with the method itself.

hughsalimbeni:
I'm looking at this now also and experimenting with using LBFGS on the full data. With py3.5, tf1.2 (GPU, float64) and M=300, ARD=False and RBF kernel with fixed variance of 1, after 1000 iterations (about 1h) I get
```
svgp, test ll: -0.082, test acc: 0.9792
svgp white, test ll: -0.086, test acc: 0.9788
```
In both cases the lengthscale optimizes to about 10, but with free kernel variance the hyperparmeters diverge. Not sure why this happens.

alexggmatthews:
@hughsalimbeni thanks for the update. Originally it was necessary to `freeze' the hypers and inducing points at the beginning of optimization because the optimization surface could lead off to bad local optima as reported in section 5 here:

http://proceedings.mlr.press/v38/hensman15.pdf

You could also ask James about this if you have started in Cambridge.

The mnist experiment actually appeared in our next NIPS paper though in many ways it would have been more natural in the AISTATS one. The code we released at that point is here: 

https://github.com/sparseMCMC/NIPS2015/blob/master/experiments/mnist/mnist_demo.py

Note the freezing. It might be a bit fiddly to resurrect this as the GPy versioning wasn't that tight. 

In this PR I was hoping to do away with freezing using a better initialization but it seems to have stopped working.

AtheMathmo:
I found in my own implementation that normalizing the input data was far more effective than rescaling between [-1,1]. If I find some time I'll see if that gives any wins for this PR too.

markvdw:
I've noticed the diverging hyperparameters as well. I've also noticed on some similar experiments which I've allowed to run for much longer that eventually, the hyperparameters stop increasing and start dropping back down again, before eventually stabilising.

A longer lengthscale with a larger variance still allows the GP to obtain a similar swing in latent function value between two points, keeping the data fit term the same. However, the variational approximation prefers longer length scales. Perhaps it eventually settles down once the variational approximation (through the inducing point locations) is good enough. It would be nice to test this hypothesis though, and perhaps do something about it. It probably doesn't help training.

I also wonder whether EP based approaches exhibit this effect. I wonder whether it's a consequence of the highly lopsided true posterior combined with a Gaussian approximate posterior. VI gets penalised heavily for placing density in regions where the true posterior is low. EP does not have this behaviour.

alexggmatthews:
@markvdw without further work that would be my hypothesis too.

alexggmatthews:
I'm going to try and get this up to date and in the next couple of days. Did anyone from the discussion want to update their comments?

markvdw:
By training with Adam for a really long time (50 gpu hrs), the error became 1.9%. Not necessarily practical, but it does show that more afressive optimisation can help. May be worth mentioning. 

More broadly, I think we don't really understand (our rather, have good rules of thumb) for stochastic optimisation of these objective functions. In NNs, so many things that don't affect the objective function (like batch size) affect performance. It makes it hard to figure out what to do if we really want to optimise the objective...

alexggmatthews:
@markvdw that seems like a very long time. Interesting thoughts though.

I believe @hughsalimbeni ran LBFGS on it which takes the stochastic aspect out of the loop. MNIST isn't that big for us anymore.

markvdw:
Agreed. I've run that experiment as well, although I don't think I ran it to convergence. With some similar models I've got the issue that I _just_ can't fit a single batch on the GPU memory in `float64`. This is needed to play well with LBFGS.

alexggmatthews:
So I find out doing this that TensorFlow currently allows non-determinism on GPU:

https://github.com/tensorflow/tensorflow/issues/2732

Therefore we can't expect exact replication here. However the initialisation needs to be sufficiently stable not to give large differences.

alexggmatthews:
Also try.contrib decided to randomised the mnist order silently but this is fixable.

alexggmatthews:
@AtheMathmo d7c7f86 restores freezing as we did in the original paper and has repeatedly given me accuracies around 98.0% . 

I would advise you to run  with float_type=float32 and jitter_level=1e-4 in your gpflowrc. I plan to make this automatic for this notebook. 

Note my comments above about exact TensorFlow determinism. 

My apologies for the delay fixing this. 

I will now make cosmetic changes to this pull request ready for merging to master.

PR #202, opened 2016-09-09
Verbosity flags for optimisation and compilation now used

markvdw:
Ready to be merged.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -10,3 +10,4 @@ exclude_lines =
     raise AssertionError
     raise NotImplementedError
     if __name__ == .__main__.:
+    print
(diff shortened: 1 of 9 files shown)


jameshensman:
Hmm, coverage fails because the print statements don't get tested. We could change the coveragerc to ignore all print statements? Writings tests for print statements seems overkill.

jameshensman:
woah, that's a bug in gpmc.py

```python
 L = tf.cholesky(K) + eye(tf.shape(self.X)[0])*settings.numerics.jitter_level
```

should be 
```python
 L = tf.cholesky(K + eye(tf.shape(self.X)[0])*settings.numerics.jitter_level)
```

jameshensman:
jitter also gets used in sgpr.py, conditionals.py, svgp.py

markvdw:
Should tests also depend on the jitter level?

jameshensman:
I thinks tests should pass with the gpflowrc that we provide. 

If users want to go off-piste and change the settings, that's fine, but running the tests for every permutation of gpflowrc is not feasible.

PR #205, opened 2016-09-09
Travis py3

jameshensman:
edits to travis.yml to get tests running on python 3


alexggmatthews:
@jameshensman thank you for your pull request. My only question is do we want code coverage on both python 2 and 3? I'm not sure.

jameshensman:
I don't see that it would add anything.

alexggmatthews:
@jameshensman OK I agree. I was slightly confused. I see 5 different "successful checks" two of them are coverage tests. What is the difference between them?

jameshensman:
If I understand correctly, travis/push tests the pushed code. travis/pr tests the code after it's been merged into master. So I guess if the pr is up to date with master, they will necessarily be identical.

https://docs.travis-ci.com/user/pull-requests

alexggmatthews:
OK thanks that explains some of what I see but there is a question which I think is orthogonal to this PR which is: "why is coveralls reporting the code coverage when we don't use it on the front page any more?"

jameshensman:
I think we left coveralls in because we were still transitioning around between services. It's easy to remove it...

alexggmatthews:
OK I'll add an issue to remind us to do it. A good end to a long day. Thanks and have a good weekend.

PR #210, opened 2016-09-14
Settings module now has ability to change during a run

markvdw:
Settings can now be modified while running. Needs a bit more work.

### DIFF
--- a/GPflow/_settings.py
+++ b/GPflow/_settings.py
@@ -1,7 +1,75 @@
 from six.moves import configparser
+import copy
 import os
 import collections
 import tensorflow as tf
+from collections import OrderedDict
+
+
+class SettingsContextManager(object):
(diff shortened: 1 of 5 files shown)


markvdw:
Ok, looks ready to be reviewed. What do you think (@jameshensman)? I think this will be pretty useful in tests. E.g.:
- Can test against float32 and float64 in a single file (if high precision is needed for a certain test)
- Can temporarily switch raise or lower jitter
- Demos / experiments can set the configs, for more guaranteed reproducibility

jameshensman:
I think this looks great!

Could you make a quick notebook explaining how to use it?

jameshensman:
I added the settings notebook to index.rst so that it show up in readthedocs. I also added a few lines to the notebook explaining where the gpflowrc file is and how to use it.

PR #217, opened 2016-10-03
Fixes for TF 0.11

icouckuy:
All tests succeed with these small changes.
- Removed the batch_ prefix from all except matmul
- Tensor class has moved to tf.

Ran 195 tests in 187.652s
OK


jameshensman:
Thanks @icouckuy !

PR #219, opened 2016-10-03
Added return statements in vgp._compile().

fujiisoup:
Fix to enable optimize_tf for vgp.


jameshensman:
with a test and everything! Thanks so much, nice work.

PR #222, opened 2016-10-04
Minor fix for gcc 6

icouckuy:
##


markvdw:
@jameshensman Any ideas why this is not reporting?

jameshensman:
Yes, it's because travis and codeship do not allow external PRs to trigger a run for security reasons. I made an internal version in #249

jameshensman:
There we go, now you can merge as you see fit, @markvdw

PR #224, opened 2016-10-08
Decrease the learning rate during stochastic learning

fujiisoup:
I think it is not a bad idea to decrease the learning rate during stochastic learning
by [tf.train.exponential_decay](https://www.tensorflow.org/versions/r0.10/api_docs/python/train.html#exponential_decay).

I modified the codes slightly to let model._optimize_tf() accept this option (and other for the future extension).
See `test_adam_option` method in `test_model.py` for the usage.

I'm looking forward to any comments.


markvdw:
I think this is a good option. Often, stochastic optimisation does fail to converge if the step size is too large.

Just curious, does the Adam test fail if the exponential decay is not used? This would be an actual functional test over just running the code.

fujiisoup:
@markvdw 

Thank you for the comment.
You are right. Test does not fail even without exponential decay.
I will update it.

fujiisoup:
I updated the test to make sure the global_step is certainly used.

+ Adopt GradientDescent rather than Adam
+ Decrease the threshold
+ Added test code to make sure this test fails without the global_step

markvdw:
@fujiisoup, Sorry that this PR has been unnoticed for a while. While it's a useful contribution, we're working on some modifications to GPflow that would make doing things like this a lot easier. I'm closing this PR for now while we come up with a more general solution.

Apologies for the long wait, we'll aim to merge PRs like this one faster in the future.

PR #226, opened 2016-10-12
added transforms for BoundedAbove and BoundedBelow

jameshensman:


alexggmatthews:
@jameshensman  Thanks for the PR. I agree that something along these lines would be useful addition.

I personally wonder if the class structure is a little confusing as it stands, though. The doc string in log1pe class doesn't explain that you can add a constant to the transform in this class. Do we really need three different classes here? In particular log1pe and BoundedBelow are really just the same class with different default parameters. Sorry to be picky but I'm thinking about getting this as clear as possible from a user perspective.

PR #229, opened 2016-10-17
Tensorboard

icouckuy:
Just some experimentation with tensorboard and GPflow. I started out with simply tracking the likelihood which, after some cleanup, is not very invasive. However, for illustration purposes I added some code to track the Param objects (only 1D atm) which required some hacks: most notable assigning a name to each Param instance.

In short, the code in Model._compile and Model._optimize_tf looks okay to me. Adding additional summary statistics for variables should be quite simple (these are oneliners in all the derived classes). If you have any ideas what to track over the optimization (or [MCMC] prediction) I can take a look at it (looking at Model.sample() in more detail next).

Regarding the cleanliness, I'm open to suggestions here (e.g., for adding a name to the summary statistics of Param object) as you know the code better than me.

(This branch is based of tf11.)


icouckuy:
Maybe I should add an option to the Param class that says how to visualise that parameter:
- as a scalar statistics (vector will be iterated over and a separate summary will be created for each
- as a distribution or histogram (does it makes sense tracking Z like that?)

For specific applications you also can store images and audios, but mostly for predicting (small) validation sets is what I would think.

icouckuy:
I noticed the code is broken in the sense that if you optimize the same model twice it will fail on adding the summary statistics (an op with the same tag/name already exist).

In fact this seems to be a more generic issue I encountered before. Calling build_likelihood twice presumably adds duplicate ops to the default graph. In our toolbox we use build_predict in our own wider TF flow and then call the necessary GPflow commands: make_tf_array, collect the feed_dict, ... just like in the autoflow decorator. However, e.g., this might get the Params in a wrong state:
- if I first call predict_y it fills in the storage cache, then use build_predict in our own code (here it already fails with unknown placeholders *only* if predict_y was called before, not sure why yet), calling predict_y afterwards will likely fail because it will take the free vars from the cache while I changed the state outside GPflow by calling make_tf_array.

I see there is some _recompile logic, is it supposed to solve this kind of problem?

jameshensman:
@icouckuy , I'm not sure I understand. Could you post a MWE as a bug, please?

icouckuy:
Actually, I'm not so sure where the error is originating from and I did not investigate it further. I skipped it for now and I am working on integrating some other functionality of GPflow in our toolbox first. I will get back to you when we also implement caching/autoflow on our side (and I still get the error)

icouckuy:
Here you go: [mweHensman.tar.gz](https://github.com/GPflow/GPflow/files/545876/mweHensman.tar.gz)

It seems simply constructing two GPflow models in the same python session triggers an undefined tensor error from tensorflow (using the tf11 branch)
`InvalidArgumentError: You must feed a value for placeholder tensor 'unnamed' with dtype double
	 [[Node: unnamed = Placeholder[dtype=DT_DOUBLE, shape=[], _device="/job:localhost/replica:0/task:0/cpu:0"]()]]`
caused on the second optimize.

icouckuy:
After understanding GPflow a little bit better (and finding out there is a self.long_name) I cleaned up the code. It is much cleaner now and provides basic support for TB.

I only have to test run the tracing again, but I remember there was another PR that generated trace files?

andreasdoerr:
I'm wondering about the current status of Tensorboard integration into GPFlow? Is there a recommended way of using Tensorboard in combination with GPFlow? Are @icouckuy changes going to be integrated in the future?

markvdw:
Personally I would love tensorboard support for GPflow. Currently I'm using a custom tracking class I published here https://github.com/markvdw/opt_tools, but tensorflow would be way better.

Currently, the main issue with this branch is that it's lacking a good minimum working example that allows us to verify its correctness. Before merging we need the code to be unittested and documented.

icouckuy:
@markvdw Just wondering, how would you verify the correctness of the tensorboard output?

This branch is quite old by now and needs updating (GPflow and TF development goes quite fast). I'll see if I can update it the coming weeks. In general, tensorboard support is pretty simple to do. One of the questions is whether to always enable tensorboard or have a switch (which would add a lot of if clauses)

alexggmatthews:
I'm thinking of closing this one as it is rather old and we have seem to have moved on. Happy to discuss. Any objections?

icouckuy:
Personally I think tensorboard support would still be useful however it can be closed for now. We can come back to it later when and if the need arises again.

alexggmatthews:
Agreed.

jameshensman:
We have some internal tensorboard support for GPflow that we will release
shortly. It's actually a fairly painless modification.

PR #238, opened 2016-10-24
Versionupdate3.2

jameshensman:
given that GPflow now requires TensorFlow 0.11, I think we should update the minor version number.


alexggmatthews:
@jameshensman thanks for your PR

PR #246, opened 2016-10-28
Derivatives of mean and var

icouckuy:
I needed the derivatives of the predictive distribution so I added an AutoFlow method for that. Feel free to pull if it is of interest.

I see it also includes the matpackops compilation commit for gcc 6 (thought that was already merged).


jameshensman:
Thanks, this looks useful!

A few small things before we can merge this. 

1) could you add a test, please? Doesn't need to be fancy, we just need to make sure we don't break this feature in future. 

2) Consider renaming the function. There's some ambiguity as to whether it's the derivative of  mean and variance of the function, or the mean and variance of the derivative of the function(!). Maybe the docstring could be a little more verbose to make this distinction too. 

3) I feel like we should have a derivative function for the latent function `f` too?
My preferred names would be `predict_f_gradients` and `predict_y_gradients`.

icouckuy:
Sure, I usually wait with cleaning up until the end.
1) Done
2) Renamed the function. Not sure what you would like for the docstring, to me it is about as clear as it gets now. Suggestions welcome.
3) Done

Bonnevie:
Just a cautionary comment, I am quite sure this does not compute the distribution of the predictive gradient (that is, the gradient of the posterior Gaussian process). If you just need these quantities for something else, then no worries, but just wanted to make sure the two weren't confused.

st--:
(superseded by #247)

PR #247, opened 2016-10-29
Add predict_f_gradients and predict_y_gradients to GPModel

jameshensman:
This is the in-house version of #246 (for continuous-integration purposes)


alexggmatthews:
This might be a good moment to ask. @icouckuy who are you? I'd like to update the readme with contributors.

icouckuy:
Ivo Couckuyt from Ghent university, Belgium. I  think I have a web page here:
http://sumo.intec.ugent.be/icouckuy
but it is a bit outdated.

robertocalandra:
It would be great if this pull request could be merged as it is very useful.
However, the function predict_f_gradients in this moment does not seem to support vectorized inputs -- all inputs after the first return 0 as gradient of both mean and variance.

Bonnevie:
Just put this comment in the original PR (and now I'm sincerely hoping I'm not wrong!):
Just a cautionary comment, I am quite sure this does not compute the distribution of the predictive gradient (that is, the gradient of the posterior Gaussian process). If you just need these quantities for something else, then no worries, but just wanted to make sure the two weren't confused.

icouckuy:
You are right that it does not compute the distribution of the predictive gradient. There was a short discussion on how to make that clear in the doc strings. I guess we failed :)

Bonnevie:
@icouckuy: ah, it's probably my mistake then - just skimmed the code. 
Is it for expected improvement?

icouckuy:
@Bonnevie yes indeed, I hooked it up in our toolbox so the complete EI (or KG, entropy, multi-objective EI, etc.) optimization is one big TF graph. Easily parallelizable (MC initialization of starting points and then n SGDs).

robertocalandra:
small note: I double-checked the gradients numerically and they are correct.

Bonnevie:
@icouckuy: it just so happens that I need BO myself, preferably with GPflow. Do you have the code in a particular branch I could take a peek at? Or is the toolbox not on GitHub?

jameshensman:
@robertocalandra , is this still the case?

"However, the function predict_f_gradients in this moment does not seem to support vectorized inputs "

I'd like to either merge or close this PR.

icouckuy:
Just a heads-up that this is still on our radar to check if vectorized inputs work. So that it doesn't get closed :)

icouckuy:
I'm not sure if I understand the issue correctly. It seems to work correctly when predicting many points at once in TF 1.2rc0.
figure_1-1

Someone pointed out you might mean multi-output GPs which indeed does not work. TF gradients returns sum(ys/dx). So if you model [y, -y] then only one set of gradients will be predicted, which in this case will likely be zero.

To solve this, tf.gradients needs to be called for each output separately or use the shorter version as in:
https://github.com/tensorflow/tensorflow/issues/4897#issuecomment-253056458

Both approaches requires the shape of self.Y (in particular, the number of outputs) to be known when constructing the graph which isn't the case. Methods like tf.split and tf.unstack raise an exception when they can't infer the shape. There are a couple of options as I see it:
- store the shape as a numpy array self._Y_shape before it becomes a DataHolder (or let DataHolder hold it/define a static TF shape).
- add an extra parameter outputIdx so the user can specify of which output he wants the gradient
- keep the current behavior as is
- ...

jameshensman:
@icouckuy I pushed what I hope is an acceptable fix. The shape of Y does not need to be stored anywhere.

 The return shape is num_points x input_dim x output_dim.

 Let me know if this works for you.

icouckuy:
@jameshensman Have to admit the tf.map_fn approach would have looked better, but I think is good for now. For some reason I like to keep controler flow structures outside of TF.

> The shape of Y does not need to be stored anywhere.
What's wrong with DataHolders returning a static TF shape?

jameshensman:
In the future, we might not know the shape of Y. For example if Y is a minibatch object, or some object that implements a tensorflow queueing system for the data. Since we don't need to know the dim of Y, it seems prudent not to use it.

st--:
Hi @icouckuy - before we merge this, could you please explain why you needed the tf.while loop? I've needed a similar column_wise gradient but implemented it as ```index_list = tf.range(tf.shape(xx)[0]);
grads = tf.map_fn(lambda i: tf.gradients(arg[i], xx)[0][i], index_list, dtype=float_type)``` where `arg` is the tensor of which you want the gradients and I hadn't noticed any issues.

st--:
We need to confirm what the best form for `columnwise_gradient()` is. There have been several requests for this feature, so it would be good to get this PR into master! Help appreciated... :)

jejjohnson:
Hello,

I was wondering if there are plans to continue with this pull request in the future for GPFlow 2.0?

jwangjie:
I wonder if there is any plan to add predict_f_gradients and predict_y_gradients to GPFlow 2.0? Thanks.

P-Mikkola:
Same here, I wonder if there is any plan to add predict_f_gradients and predict_y_gradients to GPFlow 2.0? :)

PR #251, opened 2016-11-01
improvements to setup.py

jameshensman:
minor edits to setup.py. 

When building the user ops, we need to get the include directories for TensorFlow. Previously, we were doing this by calling python from within setup.py (!) which is problematic on some clusters. This fixes that.


markvdw:
Nice, thanks James!

PR #252, opened 2016-11-01
improvements to doctring of kernels regarding slices

jameshensman:
I hope this clears up the discussion we've been having, @alexisboukouvalas , @markvdw


markvdw:
I just fixed the kernexp branch to allow a slice object to be passed into kernel. I was planning on just merging that into master once kernexp is ready, but this fixes the misunderstanding for now!

PR #253, opened 2016-11-01
active_dims fix in Kdiag

markvdw:
Linear and Polynomial kernels did not respect active_dims properly in Kdiag.

- Added slice call to Linear.Kdiag
- Fixed TestSlice to test more kernels.
- Fixed TestSlice to have the correct inputdim.

@jameshensman: I just want to double-check, I also think I fixed an existing bug in `TestSlice`, where all the `input_dim`s passed to the kernels was 1. Where actually, `input_dim` should be the number of inputs passed into the kernel, rather than the number actually used, right?

### DIFF
--- a/testing/test_kerns.py
+++ b/testing/test_kerns.py
@@ -3,7 +3,7 @@ import GPflow
 import tensorflow as tf
 import numpy as np
 import unittest
-from .reference import referenceRbfKernel, referencePeriodicKernel
+from testing.reference import referenceRbfKernel, referencePeriodicKernel
 
 
 class TestRbf(unittest.TestCase):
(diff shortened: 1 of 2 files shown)


testing/test_kerns.py
```diff
@@ -221,29 +221,38 @@ class TestSlice(unittest.TestCase):
...
+        tf.reset_default_graph()
+
         self.X = self.rng.randn(20, 2)
         self.Z = self.rng.randn(10, 2)
 
+        kernels = GPflow.kernels.Stationary.__subclasses__() + [GPflow.kernels.Constant, GPflow.kernels.Linear,
+                                                                GPflow.kernels.Polynomial]
+        self.kernels = []
+        for kernclass in kernels:
+            k1 = kernclass(self.X.shape[1], active_dims=[0])
```
jameshensman: I think this should be ` kernclass(1, active_dims=[0])`

testing/test_kerns.py
```diff
@@ -3,7 +3,7 @@
 import tensorflow as tf
 import numpy as np
 import unittest
-from .reference import referenceRbfKernel, referencePeriodicKernel
+from testing.reference import referenceRbfKernel, referencePeriodicKernel
```
jameshensman: I thing this was correct as it was? when does the relative import fail? Perhaps what's needed as python2 compatibility using `from __future__ import absolute_import`.

testing/test_kerns.py
markvdw: Ok, this is what I was really confused about, but I think it's consistent in my head now. So `input_dim` is the number of total active_dims, rather than the number of columns in `X`?

testing/test_kerns.py
markvdw: For me it fails if I try to run the test standalone as `python testing/test_kerns.py`. It works fine with nosetests. I can't really get why though.

testing/test_kerns.py
markvdw: Or to put it another way, after `_slice`, `tf.shape(X)[1]` should be equal to `active_dims`? Should we place an assertion for this in the code? I feel like currently it's really brittle, that you can pass in a set of arguments that will result in the kernel working on a different number of dimensions. This is an issue especially for the kernel expectations, where certain constants depend on the number of dimensions the kernel is acting on.

jameshensman: yes, we should assert that the arguments passed are consistent.

testing/test_kerns.py
markvdw: I can run the single test using `python -m testing.test_kerns` as well though. So I'm happy either way. Any strong opinion?

jameshensman: Urggh, I just went down a long rabbit hole on stackoverflow on this. so long as it works with nosetests and the tests pass on the serves, I'm happy :)

PR #258, opened 2016-11-03
Better deduction of `input_dim` for `kernels.Combination`

markvdw:
Deduces `input_dim` based on `input_dim` of a sub-kernel if a slice is used for `active_dims`, and the max of the `active_dims` otherwise.


jameshensman:
Beautifully pythonic.

PR #261, opened 2016-11-03
Autoflow fix

### LINKED ISSUE #256: Error when computing log likelihood with prior
The MWE below results in 
TypeError: Cannot interpret feed_dict key as Tensor: Tensor Tensor("X:0", shape=(?, ?), dtype=float64) is not an element of this graph.
If the _compile call is commented in the code works.

```import GPflow
import numpy as np
N = 10
X = np.random.rand(N, 1)
Y = np.sin(X) + np.random.randn(N, 1)
ki = GPflow.kernels.Matern32(1)
m = GPflow.gpr.GPR(X, Y, ki)
print('needs recompile', m._needs_recompile)
objInt = m.compute_log_likelihood()-m.compute_log_prior()
# m._compile()
objInt = m.compute_log_likelihood()-m.compute_log_prior()
print('needs recompile now?', m._needs_recompile)```

jameshensman:
Fixes #256 

#256 was caused by the changing of the graph behaviour. Now that all autoflow instances work on separate graphs, the `_tf_array` placeholders are overwritten each time an autoflow instance is created, because the placeholder must be on separate graphs. 

This broke the get_feed_dict mechanism, because sometimes the feed_dict contained placeholders that were outdated, i.e. on the wrong graph. 

The fix is to treat the `_tf_array` placeholders as temporary, and keep a reference to them inside AutoFlow (and in the model objective, which is really just a special case of AutoFlow). 

A regression test is included.

### DIFF
--- a/GPflow/kernels.py
+++ b/GPflow/kernels.py
@@ -34,7 +34,16 @@ class Kern(Parameterized):
     def __init__(self, input_dim, active_dims=None):
         """
         input dim is an integer
-        active dims is a (slice | iterable of integers | None)
+        active dims is either an iterable of integers or None.
+
+        Input dim is the number of input dimensions to the kernel. If the
+        kernel is computed on a matrix X which has more columns than input_dim,
+        then by default, only the first input_dim columns are used. If
+        different columns are required, then they may be specified by
+        active_dims.
+
+        If active dims is None, it effectively defaults to range(input_dim),
+        but we store it as a slice for efficiency.
         """
         Parameterized.__init__(self)
         self.scoped_keys.extend(['K', 'Kdiag'])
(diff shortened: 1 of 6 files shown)


jameshensman:
@alexisboukouvalas , could you confirm that this fixes #256 for you, please?

@alexggmatthews could you sanity-check this, please?

I this bugfix might warrant a minor version bump.

alexggmatthews:
@alexisboukouvalas no need to confirm I've checked both examples in #256 .

alexggmatthews:
@jameshensman I'm merging this as a priority bug fix.

I'd like to review param.py with you when you visit new week, please. It is necessarily complicated given everything it does, but given that it is so important I wonder if this is anything we can do to simplify it and add to the documentation.

alexggmatthews:
@jameshensman I will bump the minor version number in a PR.

PR #265, opened 2016-11-11
Student-t process

icouckuy:
Would the student-t process fit in GPflow's focus? I was experimenting a bit with it and implemented it based on GPflow. I wasn't planning to do a PR as it doesn't really fit mathematically into the design of GPflow but perhaps other people find it already useful. It is based on the work of Shah (using an inverse Wishart prior)

If someone can point me how they would like it implemented I'd be happy to make the changes. I ripped it from our own toolbox so it requires some changes, will do that the coming days. Aside from that:
- it is based on GPR because the posterior mean is the same as GPR, but perhaps it is better to derive it from GPModel or even just Model.
- the noise in a TP can't be handled by the likelihood (the sum of two t-distributions doesn't exist) so a White kernel needs to be added (and hence predict_y is probably not correct).

As an aside, I also want to marginalize (of a standard GP) the kernel variance and regression coeffients (like Kriging). However the predictive distribution in that case is a t-distribution and I would need to make some changes to the mean function classes to get model matrices back so I can solve them with least squares.


1Reinier:
Interested in using this. How is it coming along?

icouckuy:
What do you mean exactly? I don't think it will immediately be merged. At least I don't think expect so, as it just doesn't fit in the scope or design.

However, the code should be working and you can use it as is. I just tested it on a simple mathematical function where it worked as expected.

Or did you mean a GPR where the signal variance and regression coefficients are also marginalized?

1Reinier:
Okay thanks

markvdw:
@icouckuy I'm closing this for now as it doesn't really fit in the scope of GPflow. However, if you publish this somewhere as a gist or in a repo of its own, we'd love to link to it!

Issue #277, opened 2016-11-24
mean_function.Linear problem with params  


hughsalimbeni:
I've noticed that the Linear mean function has a bug that causes unintended parameter sharing. E.g.

```
m1 = GPflow.mean_functions.Linear()
m2 = GPflow.mean_functions.Linear()

print m1.b # 0
print m2.b # 0

m1.b = 1.

print m1.b # 1
print m2.b # has become 1. surely should be 0...

k1 = GPflow.kernels.RBF(1)
k2 = GPflow.kernels.RBF(1)

print k1.variance # 1
print k2.variance # 1

k1.variance = 2.

print k1.variance # 2
print k2.variance # still 1, as expected
```

This is due to the fact that b=np.array(1.) is given as a default argument, which is passed by reference to Param. Easy to fix  with a .copy(), but though I'd point this out in case it crops up elsewhere

jameshensman:
Thanks for the excellent report, @hughsalimbeni , I tried to fix this in #278. Could you take a look at that PR and tell me if the solution makes sense to you?

Thanks, 

James.

hughsalimbeni:
Looks great. Thanks!
Hugh

john-bradshaw:
Related to the above issue: I was wondering about the list default arg for list_of_params in the constructor of ParamList, ie line 923 of [param.py](https://github.com/GPflow/GPflow/blob/master/GPflow/param.py):

```

    def __init__(self, list_of_params=[]):
        Parameterized.__init__(self)
        for item in list_of_params:
            assert isinstance(item, (Param, Parameterized))
            item._parent = self
        self._list = list_of_params
```

Is this intentional here, ie do you mean to use a reference to same list each time you do not pass in an argument?

jameshensman:
Nope, this is a similar bug. Should only catch people using multiple paramlists _and_ using the default argument though! 

Perhaps the most elegant solution is to make the `list_of_params` argument non-optional. Then users can always init with `[]` if needed. Will break back compatibility, but I doubt it will actually hit any users.

alexggmatthews:
Fixed in #282 @john-bradshaw thanks for the spot.

PR #278, opened 2016-11-24
fixing bug #277, copy error in mean function constructors

jameshensman:


alexggmatthews:
@jameshensman thanks for this PR . Fixes #277

Issue #280, opened 2016-11-24
SGPR multiple columns 


hughsalimbeni:
I think SGPR may be missing output_dim factors in the last two terms of the bound. 

e.g.

```
rand = lambda s: np.reshape(np.random.randn(np.prod(s)), s)
X = rand((2, 1))
Y = rand((2, 4))
Z = rand((2, 1))

class SPGR_(GPflow.sgpr.SGPR):
    @GPflow.param.AutoFlow()
    def log_likelihood(self):
        return self.build_likelihood()

kern = GPflow.kernels.RBF(1, lengthscales=0.2)

l = SPGR_(X, Y, kern, Z).log_likelihood()
l1 = SPGR_(X, Y[:, :2], kern, Z).log_likelihood()                                
l2 = SPGR_(X, Y[:, 2:], kern, Z).log_likelihood()                                

print l
print l1 + l2 # should be equal...
```
but using GPy
                                 
```
kern_GPy = GPy.kern.RBF(1, lengthscale=0.2)
m = GPy.models.SparseGPRegression(X, Y, kern_GPy, Z).log_likelihood() 
m1 = GPy.models.SparseGPRegression(X, Y[:, :2], kern_GPy, Z).log_likelihood() 
m2 = GPy.models.SparseGPRegression(X, Y[:, 2:], kern_GPy, Z).log_likelihood() 

print m
print m1 + m2 # these are the same
```

jameshensman:
Correct, well spotted! The last two lines are missing a factor of `output_dim`. 

How about I add you to the GPflow team, and you can submit a PR? That way your name gets attached to the code. A small recognition, but these are excellent bug reports, really valuable.

jameshensman:
fixed in #283

PR #283, opened 2016-11-25
added test for multiple Y columns. Fix in SGPR and GPRFITC

hughsalimbeni:
response to #280 

Turns out there was also an error in FITC . 

I modified the method equivalence test to have two identical columns of Y. When I tried generating different columns (and higher dimensional inputs) the tests (sometimes) failed at the current tolerances.


jameshensman:
Hi @hughsalimbeni 

Yes, I'm not surprised that the tests fail numerically for more difficult problems. 

Congrats on your first GPflow PR!

Issue #281, opened 2016-11-24
minibatches with _optimize_tf


hughsalimbeni:
Apologies if I'm missing something with usage here, but it seems that minibatches are not actually changing when using a tensorflow optimizer. Is the optimize function intended to be called once per batch? This requires a recompile each time so I suspect not.

When I moved the lines
`feed_dict = {}
self.update_feed_dict(self._feed_dict_keys, feed_dict)
`
(225 of model) into the loop it works fine, however, but otherwise only the first batch is used for all the iterations.

alexggmatthews:
@hughsalimbeni I will have a look into this.

markvdw:
@hughsalimbeni: How could you tell that the same minibatch was used? I'm pretty sure I have tested this in the past and I found that repeated calls to `model._objective()` yielded noisy results, therefore I concluded that different minibatches were passed. Similar for optimisation.

hughsalimbeni:
I noticed this when trying the following sanity check:

```
from tensorflow.examples.tutorials.mnist import input_data
mnist = input_data.read_data_sets("MNIST_data/", one_hot=True)

X, Y = mnist.train.next_batch(50000)
Xs, Ys = mnist.validation.next_batch(5000)

rand = lambda s: np.reshape(np.random.randn(np.prod(s)), s)

class Linear_model(GPflow.model.Model):
    def __init__(self, X, Y, b):
        self.X = GPflow.svgp.MinibatchData(X, b, np.random.RandomState(0))
        self.Y = GPflow.svgp.MinibatchData(Y, b, np.random.RandomState(0))
        self.W = GPflow.param.Param(rand((X.shape[1], Y.shape[1])))
        self.b = GPflow.param.Param(np.zeros(Y.shape[1]))
        GPflow.model.Model.__init__(self)
            
    def build_likelihood(self):
        Y_preds = self.build_predict(self.X)
        return tf.reduce_mean(tf.reduce_sum(self.Y * tf.log(Y_preds), 1))
        
    def build_predict(self, Xnew):
        return tf.nn.softmax(tf.matmul(Xnew, self.W) + self.b)
        
    @GPflow.param.AutoFlow((tf.float64, [None, None]), (tf.float64, [None, None]))
    def accuracy(self, Xnew, Ynew):
        Y_preds = tf.arg_max(self.build_predict(Xnew), 1)
        Y_true = tf.arg_max(Ynew, 1)
[...]

PR #287, opened 2016-11-29
Improve test coverage for stochastic gradients with tf optimizer.

### LINKED ISSUE #284: Improve test coverage for stochastic gradients with tf optimizer.
#281 suggests we need better test coverage of this part of the code.

alexggmatthews:
Fixes #284 
Includes regression test for #281 
This one will need squash merging, please.

I've pulled out a lot of the minibatch functionality, tidied it and added unit tests. I've also added a deterministic minibatcher that cycles through the data.

The cyclic minibatcher enabled me to give much more precise tests of the expected behaviour of SVGP. It could also be useful in general but the interface is not intuitive. I suggest we keep it as an expert feature.

### DIFF
--- a/GPflow/minibatch.py
+++ b/GPflow/minibatch.py
@@ -28,19 +28,19 @@ class IndexManager(object):
         raise NotImplementedError
 
 class ReplacementSampling(IndexManager):
-	def nextIndeces(self):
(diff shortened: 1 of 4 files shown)


alexggmatthews:
@jameshensman this one is ready for review.

PR #289, opened 2016-11-30
fix minor bug in MultiClass likelihood when passing invlink

dougalsutherland:
The `MultiClass` likelihood allows you to pass in an `invlink` explicitly, as long as it's a `RobustMax`. But if you did that, it wouldn't actually set `self.invlink` and the object would be broken.

I doubt anyone was likely to do that, but it might as well work.


alexggmatthews:
#291 is the same as this so closing.

PR #291, opened 2016-12-02
fix minor bug in MultiClass when passing invlink

jameshensman:


alexggmatthews:
I have two pull requests that look similar. This one and #289 @jameshensman what is the difference, please? Thanks,
Alex

jameshensman:
The only difference is that this one is on an internal branch, so that the CI runs.

alexggmatthews:
@dougalsutherland  @jameshensman  thanks for the PR

Issue #292, opened 2016-12-05
Handing the random number generator down into burn in sampler


john-bradshaw:
So this is probably a really really minor issue (so maybe not worth changing) but for controlled randomness in https://github.com/GPflow/GPflow/blob/master/GPflow/hmc.py do you want to hand in the RNG object to the burn in sampler. ie change lines 61-63 in hmc.py from:

```
        samples = sample_HMC(f, num_samples=burn, Lmin=Lmin, Lmax=Lmax,
                             epsilon=epsilon, x0=x0,
                             verbose=verbose, thin=1, burn=0)
```

to:

```
        samples = sample_HMC(f, num_samples=burn, Lmin=Lmin, Lmax=Lmax,
                             epsilon=epsilon, x0=x0,
                             verbose=verbose, thin=1, burn=0, RNG=RNG)
```

john-bradshaw:
Also I think the array `logprob_track` initialised on line 55 of the same hmc.py module is one longer than it needs to be?

This is because it is filled by the following indexing operations `logprob_track[t // thin] ` (eg lines 111, 122 & 127) and t starts at 1 rather than 0 (line 77: `for t in range(1, num_samples * thin):`). 

Therefore I don't know whether it would be a good idea to change line 129-130 from:
```
    if return_logprobs:
        return samples, logprob_track
```
to:
```
    if return_logprobs:
        return samples, logprob_track[1:]
```
(if you want to keep the range of t starting at 1) the alternative maybe adding the initial logprob value into the first location in the array, similar to what is done with the samples array on line 70). 

Again this is really minor and I don't think has any serious negative effects, apart from traceplots produced from the logprobs having a confusing jump at the very beginning.

jameshensman:
Great suggestions @john-bradshaw,

It's absolutely super to have so many eyes on the code. I made a PR in #293: let me know what you think. 

Thanks,

James.

john-bradshaw:
looks great, thanks!

PR #293, opened 2016-12-05
minor updates to hmc.py

jameshensman:
as suggested in #292 by @john-bradshaw


alexggmatthews:
@jameshensman @john-bradshaw thanks for the price.

PR #304, opened 2016-12-22
Exponential distribution as prior

kiudee:
The exponential distribution here is parametrized using the rate parameter lambda and always has its mode at 0.

Any suggestion for improvement is welcome!

### DIFF
--- a/GPflow/priors.py
+++ b/GPflow/priors.py
@@ -46,6 +46,31 @@ class Prior(Parameterized):
         raise NotImplementedError
 
 
+class Exponential(Prior):
+    """
+    Exponential distribution.
+
+    Support: [0, inf)
+
+    Parameter:
+    rate -- float
+        Rate parameter (inverse scale) (rate > 0)
+    """
+    def __init__(self, rate):
+        Prior.__init__(self)
+        self.rate = np.atleast_1d(np.array(rate, np_float_type))
+        if rate <= 0:
+            raise ValueError("The rate parameter has to be positive.")
+
+    def logp(self, x):
+        return tf.cond(tf.reduce_any(x < 0),
+                       lambda: tf.constant(-np.infty, float_type),
+                       lambda: tf.reduce_sum(tf.log(self.rate) - self.rate * x))
+
+    def __str__(self):
+        return "Exp({})".format(self.rate)
+
+
 class Gaussian(Prior):
     def __init__(self, mu, var):
         Prior.__init__(self)
@@ -125,3 +150,4 @@ class Uniform(Prior):
 
     def __str__(self):
         return "U("+str(self.lower) + "," + str(self.upper) + ")"
+
--- a/GPflow/priors.py
+++ b/GPflow/priors.py
@@ -63,9 +63,8 @@ class Exponential(Prior):
             raise ValueError("The rate parameter has to be positive.")
 
     def logp(self, x):
-        return tf.cond(tf.reduce_any(x < 0),
-                       lambda: tf.constant(-np.infty, float_type),
-                       lambda: tf.reduce_sum(tf.log(self.rate) - self.rate * x))
+        scale = 1 / self.rate
+        return tf.reduce_sum(densities.exponential(scale, x))
 
     def __str__(self):
         return "Exp({})".format(self.rate)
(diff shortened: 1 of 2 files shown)


alexggmatthews:
Hi @kiudee  thanks for your interest in the software. Did you consider maybe using the Gamma prior? This has the exponential distribution as a special case.

kiudee:
Hi @alexggmatthews,

I'm planning to work more on the software in the future and that is why I
picked a simple distribution to implement and more will follow. I chose the
exponential distribution as a special case because it is used in many
examples of Bayesian inference and thus it is something many beginners
recognize.

I agree that we could mention in the doc string that it corresponds to a
gamma distribution with shape=1.

alexggmatthews:
Hi @kiudee , I guess what I meant is that we already have the Gamma distribution implemented, so you could use that instead.

kiudee:
Yes, I would have to check it out next week when I'm back from my holidays.
If the performance of both implementations is comparable, a simple wrapper
might be sufficient.

jameshensman:
I think we should include the Exponential prior for completeness.

J.

alexggmatthews:
@jameshensman I'm personally not so sure. Isn't there a risk of code creep here given that we already have the exponetial prior functionality implemented? Let's discuss it next time we talk on phone.

jameshensman:
Okay!

alexggmatthews:
@kiudee @jameshensman and I have just been chatting about this. We suggest using the simpler implementation of logp implementation of densities.py which doesn't have your defensive programming for the less than zero case. Other than that we would be happy to accept this PR.

kiudee:
@alexggmatthews @jameshensman I adjusted the code as suggested and added the necessary transformation to the test. Currently, the test is failing, but since I am on a windows pc right now I cannot verify why that is the case. The convergence to the mode is working for the following debug code:
```python
import scipy.optimize as opt
def wrapper_and_jac(x, lam=2.0):
    return -exp(np.exp(x), lam=lam), np.exp(x)/lam
result = opt.minimize(wrapper_and_jac, 1, jac=True)

np.allclose(np.exp(result.x), 0, atol=1e-4)
>>> True
```

kiudee:
@alexggmatthews @jameshensman After some debugging, I found the problem.
We are adjusting the logp using the jacobian of the transformation:
```python
return self.prior.logp(self._tf_array) + self._log_jacobian
```
This would not be the problem if the global maximum would be contained within the transformed space. But for the unconstrained exponential distribution the log probability is unbounded when x → -∞.

Without jacobian adjustment the optimizer correctly converges to a value close to 0 in the transformed space. With jacobian adjustment the optimizer converges to the real part of the logarithm of the rate:
```
real(log(-rate)) = log(rate)
```

What would you say - how can we ensure that the optimizer finds the correct mode in these cases?

jameshensman:
@kiudee @alexggmatthews 

This is an interesting and slightly-deeper-than-you-might-think question. The short answer is that there is no 'correct' mode. 

Take for example the uniform distribution on [0, 1]. Where would you say the mode was? Now, suppose we re-parameterize that as `x = probit(z)`. z can take any real value and x will be \in [0, 1]. But the density of z is N(0, 1), so has a mode at 0!

The mode of a (continuous) distribution is pretty meaningless. For the sake of tests, it's a simple thing that we can check to make sure (1) the code runs and (2) there have been no dramatic changes to the expression. For the exponential distribution, I suggest doing something different, like making sure that the density at 1 is e, say, or checking the gradient at a point against some known quantity.

kiudee:
@jameshensman Indeed, it’s definitely an interesting question to ponder.

Regarding the test: Currently, I check that the log probability at 1 for two different values of the rate. Is there a good way to check the gradient without needing to specify session, graph etc? I am just worried that the test suite is polluted with too much setup code just to test one gradient.

alexggmatthews:
@kiudee @jameshensman it would be good to either get this one in or close it. As far as I can tell there were just some straightforward things to fix with the tests?

kiudee:
Thanks for the reminder!
The failing test was caused by the missing sample function of the prior and was easy to fix. I would say the code is ready for review.

st--:
Superseded by #717

PR #307, opened 2017-01-08
Changes for tf0.12

markvdw:
TensorFlow has now released version 0.12.1, with a few breaking changes. Fixed with a simple rename though!


jameshensman:
We should definitely accept this, but I think it would be good to bump the version number also. Perhaps a note in the revisions file, and change the readme too.

markvdw:
I changed the test setup on codeship, but I couldn't figure out how to do it (or if I had the right privileges) on travis.

Issue #306, opened 2017-01-05
Double hermgauss


Bonnevie:
There appears to be two identical functions called hermgauss - one in likelihoods.py and one in kernels.py. I assume this is not intentional. Other files like ekernels.py import from kernels.py.

jameshensman:
Thanks for the report, @Bonnevie . You're right of course. 

The source of the confusion is that @markvdw implemented the multi-dimensional quadrature code for ekernels.py, and also reused the hermgauss function. It would be nice to tidy the code a little to unify these. 

Mark, since you know your way around these bits of code better than most, I've assigned you. I don't think this is urgent: perhaps we could flag it as a 'junior job' that people who wanted to start hacking on the codebase could handle, with your supervision?

markvdw:
Ah yes, I remember doing this and knowing it wasn't particularly neat. Perhaps the best solution would be to have a separate `quadrature.py`?

I agree it's not super high priority, and if you know anyone who wants to submit a patch, that would be great. Otherwise I can do it.

Bonnevie:
A separate quadrature.py would make a lot of sense. 

Related issue: I was just in the process of figuring out how mvhermgauss works and from an object-oriented perspective it seems a bit off. Since it's made to calculate N simultaneous quadratures it collects N separate grids in a single array. This is of course sensible since it allows a function to be applied to all evaluation points simultaneously, but it also means there are additional post-processing steps that need to be performed after every call to the function. Furthermore, I wouldn't say it's immediately apparent how you construct N separate quadratures from arrays with shape (herm_points^(input_dim) * N ) x input_dim in an unambigious manner without consulting the source and/or prior applications of mvhermgauss.

Copying a pattern from kernels.py, I think it would make sense to have a function of the form (maybe with a few more bells and whistles):
```
def mvquad(f, Xmu, Xcov, H, D):
'''compute N multivariate Gaussian expectations using Gauss-Hermite quadrature.
f - function object. Function to integrate.
Xmu - NxD tensor. Means of each of the N gaussians to integrate over.
Xcov - NxDxD tensor. Covariance matrices of each of the N gaussians to integrate over.
H - Integer. Number of points used for the quadrature along each dimension.
D - Integer. Number of input dimensions. 
'''
    X, wn = mvhermgauss(Xmu, Xcov, H, D)  # (H**DxNxD, H**D)
    fX = tf.reshape(f(X), (H ** D, tf.shape(Xmu)[0]))
[...]

PR #308, opened 2017-01-09
Quadrature

Bonnevie:
I have refactored the existing quadrature functions following the discussion for issue #306, and have made the necessary changes to all other files. The exception is ekernels.py, as the file's sole use of mvhermgauss seemed a bit complicated at first glance, so I thought it best to leave it for later. I have made sure that ekernels.py still works, though, but it should be fixed in the future. All existing tests finish without issue. 

EDIT: @markvdw might be interested.

EDIT2: I think I branched off at the wrong place, the multi-GP files (multilikelihoods.py and multimodels.py) are super-incomplete and shouldn't really be part of this pull request.


markvdw:
Yes, you always want to branch off the latest master for a new PR. Shouldn't be too hard to fix, though you may want to start a new PR.

As for `ekernels.py`, I think we do want to fix this as well before accepting this PR. When there's a problem, it's best to fix it fully.

That said, I don't think it's going to be all too hard. Essentially, in `ekernels.Add. quad_eKzx1Kxz2`, a quadrature is performed over a multi input, multi output function. The input is `D` dimensional, and the grid is produced by `mvhermgaussI()` accordingly. The output of the function that we want to integrate against, is NxMxM. Essentially the `f` needs to return `cKa[:, :, None, :] * cKb[:, :, :, None]`, calculated for the quadrature `X`s.

Bonnevie:
I have removed the files with an additional commit, but it might not be ideal. Is there any way to migrate the current changes to a branch off the master?

markvdw:
Argh, I don't know git well enough for that stuff. Whenever this happens to me, I just copy the edited files somewhere separate, switch to master, and then paste them back :)

mathDR:
Easiest way to do this would be to rebase against master:
https://git-scm.com/book/en/v2/Git-Branching-Rebasing

alexggmatthews:
Hi everyone, I usually avoid rebases on the main repo unless absolutely necessary and I can't really see that it is in this case. I suggest a fresh PR if this one has gone wrong as per @markvdw suggestion.

Bonnevie:
Think I got a clean branch going now, so I'll fix the ekernels.py file and then I'll make a new pull request.

markvdw:
Great, thanks a lot! I'll close this one then for the time being.

PR #309, opened 2017-01-09
Quadrature refactoring

Bonnevie:
This pull request is for a refactoring of the quadrature functions into a new file quadrature.py as discussed in issue #306 with @markvdw. All existing functions using either `hermgauss()` or `mvhermgauss()` have been rewritten to use the new framework. All existing tests finish without error.

The code has been documented in line with existing documentation.

It is quite possible that some of the implementations can be made neater, especially with respect to the individual functions, although care was taken to make the code legible and PEP-compatible.


markvdw:
Ok looks great, but I'm still wondering about whether `quadcov` is really necessary. We try to keep the number of functions down as much as possible, so adding this adds some hesitation in accepting the PR.

I removed the means from the covariance calculation, and then added them back later, since it works as a control variate in a way. The mean is strongly correlated with the integrand, and can be integrated exactly. Therefore the remaining contribution is smaller, so the error will be smaller as well.

I was thinking of calculating the integral as follows:

```
# Define this in `ekernels.Add.quad_eKzx1Kxz2()`
def integrand(X):
    cKa, cKb = [tf.reshape(
        k.K(tf.reshape(X, (-1, self.input_dim)), Z, presliced=False),
        (HpowD, N, M)
    ) - k.eKxz(Z, Xmu, Xcov)[None, :, :] for k in (Ka, Kb)]  # Centred Kxz
    eKa, eKb = Ka.eKxz(Z, Xmu, Xcov), Kb.eKxz(Z, Xmu, Xcov)

    return cKa[:, :, None, :] * cKb[:, :, :, None]  # H**D x N x M x M

cc = mvnquad(integrand, Xmu, Xcov, ...)
```

Any reason why this would not work?

Bonnevie:
I understand the hesitation. In the defense of `quadcov()`, though, I think it's a pretty sensible abstraction that might be useful for future implementations. 

There are two things that worry me about your proposal:
- It doesn't follow the same output shape format as the other cases. In the other functions, 'integrand()' outputs a matrix of collapsed shape (H**D * N x Dout) which `mvnquad()` then reshapes to (H**D x N x Dout). Your function would possibly have to be reshaped back to the collapsed shape to be consistent with `mvnquad()`. 
- Your function depends on Xmu and Xcov which means that those quantities affect the result both indirectly through 'integrand()' and directly through `mvnquad()`. I think this is undesirable as it makes the code more convoluted.

I think that maybe `mvnquad()` could be used directly with your function if we take Dout=(NxMxM) and tile `Xmu` and `Xcov` appropriately, but I feel like this would be a bastardization/hack of `mvnquad()`and thus not really result in transparent code.

My vote is to keep `quadcov()` for now for the sake of code readability since it's in line with peoples' intuitive understanding of what operation they want to perform.

One possible (future) resolution to this would be to integrate  `quadcov()'s extra functionality into `mvnquad()`. `quadcov()` already generalizes `mvnquad()` after all, as we just need to input some constant functions and zero-means to make it equivalent to standard quadrature. I'm not super-happy with `quadcov()`s function signature though, which seems a bit heavy with too many inputs.

markvdw:
Hmm, yes I see the issue now. It would have been fine if we didn't do the mean subtraction, but doing so means that the function needs to be aware of the different integration tasks that will be done with it. Not ideal.

I think the solution that I would currently favour is to just use `mvhermgauss()` within `quad_eKzx1Kxz2()`, and do the manipulations to find the correct `X` and `wn` there, instead of adding the extra function `quadcov()`.

I agree that `quadcov()` could be a useful abstraction in the future, but currently:
- It's only used in 1 place. It doesn't make sense to abstract something that's only used in one place.
- We're not quite happy with the interface yet. I don't want to add a function to the code when the interface may change in the future, especially if it's only used in one spot.
- Calculating the correct `X` and `wn` is only a few more lines in `quad_eKzx1Kxz2()`.

So my vote would be to:
- Branch off the current solution so we don't access to `quadcov()`. This way, we can easily find the code again if we need this functionality again in the future. Then we can re-assess what the correct interface should be in light of the more general use cases.
- Remove `quadcov()` from the current PR and do the cholesky for `X` manually in `quad_eKzx1Kxz2()`, using the new `mvhermgauss()`.

I think this solution has the most compact code, with the highest quality and the lowest complexity. This way we can get it merged as soon as possible.

Bonnevie:
So the only difference from the status quo in ekernels.py will be that it will use the `mvhermgauss()` from `quadrature.py` and otherwise it will follow the original implementation? 
I'll make the changes.

markvdw:
Yes that's right, since quadrature is only used in `quad_eKzx1Kxz2()` in `ekernels.py`. I was originally envisaging it to use `mvquad()`, like `kernels.py`, but I think you're right that it would be too messy.

Bonnevie:
Should be fixed now, @markvdw.

markvdw:
Amazing, this looks great. The commit history will be cleaned up when we squash & merge. We may need to wait for the `tf0.12` branch to be merged for the tests to pass again, but otherwise I reckon this is good to go!

Thanks for the great work.

Bonnevie:
Excellent!

jameshensman:
closing because this code was merged through an internal PR.

PR #310, opened 2017-01-11
Bonnevie quadrature refact

jameshensman:
in-house version of #309 so that CI runs correctly


jameshensman:
@Bonnevie Looks like the tests fail with 2.7 but pass with 3.5.

Bonnevie:
@jameshensman hmm, not really sure what I could have done to break 2.7 compatibility - but was naive enough to think that the 3.5 tests were sufficient (or maybe I was just hoping that it was magically setup to do all necessary testing across versions). Is there a development guide for GPflow with all the tests etc you'd like us to run prior to sending in a pull request? 
I'll take a look.

markvdw:
Generally we just run the tests under python2.7 and python3. It's the same testing suite and the same testing command. I think you should be able to view the error [here](https://app.codeship.com/projects/147609/builds/21604593). I have an anaconda environment set up that uses python2 by default.

I think the offending code is `(H ** Din, N, *Dout)`. I don't think this form of expanding tuples in Python2. I think an alternative is `(H ** Din, N) + Dout`.

I'm confused as to why `densities` can't seem to be imported though.

Bonnevie:
Got the `densities` thing too, although it seems to disappear if you use the innermost folder as your active folder. Can run the tests now and they seem to succeed. There was another problem with the same pattern, something like `*np.ones(...)`. 

How do I commit the changes to this particular PR?

jameshensman:
@Bonnevie I pulled the changes from your repo over to this pr (which is from within GPflow). 

@markvdw if you're happy with this PR you should be able to merge it when the tests pass. Let me know if otherwise.

markvdw:
Great, it's in! Thanks a lot @Bonnevie, a refactor like this can get pretty complicated, and it looks a lot better now than it was.

Bonnevie:
I'm glad I could help out :)

PR #316, opened 2017-01-12
Compatibility with next version of TensorFlow

markvdw:
This PR makes GPflow compatible with the next version of TensorFlow, where a few functions get renamed, and old deprecated ones are removed.

This is also compatible with the current version of TensorFlow, so it can be merged now.

### DIFF
--- a/GPflow/conditionals.py
+++ b/GPflow/conditionals.py
@@ -71,10 +71,10 @@ def conditional(Xnew, X, kern, f, full_cov=False, q_sqrt=None, whiten=False):
     # compute the covariance due to the conditioning
     if full_cov:
         fvar = kern.K(Xnew) - tf.matmul(A, A, transpose_a=True)
-        shape = tf.pack([tf.shape(f)[1], 1, 1])
+        shape = tf.stack([tf.shape(f)[1], 1, 1])
     else:
         fvar = kern.Kdiag(Xnew) - tf.reduce_sum(tf.square(A), 0)
-        shape = tf.pack([tf.shape(f)[1], 1])
+        shape = tf.stack([tf.shape(f)[1], 1])
     fvar = tf.tile(tf.expand_dims(fvar, 0), shape)  # D x N x N or D x N
 
     # another backsubstitution in the unwhitened case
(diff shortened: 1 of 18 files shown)


alexggmatthews:
@markvdw thanks for the work. this is tf1.0 correct? I think this is still in alpha. 

https://groups.google.com/a/tensorflow.org/forum/#!topic/discuss/HWjt_9zK4Qg

The tensorflow main page is still using the 0.12 version

https://www.tensorflow.org/get_started/os_setup

I suggest we wait until tf 1 has stabilized but good to know the changes aren't too horrible. Thoughts?

markvdw:
The changes here are compatible with 0.12.1, but required for 1.0. Some functions are removed in 1.0, like old batch_*, which are deprecated now anyway.

markvdw:
In case there's confusion, the tests still use tf0.12.1. There's no rush to get this patch in, although we will need the changes made here in the future.

jameshensman:
the most exciting part of tf1.0 might be that it will be available through pypi, so

`pip install tensorflow`

we should update the readme appropriately. We should also consider adding GPflow to pypi.

alexggmatthews:
Sorry was still half asleep when I read this morning.

alexggmatthews:
@markvdw I personally see no reason to wait.
@jameshensman Do you want to bump the version number for this?

markvdw:
There are no changes to the interface. We already bumped the version for compatibility with 0.12.1. Personally I'm always a little disappointed when I see a version bump of my favourite software and there not being anything new :)

That said, what's usually the version bump policy?

alexggmatthews:
@markvdw I too personally think there may be a risk of overbump here. Certainly if the external interface changes we try to increment soon after. Not sure there is much of a policy apart from that. @jameshensman will probably have a view on this.

jameshensman:
if this still works with 0.12.1, then I suggest we merge it now. But I think we should still test on 0.12.1, and point to 0.12.1 on the readme. When 1.0 hits, we should have minimal disruption.

We should bump versions if we ever change any of the api, add significant new features, or change the guts in such a way that power users might be affected.

jameshensman:
No, I don't.

My vote is to merge this now, without bumping version. When tf1.0 is released, we should check everything works and make a new version, even if the new feature is just that we're tf1.0 compatible.

alexggmatthews:
OK thanks everyone.

PR #320, opened 2017-01-17
Added a function evaluation counter

markvdw:
I thought it would be good to have a function evaluation counter in `Model`, for profiling purposes.


jameshensman:
Don't we get this in the return structure of the optimizer?

markvdw:
Yes we do, but:
- It's hard to keep track if you do multiple consecutive optimisation runs.
- You don't get it if you terminate the optimisation early (e.g. using a KeyboardInterrupt).
- The information is lost if you pickle the object.

jameshensman:
Oaky. But I think it should be called `feval_counter` or `num_fevals`.

markvdw:
Done.

markvdw:
I think this is ready to be merged. I added a small test for the counter as well.

PR #325, opened 2017-01-26
TF futureproofing: tf.mul -> tf.multiply.

markvdw:
More future proofing... Again, not crucial, but will have to happen eventually.


alexggmatthews:
@markvdw thanks for this PR. I see no reason to wait for a merge.

Issue #312, opened 2017-01-12
Support for multiple covarying GPs


Bonnevie:
A number of tasks like co-kriging, derivative observations, and Bayesian quadrature involve two (or more) GPs with different kernels. Often, there's some covariance structure between observations from different GPs in these settings (e.g. integrals and derivatives depend naturally on the function). 

The two GPs can be collected in a single Gaussian field with a block covariance matrix structure so that if we have observations x_1 ~ GP(0, K_11) and x_2 ~ GP(0, K_22)
```
k( (x_1, x_2), (x_1, x_2) )  = [[K_11, K_12], 
                                [K_21, K_22]]
```

which doesn't seem to be too far from something you could model in GPflow. I was thinking that maybe an approach like what's used in the coregional kernel could be repurposed? The framework could maybe also work by defining separate GPs and then have a function that takes the two GPs along with a covariance between them. 

Am I wrong in stipulating that this would be a pretty core thing to support? Derivative observations in particular should be plenty of reason since it integrates rather simply with all existing approaches.

jameshensman:
Hi @Bonnevie , 

Sure, multiple-output GPs fit into the GPflow framework in the way that the Coregion kernel works, exactly as you noted. The idea is that instead of specifying the blocks separately, you simply specify a kernel which takes an extra input column to specify where the blocks are. 

Thus, you can identify any multiple-output GP with a single-output GP, with an  extra indicator on the input. 

The neat thing about this approach is that one only has to implement a new kernel, and then it can plug in to any of the model classes in GPflow to do sparse GPs or VB or MCMC etc. 

I'm not sure that the working model you propose is the way to go, since it's not clear how to ensure that the resulting covariance matrix is positive-definite. Specifying an arbitrary off-diagonal block might lead to non-positive-definiteness. 


As for whether it's a core thing, I'm not sure. I think derivative GPs might be a core thing, so long as the implementation was pretty general. 

Perhaps the first thing to have a go at is a derivative kernel. Take the RBF as an example, and introduce extra columns that specify the derivative(s). That's something that I think many people would be interested in any could form a part of GPflow.

Bonnevie:
Won't you need to exploit the block structure for it to scale well? Of course, using the indicator approach, you could possibly just run a sort prior to computation, but I don't know whether that's a waste of computational effort.

[...]

PR #326, opened 2017-01-31
Block kernels, covarying GPs and gradient/Hessian observations for stationary kernels

Bonnevie:
Just wanted some extra pairs of eyes on this. (this touches on #312)

There are two files here, `multikernel.py` and `gradkern.py` where the latter is basically an application of the former. Both files predominantly define a new series of abstract classes that can be subclassed to get high-level kernels. Specifically, `multikernel.py` allows kernel matrices that are constructed block-wise according to an integer input denoting each observation's "group".

The highest level class in multikernel.py is `MultiKernel` which uses the Tensorflow functions `tf.dynamic_partition` and `tf.dynamic_stitch` to split the input into groups, then construct the kernel matrix from the chunks (left to the user), and then puts it back together and reorders to match the original ordering.

The next two classes implement two schemes for constructing kernel matrices. `BlockKernel` assumes two functions `subK` and `subKdiag` which take an index (i,j) as input. It then loops over all possible group combinations i-and-j, calls the sub functions to construct each block individually, and then collects them.
`RecursiveKernel` instead assumes a co-kriging setup with observations from group 0 being drawn from kernel 0, group 1 from 0 *and* 1, ..., group N from all prior kernels. The final covariance matrix is then a sum of nested submatrices. 

As a simple demonstration, `BlockLookupKernel` and `RecursiveLookupKernel` take lists-of-lists and lists (respectively) of kernels and offload the sub-construction function calls to those kernel functions. `RecursiveLookupKernel` works out of the box, `BlockLookupKernel` is only valid for appropriate choices of off-diagonal covariance. The kernels are stored in a new `KernelList` object to facilitate `Param` discovery.
There is also a `HeteroscedasticWhite` kernel to facilitate groups with different observation noise.

Using `dynamic_partition` and -`stitch` seems to be the only practical approach at the moment. Direct assignment to submatrices would likely be preferable, but does not seem to be supported by Tensorflow (tensors are immutable). Circumventing those two would involve assuming input that is already partitioned appropriately. 
`tf.concat` might be another bottleneck in the block-kernel setting, but is similarly inescapable.

`gradkern.py` implements support for derivative and second derivative observations for stationary kernels. There are two classes, `StationaryGradKernel` and `StationaryHessianKernel`, but they are fundamentally the same - only difference is that the latter instantiates more groups/blocks to accomodate second order derivative observations.

Since it subclasses `BlockKernel` the missing piece the class implements is the appropriate sub-kernel functions for filling in each block. To avoid doing a ton of `tf.cond()` calls I've implemented it using indexing operations. First, during initialization, an array is filled with appropriate tensorflow functions. This array is then indexed to find the appropriate function for each block. 

To generalize to as many kernels as possible, I have used the chain rule to decompose the derivative into univariate derivatives of the kernel function with respect to squared lengthscale-normalized distance and then partial derivatives of said distance with respect to the inputs. The appropriate functions are created dynamically by finding all possible combinations of the differential operators (in line with writing out the chain rule on paper, see [diffkerns.pdf](https://github.com/GPflow/GPflow/files/741860/diffkerns.pdf)). 

I could use your experience with "compiling" TensorFlow code - I have the feeling some of the graphs are being constructed more than once. I think the procedural generation is reasonable, but I'm also quite sure it must be possible to optimize it somewhat. 
    
Would like `gradkern.py` to integrate more neatly with existing `Stationary` kernels, but 1) we need a number of non-implemented derivatives of the kernel function wrt distance and 2) parameters need to be shared between derivatives. 

Ideally, we'd have used Tensorflow's automatic differentiation to build the kernels automatically, but after discussion and experimentation with a colleague this seems to not be possible (`tf.gradient` cannot do element-wise derivatives, `tf.unpack` doesn't work for dynamically-sized Tensors, you cannot use `tf.gradient` inside `while_loop`, `map_fn` or `scan`, and there's no clear route to emulating symbolic differentiation in Tensorflow using e.g. `graph_replace`).

### DIFF
--- a/GPflow/gradkern.py
+++ b/GPflow/gradkern.py
@@ -1,0 +1,240 @@
+import numpy as np
+import tensorflow as tf
+import GPflow
+from GPflow.param import Param
+from GPflow import transforms
+import pdb
+
+import matplotlib.pyplot as plt
+
+from functools import partial
+
+from multikernel import BlockKernel
+
+from GPflow._settings import settings
+float_type = settings.dtypes.float_type
+
+
+def find_partitions(collection):
+    '''Take an iterable `collection` and return a generator which produces
+    all unique partitions as lists-of-lists with every object being a member of
+    one set per partition.
+    alexis @ http://stackoverflow.com/questions/19368375/set-partitions-in-python
+    '''
+    if len(collection) == 1:
+        yield [ collection ]
+        return
+
+    first = collection[0]
+    for smaller in find_partitions(collection[1:]):
+        # insert `first` in each of the subpartition's subsets
+        for n, subset in enumerate(smaller):
+            yield smaller[:n] + [[ first ] + subset]  + smaller[n+1:]
+        # put `first` in its own subset
+        yield [ [ first ] ] + smaller
+
+class StationaryGradKernel(BlockKernel):
+    ''' Abstract kernel class which allows arbitrary derivative observations for
+    any differentiable stationary kernel. Computes each kernel matrix as a
+    composition of block matrices and automatically generates appropriate kernel
+    functions if given derivatives of the kernel matrix wrt squared distance.
+
+    To extend, inherit and replace `derivative_base` attribute with list of
+    the 0th, 1st, and 2nd derivative of desired kernel k(tau) with respect to squared
+    distance tau. Defaults to RBF/squared exponential.
+    '''
+    def __init__(self, input_dim, active_dims = None):
+        groups = input_dim
+        BlockKernel.__init__(self, input_dim, groups, active_dims)
+        self.variance = GPflow.param.Param(0.1, transforms.positive)
+        self.lengthscales = GPflow.param.Param(1.*np.ones(input_dim-1), transforms.positive)
+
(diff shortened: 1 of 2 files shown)


alexggmatthews:
@Bonnevie I think that we can't see all the tests here because this isn't a GPflow branch. If you accept the invitation I sent you then you should be able to run CI properly.

Bonnevie:
@alexggmatthews: ah, had somehow missed the important email I actually had to react to :) 
Was worried I had missed something, so just thought I'd play it safe for this pull request (the PR is mainly to get your feedback anyway). Anyway, now that I've accepted, do I just clone the true repository and push to branches in there instead of the fork?
I'm guessing code coverage isn't going to say much since none of the tests interact with the code so far.

alexggmatthews:
@Bonnevie  Yes clone the official repository and push to a remote branch that is not master and then do a PR using the Github web UI. master is in fact a protected branch as is detailed here:

https://github.com/GPflow/GPflow/blob/master/contributing.md

Suggest you do that in a new PR.

alexggmatthews:
@Bonnevie if this have lots of different components which you are suggesting may be merged to GPflow I also suggest trying to break it down into manageable chunks. Most PRs add a single feature. Doing it in this way helps the discussion stay focussed.

Bonnevie:
@alexggmatthews I have made a copy of the PR (#328) following your guidelines, except for splitting it into sub-issues since all of the features are all variations on the same new kernel superclass, so I think it's better judged as a whole.

PR #328, opened 2017-01-31
Block kernels, covarying GPs and gradient/Hessian observations for stationary kernels

Bonnevie:
This pull request is mostly to get opinions on the viability of the proposed approach. It touches on #312. There was concerns about this PR involving too many new features, but I couldn't really find a natural way to split it since the PR consists of an abstract kernel class and then various applications of it. 

There are two files here, `multikernel.py` and `gradkern.py` where the latter is basically an application of the former. Both files predominantly define a new series of abstract classes that can be subclassed to get high-level kernels. Specifically, `multikernel.py` allows kernel matrices that are constructed block-wise according to an integer input denoting each observation's "group".

The highest level class in `multikernel.py` is `MultiKernel` which uses the Tensorflow functions `tf.dynamic_partition` and `tf.dynamic_stitch` to split the input into groups, then construct the kernel matrix from the chunks (left to the user), and then puts it back together and reorders to match the original ordering.

The next two classes implement two schemes for constructing kernel matrices. `BlockKernel` assumes two functions `subK` and `subKdiag` which take an index (i,j) as input. It then loops over all possible group combinations i-and-j, calls the sub functions to construct each block individually, and then collects them.
`RecursiveKernel` instead assumes a co-kriging setup with observations from group 0 being drawn from kernel 0, group 1 from 0 *and* 1, ..., group N from all prior kernels. The final covariance matrix is then a sum of nested submatrices. 

As a simple demonstration, `BlockLookupKernel` and `RecursiveLookupKernel` take lists-of-lists and lists (respectively) of kernels and offload the sub-construction function calls to those kernel functions. `RecursiveLookupKernel` works out of the box, `BlockLookupKernel` is only valid for appropriate choices of off-diagonal covariance. The kernels are stored in a new `KernelList` object to facilitate `Param` discovery.
There is also a `HeteroscedasticWhite` kernel to facilitate groups with different observation noise.

Using `dynamic_partition` and -`stitch` seems to be the only practical approach at the moment. Direct assignment to submatrices would likely be preferable, but does not seem to be supported by Tensorflow (tensors are immutable). Circumventing those two would involve assuming input that is already partitioned appropriately. 
`tf.concat` might be another bottleneck in the block-kernel setting, but is similarly inescapable.

`gradkern.py` implements support for derivative and second derivative observations for stationary kernels. There are two classes, `StationaryGradKernel` and `StationaryHessianKernel`, but they are fundamentally the same - only difference is that the latter instantiates more groups/blocks to accomodate second order derivative observations.

Since it subclasses `BlockKernel` the missing piece the class implements is the appropriate sub-kernel functions for filling in each block. To avoid doing a ton of `tf.cond()` calls I've implemented it using indexing operations. First, during initialization, an array is filled with appropriate tensorflow functions. This array is then indexed to find the appropriate function for each block. 

To generalize to as many kernels as possible, I have used the chain rule to decompose the derivative into univariate derivatives of the kernel function with respect to squared lengthscale-normalized distance and then partial derivatives of said distance with respect to the inputs. The appropriate functions are created dynamically by finding all possible combinations of the differential operators (in line with writing out the chain rule on paper, see [diffkerns.pdf](https://github.com/GPflow/GPflow/files/741860/diffkerns.pdf)). 

I could use your experience with "compiling" TensorFlow code - I have the feeling some of the graphs are being constructed more than once. I think the procedural generation is reasonable, but I'm also quite sure it must be possible to optimize it somewhat. 
    
Would like `gradkern.py` to integrate more neatly with existing `Stationary` kernels, but 1) we need a number of non-implemented derivatives of the kernel function wrt distance and 2) parameters need to be shared between derivatives. 

Ideally, we'd have used Tensorflow's automatic differentiation to build the kernels automatically, but after discussion and experimentation with a colleague this seems to not be possible (`tf.gradient` cannot do element-wise derivatives, `tf.unpack` doesn't work for dynamically-sized Tensors, you cannot use `tf.gradient` inside `while_loop`, `map_fn` or `scan`, and there's no clear route to emulating symbolic differentiation in Tensorflow using e.g. `graph_replace`).

### DIFF
--- a/GPflow/gradkern.py
+++ b/GPflow/gradkern.py
@@ -1,0 +1,240 @@
+import numpy as np
+import tensorflow as tf
+import GPflow
+from GPflow.param import Param
+from GPflow import transforms
+import pdb
+
+import matplotlib.pyplot as plt
+
+from functools import partial
+
+from multikernel import BlockKernel
+
+from GPflow._settings import settings
+float_type = settings.dtypes.float_type
+
+
+def find_partitions(collection):
+    '''Take an iterable `collection` and return a generator which produces
+    all unique partitions as lists-of-lists with every object being a member of
+    one set per partition.
+    alexis @ http://stackoverflow.com/questions/19368375/set-partitions-in-python
+    '''
+    if len(collection) == 1:
+        yield [ collection ]
+        return
+
+    first = collection[0]
+    for smaller in find_partitions(collection[1:]):
+        # insert `first` in each of the subpartition's subsets
+        for n, subset in enumerate(smaller):
+            yield smaller[:n] + [[ first ] + subset]  + smaller[n+1:]
+        # put `first` in its own subset
+        yield [ [ first ] ] + smaller
+
+class StationaryGradKernel(BlockKernel):
+    ''' Abstract kernel class which allows arbitrary derivative observations for
+    any differentiable stationary kernel. Computes each kernel matrix as a
+    composition of block matrices and automatically generates appropriate kernel
+    functions if given derivatives of the kernel matrix wrt squared distance.
+
+    To extend, inherit and replace `derivative_base` attribute with list of
+    the 0th, 1st, and 2nd derivative of desired kernel k(tau) with respect to squared
+    distance tau. Defaults to RBF/squared exponential.
+    '''
+    def __init__(self, input_dim, active_dims = None):
+        groups = input_dim
+        BlockKernel.__init__(self, input_dim, groups, active_dims)
+        self.variance = GPflow.param.Param(0.1, transforms.positive)
+        self.lengthscales = GPflow.param.Param(1.*np.ones(input_dim-1), transforms.positive)
+
+        self.derivative_base = [lambda tau: self.variance*tf.exp(-0.5*tau), lambda tau: -0.5*self.variance*tf.exp(-0.5*tau), lambda tau: 0.25*self.variance*tf.exp(-0.5*tau)]
+        self.kerns = [[self._kernel_factory(i,j) for j in range(groups)] for i in range(groups)]
+
+    def subK(self, index, X, X2 = None):
+        i, j = index
+        if X2 is None:
+            X2 = X
+        return self.kerns[i][j](X, X2)
+
+    def subKdiag(self, index, X):
+        K  = self.subK((index, index), X, X)
+        return tf.diag_part(K)
+
+    def square_dist(self, X, X2):
+        '''Function `squared_dist` identical to method in `Stationary`.'''
+        X = X / self.lengthscales
+        Xs = tf.reduce_sum(tf.square(X), 1)
+        if X2 is None:
+            return -2 * tf.matmul(X, tf.transpose(X)) + \
+                   tf.reshape(Xs, (-1, 1)) + tf.reshape(Xs, (1, -1))
+        else:
+            X2 = X2 / self.lengthscales
+            X2s = tf.reduce_sum(tf.square(X2), 1)
+            return -2 * tf.matmul(X, tf.transpose(X2)) + \
+                   tf.reshape(Xs, (-1, 1)) + tf.reshape(X2s, (1, -1))
+
+    def square_dist_d(self, X, X2, diff_x=(), diff_y=()):
+        '''Take derivative of square_dist(x,y) wrt dimensions `diff_x` of first
+        argument and `diff_y` of second argument.
+
+        Parameters:
+        - `X` and `X2`: arguments for `square_dist`.
+        - `diff_x`: Tuple of integers. Dimensions to differentiate with respect
+            to in the first argument of `square_dist(x,y)`.
+        - `diff_y`: Tuple of integers. Dimensions to differentiate with respect
+            to in the second argument of `square_dist(x,y)`.
+
+        Return:
+        - Appropriate derivative as a Tensorflow Tensor.
+        '''
+        x_order = len(diff_x)
+        y_order = len(diff_y)
+        diff_order = x_order + y_order
+        dx = diff_x + diff_y
+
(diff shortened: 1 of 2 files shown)


alexggmatthews:
Obv. not to be merged yet. @Bonnevie I'll have a look. @jameshensman will want a look too but is on vacation this week.

Bonnevie:
@alexggmatthews: Obviously :) Hope it's not an issue that I started the PR though, just thought it'd make an easier forum for review. I would be thrilled if you'd take a look - I think it's making some fairly sensible extensions to GPflow and I've made an effort to stay within the bounds of the ecosystem.

Bonnevie:
Managed to reduce the computational load a bit by pre-computing the partitions and restructuring the computational graph a bit to avoid some redundant calculations (e.g. computed the difference between `X` and `X2` once and then fed the result to the convenience functions). 
I'll refrain from uploading the changes just yet as the code is probably a bit more transparent in its current form.

Bonnevie:
@jameshensman: I realize this is a rather large batch of code, but could you comment on whether this approach to derivatives is reasonable for GPflow? You mentioned earlier that you'd like to see derivative observations included.
It's likely that something could be done in the Kronecker setting as well, but it would be more restrictive.

gparracl:
@Bonnevie im interested in trying your code for a special case of autocovariances and cross-covariances functions, it is there any way to contact you?

Bonnevie:
@gparracl: You can contact me personally on rabo@dtu.dk. The code should be available through the blockkernel branch, and the minor performance updates are in the fast-grad branch. Me and a colleague have tested it a bit, so we are fairly confident it's working.

icouckuy:
@Bonnevie is there any update on this? It seems really useful.

Regarding tf.gradient, apparently it does work in a tf.while_loop
https://github.com/GPflow/GPflow/pull/247/commits/5ae691d154350b30926b9f7896ac51ba141bc972

Bonnevie:
@icouckuy: I haven't been working actively on it, but the code functioned perfectly well last time I tried it out. I was mostly waiting on responses as to whether this was the "right way" to go about the problem in GPflow or people had any constructive criticism. I'd gladly take up work on it again if there's a need for it.

Interesting to hear that `tf.gradient`works in `tf.while_loop`. We investigated that quite thoroughly at the time, so I'm guessing it's a new development?

Bonnevie:
Seems like it's mostly test coverage that fails - if @alexggmatthews and the rest of the core sign off on the basic design, I can write some tests. 
That said, what would be a suitable way to test this properly? I can compare predicted derivatives to true for a number of functions, but it's not clear how to turn this into a deterministic test. Maybe I should hand-evaluate a minimal example?

mrksr:
I like the idea of generic blockkernels since these tend to pop up often. A few thoughts:
* I think it might make things nicer if the index dimension is the last dimension. This would make things consistent with `SwitchedLikelihood`, the coregionalization stuff and GPy. Also, you could use non-block-kernels like a standard RBF by using a lower dimensionality of the kernel without having to manipulate the `active_dims`.
* In the same vein, would it make sense to call these things `SwitchedKernel` or rename `SwitchedLikelihood` to `MultiLikelihood`?
* It would be great if these kernels could be used in LVMs since for some of them, implementing the expected values should be fairly easy. It might be good to at least keep this in mind, not necessarily implement it right now.
* A special case: How can we make this compatible with the quadratures? They would need to honor the index dimension.

And a broader question: Index-Dimensions are fairly hacky, both on the Xs (for block kernels) and Ys (for the likelihood). Should this be handled differently?

Bonnevie:
Thanks for the response @mrksr. Hadn't noticed `SwitchedLikelihood` previously. Vindicating to see that we've come up with almost identical solutions independently. 
Coregionalization doesn't seem to use the last index, but `SwitchedLikelihood`does, so I agree that it would make sense to make the last index the default. Renaming seems sensible too. Haven't looked at how to integrate it with LVMs and quadrature.

Concerning index dimensions, it really seems to be the preferred way to implement it. On the abstract level, we want to implement a GP with a block-structured kernel so we need some way to instruct GPflow what block each datapoint belongs to. We could block-structure everything ahead of time, but this doesn't mesh with the kernels-as-objects setup. Alternatively, we could have a `SwitchedGP` class which could be a list of dependent GPs and then you'd pass observations as a list of arrays as well so that the structuring is explicit. But this goes quite a bit against how everything is set up. I think Hensman recommended indices in a discussion somewhere here in the repository.

mrksr:
The coregionalization-kernel is indeed implemented in an agnostic manner. I was just thinking of the [Tutorial Notebook](https://gpflow.readthedocs.io/en/latest/notebooks/coreg_demo.html).

I, too, think, that the index dimensions are the way to go, at least from a simplicity point of view. The only other way I see is to in some (non inband) way augment data points with metainformation. Like an additional parameter to the relevant functions. I have thought about this a bit (for other models I implemented) and could not come up with any good solution, which is why I mainly wanted to raise this as food for thought. The main argument I see against index dimensions is the fact that these dimensions always need to be handled in a special way and there is no way to recognize them. The quadratures are an instance of this problem, where it suddenly becomes invalid to assume that the inputs are continuous. I decided to stick with index dimensions, too, but they feel wrong to me.

Bonnevie:
If GPflow was graphical model-oriented with random variables as first-class citizens, handling these sort of dependencies could be done in the background, but it's troublesome when models and likelihoods are in the center - there's no obvious way to define a Hessian GP object with no data or to take the derivative of a GP object. (of course, that design choice makes a lot of other things a lot easier).

Making new data containers augmented with metadata might be a sustainable solution, though? Of course it's quite similar to using indices. On the other hand, does storing the indices in a data column make it easier to manage and access?

markvdw:
This is for the old GPflow. May be useful if these features are to be included in the future.

st--:
@Bonnevie would you be interested in updating this for current GPflow ?

Issue #327, opened 2017-01-31
LVMs slower on GPU than on CPU


mrksr:
Hi everyone!
After updating to tensorflow 0.12.1, LVMs are very slow on GPUs.
Using this code:
```python
import GPflow
import tensorflow as tf
import numpy as np

import pods
pods.datasets.overide_manual_authorize = True

data = pods.datasets.oil_100()
Y = data['X']

Q = 2
M = 20 
N = Y.shape[0]
X_mean = GPflow.gplvm.PCA_reduce(Y, Q)
Z = np.random.permutation(X_mean.copy())[:M]

k = GPflow.ekernels.RBF(2, ARD=True)

m = GPflow.gplvm.BayesianGPLVM(
    X_mean=X_mean,
    X_var=0.1*np.ones((N, Q)),
    Y=Y,
    kern=k,
    M=M,
    Z=Z
)

m.likelihood.variance = 0.01
%time m.optimize(method='L-BFGS-B', disp=True, maxiter=1000)
````

On the same machine, the optimization takes about **1 Minute using only CPUs** and about **10 Minutes when adding one GPU**.

I tried something similar for non-LVMs:
```python
import GPflow
import tensorflow as tf
import numpy as np

X = np.mgrid[0:100:10000j][:, None]
Y = np.sin(X) + 0.1 * np.random.randn(*X.shape)

num_data = X.shape[0]
num_inducing = 200

Z = X[np.random.permutation(num_data)[:num_inducing]].copy()

m = GPflow.sgpr.SGPR(X, Y, GPflow.ekernels.RBF(X.shape[1], ARD=True), Z)

%time m.optimize(method='L-BFGS-B', maxiter=1000)
```
And here, optimization using GPUs was considerably faster by about a factor of 10.

mrksr:
I initially posted this in #181 since a quick skim through the latest changes to GPFlow showed some changes concerning `batch_matmul` and `matmul`, which was discussed as a possible cause there.

alexggmatthews:
Hi guys, we need to use the recently merged #174 to profile it and that will quickly tell us the cause. In general the profiling functionality in tf is excellent so this will quickly help us get to the bottom of this.

markvdw:
@mrksr: So the code was faster in previous versions of TensorFlow?

[...]

PR #329, opened 2017-02-02
Avoid copies via tiling in eKernels

mrksr:
As discussed in #327, the tiling operations in `ekernels.py` seem to be the main reason for the bad performance of LVMs on GPUs right now. This pull request avoids these tiling operations, but currently only for the RBF-kernel, so it is not ready to merge yet. It seems like the most important change is the one in `eKzxKxz`. For some toy data, the running time on GPUs reduced from `2s` to `25ms`.

You can comparre the traces before and after the change:
* [timeline_objective_lvm_cpu.json.gz](https://github.com/GPflow/GPflow/files/744540/timeline_objective_lvm_cpu.json.gz)
* [**timeline_objective_lvm_gpu.json.gz**](https://github.com/GPflow/GPflow/files/744541/timeline_objective_lvm_gpu.json.gz)
* [**timeline_objective_lvm_gpu_new.json.gz**](https://github.com/GPflow/GPflow/files/747623/timeline_objective_lvm_gpu_new.json.gz)

There is some feedback I would like:
1. I use `tf.einsum` to avoid one tiling operation in a matrix product. This is the only use of `tf.einsum` in GPflow. Should and can the line be formulated differently? I will probably remove the extra comments later on.
2. There is still a huge allocation in [ekernels.py#L112](https://github.com/GPflow/GPflow/pull/329/files#diff-77a07191da6b8b3d69177df73dbe080dR112), but this one does not contain duplicates it think. Can it be formulated any better?

### DIFF
--- a/GPflow/ekernels.py
+++ b/GPflow/ekernels.py
@@ -77,13 +77,10 @@ class RBF(kernels.RBF):
         )  # N
 
         vec = tf.expand_dims(tf.transpose(Z), 0) - tf.expand_dims(Xmum, 2)  # NxDxM
-        smIvec = tf.matrix_solve(scalemat, vec) # NxDxM
-        q = tf.reduce_sum(smIvec * vec, [1]) # NxM
+        smIvec = tf.matrix_solve(scalemat, vec)  # NxDxM
+        q = tf.reduce_sum(smIvec * vec, [1])  # NxM
 
-        # Xsigmc: NxD*xD
-        # smIvec: NxD*xM
-        # addvec: NxMxD
-        addvec = tf.einsum('ned,nem->nmd', Xsigmc, smIvec) + tf.expand_dims(Xmup, 1)
+        addvec = tf.matmul(smIvec, Xsigmc, transpose_a=True) + tf.expand_dims(Xmup, 1)  # NxMxD
 
         return self.variance * addvec * tf.reshape(det ** -0.5, (N, 1, 1)) * tf.expand_dims(tf.exp(-0.5 * q), 2)
 
--- a/GPflow/ekernels.py
+++ b/GPflow/ekernels.py
@@ -36,11 +36,11 @@ class RBF(kernels.RBF):
         D = tf.shape(Xmu)[1]
         lengthscales = self.lengthscales if self.ARD else tf.zeros((D,), dtype=float_type) + self.lengthscales
 
-        vec = tf.expand_dims(Xmu, 1) - tf.expand_dims(Z, 0)  # NxMxD
+        vec = tf.expand_dims(Xmu, 2) - tf.expand_dims(tf.transpose(Z), 0)  # NxDxM
         scalemat = tf.expand_dims(tf.diag(lengthscales ** 2.0), 0) + Xcov  # NxDxD
-        rsm = tf.tile(tf.expand_dims(scalemat, 1), (1, M, 1, 1))  # Reshaped scalemat
-        smIvec = tf.matrix_solve(rsm, tf.expand_dims(vec, 3))[:, :, :, 0]  # NxMxD
-        q = tf.reduce_sum(smIvec * vec, [2])  # NxM
+        smIvec = tf.matrix_solve(scalemat, vec) # NxDxM
+        q = tf.reduce_sum(smIvec * vec, [1]) # NxM
+
         det = tf.matrix_determinant(
             tf.expand_dims(eye(D), 0) + tf.reshape(lengthscales ** -2.0, (1, 1, -1)) * Xcov
         )  # N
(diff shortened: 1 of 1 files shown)


markvdw:
Amazing work, I'll take a look at this in the afternoon.

GPflow/ekernels.py
```diff
@@ -76,17 +76,14 @@ def exKxz(self, Z, Xmu, Xcov):
...
 
-        addvec = tf.matmul(
-            tf.tile(tf.expand_dims(Xsigmc, 1), (1, M, 1, 1)),
-            tf.expand_dims(smIvec, 3),
-            transpose_a=True
-        )[:, :, :, 0] + tf.expand_dims(Xmup, 1)  # NxMxD
+        # Xsigmc: NxD*xD
+        # smIvec: NxD*xM
+        # addvec: NxMxD
+        addvec = tf.einsum('ned,nem->nmd', Xsigmc, smIvec) + tf.expand_dims(Xmup, 1)
```
markvdw: I would avoid using `tf.einsum()`. It's super useful because of its generality, but I don't really understand how it works under the hood. My suspicion is that if you can write things as a `tf.matmul()`, that's much more likely to be fast.

In this case, you can write it as a `tf.matmul()` operation: `tf.matmul(Xsigmc, smIvec, transpose_a=True)`, right?

GPflow/ekernels.py
```diff
@@ -112,11 +109,12 @@ def eKzxKxz(self, Z, Xmu, Xcov):
...
-                     tf.reshape(Z, [1, 1, M, D])) - tf.reshape(Xmu, [N, 1, 1, D])  # NxMxMxD
-        cmr = tf.tile(tf.reshape(cm, [N, 1, 1, D, D]), [1, M, M, 1, 1])  # NxMxMxDxD
-        smI_z = tf.matrix_triangular_solve(cmr, tf.expand_dims(vec, 4))  # NxMxMxDx1
-        fs = tf.reduce_sum(tf.square(smI_z), [3, 4])
+        vec = 0.5 * (tf.reshape(tf.transpose(Z), [1, D, 1, M]) +
+                     tf.reshape(tf.transpose(Z), [1, D, M, 1])) - tf.reshape(Xmu, [N, D, 1, 1])  # NxDxMxM
+        svec = tf.reshape(vec, (N, D, M * M))
+        ssmI_z = tf.matrix_triangular_solve(cm, svec)  # NxDx(M*M)
+        smI_z = tf.reshape(ssmI_z, (N, D, M, M)) # NxDxMxM
+        fs = tf.reduce_sum(tf.square(smI_z), [1]) # NxMxM
```
markvdw: This looks great. Exactly how I hoped it would work out!

markvdw: I can't immediately see how this allocation could be improved.

GPflow/ekernels.py
```diff
@@ -76,17 +76,11 @@ def exKxz(self, Z, Xmu, Xcov):
...
+        vec = tf.expand_dims(tf.transpose(Z), 0) - tf.expand_dims(Xmum, 2)  # NxDxM
+        smIvec = tf.matrix_solve(scalemat, vec)  # NxDxM
+        q = tf.reduce_sum(smIvec * vec, [1])  # NxM
 
-        addvec = tf.matmul(
-            tf.tile(tf.expand_dims(Xsigmc, 1), (1, M, 1, 1)),
-            tf.expand_dims(smIvec, 3),
-            transpose_a=True
-        )[:, :, :, 0] + tf.expand_dims(Xmup, 1)  # NxMxD
+        addvec = tf.matmul(smIvec, Xsigmc, transpose_a=True) + tf.expand_dims(Xmup, 1)  # NxMxD
```
mrksr: It is indeed possible to get away with a single transpose when swapping the two operands. For my toy example, `tf.matmul` and `tf.einsum` both were about equally quick, with `tf.einsum` having a small edge of about 5%.

mrksr: It is indeed possible to get away with a single transpose when swapping the two operands. For my toy example, `tf.matmul` and `tf.einsum` both were about equally quick, with `tf.einsum` having a small edge of about 5%.

markvdw: Interesting that `tf.einsum` is slightly quicker. For now I would argue we should still use `tf.matmul`, since the difference is small, and the one example doesn't give a super clear image  about what happens when matrices are much larger, or what happens on different devices. `tf.einsum` is very new, and I've recently encountered some bugs in it.

I could be convinced otherwise, but I'd have to understand how both ops work and how they're optimised.

jameshensman:
+1 for matmul over einsum.

markvdw:
@mrksr: Does this fix the whole problem you had with the GPLVM being slower on the GPU? If so, I'm happy to merge and thank you very much!

mrksr:
@markvdw since I am currently only using rbf kernels, this pr fixes the issues. The same probably exists for the RBFlinear implementation as well however. If I find the time I can make similar changes there next week. You can merge this nonetheless and I will just create a new PR. Your choice :-)

I agree on the einsum issue and think the new matmul formulation is more elegant here anyway. I was just surprised that einsum was actually a little bit faster and thought I'd share. One small example does of cause not make a trend though.

markvdw:
Ok I'll merge this now. Any more fixes in the rest of the code will be greatly welcomed!

Thanks again!

PR #330, opened 2017-02-02
Mrksr avoid tile in ekernels

markvdw:


jameshensman:
Closing due to 0 diff.

Issue #335, opened 2017-02-13
Docker file


alexggmatthews:
Some people appreciate a docker file for reproducable installation of a package.

alexggmatthews:
Done in #336

PR #336, opened 2017-02-13
Docker file.

alexggmatthews:
Implements #335

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,35 @@
+# Copyright 2016 The GPflow authors
+# 
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+# 
+# http://www.apache.org/licenses/LICENSE-2.0
+# 
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+#USAGE:
+#To run from prebuilt version use:
+#docker run -it -p 8888:8888 gpflow/gpflow
+
+#To replicate build use:
+#docker build -t gpflow/gpflow .
+
+#Uses official Tensorflow docker for cpu only.
+FROM tensorflow/tensorflow
+RUN apt-get -y update
+RUN apt-get -y install git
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+
+#USAGE:
+#To run from prebuilt version use:
+#docker run -it -p 8888:8888 gpflow/gpflow
+
+#To replicate build use:
+#docker build -t gpflow/gpflow .
+
+#Uses official Tensorflow docker for cpu only.
+FROM tensorflow/tensorflow
```
john-bradshaw: You should choose a specific tag otherwise it will default to latest which may not be what you want to do?

alexggmatthews: Seems sensible, thanks. Do you think we should also add use a specific version for our repository, or always assume we are using master? I guess we could do a couple of different images.

Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+RUN apt-get -y update
```
john-bradshaw: you should merge these run steps (\S minimise the number of layers [1])

and add some cleaning after installing to keep the container small (\S apt-get [1]), ie something along the lines of:
```
 && apt-get clean  \
 && rm -rf /var/lib/apt/lists/* \
```

[1]. https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/

Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+RUN apt-get -y install git
+RUN git clone https://github.com/GPflow/GPflow /usr/local/GPflow
+RUN cd /usr/local/GPflow && \
+    python setup.py develop
+#Clear TensorFlow demos and replace them with GPflow examples.
+RUN rm /notebooks/*.*
+RUN rm /notebooks/*
```
john-bradshaw: what does this achieve in addition to line 30?

Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+COPY doc/source/notebooks/*.* /notebooks/
```
john-bradshaw: could you not copy from your cloned repo, rather from a one outside?

Dockerfile
john-bradshaw: yes I think that would be a good idea - match the docker tags with the github tags

alexggmatthews:
@jameshensman we've had a good guest review from @john-bradshaw on this one. I've met all his comments. If you have anything to add let me know. Otherwise lets merge.

jameshensman:
Lovely work @alexggmatthews , and thank for the input @john-bradshaw .

PR #337, opened 2017-02-15
Fresh docs

alexggmatthews:
I've had a tidy up and restructure of the docs. The feedback we were getting is that people had difficulty finding things they were looking for. 

I've added some depth to the doctree in order to make this clearer.

I've made some graphical changes that make the whole thing much more approachable. 

I've fixed some errors and out of date stuff.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,8 @@ python setup.py install
 ```
 You can run the tests with `python setup.py test`.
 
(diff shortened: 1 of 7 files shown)


alexggmatthews:
@jameshensman it would be good to get this one through relatively fast as the paper version depends on it.

Issue #286, opened 2016-11-28
Inference of latent variables of new points in GPLVM


mrksr:
Hi everyone,

I was very happy to see that GPLVMs are finding their way into GPflow, which is very interesting to me. Since I would like to use them in a prediction-setting, I am wondering how they can be used to infer the latent variables of an unseen observed point `Ynew`, analogous to the `infer_newX`-method in [GPy](https://github.com/SheffieldML/GPy/blob/devel/GPy/inference/latent_function_inference/inferenceX.py). Additionally, I would love for this inference to support missing values in this new sample.

Is this already in place in GPflow? If not, what would be the correct way of implementing it? I would be happy to help.

Thank you for all your work!

markvdw:
None of this is currently supported. I'm not sure about missing data, but a bit of a hack for `infer_newX` could be to just add a new observation and initialisation to the `Y`, `X_mean` and `X_var` matrices, and then optimise. I'm not sure what GPy does, but this would technically be the correct thing to do.

alexisboukouvalas:
thanks @mrksr , glad you are finding the GPLVM support interesting!
As @markvdw says we currently do not support either of these features (missing data, inference on X). We have an issue currently ( #274)  with suggestions on how to improve the latent variable model code. 
Any help appreciated!

mrksr:
I think it might be a good idea to start with the inference without missing values. I hope the generalization to missing values will not be too hard then. If I understand the code in GPy correctly, they are essentially cloning the model and setting up the optimization @markvdw described.

Would the same thing be the way to go with GPflow, i.e. cloning the model, fixing all existing parameters and adding parameters for the new points? I think it would be more correct to optimize all parameters and integrate this into the original model, but this would probably take way too much time for every prediction. Plus, it might not really be desirable once there is missing data.

mrksr:
Just a ping to @alexisboukouvalas and @jameshensman, could you maybe give a hint as to how to implement the inference nicely into GPflow?

I think in a more general sense, the problem of which this inference is an instance of is that given some already existing model, one would have to formulate a derived one which shares some parameters (and maybe fixes them) but also adds a few new ones.

jameshensman:
Hmm, making a nice interface might be a bit tricky. 

[...]

Issue #301, opened 2016-12-19
NaNs in BayesianGPLVM likelihood-gradient


mrksr:
While playing around with GPLVM-Models I noticed that sometimes there are NaN-values in the gradients of the likelihood. I have not observed problems with the likelihood itself yet. In search of a small example, I have found the following, which one might argue is a bit of a misuse of GPLVMs, but I think the optimization should still work.

Instead of using the GPLVMs, I tried to "recreate" standard sparse regression on a 1D dataset. I have created some data which can be found on [pastebin](http://pastebin.com/4aEenLKV). The first column. the input, is a Gaussian random walk with some probability of remaining at the same position. The second column, the target, is a (slightly noisy) transform of the input with a linear component and some sine-transformations:
```python
X = gaussian_random_walk()
transform = lambda x: 0.4 * x + 3 * np.sin(x * np.pi * .0025) + 1.5 * np.sin(x * np.pi * 0.1) + 0.8 * np.sin(x * np.pi * 0.3)
Y = transform(X) + some_noise()
```
The only thing special about the data is the fact that there can be multiple observations of a single point in the input space. The problem remains if I add some noise to `X`, thereby forcing them to be different. And since the GPs can model the observations via the noise parameter, this should not be a problem anyway.

SGPR
====
This data can easily be learnt with a standard (sparse) GP regression:
```python
kernel = lambda: GPflow.ekernels.Add((
    GPflow.ekernels.RBF(1),
    GPflow.ekernels.Linear(1),
))

X = data[:, [0]]
Y = data[:, [1]]
num_inducing = 50
Z = X[np.random.permutation(X.shape[0])[:num_inducing]].copy()

m_gpr = GPflow.sgpr.SGPR(
    X,
    Y,
    kernel(),
    Z
)
```
Since the optimization works fine, the gradients are of cause well-defined:
```python
m_gpr._compile()
m_gpr._objective(m_gpr.get_free_state())

Out:
(array([ 1000.95725344]),
 array([ -2.81704226e+02,  -6.40559958e-01,  -6.40560062e-01,
         -6.40560284e-01,  -3.04001075e+00,   1.16295786e+01,
          5.82681802e+01,  -1.18798982e+01,   5.58909405e+00,
          3.17550297e+00,  -8.53468393e+00,   1.57533022e+01,
         -2.78936622e+00,  -5.84108268e+02,  -9.46332675e+00,
         -1.40194934e+00,   5.78258978e+01,   2.63095370e+00,
[...]

PR #341, opened 2017-02-15
GPLVM: infer latent X and partial predictions

javdrher:
Implements #286 based on the suggestions in the issue, and also includes partial predictions. Thanks to @nknudde for figuring out math. 
All suggestions and remarks welcome. 
Some of my own:
- We can provide a notebook like the sgpr notes 
- The inference of X can be pushed up into the GPModel class I think, implementing general inversion of GPs


markvdw:
Wow this looks really great! I'll look at it in more detail soon.

GPflow/gplvm.py
```diff
@@ -111,12 +116,26 @@ def __init__(self, X_mean, X_var, Y, kern, M, Z=None, X_prior_mean=None, X_prior
     def build_likelihood(self):
         """
         Construct a tensorflow function to compute the bound on the marginal
-        likelihood.
+        likelihood for the training data (all dimensions).
+        """
+        return self._build_likelihood_graph(self.X_mean, self.X_var, self.Y, self.X_prior_mean, self.X_prior_var)
+
+    def _build_likelihood_graph(self, X_mean, X_var, Y, X_prior_mean=None, X_prior_var=None):
```
jameshensman: please add a docstring to explain why this function is split out from the main `build_likelihood`. I guess it's because it gets reused at predict time?

GPflow/gplvm.py
```diff
@@ -187,3 +206,228 @@ def build_predict(self, Xnew, full_cov=False):
             shape = tf.stack([1, tf.shape(self.Y)[1]])
             var = tf.tile(tf.expand_dims(var, 1), shape)
         return mean + self.mean_function(Xnew), var
+
+    def build_predict_distribution(self, Xstarmu, Xstarvar):
```
jameshensman: docstring please. Preferably pointing to a notebook derivation :)

jameshensman:
This looks absolutely super.

One suggestion for a test is to use compare the predict_unobserved function with "everyting unobserved" with the plain prediction function.

javdrher:
Thanks for the review! we are creating a notebook now, will add it together with the docstrings.

regarding your test suggestion, predict_unobserved with everything unobserved probably won't work as I don't think it can infer Xmean and Xvar from nothing (perhaps I should add checking on the observed parameter?). I could however compare by calling predict_f/y_uncertain and add a diagonal variance matrix.

javdrher:
Something else we noticed: currently we assume the mean function is zero and don't add anything in build_predict_distribution. For constant mean functions we could add to the mean without issues, but anything higher order is currently troublesome, because we would have to compute the expectations w.r.t the Xmean/Xvar to account for them. These expectations do not seem to be available currently. 

Should we for now assert the mean function to be constant?

markvdw:
Personally I dislike mean functions anyway. Adding a linear covariance function gives you a linear component and gives you Bayesian inference in it for free, hence why we put a lot of effort into getting the kernel expectations for the Linear+RBF kernel.

So I would say that it's fair to assert a zero mean function for now.

javdrher:
Asserts added. In addition, the workaround of data normalization presented #301 makes the idea of having a mean function in GP-LVM a bit awkward, at least with respect to unit variance?

Any further comments?

markvdw:
@javdrher, apologies for the delay in fixing this. We discussed how to move this forward, so we can close this for now. If not, feel free to re-open!

duncanwp:
This seems to have just died, is there any interest in picking it up again? I'm looking for a GPR implementation, which I presume would be much simpler? Cheers

PR #342, opened 2017-02-16
Update to tensorflow 1.0

mrksr:
Hi everyone!

Since tensorflow 1.0 is now available via pip I quickly checked what work would be necessary to update GPflow. After renaming `tf.concat_v2` to `tf.concat`, all tests pass.

Since I do not know whether there are more subtle changes in the API not covered by the tests this PR should of cause be taken with a grain of salt, but I thought it might be a good starting point.


markvdw:
I'm pretty sure that's all that's required. I've been working off tf1.0 for ages now.

jameshensman:
I'm going to recreate this branch within gpflow so that the tests etc can run.

jameshensman:
I merged this via an internal PR.

PR #345, opened 2017-02-17
Update svi_test.ipynb

alessandrousseglioviretta:
To avoid a `TypeError: _optimize_tf() got an unexpected keyword argument 'max_iters'`, `max_iters` needs to be replaced by `maxiter`


alexggmatthews:
We have merged this using #361 thanks very much

PR #356, opened 2017-02-25
Add MLP kernel

mrksr:
This pull request adds an implementation of the MLP kernel as used in [GPy](https://pythonhosted.org/GPy/GPy.kern.src.html#module-GPy.kern.src.mlp), together with a few tests. If there is interest to merge this PR, I can add more information to the docstring or maybe more tests. This code is also not tested very thoroughly yet.

### DIFF
--- a/GPflow/kernels.py
+++ b/GPflow/kernels.py
@@ -483,6 +483,58 @@ class Cosine(Stationary):
         return self.variance * tf.cos(r)
 
 
+class MLP(Kern):
+    """
+    The multi layer perceptron (MLP) kernel
+    """
+    def __init__(self, input_dim, variance=1.0,
+                 weight_variances=1., bias_variance=1.,
+                 active_dims=None, ARD=False):
+        Kern.__init__(self, input_dim, active_dims)
+        self.variance = Param(variance, transforms.positive)
+        self.bias_variance = Param(bias_variance, transforms.positive)
+        if ARD:
+            if weight_variances is None:
+                weight_variances = np.ones(input_dim, np_float_type)
+            else:
+                # accepts float or array:
+                weight_variances = weight_variances * np.ones(input_dim, np_float_type)
+            self.weight_variances = Param(weight_variances, transforms.positive)
+            self.ARD = True
+        else:
+            if weight_variances is None:
+                weight_variances = 1.0
+            self.weight_variances = Param(weight_variances, transforms.positive)
+            self.ARD = False
+
+    def _weighted_product(self, X, X2=None):
+        if X2 is None:
+            return tf.reduce_sum(self.weight_variances * tf.square(X), axis=1) + self.bias_variance
+        else:
+            return tf.matmul((self.weight_variances * X), tf.transpose(X2)) + self.bias_variance
+
+    def K(self, X, X2=None, presliced=False):
+        if not presliced:
+            X, X2 = self._slice(X, X2)
+
+        X_denominator = tf.sqrt(self._weighted_product(X) + 1.)
+        if X2 is None:
+            X2 = X
+            X2_denominator = X_denominator
+        else:
+            X2_denominator = tf.sqrt(self._weighted_product(X2) + 1.)
+
+        enumerator = self._weighted_product(X, X2)
+        return self.variance * (2. / np.pi) * tf.asin(enumerator / X_denominator[:, None] / X2_denominator[None, :])
+
+    def Kdiag(self, X, presliced=False):
+        if not presliced:
+            X, _ = self._slice(X, None)
+
+        X_product = self._weighted_product(X)
+        return self.variance * (2. / np.pi) * tf.asin(X_product / (X_product + 1.))
+
+
 class PeriodicKernel(Kern):
     """
     The periodic kernel. Defined in  Equation (47) of
--- a/GPflow/kernels.py
+++ b/GPflow/kernels.py
@@ -491,14 +491,13 @@ class ArcCosine(Kern):
     implemented_orders = {0, 1, 2}
     def __init__(self, input_dim, variance=1.0,
                  weight_variances=1., bias_variance=1.,
-                 order=0, num_layers=1,
+                 order=0,
                  active_dims=None, ARD=False):
         Kern.__init__(self, input_dim, active_dims)
 
         if order not in self.implemented_orders:
             raise ValueError('Requested kernel order is not implemented.')
         self.order = order
-        self.num_layers = num_layers
 
         self.variance = Param(variance, transforms.positive)
         self.bias_variance = Param(bias_variance, transforms.positive)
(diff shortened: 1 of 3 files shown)


alexggmatthews:
@mrksr Thanks for the PR.

It isn't policy to include everything that appears in GPy as a matter of course. We intend that GPflow remains lean and extensible. As such we consider all new functionality carefully before including it in the core library.

All that said, I think we would be interested in this functionality, assuming it works well. You mentioned it isn't tested very thoroughly yet? Any thoughts about how we could remedy that? 

Thanks,
Alex

alexggmatthews:
Also it is worth us thinking about this paper:

http://cseweb.ucsd.edu/~saul/papers/nips09_kernel.pdf

As I understand it what we have here corresponds to n=0 in that paper. Perhaps we should also consider n=1 give the correspondence to the rectified linear unit? I personally think going further that that would be overkill. 

@jameshensman may have an opinion on this.

mrksr:
Thanks for the feedback, @alexggmatthews.
The reason for me implementing the MLP-kernel for GPflow is that it has proven to be a robust choice as a general purpose kernel with performance comparable to the RBF kernel and it is used quite regularly. The choice of whether or not you want to merge it into GPflow is yours of cause.

Concerning my comment about the code not being tested thoroughly: I am quite confident that the "real" implementation is equivalent to the reference implementation. I think the tests which are part of the PR are adequate at making sure this is the case. I have however not tested the kernel on very large models, so there might be some unexpected problems there.

If you are interested, I'd be happy to implement the n=1 case as well. This should not add too much code.

Bonnevie:
This might be a terrible misunderstanding on my part, but is the MLP kernel really n=0 in the paper? As I understand it n=0 corresponds to the ReLU setting and the standard MLP kernel is not even covered because the Williams-style MLP kernel (as given in e.g. eq 4.29 of Rasmussen and Williams or in the linked GPy doc) is based on sigmoidal units.

mrksr:
The paper mentions the connection between the Williams MLP kernel and the n=0 case just below equation 8. Looking at the math, mainly given by equations 2 and 5, the two seem to different in some constants here and there and the usage of an arccos instead of an arcsin. The two should at least be closely related. The paragraph below equation 7 gives a geometric interpretation of the different kernels.

It might be a good idea to understand this connection more thoroughly before implementing the new case :).

Bonnevie:
Reviewing Rasmussen and Williams, it's stated that the MLP kernel is derived from Gaussian expectations over the error function (used as a sigmoid), so it can't be equal to n=0. However, a sigmoidal function like the error function is basically designed to be similar to the step function, so this could help to explain their similarity. I agree it's an interesting connection! For reference, `arccos(x)=pi/2-arcsin(x)`, so I think the really central difference is the addition of biases in the denominator which means that the quantity won't normalize properly to the cosine of the angle.

jameshensman:
Perhaps a really neat solution would be to implement the Cho/Saul paper
that Alex linked, using their notation. We could also note the equivalence
of the n=0 to the equation given in GPML (in the docstring).

I for one would like to see the Cho/Saul version of the kernel in GPflow,
would make a nice addition and I think it has enough mindshare in the
community to be considered 'core'.

mrksr:
For n=0 and n=1?

jameshensman:
Why not 2 too?

Bonnevie:
@jameshensman: I'd still argue that it's not equivalent to the GPML equation :) But they are certainly very close!
Would you also want to see the multi-layer version?

jameshensman:
If it can be done elegantly, why not?

alexggmatthews:
I personally wonder if the multi-layer version might be overkill? I'm very interested to know what more about it but it has much less of a history than the non-nested family. Is there any paper that uses it apart from the one I added? What do you think @Bonnevie @mrksr ? We're always trying to strike a balance between core functionality and code creep. I don't feel very strongly about, it so if enough other users are keen to see it then I could probably be brought round to @jameshensman 's view.

Bonnevie:
@alexggmatthews since the choice of kernel has such a huge effect on GPs and other kernel machines, I think an argument can be made that a kernel package should strive to include as many kernels as possible, so as to encourage people to go beyond the standard go-to kernels. This certainly conflicts with GPflow's leanness goal, but that might just be another argument for splitting off the kernels into a separate package as has been argued previously.

alexggmatthews:
I guess that is one way of viewing it certainly. Another is to have commonly used kernels and make it straight forward to define your own either by adding a small amount of tf code or by building from other base kernels. Certainly that is closer to what we have at the moment. We could arguably make it easier to incorporate computational graphs from other tensorflow code as has also already been discussed.

alexggmatthews:
In a sense there are infinitely many kernels, of course and finite maintenance resources so obviously a balance has to be drawn somewhere :)

mrksr:
I have added a new commit which implements the ArcCosine-Kernels of orders 0, 1 and 2. I have not yet implemenhkp://pool.sks-keyservers.netted the multi-layer case.
Since some quick checks suggest that tensorflow recognizes terms of the form `some_complicated_tensor ** 0` and optimizes the calculation a constant, I have not added special code for the `n=0`-case.

Comments are welcome :).

Bonnevie:
@mrksr: since

mrksr:
@Bonnevie I guess this was some mistake?

alexggmatthews:
Echoing @mrksr question. @Bonnevie did you actually mean to close this pull request?

GPflow/kernels.py
```diff
@@ -483,6 +483,83 @@ def K(self, X, X2=None, presliced=False):
...
+    """
+
+    implemented_orders = {0, 1, 2}
+    def __init__(self, input_dim, variance=1.0,
+                 weight_variances=1., bias_variance=1.,
+                 order=0,
+                 active_dims=None, ARD=False):
+        Kern.__init__(self, input_dim, active_dims)
+
+        if not order in self.implemented_orders:
```
jameshensman: the "not in" operator would make this a little less ambiguous

GPflow/kernels.py
```diff
@@ -483,6 +483,83 @@ def K(self, X, X2=None, presliced=False):
...
+            self.weight_variances = Param(weight_variances, transforms.positive)
+            self.ARD = False
+
+    def _weighted_product(self, X, X2=None):
+        if X2 is None:
+            return tf.reduce_sum(self.weight_variances * tf.square(X), axis=1) + self.bias_variance
+        else:
+            return tf.matmul((self.weight_variances * X), tf.transpose(X2)) + self.bias_variance
+
+    def _J(self, theta):
```
jameshensman: could you add a note to reference the equation in the paper, please?

GPflow/kernels.py
```diff
@@ -483,6 +483,83 @@ def K(self, X, X2=None, presliced=False):
         return self.variance * tf.cos(r)
 
 
+class ArcCosine(Kern):
+    """
+    The arc cosine family of kernels
```
jameshensman: please add a docstring with a bibtex entry for the main paper.

GPflow/kernels.py
```diff
@@ -483,6 +483,83 @@ def K(self, X, X2=None, presliced=False):
...
+            X, X2 = self._slice(X, X2)
+
+        X_denominator = tf.sqrt(self._weighted_product(X))
+        if X2 is None:
+            X2 = X
+            X2_denominator = X_denominator
+        else:
+            X2_denominator = tf.sqrt(self._weighted_product(X2))
+
+        enumerator = self._weighted_product(X, X2)
```
jameshensman: enumerator -> numerator :)

testing/test_kerns.py
```diff
@@ -26,6 +26,60 @@ def test_1d(self):
         self.assertTrue(np.allclose(gram_matrix, reference_gram_matrix))
 
 
+class TestArcCosine(unittest.TestCase):
```
jameshensman: :+1:

testing/test_kerns.py
```diff
@@ -26,6 +26,60 @@ def test_1d(self):
...
+    def evalKernelError(self, D, variance, weight_variances,
```
jameshensman: our convention is snake_case for methods

testing/reference.py
```diff
@@ -13,9 +13,36 @@ def referenceRbfKernel( X, lengthScale, signalVariance ):
     return kernel
 
 
+def referenceArcCosineKernel( X, order, weightVariances, biasVariance, signalVariance ):
```
jameshensman: :+1:

jameshensman:
I'm going to assume that @Bonnevie closed this by mistake, reopening

jameshensman:
very nice PR, much appreciated. 

Thanks for keeping to the GPflow style, made the code nice and easy to follow. 

I made a few comments, happy to merge once these are addressed.

testing/test_kerns.py
mrksr: I copied the name from the test below. Should I change it there, too, although it is not strictly part of the PR?

jameshensman:
Okay, leave this for the moment

Bonnevie:
@alexggmatthews @mrksr no, that was definitely a mistake on my part!

mrksr:
Thanks for you comments, @jameshensman. I will add the documentation last when the layers-thing is decided.

This last commit adds a preliminary implementation of the multi-layer kernel. The tests do not pass because of some numerical issues, which seem to surface when `x = y` and (I suspect) `x` small. Before spending time on coming up with a solution, I wanted to get your opinion since the multi-layer-implementation adds considerable complexity (and I think is kind of slow). Do you think it should be part of GPflow? Any thoughts, @jameshensman, @alexggmatthews, @Bonnevie?

jameshensman:
No. If it adds considerable complexity it is best left to a more specialised project.

Bonnevie:
I agree - if it results in a significant slow-down and adds more complexity, it should probably be left out. If the main intended demographic of GPflow is people who want to apply GPs in large-scale settings it seems wonky to have certain choices of kernels that can result in inexplicable performance hits.

alexggmatthews:
@mrksr thanks for your efforts on this PR. In answer to your question I agree with @jameshensman and @bonnevie .

mrksr:
I have reverted the Multi-Layer implementation (but left the commit in for now for future reference). I have also added docstrings, trying to stick with the style in `kernels.py`. I think this should be ready for merge, now.

jameshensman:
@mrksr , thanks so much for this PR, and all the interesting discussion it has generated. 

To merge this, I need the CI to run, for which I need to move to an internal branch. I'll open a new PR and accept that.

Issue #355, opened 2017-02-25
Question about `predict_density` for MultiClass likelihood


john-bradshaw:
Hey,

I had a question about lines 445-451 of GPflow/likelihoods.py, ie 
``` python
    def predict_density(self, Fmu, Fvar, Y):
        if isinstance(self.invlink, RobustMax):
            gh_x, gh_w = hermgauss(self.num_gauss_hermite_points)
            p = self.invlink.prob_is_largest(Y, Fmu, Fvar, gh_x, gh_w)
            return p * (1. - self.invlink.epsilon) + (1. - p) * self.invlink._eps_K1
        else:
            raise NotImplementedError
```

So as I understand it this function will be called by the GPmodel class's predict_density method, which states that it returns log probabilities. I just wanted to check that this was the case and that if so that the corresponding method of the MultiClass likelihood class did log the probabilities?
(Also as an aside this method does not affect the training though I think?)

I'm asking as I'm getting odd looking density values and also cannot work out why the `variational_expectations` method ends with `return p * np.log(1 - self.invlink.epsilon) + (1. - p) * np.log(self.invlink._eps_K1)` whereas the predict_density method ends with an almost identical statement but missing the logs: `return p * (1. - self.invlink.epsilon) + (1. - p) * self.invlink._eps_K1`.

Thanks for your help!

alexggmatthews:
Good spot. That is a bug. Would you like to try fixing it? We should check why this isn't picked up in the unit tests also.

alexggmatthews:
@jameshensman I'm surprised this hasn't caused trouble sooner?

jameshensman:
Me too. Let's make sure there's a regression test. 

How about changing predict_density for predict_log_density, for clarity? With suitable deprecation warnings of course

john-bradshaw:
So I think the `testSymmetric` method of `TestRobustMaxMulticlass` was incorrect, which was why it was passing. Some of the test methods of the `TestPredictConditional` class were also indirectly testing this method too through their call of `predict_mean_and_var`, however, I assume as the variance for the f values for these calls were 0 that the p calculated would be either almost exactly 1.0 or 0.0. And then
exp(1 * ln(1-10^-3) + 0* ...) = 1* (1 - 10^-3) + 0 * ...
(although I haven't checked for sure this is actually what is happening.)

I've created a pull request:
https://github.com/GPflow/GPflow/pull/359

[...]

PR #359, opened 2017-02-28
correcting the multiclass predict density so that it gave a log density.

john-bradshaw:
* fixing the expression in the Multiclass likelihood for predict_density (when using RobustMax).
* adjusting the predict mean and variance function of multiclass to deal with predict density giving log probabilities
* fixing the corresponding test for this to correct values,
* adding another very simple test to check that it computes the log probs.


john-bradshaw:
fix for issue: https://github.com/GPflow/GPflow/issues/355

GPflow/likelihoods.py
```diff
@@ -436,7 +436,7 @@ def predict_mean_and_var(self, Fmu, Fvar):
             # To compute this, we'll compute the density for each possible output
             possible_outputs = [tf.fill(tf.stack([tf.shape(Fmu)[0], 1]), np.array(i, dtype=np.int64)) for i in
                                 range(self.num_classes)]
-            ps = [self.predict_density(Fmu, Fvar, po) for po in possible_outputs]
+            ps = [tf.exp(self.predict_density(Fmu, Fvar, po)) for po in possible_outputs]
```
john-bradshaw: so as the predict density class is now creating correct log probabilities I turn them back into normal probabilities here: do you think this is okay or is there a more stable way to do this?

testing/test_likelihoods.py
```diff
@@ -181,14 +181,45 @@ def testSymmetric(self):
             pred = tf.Session().run(l.predict_density(F, F, Y), feed_dict={x: l.get_free_state(), F: F_data})
```
john-bradshaw: note how above that line 176 there is the `l.invlink.epsilon = epsilon`. This is fine as epsilon are the same before and after. However, if epsilon does change then `_eps_K1` will not change to suit. I don't know if you want me to make _eps_K1 a property therefore, or add a setter to epsilon?

testing/test_likelihoods.py
```diff
@@ -181,14 +181,45 @@ def testSymmetric(self):
...
             variational_expectations = tf.Session().run(l.variational_expectations(F, F, Y),
                                                         feed_dict={x: l.get_free_state(), F: F_data})
-        self.assertTrue(np.allclose(mu, np.ones((nPoints, nClasses))/nClasses, tolerance, tolerance))
-        self.assertTrue(np.allclose(pred, np.ones((nPoints, 1))/nClasses, tolerance, tolerance))
+        expected_mu = np.exp(1./nClasses * np.log(1. - epsilon) + (1. - 1./nClasses) * np.log(epsilon / (nClasses - 1)))
```
john-bradshaw: I think mu now is very low in this case 0.031607..., which seems very low when the f for both points are equal, I assume it is just a characteristic of the RobustMax that it is like this though?

testing/test_likelihoods.py
```diff
@@ -181,14 +181,47 @@ def testSymmetric(self):
             pred = tf.Session().run(l.predict_density(F, F, Y), feed_dict={x: l.get_free_state(), F: F_data})
             variational_expectations = tf.Session().run(l.variational_expectations(F, F, Y),
                                                         feed_dict={x: l.get_free_state(), F: F_data})
-        self.assertTrue(np.allclose(mu, np.ones((nPoints, nClasses))/nClasses, tolerance, tolerance))
-        self.assertTrue(np.allclose(pred, np.ones((nPoints, 1))/nClasses, tolerance, tolerance))
+        expected_mu = (1./nClasses * (1. - epsilon) + (1. - 1./nClasses) * epsilon / (nClasses - 1)) * np.ones((nPoints, 1))
+        self.assertTrue(np.allclose(mu, expected_mu, tolerance, tolerance))
+        expected_log_denisty = np.log(expected_mu)
+        self.assertTrue(np.allclose(pred, expected_log_denisty, 1e-3, 1e-3))
```
john-bradshaw: had to reduce the tol to get this to work -- is this okay?

markvdw:
I usually increase the number of Gauss-Hermite points.

GPflow/likelihoods.py
```diff
@@ -446,7 +446,7 @@ def predict_density(self, Fmu, Fvar, Y):
         if isinstance(self.invlink, RobustMax):
             gh_x, gh_w = hermgauss(self.num_gauss_hermite_points)
             p = self.invlink.prob_is_largest(Y, Fmu, Fvar, gh_x, gh_w)
-            return p * (1. - self.invlink.epsilon) + (1. - p) * self.invlink._eps_K1
+            return p * np.log1p(- self.invlink.epsilon) + (1. - p) * np.log(self.invlink._eps_K1)
```
jameshensman: I'm not sure this is right. I think maybe it should be 
```python
tf.log(p * (1. - self.invlink.epsilon) + (1. - p) * self.invlink._eps_K1)
```
i.e. with the log outside the expectation. Happy to be corrected if wrong.

john-bradshaw: yes, I think I was wrong,  on my second commit (2a1892e) I have fixed this. sorry.

GPflow/likelihoods.py
jameshensman: This should be fine. In fact, since the last operation inside predict is now log, and we immediately call exp on it, tensorflow should remove both ops using its graph reduction algorithm. I wonder if it does?

testing/test_likelihoods.py
jameshensman: making eps_K1 a properts seems like an elegant solution

testing/test_likelihoods.py
jameshensman: yes, fine. 

It's always going to be tricky to get fine tolerances when the numbers we're targetting are [0, 1].

jameshensman:
@john-bradshaw this is nice, did you want to make any other changes, or ready to merge?

GPflow/likelihoods.py
john-bradshaw: Not really too relevant as we no longer do this, but I'm not too sure whether exp(lnA) is simplified to A. The other way round should occur I think (see line 325 of https://github.com/tensorflow/tensorflow/blob/master/tensorflow/compiler/xla/service/algebraic_simplifier.cc), but I could be wrong.

 I have found it difficult to ever find any documentation for the simplifications Tensorflow does and would be grateful if you had a reference.

john-bradshaw:
Thanks!, no no other changes. I can change `eps_K1` to a property in a separate PR.

jameshensman:
@john-bradshaw I'm going to merge this via an internal GPflow branch -- just so that all the CI can run. In future, you can make prs by pushing to branches inside the GPflow repo, and they will be tested/covered automatically.

Issue #349, opened 2017-02-17
Bug in multi-parental name referencing


tfwnicholson:
I encountered a bug in the naming logic. In the following, quite reasonable, code the third line will throw a `ValueError('This Param appears to be doubly referenced by a parent')`:
```
constant_mean = Constant(1)
additive_mean = constant_mean + constant_mean
constant_mean_name = constant_mean.name
```
The `additive_mean` clearly contains two references to the `constant_mean`, but this seems to me that it should be valid. 

I see a few possible solutions:
 + allowing multi parental references (why was this disallowed in the first place?)
 + detecting this case in the constructor and raising there
 + detecting this case in the constructor and replacing the additive mean with a: `Constant(2) * constant_mean`

Which method would you prefer? I'll prepare a PR with it.

jameshensman:
I think the desired behaviour is that we shouldn't allow multiple references to parameters, so the code you posted should raise an exception. 

To fix we should examine parameters (and parameterizeds) being assigned to the parameterzed class, and reject them if we already have a reference to them. In fact, we should reject them if there is a reference to them anywhere on the tree.

 This should result in the above code raising an exception in the constructor of `Add()`, which is called when the __add__ function is called using the `+` operator.

alexggmatthews:
Fixed in #364 #370

PR #364, opened 2017-03-03
Preventing multiple params in tree

tfwnicholson:
This addresses the issue of multiple parameters being in the tree (#349) which was causing ```longname``` to crash. It was decided with @jameshensman that rather than delaying until the construction of the name to detect and maybe recover from multiple parameters in the tree, that we should detect and abort as soon as someone tries to add the same param to the same tree.

This PR therefore adds the invariant to a tree that a ```Param```/ ```Parameterized``` may only appear that a single point.

Cheers.

### DIFF
--- a/GPflow/model.py
+++ b/GPflow/model.py
@@ -348,9 +348,9 @@ class GPModel(Model):
     """
 
     def __init__(self, X, Y, kern, likelihood, mean_function, name='model'):
+        Model.__init__(self, name)
         self.kern, self.likelihood, self.mean_function = \
             kern, likelihood, mean_function
-        Model.__init__(self, name)
 
         if isinstance(X, np.ndarray):
             #: X is a data matrix; each row represents one instance
--- a/testing/test_param.py
+++ b/testing/test_param.py
@@ -369,7 +369,7 @@ class SingleParamterizedInvariantTest(unittest.TestCase):
         m1.foo = GPflow.param.Parameterized()
 
         m2 = GPflow.param.Parameterized()
-        m2.foo = GPflow.param.Parameterized()
+        m2.foo = m1.foo
 
 
 class SingleParamInvariantTest(unittest.TestCase):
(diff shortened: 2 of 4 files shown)


GPflow/model.py
```diff
@@ -348,9 +348,9 @@ class GPModel(Model):
     """
 
     def __init__(self, X, Y, kern, likelihood, mean_function, name='model'):
+        Model.__init__(self, name)
         self.kern, self.likelihood, self.mean_function = \
             kern, likelihood, mean_function
-        Model.__init__(self, name)
```
jameshensman: Good catch

testing/test_param.py
```diff
@@ -318,6 +311,120 @@ def testTFMode(self):
...
+
+    def testAddingToAnother(self):
+        """
+        Adding the same Paramterized object to another tree is fine.
+        """
+        m1 = GPflow.param.Parameterized()
+        m1.foo = GPflow.param.Parameterized()
+
+        m2 = GPflow.param.Parameterized()
+        m2.foo = GPflow.param.Parameterized()
```
jameshensman: Did you mean `m2.foo = m1.foo` ?

testing/test_param.py
```diff
@@ -318,6 +311,120 @@ def testTFMode(self):
...
+
+    def testAddingToAnother(self):
+        """
+        Adding the same Param object to another tree is fine.
+        """
+        m1 = GPflow.param.Parameterized()
+        m1.foo = GPflow.param.Param(1)
+
+        m2 = GPflow.param.Parameterized()
+        m2.foo = GPflow.param.Param(1)
```
jameshensman: again, I think you meant `m2.foo = m1.foo`.

jameshensman:
The main bit of code looks good. I have added a few queries for the tests though.

jameshensman:
looks good to me, @alexggmatthews please have a look at merging this. Thanks.

alexggmatthews:
Hi all, happy to look into this. It will have to be Monday morning now though I'm afraid. Have a good weekend,
Alex

alexggmatthews:
I've moved this to #370 to allow the Continuous Integration to run.

alexggmatthews:
Merged in #370

PR #365, opened 2017-03-03
Updates to multiclass likelihood.

jameshensman:


jameshensman:
much appreciated @john-bradshaw

PR #366, opened 2017-03-03
making ekern_tests deterministic

jameshensman:


jameshensman:
These changes attach a separate random number generator to each Test class during setUp. 
Previously, a global (to the file) random number generator was used, with a fixed seed. This was non-deterministic because the order that the tests are run is not guaranteed, so differnt 'parts' of the generator could be used by different tests.

jameshensman:
I'm still looking at this. 

At least we have deterministic failures now, which I think is better than nondeterminism.

markvdw:
@tfwnicholson Yes that's the case. However, if you increase the number of Gauss-Hermite points, the tests will pass again, at the expense of the test taking longer.

jameshensman:
Hmm. I tried increasing the number of GH points to 50, but the test still failed. 

However, switching the random seed to 1 makes them work (!)

markvdw:
For some cases I needed quite a lot of GH points, up to 100 or more in some cases.

Also, thanks for the numerical optimisations as well!

jameshensman:
So, after digging through the code quite a bit, I agree with Mark. 

The test works perfectly for the Linear kernel. But for the RBF kernel the quadrature is not very effective, because the function we're trying to integrate (in eKxz) is very much not a polynomial. 

Doing as @markvdw suggested and increasing the number of quadrature points does make the error go down, but with seed 0 you need 100 quadrature points for the tests to pass (!). On my laptop that's not slow, so I'm pretty happy to leave it at that number. 

@markvdw , I think we discussed at the time of writing that the important thing was that the implementation was correct, rather than efficient. I still agree with this. But as part of my code-digging, I re-jiggled eKxz into the Cholesky form, which should make for a nicer compute graph: we now have only one trick op (cholesky) whereas before we had separate calls to `det` and `solve`.

markvdw:
Correct and efficient > only correct :)

Thanks!

jameshensman:
Great, let me just jiggle the test back into working order and would be great if you could merge.

Issue #209, opened 2016-09-13
Ability to pass through tf config proto to Session 


alexggmatthews:
It would be useful to be able to use the following tf functionality for timed experiments.

```
config = tf.ConfigProto(intra_op_parallelism_threads=1,inter_op_parallelism_threads=1)
self._session = tf.Session(config=config)
```

This limits the number of cpu threads for testing. Overwriting the model._session at a later point doesn't seem to work.

jameshensman:
Does it if work if you run _compile after replacing the session?

alexggmatthews:
Oddly no it doesn't. I've tried a couple of times. I'm not quite sure why tbh.

alexggmatthews:
I wondered if it might be some additional state that we aren't getting to when we recompile.

I notice for instance that AutoFlow creates a different session in param.py

`storage['session'] = tf.Session()`

jameshensman:
Ok, that's strange. 

I wonder if we should also be separating out our graphs in tensorflow? Then we could ensure that the autoflow and _compile session are working with different graphs. At the moment, I think everything gets done on the default graph, which is maybe why we have issues.

Tensorflow graphs are a nuance that I didn't understand when writing GPflow to start with, and I'm still not sure I understand them.

alexggmatthews:
A fair observation. I don't personally think the aspect is very well documented.

alexggmatthews:
This branch has a config proto passed through SVGP, GPmodel and Model to allow us to control the number of CPU threads when doing profiling experiments. Obviously we would need to do it for the other model daughter classes to be a full solution. It isn't actually that intrusive and allows the ability to do things like logging device placements.

https://github.com/GPflow/GPflow/tree/profiling_mods

alexggmatthews:
Actually there is also some suggestion that it might be possible to set tracing level for the whole session rather than individual run calls:

https://github.com/tensorflow/tensorflow/blob/r0.10/tensorflow/core/protobuf/config.proto

Might be worth looking into as it might make profiling less intrusive in our code base than the current plan for #174

alexggmatthews:
Unfortunately I don't think this will help with the profiling aspect after all.

kiudee:
I would like to add that setting the number of threads is also mandatory in a cluster computing setting where resource limits are enforced.

markvdw:
[...]

PR #371, opened 2017-03-07
session configuration settings

alexisboukouvalas:
This hopefully address #209
As @kiudee pointed out in a cluster setting the tensorflow session may need to adhere to a limit on the number of cores used. 
I have attempted here to give a general solution where any config.proto setting can be passed to both model and autoflow sessions.
@markvdw @jameshensman are you happy with this solution?

Any ideas how to test autoflow session? It's not stored anywhere.

If you are happy with this, I am happy to add documentation for settings.session
Also note inter_op_parallelism_threads == '0 means the system picks an appropriate number.' [see here](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/protobuf/config.proto) so I think 0 is a reasonable default.

### DIFF
--- a/GPflow/ekernels.py
+++ b/GPflow/ekernels.py
@@ -4,6 +4,7 @@ import tensorflow as tf
 from . import kernels
 from .tf_wraps import eye
 from ._settings import settings
+from .param import AutoFlow
 
 from .quadrature import mvhermgauss
 from numpy import pi as nppi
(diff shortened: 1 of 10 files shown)


markvdw:
This is great, exactly what I've been needing for a while as well.

The AutoFlow session can be accessed through a property with the name`storage_name = '_' + tf_method.__name__ + '_AF_storage'`, as defined in the `AutoFlow` decorator definition.

The only thing I would add, is to allow the user to override the config field in `kwargs`. I.e. the settings from `gpflowrc` should only be applied if no `config` option is passed to `get_session()`. My interpretation is that the priority should be defaults < settings file < overriding in the code.

alexggmatthews:
I agree with Mark re settings file otherwise looks good.

alexisboukouvalas:
I have added
1.  Caller of session.get_session can override config settings
1. Added tests for override behaviour and autoflow.
1. Added documentation in the settings notebook.

Thank you @alexggmatthews and @markvdw  for the review. 
If you are happy with the changes could you kindly merge them?

When testing the documentation changes I noticed the code was broken (missing import numpy as np).
Do you think it would be possible to run all the notebook in the documentation as part of the continuous integration tests? If so, I can open an issue.

markvdw:
@alexisboukouvalas: Great, thanks for putting that in!

alexisboukouvalas:
Should we close issue #209 ?

jameshensman:
Please open an issue to run all notebook code in tests. Perhaps with ability to blacklist heavy notebooks.

Issue #367, opened 2017-03-06
optimize_restarts and randomize


Bonnevie:
These two utility functions from GPy (possibly inherited from paramz) would be quite handy in GPflow, or is there a particular reason they were excluded? In short, `randomize` picks random hyperparameters, and `optimize_restarts` runs several optimization runs, keeping the best.

I am not sure what the preferred randomization strategy is. If there's a prior, the randomization should be with respect to that distribution, but if not it might make the most sense to sample from broad gaussians and map through each parameter's transformation to get the proper constraints.

`optimize_restarts` should be pretty straightforward to implement by just iterating the standard optimization routines, but I don't know if more can be done to exploit parallelism explicitly. 

edit: the `sample` method (which uses HMC) could also potentially be used, although I'm thinking there are situations where it's more appropriate to sample from your prior rather than your posterior.

markvdw:
I think these things are best left out of the core GPflow package. There are so many different ways of doing this, that each implemented method is likely to only be useful to a subset of users, while increasing complexity considerably.

Personally, my suggestion would be to have a GPflow_tools (or similar) package, that provided things like this. This would also allow much more sophisticated things to be done, like Bayesian optimisation. I would really like to see some form of "freeze-thaw" Bayesian optimisation on random restarts. It's implemented in https://github.com/HIPS/Spearmint.

jameshensman:
I agree with @markvdw re the optimize-restarts method. 

When it comes to setting parameters to prior draws, I actually think that's a pretty re-usable feature that bigger-scale optimizers could use, including a freeze-thaw scheme, or multiple runs of MCMC, or similar. I think we should consider that as a feature if it can be done unobtrusively, which I think it can.

Bonnevie:
@markvdw: a GPflow_tools package seems like a good idea, as it would allow you to keep the core GPflow package compact.

@jameshensman: I can try and implement it. What default randomization scheme do you think is appropriate and where should the method go? Is it better off as a `Param` method or as a separate utility function?

If a randomization scheme is implemented, wouldn't it be quite unobtrusive to have a `restarts` keyword for the `optimize` method with the default randomization?

markvdw:
@Bonnevie: I made some tools for tracking optimisation progress in GPflow https://github.com/markvdw/opt_tools/, if you're interested. It could be a start for some more GPflow tools.

Bonnevie:
[...]

PR #372, opened 2017-03-08
Randomization of parameters

Bonnevie:
As per issue #367, this pull request implements randomization functionality for parameters. The code is minimal, with parameters now being able to randomly reset themselves and models being able to randomize all children. Priors have been modified to have a sample method.   

If possible it samples from priors, otherwise it defaults to a weak but universal alternative where it samples a free unconstrained state from a standard normal and then propagates it through the forward transform to obey constraints.  

The implemented tests verify that parameters are modified, that it obeys "fixed" flags, that priors supersede default behavior, that all priors work, and that randomization can be passed down through the model hierarchy appropriately.

edit: codeship appears to be throwing an error, but I cannot view the details of it without admin privileges?


jameshensman:
This doesn't appear to work for params that have size greater than 1. 


```python
In [1]: m = GPflow.param.Parameterized()

In [2]: m.p = GPflow.param.Param(np.random.randn(3))

In [3]: print(m)
unnamed.p transform:(none) prior:None
[ 0.40870328 -0.25011942  0.03627796]

In [4]: m.randomize()

In [5]: print(m)
unnamed.p transform:(none) prior:None
0.7736654332138937
```

Bonnevie:
That's a pretty silly oversight on my part... 
What's the assumption for priors on multivariate quantities? i.i.d.?

jameshensman:
yes.

Bonnevie:
@jameshensman: got it fixed now.

GPflow/param.py
```diff
@@ -284,6 +284,30 @@ def set_state(self, x):
...
+                Defaults to True.
+        """
+        if not (skipfixed and self.fixed):
+            if self in distributions.keys():
+                self._array = distributions[self].sample(self.shape)
+            else:
+                try:
+                    self._array = self.prior.sample(self.shape)
+                except AttributeError:
+                    randn = np.array(np.random.randn(*self.shape))
```
jameshensman: Some transforms alter the size of their argument. I think you need `self.transform.free_state_size(self.shape)` to get the correct number of randns here.

GPflow/priors.py
```diff
@@ -75,6 +88,9 @@ def __init__(self, shape, scale):
     def logp(self, x):
         return tf.reduce_sum(densities.gamma(self.shape, self.scale, x))
 
+    def sample(self, shape=()):
+        return np.array(np.random.gamma(self.shape, self.scale))
```
jameshensman: shape arg is not used.

GPflow/priors.py
```diff
@@ -88,6 +104,10 @@ def __init__(self, mu, sigma):
     def logp(self, x):
         return tf.reduce_sum(densities.laplace(self.mu, self.sigma, x))
 
+    def sample(self, shape=()):
+        return np.array(np.random.laplace(self.mu, self.sigma,
+                                          size=shape if shape else None))
```
jameshensman: Why wrap this in np.array? wouldn't it be neater to have the shape default to (1,) ?

testing/test_param.py
```diff
@@ -599,6 +599,107 @@ def test(self):
...
+        m.p2 = GPflow.param.Param(1.0)
+        m.m = GPflow.model.Model()
+        m.m.p = GPflow.param.Param(1.0)
+        m.m.p2 = GPflow.param.Param(1.0)
+
+        m.p2.prior = GPflow.priors.Gaussian(0,1)
+        m.m.p2.prior = GPflow.priors.Gaussian(0,1)
+        m.randomize()
+
+        self.assertTrue(m.p.value != 1.0)
```
jameshensman: I think `self.assertFalse(m.p.value == 1.0)` is cleaner :)

jameshensman:
Hi @Bonnevie , 

This is pretty good!  I like that you've used the style used in the rest of GPflow, e.g. the abstract base classes, and you've made good use of the tree structure.  Thanks for adding some pretty comprehensive tests too. 

I added a few comments to address some specific things. Please also add a test that covers the shaping that I pointed out in the comment above (i.e. a test that would have failed in the previous version of the code).

Bonnevie:
@jameshensman: sorry for the weekend silence. I did add a few new lines to the tests to take multi-dimensional parameters into account, but I can see from review that I forgot an assert. Will look at your other comments :)

GPflow/priors.py
Bonnevie: I believe (1,) gives you a 1-d vector with one element which is ever so slightly different from a numpy scalar (e.g. you can index the latter).
I wanted scalars to be default, but I also wanted everything to be Numpy for consistency.

Bonnevie: Checked what the default is for scalar Parameters in GPflow, and that's actually (1,) so I'll modify it to fit that standard.

Bonnevie:
@jameshensman: I think I covered all of your comments. Dropped the np.array wrapping and changed the default shape.

GPflow/param.py
```diff
@@ -305,7 +305,8 @@ def randomize(self, distributions={}, skipfixed=True):
                 try:
                     self._array = self.prior.sample(self.shape)
                 except AttributeError:
-                    randn = np.array(np.random.randn(*self.shape))
+                    randn = np.array(np.random.randn(
```
jameshensman: I don't think we need `np.array(...)` here.

jameshensman:
One teeny extra `array(...)` that we dont need :)

Bonnevie:
ah, yeah, that shouldn't matter anymore either. Removed it!

Bonnevie:
all of a sudden I'm getting some peculiar errors with respect to "referenceRbfKernel" and stuff like that.

jameshensman:
The failing tests were a problem with tokens on coveralls, which is an alternative coverage site to codecov. Not your code :)

I think this is ready to merge.

thj2009:
Hi, do you have an example or test to explain how to use the restarting optimization function?
Thank you very much.

Issue #373, opened 2017-03-09
Run notebooks in tests


alexisboukouvalas:
Run all notebook code in tests. Perhaps with ability to blacklist heavy notebooks.
This issue based on discussion in #371 

What is a good way to accomplish this?

jameshensman:
http://blog.thedataincubator.com/2016/06/testing-jupyter-notebooks/

jameshensman:
this is done, thanks to you @alexisboukouvalas

PR #374, opened 2017-03-09
Testing notebooks

alexisboukouvalas:
One approach to test notebooks as proposed in #373 
Rather than having an explicit blacklist, the interrupt_on_timeout option allows 
"If execution of a cell times out, interrupt the kernel and continue executing other cells rather than throwing an error and stopping." [see docs](http://nbconvert.readthedocs.io/en/latest/config_options.html)

Tests will already fail with a python 3 error in
```
	Error executing the notebook "../doc/source/notebooks/FITCvsVFE.ipynb".
	SyntaxError: Missing parentheses in call to 'print' (FITCvsVFE.py, line 78)
```

I would appreciate your guidance on this and alternative approaches.


jameshensman:
I wish there were a way to thumbs up these commits! 

This is super work Alexis, I'm really looking forward to having our notebooks tested. Even if it's just running the code, without checking the results, I think it will help us catch bugs. It can be embarrassing when a user says that the example in the docs doesn't work!

alexisboukouvalas:
thanks @jameshensman  appreciate the thumbs up! I am hoping to get the tests passing today.

alexisboukouvalas:
How can I access the continuous-integration/codeship log? Codeship says I don't have access - thanks.
(I suspect it's a python 2.7 failure).

alexisboukouvalas:
Thanks @jameshensman all tests now pass. Please review and merge if you are happy with the changes.

jameshensman:
This looks absolutely super!

My guess is that the notebooks don't get covered by the coverage tools, but it's great to know that notebooks added in the future will be checked against the code. And of course that changes in the code aren't breaking the notebooks.

PR #376, opened 2017-03-10
removing references to tf_wraps.eye

jameshensman:
TensorFlow now has a `eye()` op, which means we can remove our implementation from `tf_wraps.py`, as was always the intention.


alexggmatthews:
@tfwnicholson did you want to look at this one? If not I am happy to.

tfwnicholson:
I'll look at this tnow

GPflow/ekernels.py
```diff
@@ -61,7 +59,7 @@ def exKxz(self, Z, Xmu, Xcov):
         ]):
             Xmu = tf.identity(Xmu)
 
-        M = tf.shape(Z)[0]
+        # M = tf.shape(Z)[0]
```
tfwnicholson: Shall we delete this as it's not needed

GPflow/tf_wraps.py
```diff
@@ -20,13 +20,16 @@
...
+import warnings
 from ._settings import settings
+import numpy as np
 
 
 def eye(N):
     """
     An identitiy matrix
     """
+    warnings.warn('tf_wraps.eye is deprecated: use tf.eye instead', np.VisibleDeprecationWarning)
```
tfwnicholson: So we don't just want to take it out and let people change their code? Typically before version 1.0 code bases reserve the right to make API-breaking changes. I know it's only a tiny piece of code, but it might be good to not let GPflow suffer from deprecation-bloat.

Bonnevie:
did anybody figure out why this doesn't pass codeship? Wondering whether it's the same issue that's plaguing my randomization PR.

GPflow/tf_wraps.py
jameshensman: there's precedent for having deprecated function warnings in GPflow. We'll leave them in for a minor version and then remove them.

GPflow/ekernels.py
jameshensman: done

markvdw:
It seems to be some error involving the repo token... Not sure what causes it, but it's not your code :)

jameshensman:
The intermittent failure on cadetship appears to have been caused by coveralls. I'm not sure anyone is using coveralls (codecov has more features, including posting coverage reports to these PRs), so I disabled coveralls on codeship. Note that we have never required coveralls to merge a PR, relying on the more mature codecov.

jameshensman:
@tfwnicholson I assume answered all your comments, okay to merge this?

tfwnicholson:
@jameshensman yep, looks good

PR #380, opened 2017-03-10
Allow naming of GPR

tfwnicholson:


jameshensman:
```
======================================================================
FAIL: test_likelihood_name (testing.test_param.TestScopes)
----------------------------------------------------------------------
Traceback (most recent call last):
File "/home/rof/src/github.com/GPflow/GPflow/testing/test_param.py", line 616, in test_likelihood_name
self.assertTrue('model.build_likelihood' in l.name)
AssertionError: False is not true
```

jameshensman:
Edit looks good, but breaks a test :)

To use the CI we have here at GPflow, you'll need to push to a branch inside the repo (you have that privilege).

PR #385, opened 2017-03-15
Remove pandas hard dependency

befelix:
I would propose to remove pandas as a requirement for the installation, since it's really only required by two export functions. Imho it's better to have a lean, minimal installation.

The corresponding tests are skipped if pandas is not installed.


Bonnevie:
As a counterpoint, I'd like to note that we also use Pandas to annotate a matrix of parameter samples correctly with `get_samples_df`. Without annotated columns, it's almost impossible to parse an output from `sample`. If we really want to ditch Pandas as a dependency we need to find an alternative to this. It could be returning a dictionary of sample vectors, or a method to return the appropriate column names so that a pandas frame could easily be constructed.

befelix:
I'm not arguing for removing pandas - it's a great library for many purposes. I just think it's a bit annoying to get another 70 odd MB of dependencies for such a limited purpose. Having an alternative method that returns a dictionary in addition to this patch sounds reasonable to me.

markvdw:
I think if those who would really care about the 70 MB of Pandas, are also going to be able to edit `setup.py` or even maintain a separate branch with the small edits to prevent it from being loaded.

Importing pandas inside the functions themselves is a good compromise though. However I would argue letting the tests fail if pandas isn't installed, so users are aware they're missing a feature.

@befelix: Could you elaborate on why this 70 MB is so detrimental for you? On most networks it should only take a few extra seconds to download, and 70 MB on a hard drive is peanuts these days.

befelix:
It's not super important - I just like to keep my docker images small. Tbh
this PR is more or less exactly the branch that you described.
I think importing inside the function and keeping it in the setup.py could
be a reasonable compromise too. I think the important bit is that tests
pass with the default installation, so as long as pandas is in the setup.py
the skip is unecessary.

jameshensman:
What about having pandas in extras-require?

befelix:
Sounds good to me, I've adapted the PR

markvdw:
@befelix, sorry for the wait, I'm merging this in an in-house PR (#486) to ensure the tests run.

PR #394, opened 2017-04-03
Fix randomize for multidimensional Params without priors

mrksr:
Hi everyone!

While trying out the convenient new randomization-feature contributed by @Bonnevie, I found a small bug: Multidimensional Params are flattened when randomized without a prior. The with-prior-case seems to work fine.
An example: The Param `x = Param(np.ones(size=(3, 3)))` with shape `x.shape == (3, 3)` would be of shape `x.shape == (9,)` after calling `x.randomize()`.

This pull request seems to fix the issue and makes the relevant test more specific in order to catch the problem. Care to have a look, @Bonnevie?

### DIFF
--- a/GPflow/param.py
+++ b/GPflow/param.py
@@ -307,7 +307,7 @@ class Param(Parentable):
                 except AttributeError:
                     randn = np.random.randn(
                         self.transform.free_state_size(self.shape))
-                    self._array = self.transform.forward(randn)
+                    self._array = self.transform.forward(randn).reshape(self.shape)
 
     def build_prior(self):
         """
--- a/testing/test_param.py
+++ b/testing/test_param.py
@@ -629,7 +629,8 @@ class TestRandomizeDefault(unittest.TestCase):
         m.pp = GPflow.param.Param(1.0, GPflow.transforms.Log1pe())
         m.pf = GPflow.param.Param(1.0)
         m.pf.fixed = True
-        m.pmd = GPflow.param.Param(np.ones(5))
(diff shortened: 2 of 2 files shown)


GPflow/param.py
```diff
@@ -305,8 +305,7 @@ def randomize(self, distributions={}, skipfixed=True):
                 try:
                     self._array = self.prior.sample(self.shape)
                 except AttributeError:
-                    randn = np.random.randn(
-                        self.transform.free_state_size(self.shape))
+                    randn = np.random.normal(size=self.shape)
```
jameshensman: @mrksr , this won't work for transforms that don't preserve size, I think @Bonnevie is right here.

The problem is below, where we need:

`self._array = self.transform.forward(randn).reshape(self.shape)`

mrksr: Good point. I have changed the commit and the new test still passes.

jameshensman:
@mrksr , could you add a test that would have failed with your previous suggestion, please? This implies that we're not covering the case where `randomize()` is called on a parameter with a transform that changes the size. See for example `GPflow.transforms.LowerTriangular`.

mrksr:
I think my understanding of the `shape`-property of a `Param` was backwards (which really only matters for transforms which change the shape). This last commit adds the test you requested and everything seems to work fine.

jameshensman:
Great, thanks. @Bonnevie , does this look okay to you? I think we should get this merged swiftly since it fixes a bug.

PR #395, opened 2017-04-03
Mrksr fix randomize without prior

jameshensman:
internal PR


jameshensman:
Thanks for this @mrksr !

Issue #383, opened 2017-03-14
Keyword Arguments for AutoFlowed Functions


mrksr:
Hello everyone!

The Problem
---
While playing with one of my models, I had a few issues with the (very convenient) AutoFlow, previously discussed in #178, namely, that a AutoFlowed function only takes arguments which are converted to tensorflow-variables. While this is usually enough, it gets a bit tricky when some function interacting with tensorflow has multiple (usually boolean) arguments which slightly alter their behaviour. An example in the current GPflow code can be seen in model.py, where the two functions
```python
    @AutoFlow((float_type, [None, None]))
    def predict_f(self, Xnew):
        return self.build_predict(Xnew)

    @AutoFlow((float_type, [None, None]))
    def predict_f_full_cov(self, Xnew):
        return self.build_predict(Xnew, full_cov=True)
```
only differ in a boolean argument.

This workaround becomes an issue in two cases:

1) A function which takes multiple boolean arguments.
2) A function which takes (say) a constant float as an argument (think quantiles).

In the first case, a rather large number of function is needed to cover all possible boolean combinations and in the second case I am not aware of a workaround other than doing things on the tensorflow side, which might incur a runtime cost.

Wrong solution
---
An obvious thing to try is to just cache the keyword arguments via a wrapper function, like this:
```python
    def predict_f(self, Xnew, **kwargs):
        @AutoFlow((float_type, [None, None]))
        def _predict_f(self, Xnew):
            return self.build_predict(Xnew, **kwargs)

        return _predict_f(self, Xnew)
```
While this seems to work on first glance, it only does so with the first set of keyword arguments, because of AutoFlow's internal caching of the tensorflow graph. Additionally, the name of the internal wrapper function (`_predict_f`) needs to be unique in the class-scope, since otherwise graphs are shared between functions with no shared visibility in the code. In short: this does not work.

My current solution
---
The problem with the wrong solution is the way Autoflow does caching. The map from function names to graphs is done using the name of the function only, see [param.py#L515](https://github.com/GPflow/GPflow/blob/master/GPflow/param.py#L515):
```python
storage_name = '_' + tf_method.__name__ + '_AF_storage'
```
[...]

Issue #178, opened 2016-08-24
Automatic AutoFlow


markvdw:
AutoFlow is great, but currently we require two different functions to compute the result of some function that returns a tensorflow variable, and one that actually constructs the variable. Would it be possible to get AutoFlow to wrap a function that returns a tensorflow variable directly, and perform the computation if `tf_mode()` is false? In this case, don't we know that the model is *not* currently constructing the graph, and so we want to get the result of the computation?

The main issue I see here is that some functions can receive a different number of variables, or variables of a range of ranks, while AutoFlow seems to require the shapes and types beforehand. Are there any more barriers?

I think this would make testing a lot nicer. Currently I'm writing a whole bunch of tests, and I need to explicitly invoke a TensorFlow each time. It would be a lot nicer if I could just evaluate functions directly...

jameshensman:
Hi Mark, 

> some functions can receive a different number of variables

Yeah, autoflow can't really handle optional arguments at the moment

> range of ranks

You don't have to specify the rank in autoflow:
```python
@Autoflow((tf.float64,))
def f(x):
    return x + 2
```

The pattern I've used for this is to use a build/compute pattern. So
```python
class Boop(GPflow.model.Model):
    def build_foo(self, x, bool=False):
        if bool:
            return x + 2
        else:
            return x + 4

    @Autoflow((tf.float64,))
    def compute_foo(self, x)
        return build_foo(x)

    @Autoflow((tf.float64,))
    def compute_foo_True(self, x)
        return build_foo(x, True)
```

It's not perfect, but it gets the job done. You'll find the pattern in `build_predict/predict_f/predict_y`. 

It's perhaps not as arduous as it may seem, because you can reuse the Autoflow functions across inherited classes. For example, in the Kern class we define an AutoFlow `compute_K` function, which calls the kernel function(called `K`, but perhaps would be better called `build_K`), which is then defined differently in all the inherriting classes. 

If you still have a use-case for a more complex AutoFlow, we can certianly think about it.

markvdw:
Hmm ok. This is the solution I've gone for in the end.

PR #398, opened 2017-04-13
Allow both tf and non-tf arguments in AutoFlow

mrksr:
Hello everyone!

This pull request accompanies issue #383, see some preliminary discussion there. As is, this PR is not ready to merge, see below.

The Problem
===
While playing with one of my models, I had a few issues with the (very convenient) AutoFlow, previously discussed in #178, namely, that a AutoFlowed function only takes arguments which are converted to tensorflow-variables. While this is usually enough, it gets a bit tricky when some function interacting with tensorflow has multiple (usually boolean) arguments which slightly alter their behaviour. An example in the current GPflow code can be seen in model.py, where the two functions
```python
    @AutoFlow((float_type, [None, None]))
    def predict_f(self, Xnew):
        return self.build_predict(Xnew)

    @AutoFlow((float_type, [None, None]))
    def predict_f_full_cov(self, Xnew):
        return self.build_predict(Xnew, full_cov=True)
```
only differ in a boolean argument.

This workaround becomes an issue in two cases:

1) A function which takes multiple boolean arguments.
2) A function which takes (say) a constant float as an argument (think quantiles).

In the first case, a rather large number of function is needed to cover all possible boolean combinations and in the second case I am not aware of a workaround other than doing things on the tensorflow side, which might incur a runtime cost.

There are two major things to consider when altering the current implementation:

1) How to handle graph memoization dependent on non-AutoFlowed arguments
2) Which syntax to use to specify AutoFlowed arguments (and how to handle backwards compatibility)

Proposed Solution
===
The main difference between the current implementation and my proposed new one is the fact that there are two kinds of arguments to AutoFlowed functions: Those which should be replaced by Tensors (and therefore be fed into a finished graph) and those which can influence the graph generation. The syntax is used to differentiate the two sets which are later on handled differently to do graph memoization and the replacement of arguments at runtime.

New Syntax
---
I'm quoting @jameshensman here:
> I think I would probably choose a differnt syntax, though. Perhaps this:
> 
> ```python
> @GPflow.AutoFlow(X=(tf.float64, [None]), Y=(tf.int32, [2, 2]))
> def myfunc(X, Y, Z, W):
>     if Z:
>         return foo(X)
>     if W:
>        return foobar(Y)
> ```
> 
> In this (hypothetical) example, X and Y get "autoflowed", the other arguments Z and W do not. I think would create a really nice syntax for autoflow.

I like this syntax as it is very pythonic, but feel free to discuss. There are however a few things to consider. First, I think the syntax should be backwards compatible if possible, so `AutoFlow` should accept both normal arguments and keyword arguments. A normal argument can be thought to be "lifted" to a keyword argument by associating it with the corresponding name. 

Varargs are a special case here. Consider this function:
```python
@AutoFlow((tf.float64,), b=(tf.float64,), x=(tf.float64,))
def func(a, b, *cs, x='x', y='y', **zs):
    print(a, b, cs, x, y, zs)
```
The proposed syntax offers a way to specify the replacement of the variabes `a, b, x, y`. I think in the current master, it would in principle be possible to specify that a constant number of elements in `cs` should also be replaced by Tensors just by adding more arguments to `AutoFlow`. I cannot really come up with a use case for this and have not tried to support it. A more interesting case might be to replace every element in `cs` by its own Tensor (e.g. for models using multiple spaces) and I think this is doable at least in python 3. I can add this if there is interest.

Also note that the semantics of function definitions are different in python 2 and python 3. Namely, in python 3 it is possible to define keyword only arguments such as `x` and `y` in the example above. The example is not valid python 2. This is one of the reasons why the `inspect`-API is considerably different in the two pythons and it will be neccessary to write special code (say ~20 lines) for both versions. I have implemented the python 3 version because it is much easier to formulate. Because of this, this PR currently works with **python 3 only**. Using python 3, everything should be fine however and GPflow should work as usual.

Graph Memoization
---
Bacause of the non-tf arguments, it is no longer enough to memoize one graph per function. Instead, we want to memoize one graph per function and combination of non-tf arguments. My solution to this is inspired by [functools.lru_cache](https://docs.python.org/3/library/functools.html#functools.lru_cache) and calculates the hash values of all non-tf arguments to generate a deterministic but unique `storage_name`. This works fine as long as the arguments are hashable. Some notable none-hashable things are lists, dicts, sets and numpy arrays. A tuple is hashable however and can be used to represent these things if necessary.

Suppose the set `hash_params` contains the names of all arguments to be hashed, then the relevant code looks something like this:
```python
hash_args = tuple(
    (name, argument)
    for name, argument in all_args.arguments.items()
    if name in hash_params
)

storage_name = '_{}_{}_AF_storage'.format(tf_method.__name__, hash(hash_args))
```
The real code contains a not-so-complicated special case for keyword-varargs.

Argument replacement
---
Currently, in order to generate the memoized graph, all arguments to the original function (which will generally be numpy arrays) are replaced by tensorflow placeholders in a simple manner. In my solution, there are more things to consider since the tensorflow placeholders have to be merged with the original non-tf parameters. In an ideal world, python should take care of this merge since it has to be implemented in CPython anyway (think, for example, of `functools.partial`). And indeed, starting with python 3.5, this merge is exposed via [inspect.Signature.bind_partial](https://docs.python.org/3/library/inspect.html#inspect.Signature.bind_partial). I think in python 2 this has to be implemented manually. See #383 for some (non-final) ideas about how it can be done.

In order to make the code more understandable, here is a simpler example without tensorflow in the picture. This implements a decorator which simply replaces arguments at call-time.
```python
import inspect
from functools import wraps


def find_varkw(signature):
    varkw = None
    for name, parameter in signature.parameters.items():
        if parameter.kind is inspect.Parameter.VAR_KEYWORD:
            varkw = name
    return varkw


def flat_bound_arguments(signature, bound):
    varkw = find_varkw(signature)

    for name, argument in bound.arguments.items():
        if name == varkw:
            for kwarg in argument.items():
                yield kwarg
        else:
            yield (name, argument)


def replace(*arg_replacements, **kwarg_replacements):
    def decorator(method):
        signature = inspect.signature(method)
        varkw = find_varkw(signature)

        replace_params = dict(flat_bound_arguments(
            signature, signature.bind_partial(*arg_replacements, **kwarg_replacements)
        ))
        print('replace_params', replace_params)

        @wraps(method)
        def wrapper(*args, **kwargs):
            all_args = signature.bind(*args, **kwargs)
            all_args.apply_defaults()

            print('would call:', all_args)

            hash_args = tuple(
                (name, argument)
                for (name, argument)
                in flat_bound_arguments(signature, all_args)
                if name not in replace_params
            )

            print('hash args:', hash_args)
            print('hash:', hash(hash_args))

            for name, argument in replace_params.items():
                if name in signature.parameters.keys():
                    all_args.arguments[name] = argument
                else:
                    if varkw is not None:
                        all_args.arguments[varkw][name] = argument
            print('will call:', all_args)

            method(*all_args.args, **all_args.kwargs)

        return wrapper
    return decorator


@replace(5, b='hithere', x=None, q='qq', r='rr')
def func(a, b, *cs, x='x', y='y', **zs):
    print(a, b, cs, x, y, zs)


func(2, 3, 4, 5, 6, y='test', q='ishouldbereplaced', s='iop')
```
This outputs
```
replace_params {'a': 5, 'b': 'hithere', 'x': None, 'q': 'qq', 'r': 'rr'}
would call: <BoundArguments (a=2, b=3, cs=(4, 5, 6), x='x', y='test', zs={'q': 'ishouldbereplaced', 's': 'iop'})>
hash args: (('cs', (4, 5, 6)), ('y', 'test'), ('s', 'iop'))
hash: -1563578292665240278
will call: <BoundArguments (a=5, b='hithere', cs=(4, 5, 6), x=None, y='test', zs={'q': 'qq', 's': 'iop', 'r': 'rr'})>
5 hithere (4, 5, 6) None test {'q': 'qq', 's': 'iop', 'r': 'rr'}
```

Things to do
===
In order to make this PR ready for merge, there are quite a few things left to be done. I would love to hear your thoughts nonetheless :). Here is a list of things to do:
- [x] Fix tests which depend on current AutoFlow (like `test_session`)
- [x] Python 2 compatibility
- [x] Thoroughly test the new syntax and possibilities
- [x] Update current GPflow functions, i.e. get rid of `predict_f_full_cov` and the like. This likely changes the API of GPflow
- [x] Update Docstring


mrksr:
I have considered the alternatives for python 2 compatibility and I think the easiest way to move forward is to reimplement the relevant parts of the new `inspect.signature`-API in python2. I see two ways to achieve this:

1. Use [funcsigs](https://github.com/testing-cabal/funcsigs), a library which backports this API
2. Implement a simplified version of the API in GPflow

The main argument for the first alternative would be that the library gets things right. It does however add a dependency. I have explored the second alternative and think we can get most of the python 3 features with comparatively little code. I will therefore go the second route for now. Here is a preliminary implementation of the `replace`-example in python 2. Note that python 2 does not know keyword-only arguments and for the sake of brevity I have not implemented proper `varargs`-handling.

```python
from __future__ import print_function

import inspect
from functools import wraps


class BoundArguments(object):
    def __init__(self, arguments):
        self.arguments = arguments
        self.args = ()
        self.kwargs = arguments

    def apply_defaults(self):
        pass


class Parameter(object):
    VAR_KEYWORD = True
    def __init__(self, kind):
        self.kind = kind


class Signature(object):
    def __init__(self, func):
        self.func = func
        self.argspec = inspect.getargspec(func)
        if self.argspec.varargs is not None:
            raise TypeError('Cannot handle varargs in AutoFlow in python 2')

        self.parameters = {arg: Parameter(False) for arg in self.argspec.args}
        if self.argspec.keywords is not None:
            self.parameters[self.argspec.keywords] = Parameter(True)

    def bind(self, *args, **kwargs):
        return BoundArguments(inspect.getcallargs(self.func, *args, **kwargs))

    def bind_partial(self, *args, **kwargs):
        if len(args) > len(self.argspec.args):
            raise TypeError('Passed too many arguments')

        args_dict = dict(zip(self.argspec.args, args))
        if not set(args_dict.keys()).isdisjoint(set(kwargs.keys())):
            raise TypeError('Passed duplicate arguments')
        args_dict.update(kwargs)

        return BoundArguments(args_dict)

# Add the code in the first post here, sans the keyword only arguments.
```

mrksr:
So this last commit adds some preliminary python2 support. I think there are still some bugs in there which will be found once I write some more tests, but in principle, the code is taking shape. Compared to the initial implementation, the code actually got a bit simpler (and more correct).

I would love to hear your opinion on this, @jameshensman, since the python2 stuff makes the code quite a bit more complex, so I think it is important to make sure it is not too much.

mrksr:
I have now fixed things some more and added tests for the new AutoFlow functionality. These tests pass on both python 2 and python 3, but the one testing varargs is only run on python 3. Concerning coverage, the python2 API shim should probably be excluded from codecov?

Additionally, it turns out that `predict_f_full_cov` is the only AutoFlow-function in GPflow right now which only exists to pass a kwarg, so I deprecated it. It might be nice to expose some more kwargs in other AutoFlow functions.

Lastly, I did adjust the AutoFlow docstring but I am not sure where else to put documentation for this. Basically, AutoFlow functions now behave like normal python functions and the replacement-API should be quite intuitive, besides the fact that it does not handle varargs. Probably this should still be documented somewhere besides here.

st--:
@mrksr still interested in getting this PR into gpflow ? it'd require some work due to the gpflow-1.0 changes.

mrksr:
If you think it still fits into the project (I think it does :)), I will gladly rework the PR for gpflow-1.0. After talking to @jameshensman last September we decided to wait and see how the new gpflow will shape out. I think now would indeed be a good time to decide if we want it.

Compared to the current implementation, the code will actually become _much_ simpler because we do not need a python2 shim anymore.

st--:
I think that'd be great.

st--:
@awav is probably the best person to discuss with.

markvdw:
Probably best to do a new PR after chatting to @awav about this. Thanks for the efforts!

PR #410, opened 2017-05-03
Fix pickle idempotence

mrksr:
In the current master, it is not possible to pickle a model which has been loaded from a pickle itself, i.e. the operation `dumps(loads(dumps(model)))` fails. The problem is that a freshly unpickled and untouched model does not have `_tf_array`s for all Param objects und the corresponding `__getstate__`-method tries to remove it resulting in an exception.
There is a fix for this in in `model.py` which ensures that things like `model._graph` are only removed when they actually exist. This PR does the same in the `Param`-class which seems to fix the issue.

### DIFF
--- a/GPflow/model.py
+++ b/GPflow/model.py
@@ -108,10 +108,7 @@ class Model(Parameterized):
         """
         d = Parameterized.__getstate__(self)
         for key in ['_graph', '_session', '_free_vars', '_objective', '_minusF', '_minusG', '_feed_dict_keys']:
-            try:
-                d.pop(key)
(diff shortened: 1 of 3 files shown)


jameshensman:
I have a suggestion for simplification. I think this:
```python
try:
    d.pop(key)
except:
    pass
```
is equivalent to
```python
d.pop(key, None)
```

mrksr:
It is, I just wanted to keep with the style of `models.py`. I can change it at both places since I prefer the default-value version, too :).

Issue #409, opened 2017-05-03
MLP kernel gradients


markvdw:
Hi all, (specifically @mrksr, since you did the PR :) )

I'm trying to use the MLP kernel, but I get nans in the gradients! Has anybody used them before (i.e. am I doing something wrong?) or is there a bug?

```python
import numpy as np
import GPflow
X = np.random.rand(4, 1)
Y = np.random.randn(X.shape[0])[:, None] * 0.1 + np.sum(np.sin(3*X), 1)[:, None]
full = GPflow.gpr.GPR(X, Y, GPflow.kernels.ArcCosine(X.shape[1]))
full._compile()
print(full._objective(full.get_free_state())[1])
```

Output: `[ 0.81873344         nan         nan  0.25598061]`

mrksr:
If I alter your example to be reproducable:
```python
import numpy as np
import GPflow

rng = np.random.RandomState(23)
X = rng.rand(4, 1)
Y = rng.randn(X.shape[0])[:, None] * 0.1 + np.sum(np.sin(3*X), 1)[:, None]
full = GPflow.gpr.GPR(X, Y, GPflow.kernels.ArcCosine(X.shape[1]))
full._compile()
print(full._objective(full.get_free_state())[1])
```
my result using float64 is `[        nan  0.22290395         nan  0.81633098]`.

My first guess would be that the culprit is the `acos`-operation and indeed, If I apply this patch:
```patch
diff --git a/GPflow/kernels.py b/GPflow/kernels.py
index 36adb26..a9f12b9 100644
--- a/GPflow/kernels.py
+++ b/GPflow/kernels.py
@@ -573,10 +573,11 @@ class ArcCosine(Kern):
             X2_denominator = tf.sqrt(self._weighted_product(X2))

         numerator = self._weighted_product(X, X2)
+        acos_bound = 1 - settings.numerics.jitter_level
         theta = tf.acos(tf.clip_by_value(numerator / \
                                          X_denominator[:, None] / \
                                          X2_denominator[None, :],
-                                         -1., 1.))
+                                         -acos_bound, acos_bound))

         return self.variance * (1. / np.pi) * self._J(theta) * \
                X_denominator[:, None] ** self.order * \
```
the result becomes `[-0.01297854  0.22256484  0.01297854  0.81661298]`. Could you check if this fixes the issue? If so, we can maybe come up with a better way of handling the `nan`s.

markvdw:
[...]

PR #411, opened 2017-05-03
Fix MLP acos numerics

mrksr:
This PR accompanies #409.

Adding a jitter to the `tf.acos`-operation fixes the gradient problem, but introduces a new one: Now, the kernel does not pass the correctness-tests anymore. We can make the closness-test more lenient, but I wanted to ask for your opinion: Using a jitter larger than `1e-10` results in the diagonal of `kern.K(X)` not being equal to `1` anymore. Will this be a problem?

### DIFF
--- a/GPflow/kernels.py
+++ b/GPflow/kernels.py
@@ -574,7 +574,7 @@ class ArcCosine(Kern):
 
         numerator = self._weighted_product(X, X2)
         cos_theta = numerator / X_denominator[:, None] / X2_denominator[None, :]
-        jitter = settings.numerics.jitter_level
+        jitter = 1e-15
         theta = tf.acos(jitter + (1 - 2 * jitter) * cos_theta)
 
         return self.variance * (1. / np.pi) * self._J(theta) * \
--- a/GPflow/kernels.py
+++ b/GPflow/kernels.py
@@ -573,10 +573,9 @@ class ArcCosine(Kern):
             X2_denominator = tf.sqrt(self._weighted_product(X2))
 
         numerator = self._weighted_product(X, X2)
-        theta = tf.acos(tf.clip_by_value(numerator / \
-                                         X_denominator[:, None] / \
-                                         X2_denominator[None, :],
-                                         -1., 1.))
+        cos_theta = numerator / X_denominator[:, None] / X2_denominator[None, :]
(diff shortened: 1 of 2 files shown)


markvdw:
I wouldn't technically use the jitter_level here, as the numerical issue is different to where jitter_level is usually used (i.e. in inversions). I expect that a very small rescaling should be enough. I found that the offending value was essentially a single machine epsilon above 1.

I suggest we just take the smallest possible value for now, and see whether it causes any problems in the future. If it does, we know where the cause is.

mrksr:
Okay, I now chose the smallest power of ten for which the tests still pass. This seems to work for both `float32` and `float64` interestingly. But I suspect this number will have to be tweaked a little bit at some point.

mrksr:
So I have now chosen this jitter to be as close to the `float64` machine epsilon as possible. And the `TestArcCosine`-tests all still pass when I change the float_type to `float32` in `gpflowrc`. The complete test-suite does not, but this is due to the fact that `float64` is hardcoded at several places in `test_kerns.py`. If I force `float32`-calculations, the correctness-tests no longer pass if I choose a jitter which prevents the NaNs (`1e-7`).

Should we choose the jitter dependent on the float type and live with the fact that using `float32`, the diagonal will not be equal to one? If so, is there a better way than this:
```python
if float_type is tf.float32:
    jitter = 1e-7
else:
    jitter = 1e-15
```

markvdw:
@jameshensman: How do you usually move a PR into an "in-house" PR that gets tested and can get merged?

jameshensman:
make a new local branch, pull the remote onto it, and push to GPflow/origin. There are some helpful hints under 'view command line instructions' below. Here, they suggest:

```bash
git checkout -b mrksr-fix-mlp-acos-numerics master
git pull https://github.com/mrksr/GPflow.git fix-mlp-acos-numerics
```
Now fix any merge conflicts. Hopefully there are none. 

But then, instead of pushing to master like they suggest (which is disabled in GPflow), you should do 

```git push --set-upstream origin mrksr-fix-mlp-acos-numerics```

to push your new branch to GPflow. Then the tests should run and you can merge.

mrksr:
Thanks for the quick merge :). So we need to keep in mind that currently the kernel will not work with `float32`.

PR #413, opened 2017-05-05
Beta prior

Bonnevie:
Felt that the Beta prior was missing from the arsenal of priors. 
Piggybacked on the existing beta density in densities.py.
Added a test in line with the existing prior tests.


markvdw:
This looks great to me. I'll merge once the tests pass.

Issue #416, opened 2017-05-15
Reproducibility using seeds


Bonnevie:
Been running some experiments, and it seems I cannot get the same reproducible result even when setting numpy and tensorflow seeds at the top of the script like

```
np.random.seed(72)
tf.set_random_seed(72)
```

Naturally, either I'm setting them wrong or I'm missing a RNG. I call Param.randomize() and Model.optimize(), and they seem to be the only sources of randomness at the moment (and I'm not even sure about the latter) so one of them is likely the culprit. Is there any way to either set the seed appropriately without messing with the source code or maybe adding an option allowing seeds to be set in the future?

mrksr:
A shot in the dark: `Parameterized.randomize` depends on the order of parameters in `sorted_params` which, in turn, depends on the order in [`Parameterized.__dict__`](https://github.com/GPflow/GPflow/blob/master/GPflow/param.py#L790). I do not think that python makes a guarantee about the order of members in `__dict__` (maybe in python 3.6?) and therefore it might happen that while the seed is the same, `Param.randomize()` gets called in a different order each time, thus resulting in different initializations.

Bonnevie:
uh, that could definitely be problematic. But running some tests it does not seem to pose any immediate problems - I can set the seed and run randomize, quit the session, repeat the process, and get the same result. 
It almost seems like the problem occurs in `Model.predict_f`, but I don't think there's anything random going on there.

Bonnevie:
No, wait, I think you're right, it just only occurs in the lowest level of my model hierarchy! Good catch!

mrksr:
There is a call to `sorted` in the function I linked which uses the `id` as a key. This `id` will not be the same for every run. Thinking about it, maybe the order of items in `__dict__` does not actually matter.
Maybe the Params could be sorted by their long names instead? I guess @jameshensman will have thought about this when implementing `sorted_params`.

As a quick fix without patching GPflow: You could iterate over the Params yourself and call randomize more carefully?

Bonnevie:
seems like sorting by `Param.name` does the trick - I can submit a micro-PR if the sorting key is a bit arbitrary anyway?

john-bradshaw:
hey although randomize does not use Tensorflow(?) so you're okay setting only the numpy seed I think `tf.set_random_seed` works on a graph level, so you want to call it in a particular graph context rather than at the start of the script.

More info here: https://www.tensorflow.org/api_docs/python/tf/set_random_seed

[...]

PR #417, opened 2017-05-15
Fix parameter ordering so that it's consistent across sessions.

Bonnevie:
Fixes issue #416 by changing the sorting key (currently based on `id`) to something that is static between sessions (`Param.name` in this PR, picked slightly arbitrarily).


GPflow/param.py
```diff
@@ -796,7 +796,7 @@ def sorted_params(self):
         params=  [child for key, child in self.__dict__.items()
                   if isinstance(child, (Param, Parameterized)) and
                   key is not '_parent']
-        return sorted(params, key=id)
+        return sorted(params, key=lambda x: x.name)
```
john-bradshaw: I think you want to use `long_name`?

Bonnevie: yup, was just fixing it.

Bonnevie:
From the issue thread:

Running a timing experiment in IPython (using %magic) of the form

```
class A():
    def __init__(self, N = 20):
        self.long_name = ''.join(random.choice(string.ascii_uppercase +
                                 string.digits) for _ in range(N))

L = [A() for _ in range(100)]
test_id = lambda: sorted(L, key=id)
test_name = lambda: sorted(L, key=lambda x: x.long_name)
%timeit test_id
%timeit test_name
```
results in timings with almost no difference, and if anything, the timings are in favor of the long_name key.

edit: e.g. one run gives 
```
10000000 loops, best of 3: 29.1 ns per loop
10000000 loops, best of 3: 24.8 ns per loop

```

jameshensman:
You need to add parentheses after the functions in %timeit. 

I get 
```python
In [14]: %timeit test_name()
10000 loops, best of 3: 27.7 µs per loop

In [15]: %timeit test_id()
100000 loops, best of 3: 10.3 µs per loop
```

Bonnevie:
@jameshensman: ah, sorry about that. I can see how looking up the function name alone would be a pretty fast operation :)
Still doesn't seem like a huge time difference?

javdrher:
This would be a helpful addition for GPflowOpt. Yesterday I was trying to implement BO with MCMC for the hypers by copying, using set_state(), and then adding all scores together. However in a (deep)copy of a model the order also differs.

Bonnevie:
@jameshensman: there seems to be a few minor PRs stuck in limbo - is there anything further I should fix before this can be merged?

jameshensman:
This PR seems good, but I think it requires a version bump. 

See https://github.com/GPflow/GPflow/blob/master/contributing.md

Could you handle that, @Bonnevie and then I will merge this.

Bonnevie:
@jameshensman: was this what you had in mind?

PR #419, opened 2017-05-22
Rescale transform

markvdw:
In situations where the gradients are very poorly scaled it can help to rescale the parameter space to make the gradients have a similar size to those of other parameters. This rescale transform can do this rescaling, while applying other transforms afterwards.

I think it definitely can be useful. Let me know if you agree and I'll implement some unittests.


ialong:
I also need this quite badly!

jameshensman:
Yep, sounds like a good feature.

GPflow/transforms.py
```diff
@@ -195,6 +197,28 @@ def __str__(self):
         return '[' + str(self.a) + ', ' + str(self.b) + ']'
 
 
+class Rescale(Transform):
```
jameshensman: Please add a docstring here.

ialong:
Sweet! Thank you!

PR #420, opened 2017-05-26
Robust numpy softplus

javdrher:
Following the discussion in GPflow/GPflowOpt#17 this implements a backward/forward transform which should avoid overflows and divisions by zero. Also adds tests to verify the warnings are no longer popping up.

### DIFF
--- a/GPflow/transforms.py
+++ b/GPflow/transforms.py
@@ -126,6 +126,10 @@ class Log1pe(Transform):
         self._lower = lower
 
     def forward(self, x):
+        """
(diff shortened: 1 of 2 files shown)


icouckuy:
Can we document these little 'tricks' a bit? Just a doc string with the original expression and why it has been implemented as below? I think this would greatly benefit the readability for newcomers.

jameshensman:
Looks good, but agree with @icouckuy that dictating would be helpful.

jameshensman:
*docstrings!

PR #422, opened 2017-06-03
added the bounds option and the Gabor kernel.

yangnw:
added the bounds option to `optimize()`.
added the Gabor kernel, which is the 1-unit special case of the Spectral Mixture (SM) kernel.


jameshensman:
I suggest that this kernel is redundant. You could make this kernel by doing

```python
k = GPflow.kernels.RBF(1) * GPflow.kernels.Cosine(1)
```

The 'lengthscale' of the Cosine kernel relates to the frequency. 

Closing: please re-open if there are any objections.

PR #424, opened 2017-06-07
Add explanation of arguments to AutoFlow decorator docstring

st--:
The arguments required by AutoFlow are not immediately obvious. Adds a paragraph to the docstring to explain with a bit more detail.


jameshensman:
I merged this with #426

PR #426, opened 2017-06-07
Provides metaclass to simplify GPflow Model creation workflow

st--:
In my work with GPflow, constructing a new model often has a workflow similar to the following:

- subclass `GPflow.models.Model` and create a `build_likelihood` function
- all the terms that need to go into `build_likelihood` are too complicated, maybe have some bugs ... so split into smaller `build_foo` functions that compute individual terms and are combined in `build_likelihood`
- to be able to interact with the individual terms, I need to write separate `compute_foo` functions, that do nothing other than calling build_foo but are wrapped using `@AutoFlow()`.

E.g.

```python
import tensorflow as tf
import GPflow

class Demo(GPflow.model.Model):
    @GPflow.param.AutoFlow(tf.float64, [None, 2])
    def compute_foo(self, x):
        return self.build_foo(x)

    def build_foo(self, x):
        return tf.reduce_sum(x)
```

With the metaclass in this module, the code can be simplified to the following:

```python
import tensorflow as tf
import GPflow, GPflow.metaautoflow

class Demo(GPflow.model.Model, metaclass=GPflow.metaautoflow.MetaAutoFlow):
    @GPflow.param.AutoFlowify(tf.float64, [None, 2])
    def build_foo(self, x):
        return tf.reduce_sum(x)
```

and the `compute_foo` function will be constructed automatically. This saves typing, making mistakes in writing compute_... functions, and leads to more readable code, particularly if there is a large number of build_... functions...

As it stands, this would not change any of the existing codebase, so would be completely backwards-compatible. If this would turn out to be useful to many other people, it would eventually be possible to add the metaclass directly to `GPflow.model.Model` so that the functionality would be inherited by all subclasses, but it can also stand alone and anyone who does find it useful can add it to their own models.


mrksr:
Hello and welcome, @st--!
I like the idea of eliminating the need for both an "original" Tensorflow function and an AutoFlowed alternative which simply calls the original. I have thought about this a bit in the course of implementing #398 and I think it might be nicer to make `AutoFlow` smater instead of adding another decorator: It should be doable to simply use tensors supplied as arguments and only do the magic replacement of numpy arrays by tensorflow variables if neccessary. I did not (yet) implement this since #398 is quite complicated already and I thought it would probably be a good idea to tackle this as a separate PR. In terms of code the change should not be that significant.

I do however think that both your method and an even more magic AutoFlow are at risk of being quite unpythonic, since so much logic is created implicitly and it might be hard to discover this functionality as a new user. What do you think?

st--:
Hi, thanks for pointing out #398 / #383 - that had come up for me, too, and would be really nice to have that sorted out! (I like the proposed keyword syntax.)

I guess it should be possible to add a clause inside `AutoFlow` that checks whether arguments are tf Tensors (in that case leave them alone and just call the wrapped method and pass them on directly) or np arrays (in that case do all the AutoFlow-magic before calling the wrapped method with the tf variables), though might want an extra check to ensure that all "free" arguments are *either* np *or* tf, not mixed... (On a related but separate note, AutoFlowed functions also seem to give rather unhelpful error messages e.g. when called with the wrong numbers/types of arguments, so something else worth looking at.) Don't know how long that'd take to implement, test & merge though.

This metaclass has the advantage of not changing any functionality, at the moment a user would have to explicitly add this metaclass to their model for it to work, and I'd hope they'd then read through the documentation as well... I think this is a perfectly pythonic use of metaclasses, actually:) Discussed this with @jameshensman who seemed in favour.

Obviously if/once AutoFlowed functions can cope both with being called inside a tensorflow graph and "from the commandline" with numpy arrays, this module becomes redundant and could be removed again - until then, it might make life easier for others.

I've split this up into issue #427 ...

mrksr:
I agree that for the development process, this PR would sure be convenient. My thoughts concerning discoverability were directed towards essential functions like predict_f or compute_log_likelihood which I think should probably be defined explicitly to make the life of both new developers and IDEs as easy as possible. This new functionality being optional is a strong argument however since, as you say, people basically need to read the documentation to use it anyway.

st--:
NB- commit 797a442793dfe9428e07a38e707a9e96c2661a6b only adds the file `metaautoflow.py` and does not import this module in the `__init__.py`. So, at the moment, one would have to explicitly `import GPflow.metaautoflow` (rather than `import GPflow` and being able to use `GPflow.metaautoflow` out of the box). Should this be added to the pull-request?

Issue #389, opened 2017-03-27
vgp accuracy 


hughsalimbeni:
While the current method equivalence test passes, I've noticed that vgp becomes quite inaccurate when N is large. I find that the svgp/fitc/sgpr methods with Z=X are pretty close to gpr with a Gaussian likelihood for any N in the hundreds, but not so with vgp. Could the tests be made more stringent?

e.g. here is an example modified from the testing code with fixed hypers and more points:

```python
Ns = np.arange(1, 20)*10
Ls = [[] for _ in range(6)]

for N in Ns:
    X = np.linspace(-10, 10, N)[:, None]
    Y = np.sin(X) + 0.9 * np.cos(X*1.6) + np.random.randn(*X.shape) * 0.8

    m1 = GPR(X, Y, RBF(1))
    m2 = VGP(X, Y, RBF(1), Gaussian())
    m3 = SVGP(X, Y, RBF(1), Gaussian(), X.copy(), q_diag=False, whiten=False)
    m4 = SVGP(X, Y, RBF(1), Gaussian(), X.copy(), q_diag=False, whiten=True)
    m5 = SGPR(X, Y, RBF(1), X.copy())
    m6 = GPRFITC(X, Y, RBF(1), X.copy())

    models = [m1, m2, m3, m4, m5, m6]
    
    for m, L in zip(models, Ls):
        m.kern.fixed = True
        m.likelihood.fixed = True
        if hasattr(m, 'Z'):
            m.Z.fixed = True
        if len(m.get_free_state())>1:
            m.optimize()
                
        L.append(float(m.compute_log_likelihood()))
        
Ls = [np.array(L) for L in Ls]

plt.plot(Ns, Ls[0]-Ls[1], label='vgp')
#plt.plot(Ns, Ls[0]-Ls[2], label='svgp') # very wrong for N>30
plt.plot(Ns, Ls[0]-Ls[3], label='svgp white')
plt.plot(Ns, Ls[0]-Ls[4], label='sgpr')
plt.plot(Ns, Ls[0]-Ls[5], label='fitc')

plt.title('Difference in log marginal likelihood from exact model')
plt.axes().set_xlabel('number points')
plt.axes().set_ylabel('error')
plt.savefig('err.pdf', bbox_inches = 'tight')
plt.legend(loc='upper left', bbox_to_anchor=(1., 1.))
plt.show()

```
[err.pdf](https://github.com/GPflow/GPflow/files/873219/err.pdf)
[...]

PR #430, opened 2017-06-19
Vgp as svgp

hughsalimbeni:
The vgp after opper and archambeau is more parameter efficient than svgp with X=Z, but suffers from non-convexity #389. This PR retains the old vgp as `VGP_opper_archambeau`, but replaces `VGP` with the (whitened) `SVGP` model with X=Z, with unnecessary computation removed.


alexggmatthews:
So the issue in #389 was diagnosed as resulting from non-convexity in the end? Perhaps I missed a conversation. If so do we know that svgp doesn't suffer from the same problem? I think I remember Emtiyaz Khan writing something on this topic.

hughsalimbeni:
[Khan, Mohamed and Murphy](https://papers.nips.cc/paper/4535-fast-bayesian-inference-for-non-conjugate-gaussian-process-regression.pdf) analysed the Opper/Archambeau parametrisation and found it non-convex even in the case of a single point (section 5, final two paragraphs). Their convex coordinate ascent scheme shows improved results over Opper/Archambeau parametrisation. Further, [Challis and Barber](http://proceedings.mlr.press/v15/challis11a/challis11a.pdf) looked at the lower triangular parametrisation and showed that the optimisation is convex for Bayesian linear models. 

I'd suggest that the saving in parameters is insufficiently useful, given that the optimisation is harder and there aren't any computation savings. 

Experimentally:

```
N = 200
X = rng.rand(N, 1)*10
Y = np.sin(X) + 0.9 * np.cos(X*1.6) + rng.randn(*X.shape) * 0.8

m1 = GPflow.gpr.GPR(X, Y, kern=GPflow.kernels.RBF(1),
                    mean_function=GPflow.mean_functions.Constant())
m2 = GPflow.vgp.VGP(X, Y, GPflow.kernels.RBF(1), likelihood=GPflow.likelihoods.Gaussian(),
                    mean_function=GPflow.mean_functions.Constant())
m3 = GPflow.vgp.VGP_opper_archambeau(X, Y, GPflow.kernels.RBF(1), likelihood=GPflow.likelihoods.Gaussian(),
                    mean_function=GPflow.mean_functions.Constant())
print(m3)
Ls = []
for m in [m1, m2, m3]:
    m.optimize()
    Ls.append(m.compute_log_likelihood()) 
print(Ls[1:] - Ls[0])  # [ -4.85780620e-06  -4.30471057e-03]
```

alexggmatthews:
Ok @hughsalimbeni sounds reasonable to me.

jameshensman:
I merged the internal dup.

Issue #377, opened 2017-03-10
BGPLVM error in expectation when passing in slice


alexisboukouvalas:
Run into this when debugging a failure in the GPLVM notebook.
When passing in slice for active dimensions code fails with cholesky error.
When passing in active dimensions as lists, the code works. (change fHmmm to False)
@markvdw any idea on how to debug this further?
```
import GPflow
from GPflow import ekernels
from GPflow import kernels
import numpy as np
import pods
pods.datasets.overide_manual_authorize = True  # dont ask to authorize
np.random.seed(42)
GPflow.settings.numerics.quadrature = 'error'  # throw error if quadrature is used for kernel expectations
data = pods.datasets.oil_100()
Y = data['X']
Q = 5
M = 10  # number of inducing pts
N = Y.shape[0]
X_mean = GPflow.gplvm.PCA_reduce(Y, Q) # Initialise via PCA
Z = np.random.permutation(X_mean.copy())[:M]

fHmmm = True
if(fHmmm):
    k = ekernels.Add([ekernels.RBF(3, ARD=True, active_dims=slice(0,3)),
                  ekernels.Linear(2, ARD=False, active_dims=slice(3,5))])
else:
    k = ekernels.Add([ekernels.RBF(3, ARD=True, active_dims=[0,1,2]),
                  ekernels.Linear(2, ARD=False, active_dims=[3, 4])])
    
m = GPflow.gplvm.BayesianGPLVM(X_mean=X_mean, X_var=0.1*np.ones((N, Q)), Y=Y,
                                kern=k, M=M, Z=Z)
m.likelihood.variance = 0.01
m.optimize(disp=True, max_iters=100)
```

markvdw:
Ok, the code is a bit complicated here, since there are quite a few cases. I suspect the error is in `Add.eKzxKxz()`.

Your kernels work on separate dimensions and the GPLVM is initialised as with Xcov being factorised over dimensions. In the case where `active_dims` is a list, we can determine that the kernel works on separate dimensions, and so ekernels.py:191 should be taken. This is simple and works as you mention.

When `active_dims` is a slice, things are more complicated. There is an analytic expression for the cross terms only for the case where _all dimensions are shared_. Currently we test for this by checking `active_dims` is a slice.  This is where the error is. We should be checking that it's a slice that covers all dimensions, and falls back to quadrature if not.

So there are two solutions:
- Check whether `active_dims` covers all dimensions for _both_ kernels in ekernels.py:221
[...]

PR #432, opened 2017-06-20
Test speeds

jameshensman:
I added attributes to the tests to mark the ones that are slow. You can now skip them with

`nosetests -A "speed!='slow'" `

note that all the tests still run on Travis/Codeship. If anyone has any good ideas for speeding up the turnaround time there, I'd be happy to hear them.


javdrher:
The GPflowOpt tests are getting lengthy on travis so I have been looking at this. I'm currently adding caching of dependencies, should save about 30s

You could add slow/!= slow as envs and split the build. Those two can then be parallelized as described here: https://blog.travis-ci.com/2012-11-28-speeding-up-your-tests-by-parallelizing-them/
Both builds will still run, but the fast build will provide feedback faster.

jameshensman:
So, in investigating the tests, I discovered that the notebook tests were not actually running. This is because the path to the notebook directory was hard coded relative to the working directory, which changes on the Travis/Codeship server. 

I fixed the relative path issue, so now the notebooks get tested as part of the CI. 

I has to blacklist the regression notebook, because the MCMC cell times out on Codeship (the free server is very underpowered). But this PR means that more is being tested, not less.

jameshensman:
Thanks @javdrher , would be great if you could put those changes in another PR. 

@alexisboukouvalas , could you give the PR a quick check to make sure it makes sense? Thanks.

javdrher:
@jameshensman I'll give it a try for GPflow. I'm not very happy about the speedup achieved for GPflowOpt. Most of the time seems to be lost while the build is waiting to be started.

alexisboukouvalas:
I am away this week but can happily take a look this weekend

jameshensman:
@alexisboukouvalas marvellous, thank you.

alexisboukouvalas:
As mentioned in #437, a big speedup could be achieved if the notebook tests was parallelised. Conceptually this is simple because the each notebook test is independent. So @javdrher it seems you already have a good implementation for this. Could you create a pull request or even merge with this one?

I would prefer to remove the 'blacklist' feature if possible.
Is the problem with the GPLVM notebook, import pods? Can we add this to tests_require or I could change the script to use local csv file. I have found at least on bug using the GPLVM notebook so I think it would be worth keeping in (#377).
Could we add a 'very slow' attribute that would run all the notebooks?

javdrher:
The conclusion of #437 is that hardly any speed is gained by running tests in parallel on Travis. I believe a build has access to something between 1 and 2 cores. Because calculations occur in Tensorflow this capacity is more or less consumed by a single test. 

At least, thats for Travis. Locally you can gain quite some speed. If interested I can apply the test parameterization to run a notebook one test each.

alexisboukouvalas:
thanks @javdrher 
I was thinking about the local case to enable running all tests quickly including the notebook tests.
If you upload your local tests, I can give it a go and see the performance difference on a couple of different machines/configurations.
Just making the notebook tests faster should not significantly increase test complexity right?

jameshensman:
@alexisboukouvalas , if this ticket is all correct, could we merge? Improving the notebook tests further seems out of the scope of this ticket. This does provide a means to run most tests quickly, and notebooks will still be tested on travis/codeship. Thanks.

alexisboukouvalas:
Of course these fixes improve the current tests

PR #433, opened 2017-06-20
Hughsalimbeni vgp as svgp

jameshensman:


jameshensman:
@hughsalimbeni this PR duplicates yours, but from within GPflow so that the CI runs.

alexggmatthews:
I  posted a question on the dup #430

PR #435, opened 2017-06-23
Fix for using tf optimiser with GPMC

dtpc:
Add optimizer keyword argument and return result of call to GPModel._compile()


jameshensman:
Hi @dtpc , and thanks for the fix!

I guess we missed this because it seems an unusual use-case to optimize that model using a tensorflow optimizer: it's designed to be MCMC'd.

Would you mind adding a test that covers this change, please?

dtpc:
@jameshensman, I see, thanks. Well I've GPMC to the list of models which test `_compile` here now.

PR #437, opened 2017-06-26
Parallel testing on Travis

javdrher:
Following #432 , I tried some of the Travis features on GPflow for running tests in parallel and split the slow and fast tests into separate builds.
This PR implements the following: 
- Cache for pip
- I merged the test_speeds into my branch in order to distinguish the slow/fast tests based on the attributes. These are split into two separate builds now based on environment variables.
- I'm using the nosetests multiprocess plugin, and launch two processes to distribute the tests
- I tried switching to the container based travis API (still in beta) which should launch builds faster and offer some more hardware. However it stalls completely (you can see a complete failure in the travis build history) and seemed to run slower so I got rid of it for now.

In general, the total speed up of this operation is not spectacular. When travis is overloaded and builds have to wait to be launched, this is by far the biggest delay. There is one more speedup we can achieve (I haven't committed that yet): currently the notebooks all run in 1 test. Using the parameterized library (https://pypi.python.org/pypi/parameterized) this test can be easily split into 1 test per notebook. This then allows the multi-processing plugin to be used again. On my laptop, running the notebook tests went down from 285s using 1 big test, to 171s using a test per notebook and permitting 2 processes. My laptop however has extra cores available hence it has additional cores to parallelize the Tensorflow operations, so I'm unsure if the same speedup can be achieved on Travis. If interested I can easily add this though.

So in conclusion:
- Is the additional complexity worth the speedup?
- Is the split of the fast/slow tests in two builds worth it? If you keep your eye on the travis interface, you get faster feedback for the fast tests. On Github, there probably won't be a notable difference.


jameshensman:
@javdrher Thanks for this, really impressive. 

It's a shame we don't get more speed up, it would be nice if it was faster. I guess to make a difference on github, we'd need it to be faster on codeship too. Recall, the reason we use codeship and travis is so that we can run the python 2 and python 3 tests in parallel. 

To answer your question, I think the extra complexity is probably not worth if for the minor speed up. Users who want faster tests can run them locally (ignoring slow ones).

javdrher:
I agree.

Actually a very useful addition to Travis would be an option to report the status of each sub-build individually to github. That way you'd get a green check here when the fast tests complete (around 3 minutes), while the slow tests still run. As far as I know, Travis can only report when the full build is over, making the seperate build mostly overhead.

markvdw:
We'll have someone look into speeding up tests soon. Perhaps this PR will serve as a good help, but for now I'll close it, given its age.

PR #441, opened 2017-07-05
Likelihood dimension fixes

hughsalimbeni:
This PR clears up the AutoFlow issue for MutliClass likelihood, and also adds some checks for Y for all likelihoods. While helping the inexperienced user is not the emphasis of GPflow, tensorflow often gives rather unhelpful messages in this situation so I think we may as well catch the obvious errors where we can.


markvdw:
@hughsalimbeni, are you still interested in this PR? I'm happy to review and merge once the tests are running again.

jameshensman:
looks to me like the _check_targets function does not get called anywhere.

hughsalimbeni:
Sorry for abandoning this... I'll have a look tomorrow and sort it

hughsalimbeni:
Ok I hope it's working now. (I was only running the tests I thought were relevant, so I missed the exponential..)

Regarding the Bernoulli: the check currently insists on {1,0} or {1,-1}, but actually anything is valid as the code treats as 1 as the positive class and everything else as the other. Perhaps just a warning might be more appropriate

hughsalimbeni:
Thanks @jameshensman for fixing whatever was broken.

markvdw:
Looks good. Can you just confirm that there is no change in behaviour when everything works? And that the only change is a warning when passing incorrect types / shapes into the likelihood?

hughsalimbeni:
Should do. The only change to functionality is ints vs floats. This can break old code (e.g. two of the notebooks), and perhaps it looks strange to pass [1., 0.] for classification rather than [1, 0], but unless AutoFlow is updated to cope with different types then I don't see any other way.

markvdw:
Ah yes, I remember bumping into this as well. Why is this important though? In the past I at least felt that there was some consistency, that all inputs were expected to be floats.

I'm happy to merge, I just want to fully understand the problem before pressing the button. Also, is there a related issue on the issue tracker for this?

hughsalimbeni:
Other way: now everything must be floats. As it was before the multiclass required ints, which didn't play with autoflow

markvdw:
Great, ok I think your current solution is best. Thanks for explaining in detail. Will merge.

PR #444, opened 2017-07-12
Robustmax epsilon learnable

john-bradshaw:
* Making the robustmax epsilon a param. This allows it to be learnable.
* Switching epsilon_K1 of robustmax class to be a property. This means that a user does not need to remember to change epsilon and epsilon_K1 if they want to make changes after initializing the class. Wrote a new test for this too.
* Moving Parentable class to its own module and unlinked this circular dependency with it having to know about its children: there is a comment on the code about how this changes slightly how ParamList works, we should probably discuss this here. Also, this is kind of a little seperate to the robustmax stuff so let me know if you want me to remove it from this PR.


markvdw:
This looks cool, John. I've been wondering about this for a while. In the notebook, the epsilon doesn't seem to change much. It should only makes a big difference when classes overlap right?

Is there a significant difference in behaviour between adding white noise to the kernel and changing epsilon?

jameshensman:
@john-bradshaw I'm afraid this can't be merged until the coverage in improved.

alexggmatthews:
@john-bradshaw this would be a nice addition and shouldn't be too much work to finish. Do you think you will be able to find the time?

john-bradshaw:
@markvdw 
Sorry for the slow reply. So yes I've found that the epsilon makes a difference when the classes overlap (I guess this could be due to not having a powerful enough kernel or label noise). I think sometimes it makes a useful difference early in training, as here the epsilon will bound the log likelihoods and so will stop the gradients being too large.
I'm not sure whether it is exactly equivalent to the white noise kernel, however, I have found epsilon to be an easier parameter to optimise at least.

john-bradshaw:
updated the PR with a test on Parentable.

GPflow/parentable.py
```diff
@@ -0,0 +1,67 @@
...
+    @property
+    def name(self):
+        """An automatically generated name, given by the reference of the _parent to this instance"""
+        if self._parent is None:
+            return 'unnamed'
+        # NB delinked this parent class from knowing about child implementations,
+        # which generally is probably a good idea. However, before a param list could contain two members of the same
+        # class and only return the first one (due to index). Now it will just fail. I think this is probably more
+        # sensible. But need to ensure others don't disagree
+        #TODO: discuss in review
```
john-bradshaw: # Dont yet merge

**We should discuss this before merging**

john-bradshaw:
# Dont yet merge -- todo comment in code to discuss first

john-bradshaw:
superseded by #635

Issue #454, opened 2017-07-20
mean functions are initialized in arguments


jameshensman:
If we have two GPR instances, they share the same mean_function object. This is a problem, because recompilation of one triggers recompilation of the other!

A solution is to make the default argument None, and then instantiate the mean function in the init. 

This may apply to other models as well as GPR.

PR #455, opened 2017-07-21
Avoid recompilation when creating new model instances

javdrher:
As reported in #454 , the Zero mean function as default argument in the constructor of the 6 models results in recompilations when new model instances are created. This should solve this, also added a regression test.


jameshensman:
Nice, thanks!  We could probably get rid of the line

```python
mean_function = mean_function or Zero()
```

from vgp, svgp, gpr etc and just have it happen once in GPmodel. What do you think?

javdrher:
Oops, that makes more sense indeed, modified.

jameshensman:
Perfect. Merged.

PR #458, opened 2017-07-30
External TensorFlow Session and Graph control

awav:
Hello everyone,

This PR contains changes to tf.Session and tf.Graph management within GPflow models and autoflow functions. Now, a user can pass either session or graph into compile method or autoflow-wrapped methods. This extremely useful feature, when GPflow models have to be incapsulated into external computation graph. It also adds better integrity with TensorFlow, as a user can specify default graphs via 
```python
with tf.Graph().as_default():
    gpflow_model.compile()
```
, so the graph is passed to GPflow model implicitly.

I couldn't resist to sort out the mess in some function nearby modified modules and classes, as `pylinter` alarmed me with about ~200 errors, warnings and infos:
gpflow-pylinter

Main modification touched `session.py`, `model.py`, and `param.py` files.
Check examples at `test_model.py` and `test_autoflow.py`.


awav:
An output for simple `setup.py test`:
```
$ python setup.py test
...
----------------------------------------------------------------------
Ran 280 tests in 1370.959s

OK
```

GPflow/param.py
```diff
@@ -519,54 +526,67 @@ class AutoFlow:
...
-                    each_time=settings.profiling.each_time
-                )
-                with storage['graph'].as_default():
-                    storage['tf_args'] = [tf.placeholder(*a) for a in self.tf_arg_tuples]
+                if session is None:
+                    filename = ''.join([settings.profiling.output_file_name,
+                                        '_', tf_method.__name__])
+                    session = session_mngr.get_session(
+                        graph=graph, output_file_name=filename)
+                graph = session.graph
```
john-bradshaw: If someone passes a session and a graph that doesn't belong to that session what happens?
Does it just use the session's graph and ignore this argument?

awav: AutoFlow behaviour replicates model behaviour regarding user defined session and graphs. https://github.com/GPflow/GPflow/pull/458/files#diff-43f82af73d60224cbbb560f1ea85cc1dR128

`session` value is always considered first, if it is non None, then `session` will be used. Otherwise autoflow will create new custom session using `graph` argument. If `graph` in turn is None, then session will pick up default tensorflow graph.

alexggmatthews:
Thanks for the PR. I'm very interested in the proposed Session changes. 

As an aside, I personally don't think it it a good idea to mix two pull request functionalities when their have quite different aims that both touch the whole code base. It makes it hard to quickly see what is new substance and what is tidying.

awav:
@alexggmatthews, I'm of the same opinion. It had to be different PR. For now, It will be hard to filter and revert code back. If you are okay with it I will leave this PR as is and create new one with code after cleaning with [pylint](https://www.pylint.org/). It is really hard to ignore `pylint` messages, hence it would be wise to get rid of all of them once and forever. Also, `pylint` can be added to travis as code smell checker in future.

alexggmatthews:
See also #459

alexggmatthews:
Not to be merged yet.

PR #465, opened 2017-08-03
Fix GPLVM for GPU/float32

alexisboukouvalas:
The GPLVM code was failing when specifying 
float_type = float32
in settings.

Small fix use the correct type as is done in Param.

Testing question: Are we testing the tf.float32 scenario?


jameshensman:
I think we're not testing float32.

markvdw:
Sorry for the wait, thanks for the fix @alexisboukouvalas.

Issue #407, opened 2017-05-02
Page or section with GPflow usages


alexisboukouvalas:
Could we put somewhere a list of projects that are using GPflow?
I am curious what people are doing with GPflow and it would help to advertise other projects building on GPflow.

If you think this is a good idea, perhaps the simplest way would be to add a section at the bottom of the readme.md page?

@jameshensman @alexggmatthews @markvdw  thoughts?

jameshensman:
I am 100% behind this and would happily merge a pr.

AjayTalati:
Dear @alexggmatthews, @jameshensman, I'm very new to this code base, but would to love to work on some simple continuous control applications of it - that would be really fun :+1: 

Is there anyone working on [MuJoCo](https://github.com/openai/mujoco-py) applications? If its not already done I could benchmark against an A3C implementation I have, (which sort of sucks at the moment), and also [OpenAIs GA](https://github.com/openai/evolution-strategies-starter), (which rocks)

Very open to any suggestions you guys might have :+1: Also I've got some free time to do some documentation if I can get it to work, and to port very basic parts of the repo over to PyTorch.

All the best,

Ajay

AjayTalati:
Any code to accompany @jameshensman your paper, 

[Identification of Gaussian Process State Space Models](https://arxiv.org/abs/1705.10888)

would be really insightful

PR #468, opened 2017-08-04
List projects using GPflow on main page

alexisboukouvalas:
List projects using GPflow. This address #407


jameshensman:
nice work, thanks! will merge asap.

PR #469, opened 2017-08-05
fix reshape issue in Param's get_samples_df

ialong:
**UPDATED**

When a Param's transform reshapes the input tensor (cf. DiagMatrix and LowerTriangular), the Param's self.size and self.shape refer to the "foward" tensor which are not the same as the "backward"/"free" ones.

The problem arises here (param.py - lines 198 onwards):

```python
def get_samples_df(self, samples):
        """
        Given a numpy array where each row is a valid free-state vector, return
        a pandas.DataFrame which contains the parameter name and associated samples
        in the correct form (e.g. with positive constraints applied).
        """
        if self.fixed:
            return pd.Series([self.value for _ in range(samples.shape[0])], name=self.long_name)
        start, _ = self.highest_parent.get_param_index(self)
        end = start + self.size
        samples = samples[:, start:end]
        samples = samples.reshape((samples.shape[0],) + self.shape)
        samples = np.atleast_1d(self.transform.forward(samples))
        return pd.Series([v for v in samples], name=self.long_name)
```

```samples``` are in the "free state" so the reshape can fail.
I suggest we replace the last 5 lines above with:

```python
        free_state_size = self.transform.free_state_size(self.shape)
        end = start + free_state_size
        samples = samples[:, start:end]
        samples = [np.atleast_1d(self.transform.forward(s).reshape(self.shape))
                   for s in samples]
        return pd.Series(samples, name=self.long_name)
```

### DIFF
--- a/GPflow/param.py
+++ b/GPflow/param.py
@@ -204,11 +204,11 @@ class Param(Parentable):
         if self.fixed:
             return pd.Series([self.value for _ in range(samples.shape[0])], name=self.long_name)
         start, _ = self.highest_parent.get_param_index(self)
-        free_state_shape = self.transform.free_state_shape(self.shape)
-        end = start + np.prod(free_state_shape)
+        free_state_size = self.transform.free_state_size(self.shape)
+        end = start + free_state_size
         samples = samples[:, start:end]
-        samples = samples.reshape((samples.shape[0],) + free_state_shape)
-        samples = [self.transform.forward(s) for s in samples]
+        samples = [np.atleast_1d(self.transform.forward(s).reshape(self.shape))
+                   for s in samples]
         return pd.Series(samples, name=self.long_name)
 
     def make_tf_array(self, free_array):
--- a/GPflow/param.py
+++ b/GPflow/param.py
@@ -204,12 +204,11 @@ class Param(Parentable):
         if self.fixed:
             return pd.Series([self.value for _ in range(samples.shape[0])], name=self.long_name)
         start, _ = self.highest_parent.get_param_index(self)
-        free_state_size = self.transform.free_state_size(self.shape)
-        end = start + free_state_size
+        free_state_shape = self.transform.free_state_shape(self.shape)
+        end = start + np.prod(free_state_shape)
         samples = samples[:, start:end]
-        if self.size == free_state_size:
-            samples = samples.reshape((samples.shape[0],) + self.shape)
-        samples = [np.atleast_1d(self.transform.forward(s)) for s in samples]
+        samples = samples.reshape((samples.shape[0],) + free_state_shape)
+        samples = [self.transform.forward(s) for s in samples]
         return pd.Series(samples, name=self.long_name)
 
     def make_tf_array(self, free_array):
--- a/GPflow/param.py
+++ b/GPflow/param.py
@@ -204,11 +204,13 @@ class Param(Parentable):
         if self.fixed:
             return pd.Series([self.value for _ in range(samples.shape[0])], name=self.long_name)
         start, _ = self.highest_parent.get_param_index(self)
-        end = start + self.size
+        free_state_size = self.transform.free_state_size(self.shape)
+        end = start + free_state_size
         samples = samples[:, start:end]
-        samples = samples.reshape((samples.shape[0],) + self.shape)
-        samples = np.atleast_1d(self.transform.forward(samples))
-        return pd.Series([v for v in samples], name=self.long_name)
+        if self.size == free_state_size:
+            samples = samples.reshape((samples.shape[0],) + self.shape)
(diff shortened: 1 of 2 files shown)


markvdw:
Thanks for flagging this. I've got a similar fix in my own code. I think it may avoid turning things into a list. Although I doubt this will ever be a practical issue, it technically should be faster and a bit neater.

ialong:
actually, just realised this is not quite general as it does not work for "free states" which are not 1-dimensional (i.e. vectors) (e.g. your TriDiagonalBlockRep @markvdw). I think it's probably best to refactor that one though since all the other ones collapse the free state to a vector

markvdw:
@ialong, is this ready to be merged? Could you make a regression test for this, to ensure that it doesn't creep back in later? I think it could be as simple as running the method on a transform that changes the shape.

markvdw:
I have this for lines 210:
```
subsamples = samples[:, start:end]
subsamples = subsamples.reshape((subsamples.shape[0],) + (self.transform.free_state_size(self.size),))
        subsamples = np.atleast_1d(
            np.concatenate([self.transform.forward(s).reshape((1,) + self.shape) for s in subsamples], 0))
return pd.Series([v for v in subsamples], name=self.long_name)
```
This avoids the `if` statement. However, looking at this (and your code) makes me think this would be the best of both worlds. Isn't the first reshape superfluous? And couldn't the series be made from the first list comprehension? Something like this:
```
subsamples = samples[:, start:end]
return pd.Series([self.transform.forward(s).reshape((1,) + self.shape) for s in subsamples], name=self.long_name)
```
This reshapes everything _after_ the transform is applied. I think this is necessary if the transform doesn't go from flat vectors (e.g. `TriDiagonalBlockRep`), and also fixes the case that you have an `if` statement for. Thoughts?

ialong:
About your code:

1)
```python
subsamples = samples[:, start:end]
subsamples = subsamples.reshape((subsamples.shape[0],) 
    + (self.transform.free_state_size(self.size),))
subsamples = np.atleast_1d(np.concatenate([self.transform.forward(s).reshape((1,) 
    + self.shape) for s in subsamples], 0))
return pd.Series([v for v in subsamples], name=self.long_name)
```
2)
```python
subsamples = samples[:, start:end]
return pd.Series([self.transform.forward(s).reshape((1,) + self.shape) 
    for s in subsamples], name=self.long_name)
```

Line 2 of snippet 1, as far as I can tell, leaves ```subsamples``` unchanged (assuming ```end``` has been defined appropriately as in my code - if not then things would go wrong when the free state has size larger than the constrained state)

I think my code is ready to be merged, **considering the following caveats**: 
It covers these cases (**note**: [change in size --> change in shape] not necessarily vice versa):
1) forward does not alter shape (i.e. both free and constrained states are vectors or have the same shape and hence the same size) - e.g. log1pe
2) forward alters the **size** (and thus also the shape) but operates on a vector  - e.g. LowerTriangular

It does **not** cover these cases: 
3) forward alters the shape (but not the size) - e.g. TriDiagonalBlockRep (this fails because we are reshaping to the constrained shape if the sizes are the same **before** applying forward)
4) forward alters the size (and hence the shape) and does not operate on a vector - e.g. a different version of TriDiagonalBlockRep which has an over/under parameterisation in the free state

The only way to address 3 and 4 is if we had knowledge of the free state's *shape*, not just its *size*. This would mean either longer and inelegant Transform constructors or adding a special-purpose ```free_state_shape``` method to each transform. Personally I don't think there's a lot of urgency to address these cases since they don't currently cause any trouble (in the GPSSM we use TriDiagonalBlockRep externally, not when we construct the Xcov Param so there is no problem there). Keen to hear your thoughts though.

About the **test**, where should I put it? I could just use the LowerTriangular transformation for one that has a difference in **size** between free and constrained, and also a simple positive transformation (which leaves shapes unchanged) on a parameter whose constrained state is not a vector (i.e. where a reshape is required beyond the forward() method if we want to recover the original shape). Those would cover the two handled possibilities.

ialong:
sorry for the long post @markvdw

markvdw:
Wow, thanks for the post, this is a super clear summary. Let me look at the code.

ialong:
@jameshensman @alexggmatthews 
**TL;DR**

This pull request addresses issues which might arise when using transforms that require specific free state shapes (i.e. not vectors) and ones where free state shape (and potentially size) differs from constrained state shape. 
It does so by adding a method ```free_state_shape``` to each transform, leaving ```free_state_size``` and ```backward``` functionalities unchanged. By default ```free_state_shape``` is assumed to be the same as the constrained shape (with exceptions for e.g. ```DiagMatrix``` and ```LowerTriangular```).

markvdw:
We'll continue discussion in #482.

Issue #463, opened 2017-08-02
Titsias evidence upper bound for SGPR


alexggmatthews:
Michalis Titsias derived an evidence upper bound for SGPR in slide 58 

http://www2.aueb.gr/users/mtitsias/papers/titsiasNipsVar14.pdf

Proved in slide 85 . Used in this recent paper by Kim and Teh:

https://arxiv.org/pdf/1706.02524.pdf

I think this would be a useful addition to GPflow and am willing to implement it with the advice of the community. @jameshensman thoughts?

markvdw:
I've got an implementation of this. Happy to contribute.

thangbui:
I've played with this upper bound but found that it's not as tight as the lower bound, i.e. for a decent M, the lower bound can be very good and close to the exact marginal likelihood but the upper bound is still way off the exact.

This is however potentially very useful in diagnosing how many pseudo-points are sufficient for a given dataset.

jameshensman:
A simpler way to work out if you have enough inducing points in to compute
trace(Kff - Qff).

That's actually an upper bound on det(Kff-Qff), which is the entropy of the
conditional p(f|u). If that entropy is zero, then f is completely specified
by u and the approximation is exact.

 Perhaps we should add that as a feature also?

alexggmatthews:
@jameshensman Could you say a bit more. Do you mean the differential entropy of the conditional p(f|u) ? Can't that be negative?

jameshensman:
The matrix of interest is Kff-Qff. This is the variance of the conditional distribution p(f|u). The entropy of that distribution is log det Kff-Qff plus constant.  If there is no entropy in f when we know u, then u captures f perfectly, which is what we want. 

The log-trace is a simple upper Bound on the log det. The first is the log of the sum of the eigvals, the second is the sum of the log eigvals. 

Is that helpful?

The upper bound is appealing, but this trick is simpler and cheaper. This also applies in the non conjugate case, though it can't help you out with slack in the bound due to non-Gaussianity.

alexggmatthews:
Ah OK. You've slipped in a logarithm, but what's a logarithm between friends :)

In the case where Kff = Qff we get log det 0 which equals -Inf . So the `ideal' differential entropy of p(f|u) in question is not zero it is -Inf.  

It seems though that when Kff=Qff  we get exp(Differential entropy) = 0 which corresponds up to a multiplicative constant to the det(Kff - Qff ) you started with. 

[...]

PR #456, opened 2017-07-21
Recompilation of nested models

javdrher:
As reported earlier (#442) the current code assumes the highest_parent is a model, which can be recompiled. If this is isn't the case (in gpflowopt it usually isn't) a recompile flag is set on objects that can not recompile. Because of this, everything which should result in a recompile in gpflowopt is broken (and for instance VGP can not be used as a model as it requires recompilation after data updates).

This code should solve this. I added a rather generic method to parentable to perform some queries on the parent nodes. The behavior is as follows:

-  when a recompile is required, the highest_parent which has a _needs_recompile flag is found and set to recompile. 
- kill_autoflow still starts at the highest_parent as the graph of autoflow methods might contain parts of children which were recompiled. 

I also looked at the reassignment of Params and tried to cover all cases I could think of and added some tests.

PR #470, opened 2017-08-06
Upper bound model and accompanying notebook added.

markvdw:
I added a model implementing Titsias' upper bound, as referenced in #463 (plus accompanying test). I also added a notebook showing some results of messing around with the bound. These will also be part of my thesis, but I think they'd be useful here as well.

The code essentially implements the upper bound, and also allows for minimising the upper bound. For fixed hyperparameters, this is a good way to find the tightest upper bound. It doesn't make sense to minimise this upper bound with free hyperparameters as there are always poor hyperparameter settings where the upper bound is tight. Currently, it's up to the user to use this correctly. This should all be extensively shown in the notebook.

The notebook also discusses using the upper bound to prove when FITC is over-estimating the marginal likelihood, without needing to calculate the marginal likelihood.


markvdw:
This is strange, codeship seems to fail on a notebook I didn't touch, and on a `KeyboardInterrupt`?

markvdw:
Strange, a simple re-run of the test made it work. Perhaps this is non-determinancy? Probably good to keep an eye out for this in the future, but I don't think it's relevant to the current PR. As far as I'm concerned this is ready to go.

alexggmatthews:
Hmm that is annoying. Seems to be happening relatively often at the moment. @javdrher has a similar problem on #456 . Possibly to do with a timeout?

jameshensman:
The notebooks take a long time to test. My guess is that they sometimes time out if codeship is under a heavy load.

awav:
@markvdw I had same issue in #471 with unstable notebook tests. The problem is that notebook error output doesn't tell what exactly happen at a computation cell. I added in-depth error message as the aftermath, but tests had started passing after that :)

markvdw:
A simple re-run of the notebooks worked fine here. I think it was a timeout issue.

Anyway, as far as I'm concerned, this is ready to be merged.

Issue #459, opened 2017-08-01
Discussion on general level of PEP8 compliance. 


alexggmatthews:
Branched from #452 which is a more specific question at the suggestion @javdrher 

I've personally always felt that extreme adherence to a coding standard could slow down development. I'm pretty happy with the compromise level we've existed at since inception which really has come from a common sense agreement between the main developers. 

I recognise as the GPflow community grows and branches we need to keep these things under review.  

Despite the fact the GPflow community is growing I still think there is a danger of distracting effort and attention from more importance questions about code content. 

I don't think we should block PRs unless they have drastically the wrong coding style. There are after all automated tools we could periodically run if people wish.

The specific case of variable names mentioned by @javdrher in #452 is a good example. I'm totally happy for variable names to be chosen to match papers if that it was the contributors feel is most sensible.

To summarize, I think coding standards such as PEP8 can be helpful in a project of our size, but extreme adherence for its own sake it literally style over substance. 

Thoughts?

awav:
Hello everyone!

The code style guides exist to simplify maintenance, get better readability, hence make your life easier as contributor and whole community as well. Different tools can relief development process, something like `autopep8`, `pep8fy`, `YAPF`, plugins to editors: `python-pylint`, `python-tools` and etc.
@javdrher gave extremely good example, but various authors can have divergent preferences for variable names, e.g. `Xnew`, `xNew`,  `x_New`, such flexibility makes code look like a mess. All of these name versions can be adapted as either `xnew` or `x_new` without loss of meaning. Most important part here is a context - why this variable was called in this way, if it is a replication of paper codestyle, where is a paper link and short description what it means. Here is an example from kernels.py:
```
    @AutoFlow((float_type, [None, None]))
    def compute_K_symm(self, X):
        return self.K(X)

    @AutoFlow((float_type, [None, None]), (float_type, [None, None]), (float_type,))
    def compute_eKxz(self, Z, Xmu, Xcov):
        return self.eKxz(Z, Xmu, Xcov)

    @AutoFlow((float_type, [None, None]), (float_type, [None, None]), (float_type, [None, None, None, None]))
    def compute_exKxz(self, Z, Xmu, Xcov):
        return self.exKxz(Z, Xmu, Xcov)
```
[...]

PR #471, opened 2017-08-07
External TensorFlow Session and Graph control

awav:
This is a copy of #458. Please, check description there.
Code coverage doesn't work for forked repos. Discussion can be continued here.


awav:
JFY, tests were not passing only on travis infrastructure, so I had to simulate their environment with docker, (ubuntu trusty, 7.5GB, 2 CPU):
https://gist.github.com/awav/4a9620f2a9a37ce13ab9a2dad96f5b0e

RELEASE.md
```diff
@@ -25,7 +25,7 @@
 
 # Release 0.3.2
  - updated to work with tensorflow 0.11 (release candidate 1 available at time of writing)
- - bugfixes in vgp._compile
+ - bugfixes in vgp.compile
```
markvdw: A bit too enthusiastic? :)

alexggmatthews:
As in #458 not to be merged yet.

javdrher:
Small question regarding the modification of AutoFlow:
- what about calling the same method twice using different kwargs? e.g.
```python
m.predict_f(X)
m.predict_f(X, graph=tf.Graph(), session=tf.Session())
```
it seems that because the method is compiled already the graph will be taken from the storage?
- if unspecified, it seems that a new graph is created for each autoflow (like is now the behavior in GPflow). I think at some point we wanted to enable to re-use the graph and session from a model. This PR would definately enable such scenario but the syntax would be slightly verbose. Am I right to assume this isn't the main purpose of these changes?

awav:
@alexggmatthews, Hello Alex, is this PR cannot be merged because of #459, and you want to discuss code style rules first? Current PR doesn't rename model variables or functions. It does have minor modifications in renaming like `_compile` to `compile`, `c` (which is config) to `cfg`, but nothing critical to model descriptions.

`codeship` test has failed for some reason, and I can't check the error log. How it can be fixed?

awav:
@javdrher, Thanks! You I helped me to find another scenario, which was not covered. Please check AutoFlow docstring,  https://github.com/GPflow/GPflow/pull/471/commits/bf770cdc84170f13ff744754c10835f301212b2c#diff-da2a47880fa50096cd5c579c2f45c874R554

> if unspecified, it seems that a new graph is created for each autoflow (like is now the behavior in GPflow). I think at some point we wanted to enable to re-use the graph and session from a model. This PR would definately enable such scenario but the syntax would be slightly verbose. Am I right to assume this isn't the main purpose of these changes?

Hm, AutoFlow can wrap `Parameterized` object functions which might not have session and graph fields, otherwise I will have to change it and adds more complexity. But, you still can use model session using this PR, for e.g.:

```
m.compile()
m.predict(X, session=m.session)
```

> Am I right to assume this isn't the main purpose of these changes?

This PR will allow you to integrate your model into any existing graph and using sessions created outside of GPflow.

javdrher:
@alexggmatthews I'm interested in this (personally, not for GPflowOpt). Two weeks ago I was looking at running GPflow on GCE cloud ML: this is a prime example of a service with an externally managed graph. I succeeded in pulling out the right graphs from GPflow but it wasn't done very nicely. 

From the GP session yesterday I remember that there are quite some approaches around now for scaling GPs into clusters and about 2 papers also did this. I think this PR would be a nice step for GPflow to extend the capabilities in this direction. With respect to the way it is done in this PR, I'll have a closer look next week. Actually thinking I might try running it on the cloud and see how it goes :)

awav:
> From the GP session yesterday I remember that there are quite some approaches around now permitting scaling GPs into clusters (including VGP of course) and some interest in this. About 2 papers also did this in practice and ran on very large datasets.

@javdrher Could you share links to these papers?

javdrher:
Sure,
http://proceedings.mlr.press/v70/peng17a/peng17a.pdf
http://proceedings.mlr.press/v70/cutajar17a/cutajar17a.pdf
(I still have to study them in more detail)

alexggmatthews:
@awav Originally delayed because of pep8 dependency. Now because I think this is an important and useful change which as I understand it might be the first of many. Therefore keen to discuss in person. The delay is not a reflection on the quality - just that James was unavailable for lunch this week.

alexggmatthews:
This is currently blocked by the issues around #472 and is also awaiting a review from @jameshensman  .Having discussed it with @awav in person I am happy for it to be merged when those two dependencies are done.

GPflow/param.py
```diff
@@ -519,54 +526,118 @@ class AutoFlow:
...
+
+            :raises ValueError: when `**kwargs` contains unknown key-valued
+            parameters or user passed both session and graph parameters and
+            session references to different graph.
+            """
+
+            if not set(kwargs.keys()).issubset(['session', 'graph']):
+                raise ValueError('Unknown arguments passed.')
+
+            session = kwargs.get('session')
```
jameshensman: why not just `kwargs['session']` ?

GPflow/param.py
```diff
@@ -519,54 +526,118 @@ class AutoFlow:
...
+            graph = kwargs.get('graph')
+
+            if session and graph and session.graph != graph:
```
jameshensman: perhaps: `if (session and graph) and (graph is not session.graph)`  -- same logic but clearer?

jameshensman:
Review: this PR makes a lot of sense. There are many changes that are cosmetic, but the main ones are:

 -- improvements to Autoflow to allow passing sessions and graphs
 -- tests accordingly
 -- some changes to session management
 -- edits to the tests to allow for use of the defauilt graph withot running out of memory (the teardown now resets the default graph). 

I'm really happy with the changes, and ready to merge this. @awav , I added you to the codeship team, you should be able to log in and see the error now.

awav:
@jameshensman, Hello James. Thank you for code review. I'm sorry for incoherence, but I have updated this branch once more time. As we discussed on Monday, @alexggmatthews and I investigated the problem with arbitrary crashes in this particular branch which caused appearance of new script './run_tests.sh'  intended to solve them. In #472 you can find description why it happened.
Could you review this PR once again? Others are welcome.

awav:
@jameshensman, new commits mostly modify tests for `tensorflow.test.TestCase` integration and add feature for using tensorflow default session.

javdrher:
Small question, following up on the need to close sessions:
if no external session/graph is supplied and compile is called twice, (due to e.g. a recompile): `self._session` does not seem to be closed while a new one is created?

awav:
@javdrher, that's good question and big topic for discussion. It would be wise to leave session life control to the user, because he/she still has access to the session via property `m.session.close()`. Maybe, it would be better to force user to create default session or use context manager and never try to create session inside model for sake of safety at all.

Particularly, the case when user modifies model parameters after first compilation, causes some troubles: 1. we rebuild whole graph from scratch and leave dangled session from previous usage, 2. use default graph and session, but still we add new variables and operation to the graph, hence consume more resources. That's good feature (changing parameters on the fly), but costs a lot for now. The tensorflow works differently, all things in it are immutable. Right hacky solution fot it is to outwit the system and use `assign`, `condition` operations to make tensorflow variables modifiable on demand.

The current PR is just small step towards better solution and is not intended to solve all problems or cover all features. But, session discovery has big impact on whole GPflow architecture, and, possibly, we need to reconsider some mechanics in it.

javdrher:
I'm guessing there are two distinct cases: 
- GPflow manages the session (like it does today), then a close() and spawning a new session should do the trick to avoid dangling resources?
- External session: then the story is a lot more complicated, it seems natural to make the user responsible in this case. Note that this makes things a more involved (when changing parameters, you'd need to be very familiar with the way TF manages resources). So I guess in terms of usability keeping the GPflow managed option seems a good thing?

The session management is potentially also an issue for AutoFlow: I guess the following is potentially problematic:
```python
with tf.Session().as_default():
           m.method(x)
m.method(x)
```
I don't think that would be really a bug, just something to be very well aware of.

PR #472, opened 2017-08-07
Script for running GPflow tests in sequential and parallel modes.

awav:
Hello all,

This script runs tensorflow based tests in distinct processes. It prevents bad memory accumulations which can lead to crashes or slow runs on resource limited hardware.
In #471, I found that adding variables to default graph can cause significant slowness, even after calling graph resetting, `tensorflow.reset_default_graph()`. Previously, this issue was not present as it was mitigated by creating new graphs, instead of reusing default one.
The script has two modes parallel and sequential. In parallel mode it detects number of cpus and spawns `nosetest` processes against of each test files.
This script is used at #471. Parallel mode doesn't cut the test time execution significantly, only 1-2mins.


javdrher:
Specific reason to have a seperate script? Nose has by default a multiprocess plugin, it allows executation as follows:
`nosetests --nologcapture --processes=4 --process-timeout=120`
It requires a static directive in the test class to execute tests in parallel which I added in #437. The problem is we saw no actual performance improvement on travis. However locally it makes a big difference.

awav:
@javdrher, `nosetests` distributes load across processes=workers. Each worker runs all tests at same memory scope and thereafter tensorflow will do allocation within it, it caused #471 failing badly. The script runs each test suit (test file) as distinct `nosetests` process and alleviates memory issues of using default graph.

javdrher:
did you add the `_multiprocess_can_split_` directives in the test classes? If not the multiprocessing plugin starts N processes but runs all tests in the same process. Nonetheless I agree it wouldn't solve your issues in case of a sequential build.

awav:
Yes. The problem with `nosetests` is that it distributes tests across N processes, it does make sense to have N being equal to the number of cpus, 2 cpus in case of travis. So, you have N=2 memory layouts. Even in that case tensorflow struggled with deallocation, so that process either got killed or hung out.

awav:
@javdrher, for example, you can check this run with `$ nosetests --nologcapture --processes=4 --process-timeout=600`, `_multiprocess_can_split_` is turned on: https://travis-ci.org/awav/GPflow/builds/262027107,
result: _No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself. _

awav:
@javdrher any thoughts/objections? :)

javdrher:
@awav this week I'm at a conference next week I'll have a closer look when I'm back. I find it surprising TF is having issues with GC causing extreme loss of performance so I'm definitely going to reproduce it. 

We got rid of the multiprocessing plugin because of sudden extreme Travis slowdowns combined with the fact that nose only gives any feedback when all processes have finished for me with python 3.x builds (do you have any idea why it did in the build you posted?). Hence our builds got terminated because after 10 minutes nothing got reported: e.g. [this build](https://travis-ci.org/GPflow/GPflowOpt/builds/256264395) for instance.
From your Travis log that clearly isn't the issue here. However, it seems that most tests execute quickly (3 minutes) and the problem is caused by the notebook tests: those are not running in parallel in your build because they are launched from a single test function. In GPflowOpt we are [splitting](https://github.com/GPflow/GPflowOpt/blob/1b38e37ee10550f00bb3291443a1bcb45163ec62/testing/test_notebooks.py) this up.

alexggmatthews:
@awav to what extent are the memory issues with the default graph alleviated by using tf.reset_default_graph() ? We had a related problem early on in development which was largely fixed by calling the reset in the unit test constructor. If there is still and issue with the default graph I'd like to establish exactly what it is. If we can't fix it maybe we shouldn't be using the default graph.

alexggmatthews:
@awav @markvdw @jameshensman I had into this issues surrounding this as discussed. 

TensorFlow does indeed have its own unit test class that inherits from the class we are currently using:

https://www.tensorflow.org/versions/r1.3/api_docs/python/tf/test/TestCase

https://www.tensorflow.org/versions/r1.3/api_guides/python/test

This first linked page contains the following quote:

> This method behaves different than session.Session: for performance reasons test_session will by default (if graph is None) reuse the same session across tests. This means you may want to either call the function reset_default_graph() before tests, or if creating an explicit new graph, pass it here (simply setting it with as_default() won't do it), which will trigger the creation of a new session.

This implies that they deliberately use the same graph across tests. I confess myself slightly confused as to why they want to do that but obviously it has significance for the current discussion.

awav:
@alexggmatthews, It does make more sense now! The graph keeps only interaction definition of variables and operations upon them. But session is a real culprit in this case, as it performs memory allocations and computation itself. If you look at https://github.com/tensorflow/tensorflow/blob/r1.3/tensorflow/python/framework/test_util.py#L275, you can see that session is properly closed in the end of the test case and GPflow tests/code doesn't do it at all. It means that GPflow abandons one session and creates another one without release resources, this is kind of a leak. Here is a quote from session documentation:
> A session may own resources, such as variables, queues, and readers. It is important to release these resources when they are no longer required. To do this, either invoke the close() method on the session, or use the session as a context manager.

alexggmatthews:
Good detective work. Do you think the right fix is therefore to move over to the TensorFlow unit test class?

awav:
@alexggmatthews, Yes, it is worth to try. I'm experimenting in parallel branch 'artemav/session-graph-assign-experiments'. I haven't finished it yet though. I will publish results as soon as possible.

awav:
@alexggmatthews, @jameshensman Okay, I did it! https://travis-ci.org/GPflow/GPflow/builds/265325416.
The 'artemav/session-graph-assign-experiments' alters tests using proper session closing via `tensorflow.test.TestCase`. It helped avoid test crashes in #471. This PR with `./run_script.sh` script was created only because of these crashes.
Recap:
- Once created session always must be closed to release resources. It can be done via using context manager or explicitly closing session. `tensorflow` provides nice wrapper for TestCase and `test_session()` function which turns default session on and closes it in the end of each test.
- 'artemav/session-graph-assign-experiments' changes only make sense in context of #471, as it has capability to use default graph and default session bypassing tedious hacks.
- current tests are flawed and 'broken'

I'm about to merge 'artemav/session-graph-assign-experiments' into #471. @jameshensman and @alexggmatthews could you review this code once more time? :)

markvdw:
fixed by #471

PR #475, opened 2017-08-09
Binned poisson notebook

mathDR:
This is a notebook demonstrating the binning capability for the Poisson likelihood.  Two datasets are included:  coal.csv and pines.csv.  The pines (2D) section is heavily borrowed from @jameshensman VFF notebook (can you provide a source for the pines dataset?)


markvdw:
@mathDR: There seems to be a type issue in the tests, a likelihood uses float32 when float64 is expected. Make sure to use the types that are set in `GPflow.settings`. Are you still interested in getting this PR in?

markvdw:
Also, it looks like this PR provides two things, a point process model and the BinnedPoisson likelihood. Was there any consensus in a discussion about how these two extensions fall within the scope of GPflow?

jameshensman:
@mathDR I think there might be some git-based confusion here. Did you mean to add the vbpp model? It doesn't look like it implement the Lloyd et al method. Also, the binnedPoisson class looks like it might be old code that is superceded by the current Poisson class (which does have a binsize attached).

Could you tidy up the PR, please? Remove the commits that add small changed to the notebooks, remove the notebook checkpoints (we don't want to git those), remove the binned Poisson likelihood, and the vbpp model, but keep the notebook, which is starting to look good!

The notebook has lots of room for some more analysis: you're hinting to the user that the binsize is important, could you explain why?

mathDR:
Hmmm, I don't know why the PR has old commits.  As for added analysis, I will extend the notebook more.  @jameshensman you are right:  I am mixing up some vbpp models and this.  I will separate them.  I have time later this week to revisit.

markvdw:
@mathDR: Currently this PR contains too many different bits and pieces. It's probably best to separate the minimal contribution and place this in a new PR. Could you open a new one once you've separated them out? We're trying to keep things a bit more updated and tidier.

mathDR:
Will do.  Thanks for the ping.  This fell off of my plate...

Issue #452, opened 2017-07-18
GPflow python package name doesn't follow PEP8


awav:
Hello everyone,

According to PEP8 - [packages and modules](https://www.python.org/dev/peps/pep-0008/#package-and-module-names) should have short and lowercase names. The irony is that [contribution](https://github.com/GPflow/GPflow/blob/master/contributing.md) markdown file tells that GPflow code should follow PEP8. But it's not too late to change, as GPflow is not present on public pip/anaconda repositories yet and version is below `1.0.0`. 

The transition can be done in three primitive phases:
1. Inform people that current version will be deprecated and replaced with new one on XX/XX/2017. (Warn of deprecation in Readme file or issue on Github).
2. Provide a script which updates the code with GPflow.
```bash
#!/bin/bash
PROJECT=$1
find "${PROJECT}" -type f -exec sed -i -e 's/GPflow\./gpflow./g' {} \;
find "${PROJECT}" -type f -exec sed -i -e 's/^import GPflow/import gpflow/g' {} \;
find "${PROJECT}" -type f -exec sed -i -e 's/^from GPflow/from gpflow/g' {} \;
```
3. Merge PR into the master.

I have created #451, where I renamed only code/module/package related stuff. The documentation, citings and docstrings refer to GPflow name as before.

alexggmatthews:
I personally think the GPflow typesetting is now too entrenched to change this.

awav:
@alexggmatthews, It is fair enough, and it is also a reason why the script for updating must be available for everyone during transition. I think the renaming will be harmless as GPflow is not available through pip and everyone who wants to get a copy either go to github and can see warning message or download specific SHA of GPflow. I would change package name now and make it standard compliant. But, community opinion is more important and I would like to hear what people think of it.

Example: `ternsorflow` transition from `<1.2` version to `1.2`, when some function interfaces were changed and Google provided a script for updating user code.

alexggmatthews:
An example of a heavily used package that doesn't use this convention is IPython. It doesn't seem to have hurt them to badly.

Another example is GPy. 

Changing the naming convention would clash with the published paper on the package:

http://www.jmlr.org/papers/volume18/16-537/16-537.pdf

awav:
[...]

PR #451, opened 2017-07-18
GPflow python package name doesn't follow PEP8

awav:
Disscussion around this topic is in #452

PR #477, opened 2017-08-09
GPflow python package name doesn't follow PEP8

awav:
Copy of #451. Codeship and code coverage test didn't work for fork.
Disscussion around this topic is in #452.


alexggmatthews:
@awav thank you for your PR. On the basis of the consensus of the community in the discussion in #452 I am happy to merge this when it is ready to go.

@awav We'll need to resolve the merge conflicts here obviously. Maybe re-run the automated script on new master?

The codeship failure is an interesting one. It relies on import GPflow but this command is unfortunately shared across all branches, including ones that use the old style. @jameshensman and I may need to treat this one in a special way to get it in. @jameshensman thoughts?

awav:
@alexggmatthews, I've updated PR. `codeship` is failing in other branches too, since yesterday (August 9th). I do not have an access to take a look at the failure though. I see this at `codeship` "Oops! You can't access this page."

awav:
@alexggmatthews, @jameshensman Could you give me an access to `codeship`, I would try to fix this thing, whatever it does :)

alexggmatthews:
Because this requires some special treatment the GPflow team have decided it will be merged at a specific time:

**29th August at 1pm UK time.** All PRs afterwards will need to use the new name in the code. All PRs before will use the old naming convention. We don't anticipate that changing between the two naming conventions will be too difficult.

PR #494, opened 2017-08-29
Added upper bound to regression models.

markvdw:
New version of #470 (addressing #463) that adds the upper bound as a method that simply computes it.


alexggmatthews:
Thanks for the PR @markvdw looks good to me.

markvdw:
Great, I'll merge once the tests are confirmed.

alexggmatthews:
Looks like we might have a few issues from the merge.

PR #498, opened 2017-09-03
One parameter - one variable

awav:
This PR is a big milestone towards `GPflow 1.0`. It contains drastic changes in a way how GPflow operates with tensorflow tensors and operations. The motivation of this work is a GPflow integration with external tensorflow models and extension user capabilities to define their tensors and assign them to parameters.

Major features and updates:

* Clean abstraction classes and core modifications in Param and Parameterized classes. Current class hierarchy implementation has web-like structure and causes error prone code changes.
* Model does not flat parameters anymore, just because of that, a lot of code can be removed.
* Parameters accept external Tensors.
* In fact, each parameter distinct trainable (or not, if fixed parameter were set up) tensor or tensorflow variable.
* Parameter can be compiled, read as 'build into tensorflow graph', separately from the model and assigned afterwards.

By this PR, I would like to spawn a discussion around the topic, hear your objections, opinions and new ideas. Everybody is welcome!

NOTE: This PR is under development and code may change over time.


markvdw:
I'll need to look over this a few times to fully grasp it, but it looks super interesting. Is the code in working state? Not that I'm expecting it, but I can't see how `compile()`can work, given that it doesn't define `self._minusF` anywhere.

One thought that I had on this matter, was that perhaps we should drop the whole `compile()` functionality. All it really does is create a numpy interface for the objective function. Could we not do this using AutoFlow (+ maybe some extra processing) directly? It seems strange to have two separate methods for compiling functions.

awav:
> Not that I'm expecting it, but I can't see how compile()can work, given that it doesn't define self._minusF anywhere.

You are right, it doesn't work for now, because I am porting autoflow, and some other stuff are 
 still changing. But param structure and base interfaces are settled.

`compile` function will have a bit different meaning, it will be a trigger for any compilable object: from mutable to semi-immutable tensor representation. If fact you are right, we don't need compile function, if we do not care about aligning tensorflow variable names with model tree structure. For e.g. we have model `m` with kernel `k`, which in turn has `xmu` and `xcov` parameters. Tensors for `xmu` and `xcov` could be named as `m/k/xmu` and `m/k/xcov` respectively. This approach has a few benefits like readability and handy diagrams, charts at _tensorboard_. If we need that feature we must have single compile point, which will build all tensors in the graph and name them appropriately. 

>  It seems strange to have two separate methods for compiling functions.
Could you point out to the exact code, I am not sure that I understood you correctly. By `compiling` in two separate methods, do you mean autoflow wrapper and `compile` method?

markvdw:
I'm not sure I understand. Within AutoFlow, wouldn't the same namespaces be applied to the Tensors as well?

awav:
@markvdw, @jameshensman, @alexggmatthews, @javdrher, @john-bradshaw 

It depends. First of all, I think it is important to recall that in current implementation parameter and module names are virtual, i.e. they have no "name links" to flat tensorflow variable. Hence it doesn't matter where and when you run compile function, full name of assigned parameter to the model will look as `model.param`, or `model.kernel.param`. Once GPflow parameter becomes distinct tensor in the tensorflow graph, according with this PR, you can't modify it in most cases, including tensor name. Let's go back to the previous example: we have model `m`, kernel `k` with params `xmu` and `xcov`. E.g. you have created the kernel and compiled it, parameters transformed to tensors with names `k/xmu` and k/xcov`. Then you decide to use this kernel in model `m`, and you assigned it, as shown here: `m.k = k`. The problem is that the model can't change tensor names, hence you can't depict tree structure of the model like `m/k/xmu` and `m/k/xcov`.

We need to decide how GPflow should build the graph's tensors, @markvdw, @jameshensman, @alexggmatthews, @javdrher, @john-bradshaw:

* Compile-in-time. Once parameter constructor is called, i.e. `__init__`, it creates all related tensors in a tensorflow graph. By default it can use `tf.get_default_graph()` or accept custom graph as parameter. The benefit of this approach is that model, parameter or any other parameterized objects can be executed with any session based on the same graph.
* Post-compile, viz., when user defined the model, he/she calls compile and builds all operations and variables, which later can be used in optimisation and cetera. But, user is still able to create each peace of the model (parameter, kernel) separately, compile, compute and whatever he/she wants to do with it and then reuse it as a part of another model, consciously breaking tensors naming structure. The benefit is that user can transfer tree structure definition into tensor names and observe better visualisation in tensorboard, it also improves readability and model comprehension.

alexggmatthews:
Thanks for PR which I think is a particularly important direction for the codebase.

Looks like the GPflow team have some homework to do discussing this.

.travis.yml
```diff
@@ -3,21 +3,5 @@ dist: trusty
...
-  - conda info -a
-
-  # Replace dep1 dep2 ... with your dependencies
-  - conda create -q -n test-environment python=3.5 numpy scipy pandas nose nbformat nbconvert jupyter_client jupyter matplotlib
-  - source activate test-environment
-  - pip install https://storage.googleapis.com/tensorflow/linux/cpu/tensorflow-1.2.1-cp35-cp35m-linux_x86_64.whl
-  - python setup.py install
 script:
-  - nosetests -v --nologcapture testing
+  - echo "Hello world!"
```
alexggmatthews: Is this just here to disable travis for the time being?

alexggmatthews:
@awav  Looks like contains some style changes? I still personally think that if we can separate that sort of thing out it makes it easier to get at the content of complex PRs like this. Perhaps this is because it is an early stage PR.

To help me get to the meat without us having to do a new PR, which files are the functional changes in at the moment, please?  Thanks.

markvdw:
I would be up for having another "round table" meeting to discuss this PR. I have a whole host of questions, which I think would be useful to others, and which I think would be easier to discuss face-to-face.

.travis.yml
awav: yep, I didn't want to run travis agains raw branch

awav:
I would love to discuss these changes as soon as possible. Unfortunately, I can meet you face-to-face in Cambridge only after Monday next week. But, we could setup skype, hangouts conference call on this week.

@alexggmatthews, It does have some code style changes, but only in `_settings.py`. Other things   are relevant to the task. This branch also includes `.pylintrc` from #499, which I hope will be in master before this PR.

javdrher:
I am going to need some time to get up to speed with these changes. I'll try to free up some time for this. 

Regarding your question: from my current understanding it seems that compile-in-time is a serious change and terminates many use cases which are currently possible. I'm assuming here that any change to what is now the free state requires a new model constructed from scratch. However, the constructor is not a standard interface which means an unattended model scenario requires an extra layer of abstraction. Right now, the GPflow model can be modified and the recompilation can be regarded as creating a new model. Should the compile-in-time option be chosen it would break a lot of things in GPflowOpt.

I guess the post-compile option is more in line with the current behavior? Therefor I think approach should be favored.

alexggmatthews:
This is a complex and important structural change. 

Agree with @markvdw that this one therefore would be worth a meeting. 

Another possible for format would be for @awav to present slides including diagrams. Given the interest we could video the talk and questions and put it online for everyone else.

If this is the way we decide to go I'm happy to organize a room here in CBL. 

I also am away next week.

awav:
@alexggmatthews sounds like a plan. I will prepare slides then. Let's define the date. I'm available next week, Wednesday-Saturday, around 12pm-3pm or after 5pm.

awav:
I'm sorry @alexggmatthews, I didn't notice that you will away next week. Will you be able to attend meeting online? ;)

awav:
@markvdw, @alexggmatthews, @jameshensman, @javdrher, @john-bradshaw I've created a branch for release candidate `GPflow-1.0-RC`. This PR is no longer relevant.

PR #499, opened 2017-09-04
Pylint configuration file

awav:
Hello everyone,

This PR is a consequence of discussion at #459 issue. I added modified skeleton of `.pylintrc` configuration file to the GPflow.

Modifications:

* 4 jobs by default.
* 100 chars per line.
* Uppercase and lowercase 1 char size variables are allowed. Example: `X`, `K`, `k` and cetera.
* Extended list of good names or exceptions. If code is a copy-paste from paper or a book and it does makes sense to use abbreviation names, but they don't fit PEP8 rules, the coder can add them to that list. Examples: `AAT` , `LB`, `Aerr`, `Luu`, `Lvec`, `Kuu`.

Any feedback is welcome here!


markvdw:
Looks great. Feel free to merge!

For information: the CI tests don't yet require pylint to pass, do they?

alexggmatthews:
Thanks for the PR. 

A discussion point that occurs to me. If I understand this correctly we need to maintain a complete list of variable names that don't match pep8 in the style file? How labour intensive is entering all variable name exceptions going to be? How do we feel this compare in terms of cost/benefit of just allowing a very broad class of variable names?

alexggmatthews:
Second discussion point. Given that this will be merged as non-binding on tests at first, to what extent is it worth getting the style as good as possible now rather than allowing a trial period and then amending the style as issues emerge? 

If we are going to spend time on it now I think @jameshensman should have a look too.

alexggmatthews:
If we could delay merge until discussion done that would be good. If we plan to amend as we go along discussion shouldn't need to be very long. My apologies @awaw @markvdw .

awav:
@alexggmatthews 

>  If I understand this correctly we need to maintain a complete list of variable names that don't match pep8 in the style file? How labour intensive is entering all variable name exceptions going to be? How do we feel this compare in terms of cost/benefit of just allowing a very broad class of variable names?

Glad you asked about it. I have same concern, the name list can grow big over time. I see two solutions to the naming problem, i.e. _how to allow only meaningful variable and function names and alert names which doesn't fit PEP8 rules and doesn't bare much information from scientific resources_:

1. is in this PR - keep list of "good" names. If `pylint` is included as travis testing stage, then PR reporter and reviewers will see that mismatch and such style errors will never pass through.
2. regexp rules allows vague range of names. The reviewers must carefully read the code and filter broken names.

awav:
@alexggmatthews 

> Given that this will be merged as non-binding on tests at first, to what extent is it worth getting the style as good as possible now rather than allowing a trial period and then amending the style as issues emerge?

1. I would create a PR with only style changes (and merge it into master). 2. Maintain code style as close as possible to `.pylintrc`, it must be short period of time. 3. Create `.travis` test stage with pylint. 4. Fix rest style inconsistencies which could appear during step (2).

alexggmatthews:
@awav sorry my meaning wasn't at all clear- I meant the style file rather than the style. i.e how much time shall we spend on this style file PR before we start the test run? The test run might go better for some discussion at this point. My apologies.

jameshensman:
I think that a good solution is to ignore variable names. 

I don't see that variables names cause a huge problem in the current codebase. People have been very good at applying the pep8 style where appropriate, and using mathematically derived variable names when that makes the code clearer. 

I'd be happy to merge this if the variable name checking were disabled.

jameshensman:
So, I just had a look at which names the linter doesn't like. I only looked at variable names, and going through the list, there were only a few that I agreed with. 

There were some camel-case names that should be snake case:
log_determinant_term, not LogDeterminantTerm

And there were some short names that might be okay in context, but I didn';t recognise:
no, ps, ka, kb. 

But those might be okay in the context of the function if they correspond to mathematical quantities.

But the vast majority of the hits were things where it's clearly good practise to match the variable name to the mathematical name, e.g.:
LiLq, cholXcov, X_denominator, etc.

jameshensman:
I also looked at argument names: each time where pylint complains about an argument name, I think it's actually a good name:
X, X2, Fmu, Fvar, etc.

awav:
@jameshensman, what's about function and attribute names?

awav:
No need for adding it to the master. GPflow 1.0 RC has this file already.

PR #500, opened 2017-09-14
Fix combined (sliced) kernel input_dim 

dtpc:
Fix for a small bug where the input_dim of a combined kernel is incorrectly calculated if one one of the kernels active_dims was specified as a slice.


markvdw:
Continuing the discussion in #515, so the unit tests run. I think it should be able to go in as-is. Many thanks!

PR #501, opened 2017-09-14
Kernels between derivative and regular observations.

john-bradshaw:
Allow GPs between regular observations and derivative observations (up to second order derivatives).

Currently this script implements a general kernel derivative wrapper that uses tf.gradients to calculate the derivatives and so should be general over many different kernels. It also includes a specialized RBF derivatives kernel. This uses the analytical derivatives of the RBF kernel as the tf.gradients approach is slow.

See below for the output of the derivative_gps script I put in the doc notebooks:
all_images

In this picture we have regular RBF GP and the DerivRBF GP (using only the observations; the observations and first derivatives; and the observations, first derivatives and second derivatives). We also use the  general kernel derivative wrapper. The time taken for each to run is given in the subplot title.

_Note_
This current code will not work with the inducing point GPs. This is because we pack the derivative information in with the x data (or z in case of inducing points). These are restricted integer values indicating the order of the derivative. They should therefore either be fixed and not optimized along with the rest of the inducing point locations, or optimized in a special manner if one wants derivative inducing points.


john-bradshaw:
The CI is currently failing. Travis is because I'm giving the  keyword argument `output_type` to tf.argmax.
This was only introduced in TF 1.3 and I assume this is not used by Travis.

I will remove the uses of this argument.

alexggmatthews:
@john-bradshaw Nice work. Thanks for the PR.
@john-bradshaw @jameshensman how difficult would it be to support inducing points? Would it be potentially a bit messy to have this in GPflow main without supporting that?

markvdw:
I think making inducing points work would fit nicely with a way we discussed to make inter-domain inducing variables work (right @jameshensman?).

If I understand @john-bradshaw correctly, the problem now is that the input takes some non-differentiable inputs. The way I see it, while the kernel supports all kinds of derivative inputs, we could happily live with a variational approximation which only uses inducing points (and *not* derivative observations) as inducing variables. The Kxz matrix would just be covariances between derivatives and points (as already implemented), and Kzz would be the covariance for points (as usual). This would fit nicely in the "inducing features" framework we discussed and can be implemented with code that @john-bradshaw has already written.

john-bradshaw:
Hey @alexggmatthews 
So yes @markvdw is right the problem is that the input is concatenated with derivative information (we'll call these the DI-columns for the derivative information columns). So for instance say I have two observations on a one dimensional problem:

derivative_information

The derivative kernel expects the information about what kind of derivatives the corresponding observations are to come in with the data as an extra appended set of DI-columns. So for instance in my simple image above I'm saying the first observation is a first derivative of the function at x_a and the second is a normal observation at x_b.

So the problem is when you want to use inducing points the kernel will expect the same DI-columns with the inducing points. If you want inducing points (not inducing derivatives) then this will mean that when you feed in the Z matrix you would do something like `np.concat([z, tf.zeros_like(z)], axis=1)`.  Which should work fine. (and also you could easily choose to have inducing derivative points by concatenating a matrix of one hot values instead of the zeros as the DI-columns).

The problem is with the training. When calling `model.optimize` GPflow will try to optimize these extra columns in a continuous way that does not make sense. 

I haven't tested the following. But maybe something like this would allow you to use the scalable GPs with this code (as long as you did not fix the inducing points as otherwise the `get_feed_dict_keys` method would also need to be overwritten)?:
```
# model has been created  as model as normal

class InducingPointsParam(gpflow.param.Param):
    def __init__(self, *args, **kwargs):
         super(self).__init__(self, *args, **kwargs)
         self._actual_tf_array = None
    
    @property
    def _tf_array(self):
        return tf.concat([self._actual_tf_array, tf.zeros_like(self._actual_tf_array)], 1)

    @_tf_array.setter
    def _tf_array(self, value):
        self._actual_tf_array = value
   
model.Z = InducingPointsParam(model.Z.value)
```

markvdw:
Yes, that's exactly the picture I had in mind too.

Like I said, there's some thought going into how to handle inducing variables in a more general way, which is actually quite similar to your solution. I think the question is whether you want to get this in now, or if you want to wait for the more general inducing variable framework. I tend to agree with @alexggmatthews that it would be weird to have kernels which don't support sparse approximations.

Bonnevie:
Exciting!
Not to beat my own drum, but do you have any ideas about how it relates to PR #328 which also added gradient functionality? One central difference is that I never found out how to use `tf.gradients` to calculate the kernel derivatives - at the time, it seemed that it wasn't quite possible. 
Could you briefly relay how derivative kernels are computed and then combined with regular kernels/observations with appropriate parameter sharing?

john-bradshaw:
@Bonnevie: hey that's really cool! So you have the gradients for all stationary kernels?
Well I guess this could just be plugged into the framework above. I imagine its quicker than the generic version I wrote which uses`tf.gradients` so would replace this for stationary kernels, and so this fallback version would only be needed for the other kernels. 

> Could you briefly relay how derivative kernels are computed and then combined with regular kernels/observations with appropriate parameter sharing?

So these derivative kernels wrap the kernel you want to use. You can then specify (as an extra matrix concatenated onto the end of the regular data matrix) whether the corresponding observation at that point is an observation of a derivative. There are two versions. One a generic derivative calculator that uses `tf.gradients`, and one a specialised RBF derivative kernel.

For the generic derivative wrapper class (`DifferentialObservationsKernelDynamic`) it uses a TensorFlow map (`tf.map_fn`) and a TensorFlow switch (`tf.case`) to go through each element of the kernel matrix and sub in the correct differentiated kernel (calculated using `tf.gradients`. Unfortunately we have to use the map function as `tf.gradients` sums the `ys` when taking gradients. It is therefore rather slow.

The `RBFDerivativeKern` class is designed to be much faster. Here I have hand-coded the analytic derivative expressions. There are 9 possibilities when you limit yourself up to second order derivative observations. All are computed and the correct one for each place in the kernel matrix is selected at the end using a `tf.gather_nd`.

Because the derivative kernel wraps the original kernel, it has access to its hyper parameters.

NicolasDurrande:
I had a look at the code with @markvdw  and we were wondering about one thing... what happen if you try to add derivative observations with a Matern 1/2 kernel? We would expect it to fail since the GP derivative do not exist... 

Similarly it should run smoothly for a Matern 3/2 for the first derivative but not for the second...

alexggmatthews:
Will the move to Python 3 only help with this PR also?

markvdw:
We should probably re-open this if we want to include this. Perhaps it needs a bit more of a general think on how to include this.

markvdw:
Actually, on more thoughts, do you think you could turn this into a separate package? I think it's too big a feature (and slightly outside the scope of GPflow), but it would be awesome to have usable. A separate package (that we could link to) would be ideal for this.

PR #506, opened 2017-09-22
More flexible Inducing Features

st--:
Following discussion with @jameshensman and @markvdw it seems that there is wider interest in more flexible inducing features beyond real-space inducing points, such as Fourier features, linear combinations and perhaps others.

This PR implements an `InducingFeature` abstract base class which provides methods for calculating covariances between features (`Kuu`) and between features and data points (`Kuf`), and the length of a feature is defined to be the number of rows in the Kuu matrix. I've changed the SVGP and SGPR models to use the features instead of inducing points directly.

I included a little wrapper function that makes it mostly backwards compatible, though I had to change all references to `model.Z` in the test to `model.feat.Z`.

Open for discussion / changing the interface!


gpflow/features.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+@singledispatch
+def conditional(feat, kern, Xnew, f, full_cov=False, q_sqrt=None, whiten=False):
+    """
+    Note the changed function signature compared to conditionals.conditional() to allow for dispatch.
+    """
+    raise NotImplementedError("No implementation for {} found".format(type(feat).__name__))
+
+@conditional.register(InducingPoints)
```
markvdw: Does GPflow need the singledispatch construction here? Can't all `InducingFeature` objects not use the same `feature_conditional` function?

I do remember discussing some things about possibly supporting matrices that have more structure, which makes them easier to invert. Is that the rationale behind this decision?

gpflow/features.py
```diff
@@ -0,0 +1,95 @@
...
+
+from .param import Param, Parameterized
+from . import conditionals
+from ._settings import settings
+
+
+float_type = settings.dtypes.float_type
+
+
+class InducingFeature(Parameterized, metaclass=ABCMeta):
```
markvdw: This fails our python 2.7 tests...

markvdw:
This looks great. I think it's a super neat solution for inducing features. I'm happy to add another one from the original inter-domain gaussian processes paper. I think it'll be good to have another example to showcase the generality of the code.

The code is currently failing on the `metaclass` argument, which seems to be Python 3 only. This is almost funny, as yesterday I was defending continuing to support Python 2.7. How much less elegant will the code be for a fix that also supports py2?

gpflow/features.py
```diff
@@ -0,0 +1,95 @@
...
+    Real-space inducing points
+    """
+
+    def __init__(self, Z):
+        """
+        :param Z: the initial positions of the inducing points, size M x D
+        """
+        super().__init__()
+        self.Z = Param(Z)
+        self._Mfeat = len(Z)
```
markvdw: Users are generally allowed to re-assign `Param` objects. This would render `self._mfeat` outdated. `__len__` needs to be derived either from `self.Z` directly, or updated whenever `self.Z` is re-assigned.

gpflow/sgpr.py
```diff
@@ -121,21 +121,20 @@ def build_likelihood(self):
...
+        Kuf = self.feat.Kuf(self.kern, self.X)
+        Kuu = self.feat.Kuu(self.kern)
         L = tf.cholesky(Kuu)
         sigma = tf.sqrt(self.likelihood.variance)
 
         # Compute intermediate matrices
         A = tf.matrix_triangular_solve(L, Kuf, lower=True) / sigma
         AAT = tf.matmul(A, A, transpose_b=True)
-        B = AAT + tf.eye(num_inducing, dtype=float_type)
+        B = AAT + tf.eye(self.num_inducing, dtype=float_type)
```
markvdw: Similarly, here, the number of inducing points can also change after object construction.

markvdw:
Perhaps the best solution to the `__len__` problem is to change the behaviour when tf_mode is on? This will allow `tf.shape()` to be used in tf_mode, while `len(self.Z)` is used otherwise.

gpflow/features.py
john-bradshaw: you can use `six`: `from six import with_metaclass` and change it to:
`class InducingFeature(with_metaclass(abc.ABCMeta, Parameterized)):`
to get Pytthon2/3 metaclasses

markvdw: Ok, bit ugly, but it'll have to do for now. At least it's only a one line change.

markvdw:
I just added the `Multiscale` inter-domain feature. It's not tested yet (I will write tests), but I think we need at least one other example of an inter-domain inducing feature in order to show the justification for the generality of the code.

gpflow/features.py
st--: Yes, depending on the feature, the structure of the code in `conditional()` may be more efficient one way or another. If you only have a single `feature_conditional` function, it won't be possible to do that, and I'd have to end up maintaining my own branch of code. This way, with a new feature you can simply register another function.

We could change the function name in the following row to something else, then you can just re-register the same function for different types. Or, if they're in here already, just add a `@conditional.register(Multiscale)` here.

gpflow/features.py
```diff
@@ -66,13 +66,64 @@ def Kuf(self, kern, Xnew):
...
+        self._M = len(Z)
+
+    def __len__(self):
+        return self._M
+
+    def _cust_square_dist(self, A, B, sc):
+        return tf.reduce_sum(tf.square((tf.expand_dims(A, 1) - tf.expand_dims(B, 0)) / sc), 2)
+
+    def Kuf(self, kern, Xnew):
+        if isinstance(kern, kernels.RBF):
```
st--: Maybe we should depend on the multimethod package after all; this would be a good example - even if you write the feature as a class, you could have a `@multimethod` on the method, then other packages could extend this for other kernels...

gpflow/features.py
markvdw: Great, that confirms why it's implemented like that.

Why does the function name need to be changed to something other than `_`? Can we not just do the following?
```
@conditional.register(InducingPoints)
@conditional.register(Multiscale)
def _(feat, kern, Xnew, f, full_cov=False, q_sqrt=None, whiten=False):
```

st--: That's what I meant. You only need to change it if you want to be able to register from outside this module, e.g. if some third-party module wants to do `@gpflow.features.conditional.register(MyOwnFeature)(gpflow.features.plain_features_conditional)` ...

gpflow/features.py
markvdw: True, true. Is there a package included in Python that supports this? A quick search yields several 3rd-party packages.

gpflow/features.py
markvdw: Ah yes! Looks like a great change.

gpflow/features.py
st--: No, there is no "standard library" version, but I've used this one, which seems to do the job well, and has comparatively succinct and tidy code: https://pypi.python.org/pypi/multimethod/0.6

gpflow/features.py
markvdw: Yes, I understand. This is a slightly difficult situation, but we have to support re-assigning `Param` objects. What do you think of the suggestion below, to maken `__len__` return a tensor when in tf_mode?

gpflow/features.py
markvdw: Are you ok with deferring this to a later PR? I think GPflow is rightly a bit conservative about using other packages. Everybody agrees that inter-domain features should go in, so I suggest we do that first, and then do enhancement.

I do agree that it's the best solution, and I'll happily push for it to get in.

st--: Sure, no problem. The multimethod implementation is less than 100 lines of actual code; if necessary, we could also simply include that file..

gpflow/features.py
st--: Yeah, if that works reliably - agree that it's cleaner to not carry around potentially stale constants if it's easy to compute them on demand.

What do you mean by re-assigning Param objects? someone doing model.feat.Z = an array with more inducing points?

gpflow/features.py
markvdw: I think we can depend on the library. I just think we should split up these discussions a bit.

gpflow/features.py
markvdw: Yes exactly. Although I think the correct syntax would be `model.feat.Z = Param(array_with_more_points)`.

It is quite useful actually. One may want to optimise inducing points, and then add more afterwards.

st--: What's the best way to check whether GPflow is in tf mode or not ?

markvdw:
```
if self._tf_mode:
    stuff
```

@jameshensman?

markvdw:
I'm confused why the tests aren't running. They did run when you first committed the branch, right?

awav:
@markvdw, code coverage can't be run for non-gpflow branch (read as fork).

alexggmatthews:
we are dropping python 2 support so feel free to use any python 3 feature that helps here:

https://github.com/GPflow/GPflow/issues/516

st--:
Superseded by #530

PR #508, opened 2017-09-23
Clean up kullback_leiblers

st--:
This PR replaces the four functions in the `kullback_leiblers` module (whether the covariance of q is diagonal or not and whether the covariance of p is the identity or not) with a single function, similar to `conditionals.conditional()`: Now there is just a single function `gauss_kl(q_mu, q_sqrt, K=None)`, which automatically detects whether `q_sqrt` has 2 or 3 dimensions, and treats `K=None` as the white case. Deprecates the other three functions.


st--:
Would be good to also have test coverage of `gauss_kl()` for the following:
- check that diagonal representation and dense diagonal covariance matrix of q(x) give same answer
- check that K = identity matrix and K = None for covariance of p(x) give same answer
- check that KL divergence between distributions with same covariance (and zero mean) is zero
- maybe some tests for small-dimensional cases where we can compute KL divergence by hand?
Could be modelled on `test_conditionals` (which I've cleaned up a bit in the process).

st--:
Test coverage added. Would be great if someone could merge this (-:

jameshensman:
I really like this PR, I think it cleans up the code nicely.

It does deprecate some functions, so perhaps bumpy the version number? 

Otherwise I think @markvdw is about to merge

markvdw:
Happy to take care of this. Will create an in-house branch and merge soon.

markvdw:
Let's continue discussion in the in-house version #514, so the tests run. I think I can merge this as soon as the tests have finished.

PR #513, opened 2017-10-02
added correct expectation for exKxz. all tests pass

stefanosele:
Hi everybody,

This is a pull request for computing the expectation `<x_t k(x_t, Z)>_q(X)`. Currently `exKxz` computes `<x_{t-1}k(x_t, Z)>_q(X)` which is basically the pair-wise expectation of two consecutive time steps. I believe this is useful in some dynamic systems that need to compute the expectation between two time steps. However the name of the function is a bit misleading and also the functionality of it is not in alignment with the general expectation terms.

This pull request handles both cases. The old function has been renamed to `exKxz_pairwise` and now the `exKxz` computes the `<x_t k(x_t, Z)>_q(X)`. I could have opted for the reverse naming, e.g., `exKxz` and `exKxz_non_pairwise` but that would be misleading. Tests are modified in order to pass with both cases now.

Currently there seems to be a repetition of code but I tried to be the least intrusive possible.

Let me know what do you think.

Cheers

### DIFF
--- a/gpflow/ekernels.py
+++ b/gpflow/ekernels.py
@@ -44,7 +44,7 @@ class RBF(kernels.RBF):
 
         return self.variance * tf.exp(-0.5 * q - tf.expand_dims(half_log_dets, 1))
 
-    def exKxz(self, Z, Xmu, Xcov):
+    def exKxz_pairwise(self, Z, Xmu, Xcov):
         """
         <x_t K_{x_{t-1}, Z}>_q_{x_{t-1:t}}
         :param Z: MxD inducing inputs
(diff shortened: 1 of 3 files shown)


markvdw:
Hi @stefanosele, this looks great. Could you elaborate a bit on the exact use-case for these extra kernel expectations?

Otherwise I agree with the PR. The code repetition doesn't seem too severe, and it's hard to see how general this is exactly. I think some of the kernel expectation code can be improved in terms of both clarity and speed (@ialong has opinions about this, I believe).

stefanosele:
Hi @markvdw,
A clear use-case is when you want to do moment matching for propagating uncertainty towards future samples from the following dynamic system:
`f_t = x_{t-1} + f(x_{t-1})`

For computing `Var[f_t]` you will need to compute at some point the term `<x_{t-1} f(x_{t-1})>_q(X)` which would require computing 
`<x_{t-1} k(x_{t-1}, z)>_q(X)`.

Does this make sense to you?

markvdw:
Great, I think that's an important use case. Again, I do think that some of this code can be cleaned, sped up and compactified, but that's a bigger job.

ialong:
@stefanosele @markvdw hey guys,

the PR looks good, though I feel it is important to point out that the old `exKxz` can be used to calculate the new one (*at the same computational cost* - which as you noted, Mark, can be improved upon, I'll get around to it at some point). 

Here's how.
The old `exKxz` essentially has two extra parameters for each index `t-1`:
`\mu_t` and `\Sigma_{t, t-1}`. The second one is passed in the second index of Xcov's first dimension.

Simply set:

`\mu_t`                  <------- `\mu_{t-1}`
`\Sigma_{t, t-1}`   <------- `\Sigma_{t-1}`

Before passing them to `exKxz`, and you should get the same values as if you used the new `exKxz`.

Intuitively this is the same as saying that q(x_t, x_{t-1}) is a degenerate Gaussian where x_t is really just x_{t-1} (and thus their cross-covariance is just the covariance of x_{t-1}).

All that would be needed would be to have separate arguments for `\mu_{t-1}` and `\mu_t` (and `\mu_t` would default to `\mu_{t-1}` if not passed, so no code would have to change). I.e. you would have:

```python
def exKxz(self, Z, Xmu, Xcov, Xmu_next=None):
```

**This would also be super useful when stitching together multiple (variable length) trials as it could handle boundaries between trials, rather than simply looking at the next entry in `Xmu`.**

So in general I'm in favour of having just one, slightly more general, method, and I'd be happy to change the code to do it if you agree.

stefanosele:
Hi @ialong,

I agree with your post, this is how I did it in the beginning and I guarantee that it gives same results.

The only annoying thing is that you need to duplicate `Xcov` for making it of rank 4 before calling the `exKxz`.

Another option (if we want to stick with the existing arguments) we can check the rank of the `Xcov` and if 4 proceed with the pairwise case, otherwise with the normal one.

ialong:
@stefanosele yes I like this solution, it saves a lot of code reuse since the logic for `exKxz` and `exKxz_pairwise` is very similar.

The extra argument for multiple trials belongs to another discussion/PR I guess.

ialong:
For future reference here's how one could merge the two functionalities into one method (don't think it's worth changing stuff yet as I need to refactor the kernexps anyway so I'll just do one PR for everything).

Just add these lines:

```python
is_cross_exKxz = tf.rank(Xcov) > 3
Xmu = tf.cond(is_cross_exKxz, Xmu[:-1], Xmu)
Xmu_next = tf.cond(is_cross_exKxz, Xmu[1:], Xmu)
Xcov = tf.cond(is_cross_exKxz, Xcov, tf.tile(Xcov[None, ...], [2, 1, 1, 1]))
```
and replace occurrences of `Xmu[:-1]` with `Xmu` and of `Xmu[1:]` with `Xmu_next` in the old code.

@stefanosele @markvdw does that seem reasonable or would you prefer two separate methods?

markvdw:
Interesting, I think this looks elegant. Is there any downside to passing a rank 3 tensor? Like memory usage?

stefanosele:
@ialong This is more or less the solution I had in mind but the other way around.
use the condition to check whether you need to compute the additional tensors that you require for the pairwise case and modify appropriately the call to `mvnquad`.

That way you don't need to pass the rank 3 tensors around in the case you want to compute the simpler expectation as @markvdw is worried about.

ialong:
@stefanosele do you mean rank 4? Xcov is at least TxDxD and potentially 2xTxDxD.

@stefanosele @markvdw you guys are probably right that it's worth worrying about the extra memory overhead involved with an unnecessary rank 4 tensor. There's an easy fix for that of course (just use tf.cond where Xcov is used rather than to define Xcov). My post was about clarifying the logic for the RBF exKxz, I would need to check the mvnquad-based exKxz's if there can be an equally parsimonious refactoring.

stefanosele:
@ialong Yes, I meant rank 4.

PR #514, opened 2017-10-03
Clean up kullback_leiblers -- in house

markvdw:
In house branch of #508.


st--:
Should the encoding not be unicode (utf-16 ?) instead of iso-8859-15?

markvdw:
I was unsure about it, my editor somehow said iso... Could you edit?

Yet another "nice thing" that Python 3 does without complaining...

markvdw:
The encoding seems to be ok, but now it's complaining about the use of the `@` for matrix multiplication. For now, we agreed to support python 2. I'm opening a topic for discussion on this. I'm sure it would be enriched by your opinion.

st--:
I've fixed the `@` in my own branch - I can't push to the one on GPflow. A bit annoying, really, that the build doesn't work for branches on forks, makes it harder to contribute!

markvdw:
I agree. I don't know GitHub well enough to find a fix. The reason is to prevent unchecked code from being able to be executed on the test servers. I could do bitcoin mining that way!

st--:
I would've if I could've...

Well, same is true for the Travis-CI (which *does* work for fork-PRs), they just put a hard time limit on how long the tests are allowed to run.

markvdw:
Testing is failing for the deprecated methods.

st--:
The deprecated methods were never tested in the first place! (Only implicitly through the test of the models relying on `gauss_kl*`)

markvdw:
Yeah, I'm aware. The changes means the code isn't run any more. This is kind of annoying, but we need these kind of checks to make sure that we keep discipline to keep test coverage high. Even if it's only run implicitly by other code will stop it at least from being obviously broken.

Long story short, I can't merge until the testing coverage of the diff doesn't decrease the total coverage.

If writing a test is really superfluous, you can `#pragma nocover` them (as some other lines) or delete them (after checking with James). They've been deprecated for quite a while now.

PR #515, opened 2017-10-03
Fix inferred input_dim for combination kernels (sum, product) with slices in active_dims

markvdw:
In house version of #500.


markvdw:
@dtpc: According to codecov, lines 759 and 760 of `kernels.py` are not getting run during the test. As soon as this is fixed, it should be able to go in! Can you push a fix to this branch?

dtpc:
@markvdw I've pushed a commit to #520. I wasn't sure how to add it here.

markvdw:
We're thinking about a way of doing this more consistently.

PR #517, opened 2017-10-04
Linear features for kernels

john-bradshaw:
This adds a method to kernels that returns their corresponding feature map.
In the case of stationary kernels where the feature map has infinite dimension an approximation of it is returned instead, following the method of https://people.eecs.berkeley.edu/~brecht/papers/07.rah.rec.nips.pdf.

Methods have also been added to the `GPR` and `SVGP` classes to allow them to return their respective posterior over weights if we were viewing the GP as a Bayesian linear model on the features induced by the kernel.

This is useful for sampling from the function and evaluating at many points. See Section 2.1 of _Hernández-Lobato, J.M., Hoffman, M.W. and Ghahramani, Z., 2014. Predictive entropy search for efficient global optimization of black-box functions. In Advances in neural information processing systems (pp. 918-926)._
where such a method is used.

Heres an image demonstrating sampling via random features approach versus from the GP in the usual way:
figure_1-7
(dashed grey lines are samples via random features and solid grey are from the GP. We're using an SVGP with inducing point locs shown by blue circles. Blue is GP posterior) 

And heres a figure showing how to use this to perform Thompson sampling of the minimum (although always finds local minima so may want to improve this):
figure_1-6
(the grey circles show Thompson samples of the minimum made by using the random features samples)

The script to produce these plots is included in doc/source/notebooks/random_features_approx_demo.py

The current issues which may prevent merging:
* Need to investigate the periodic kernel and ensure that I have implemented the sampling from the spectral density correctly.
* Only allow the posterior over the Bayesian linear model weights to be calculated for one latent function. I think it should scale to more I just need to carefully go through and check everything broadcasts in the correct way.
* Due to the fact that this currently only works on stationary kernels and those with finite feature maps you may decide it is not general enough to go in and not worth the maintaining. But anyway hope the code is useful here!


john-bradshaw:
Think the build is currently failing due to an `NotImplementedError` exception I put into the periodic kernel to remind myself that it needs checking.

john-bradshaw:
closed as superseded by https://github.com/GPflow/GPflow/pull/523

PR #520, opened 2017-10-08
Fix code coverage for #515

dtpc:


alexggmatthews:
@markvdw think we can merge this one, no?

markvdw:
Is the base branch correct? It's not currently merging into master? Otherwise, this is a fix we need!

alexggmatthews:
Good point re base branch

PR #521, opened 2017-10-10
EP for Binary Class GPs with Probit Likelihood

john-bradshaw:
Implementation of a GP using EP for binary classification (as detailed in Section 3.6 of "Gaussian processes for machine learning." http://www.gaussianprocess.org/gpml/chapters/RW.pdf), which we have called `EPBinClassGP`.

EP is run completely in TensorFlow using a TF while loop.
The PR is against the current GPflow-1.0-RC as this offers a nice way to cache the tau_tilde and nu_tildes (the natural parameters of the EP approximation) as TF untrainable variables.

## Do not Merge
This may be better suited as demonstration of GPflow that lives elsewhere...?
If it was considered for a merge then there are a few things that should be improved first:
* No tests for EPBinClassGP
* The class is too tied into using a Probit likelihood. The likelihood moment calculations needed for EP should be removed from this class and added to likelihood class to make it more general.
* No automatic running of EP whenever parameters are stale -- to make it easier for users this should be dealt with by the class and not externally by a user (see demo script for more details)
* Some hacking to access params whilst under `TensorConverter.__tensor_mode__` which could perhaps be written more nicely.
* repeat code between `_build_predict` and `_build_likelihood` that could be drawn out into separate functions.

There is a port of some C code to TF code from https://github.com/mseeger/apbsint/blob/master/src/eptools/potentials/SpecfunServices.h. This is for computing StdNormalPDF(z)/StdNormalCDF(z) in a stable manner. The idea of using this approach came from GPy at     https://github.com/SheffieldML/GPy/blob/devel/GPy/util/univariate_Gaussian.py. The TF version that I have written will work on TF tensors and I have written tests for it. This part may be worth cherrypicking into main GPflow if considered useful elsewhere.

## Demo script comparing VGP and EP for binary classification tasks
In the PR we have included a demo script running VGP and EPBinClassGP on a 2d classification class.
The output below:

ep_binary_classification

When running VGP and EP from this script we get the following output (this is using ScipyOptimiser):
```
Running VGP model.
VGP model's initial model log likelihood: -397.63106416405145
10 levels between 0 and 10
VGP model's final model log likelihood: -106.67003522163016
VGP model's final kernel variance: 14.523382622580367
VGP model's final kernel lengthscale: 0.7968240058107315
=================================


Running Binary EP Classification Model
Number iterations pre optimisation for EP convergence 5
EPBinGP model's initial model log likelihood: -148.37775237941437
Number iterations post optimisation for EP convergence 16
10 levels between 0 and 10
EPBinGP model's final model log likelihood: -106.3756120082025
EPBinGP model's final kernel variance: 15.639499248336698
EPBinGP model's final kernel lengthscale: 0.8065270393449602
=================================
```

### ADAM optimizer
Above we used a scipy optimizer but can also use ADAM, although here VGP seems to perhaps be slower to converge:
figure_1-7


john-bradshaw:
Just updated this PR

Re-factored the code with general tidy-ups. 
EP loop has been broken out so that it can be used more generally.
Same for the central moment calculators. 
Added tests for these parts.

alexggmatthews:
When you think it is ready, it might be interesting to also ask Carl Rasmussen or José Miguel Hernández-Lobato for some feedback on this @john-bradshaw  . Will talk to @jameshensman about this tomorrow.

st--:
@john-bradshaw would be great if you could put this as a repository into your own github instead of just a PR to GPflow - then we could link to it from the README!

The only change to core-GPflow that your PR makes is to add a method to the likelihoods - but you can just have that in your own module by using multiple dispatch as for features/expectations/conditional - let me know if you want to discuss. :)

PR #534, opened 2017-10-30
Gpflow 1.0 svgp args

jameshensman:
This tiny PR allows the user to pass num_data as an optional argument to the SVGP class. this is helpful when the X, Y pair are not numpy arrays, so we cannot get the num_data from them.


st--:
The models in GPflow generally seem to rely on the data being passed in as a numpy array, and wouldn't work with, say, a list of lists. However, most of them start out by converting the data to a `DataHolder` object: if you want the base models to support non-numpy-array data types, would it not be better to change the svgp to also use DataHolder, and then deal with non-numpy arrays in the DataHolder class once and for all?

PR #536, opened 2017-11-01
Moment matching for uncertain conditional

vdutor:
The PR implements moment matching for conditionals with uncertain inputs. This is generally useful for GP-SSMs, Deep GPs, Bayesian GP-LVMs, etc. The code supports whitened and unwhitened models, multi-dimensional inputs and multi-dimensional outputs.

Two different unittests are added to check numerical correctness and stability. 
1) Comparison with `gpflow.conditional` when the new inputs are deterministic, i.e. `X* ~ N(mu, 0)`.
2) Comparison with a Monte Carlo solution. The same MC strategy is used to create the plots below.

### Credits

This work is done in collaboration with @markvdw.

### Demo

1) x* is Gaussian with center 3.7 and std. dev. 0.3.

uncertain_conditional_2

2) x* is Gaussian with center 5.5 and std. dev. 0.12.

uncertain_conditional_3


markvdw:
This looks great! I need to get a handle on how the unit tests are running on the other branches before merging, but hopefully we can do this soon.

gpflow/conditionals.py
```diff
@@ -124,6 +125,68 @@ def base_conditional(Kmn, Kmm, Knn, f, full_cov=False, q_sqrt=None, whiten=False
...
+    Li_eKuffu_Lit = tf.matrix_triangular_solve(Luu_tiled, tf.matrix_transpose(eKuffu), lower=True)
+    Li_eKuffu_Lit = tf.matrix_triangular_solve(Luu_tiled, tf.matrix_transpose(Li_eKuffu_Lit), lower=True) # N x M x M
+
+    cov = tf.matmul(q_sqrt_r, q_sqrt_r, transpose_b=True) # D x M x M
+
+    fvar = (
+            tf.matrix_diag(tf.tile((eKff - tf.trace(Li_eKuffu_Lit))[:, None], [1, num_func])) +
+            tf.matrix_diag(tf.trace(tf.einsum("nmj,djk->ndmk", Li_eKuffu_Lit, cov))) +
+            # tf.matrix_diag(tf.trace(tf.matmul(Li_eKuffu_Lit, cov))) +
+            tf.einsum("kd,nkj->ndj", q_mu, tf.einsum("nmk,kd->nmd", Li_eKuffu_Lit, q_mu)) -
```
hughsalimbeni: A single einsum here might be faster

gpflow/conditionals.py
```diff
@@ -124,6 +125,68 @@ def base_conditional(Kmn, Kmm, Knn, f, full_cov=False, q_sqrt=None, whiten=False
...
 
 
+@name_scope()
+def uncertain_conditional(Xnew_mu, Xnew_var, feat, kern, q_mu, q_sqrt, whiten=False):
+    """
+    Calculates the conditional for uncertain inputs Xnew, p(Xnew) = N(Xnew_mu, Xnew_var).
+    See ``conditional`` documentation for further reference.
+
+    :param Xnew_mu: mean of the inputs, size N x D
+    :param Xnew_var: covariance matrix of the inputs, size N x D x D
```
hughsalimbeni: Perhaps a bit of pain to implement, but it would be nice to support diag Xnew_var

gpflow/conditionals.py
```diff
@@ -124,6 +125,68 @@ def base_conditional(Kmn, Kmm, Knn, f, full_cov=False, q_sqrt=None, whiten=False
...
+    eKuffu = feat.eKufKfu(kern, Xnew_mu, Xnew_var)  # N x M x M
+    Luu_tiled = tf.tile(Luu[None, :, :], [num_data, 1, 1])  # remove this line, once issue 216 is fixed
+    Li_eKuffu_Lit = tf.matrix_triangular_solve(Luu_tiled, tf.matrix_transpose(eKuffu), lower=True)
+    Li_eKuffu_Lit = tf.matrix_triangular_solve(Luu_tiled, tf.matrix_transpose(Li_eKuffu_Lit), lower=True) # N x M x M
+
+    cov = tf.matmul(q_sqrt_r, q_sqrt_r, transpose_b=True) # D x M x M
+
+    fvar = (
+            tf.matrix_diag(tf.tile((eKff - tf.trace(Li_eKuffu_Lit))[:, None], [1, num_func])) +
+            tf.matrix_diag(tf.trace(tf.einsum("nmj,djk->ndmk", Li_eKuffu_Lit, cov))) +
```
hughsalimbeni: is this equivalent to something like tf.einsum("nmj,djm->nd", Li_eKuffu_Lit, cov))?

hughsalimbeni:
Great work! Perhaps quadrature test too? Should be easy to run for D\le3

testing/test_uncertain_conditional.py
```diff
@@ -0,0 +1,101 @@
...
+        model = MomentMatchingSVGP(self.X, self.Y, k, gpflow.likelihoods.Gaussian(), Z=self.X.copy(), whiten=True)
+        model.compile()
+        gpflow.train.AdamOptimizer().minimize(model)
+
+        mean1, var1 = model.uncertain_predict_f_moment_matching(self.Xnew_mu, self.Xnew_covar)
+
+        for n in range(self.Nnew):
+            mean2, var2 = model.uncertain_predict_f_monte_carlo(self.Xnew_mu[n,...], self.Xnew_covar[n,...])
+            np.testing.assert_almost_equal(mean1[n,...], mean2, decimal=3)
+            np.testing.assert_almost_equal(var1[n,...], var2, decimal=2)
```
hughsalimbeni: Just a thought: I've been caught out before using MC unit tests where the result is numerically close but still wrong. Really the test should make sure that the convergence is 1/sqrt(S). Numerics can be a pain though when it gets close though

gpflow/conditionals.py
```diff
@@ -124,6 +125,68 @@ def base_conditional(Kmn, Kmm, Knn, f, full_cov=False, q_sqrt=None, whiten=False
...
+    Kuu = feat.Kuu(kern, jitter=settings.numerics.jitter_level) # M x M
+    Luu = tf.cholesky(Kuu) # M x M
+
+    if not whiten:
+        q_mu = tf.matrix_triangular_solve(Luu, q_mu, lower=True)
+        Luu_tiled = tf.tile(Luu[None, :, :], [num_func, 1, 1]) # remove line once issue 216 is fixed
+        q_sqrt_r = tf.matrix_triangular_solve(Luu_tiled, q_sqrt_r, lower=True)
+
+    Li_eKuf = tf.matrix_triangular_solve(Luu, eKuf, lower=True) # M x N
+    fmean = tf.matmul(Li_eKuf, q_mu, transpose_a=True)
```
st--: Can you add the more "mathy" linear algebra expressions for mean and variance as comments?

gpflow/conditionals.py
```diff
@@ -124,6 +125,68 @@ def base_conditional(Kmn, Kmm, Knn, f, full_cov=False, q_sqrt=None, whiten=False
...
+    """
+    Calculates the conditional for uncertain inputs Xnew, p(Xnew) = N(Xnew_mu, Xnew_var).
+    See ``conditional`` documentation for further reference.
+
+    :param Xnew_mu: mean of the inputs, size N x D
+    :param Xnew_var: covariance matrix of the inputs, size N x D x D
+    :param feat: gpflow.InducingFeature object, only InducingPoints is supported
+    :param kern: gpflow kernel or ekernel object.
+    :param q_mu: mean inducing points, size M x D
+    :param q_sqrt: covariance inducing points, size M x M x D
```
st--: presumably the Cholesky factor of the covariance between inducing features?

testing/test_uncertain_conditional.py
```diff
@@ -0,0 +1,103 @@
...
+
+from gpflow.test_util import GPflowTestCase
+from gpflow import settings
+from gpflow.conditionals import conditional, uncertain_conditional
+
+float_type = settings.dtypes.float_type
+
+np.random.seed(0)
+
+class MomentMatchingSVGP(gpflow.models.SVGP):
```
st--: Would be great to have some independent equivalence tests between `uncertain_conditional` and the default `conditional` for the case of variance going to zero - independent of external models.

st--: See https://github.com/GPflow/GPflow/blob/master/testing/test_conditionals.py and https://github.com/GPflow/GPflow/blob/master/testing/test_kldiv.py for similar tests.

gpflow/conditionals.py
```diff
@@ -124,6 +125,68 @@ def base_conditional(Kmn, Kmm, Knn, f, full_cov=False, q_sqrt=None, whiten=False
...
+    :return fmean, fvar: mean and covariance of the conditional, size N x D and N x D x D
+    """
+
+    # TODO: Tensorflow 1.3 doesn't support broadcasting in``tf.matmul`` and
+    # ``tf.matrix_triangular_solve``. This is reported in issue 216.
+    # As a temporary workaround, we are using ``tf.einsum`` for the matrix
+    # multiplications and tiling in the triangular solves.
+    # The code that should be used once the bug is resolved is added in comments.
+
+    assert isinstance(feat, InducingPoints)
```
awav: raise exception instead of assert

PR #541, opened 2017-11-06
Chain transform

st--:
Currently, only the `Rescale` transform allows for chaining with other transforms (such as `positive`). However, it's the wrong order of chaining for MCMC (the docstring is misleading), and it'd be beneficial to be able to chain arbitrary transformations together. This is enabled by the `Chain` transform introduced in this PR. Note- to keep the code clean and understandable, I've only written up a two-transform chain; obviously, people can call `Chain(t1, Chain(t2, t3))`. Still need to write some tests - the defaults probably will break.

EDIT: I've also introduced `__call__` on Transform subclasses, which should make chaining of several transforms even easier.


gpflow/transforms.py
```diff
@@ -23,7 +23,8 @@
 
 
 class Transform(ITransform): # pylint: disable=W0223
-    pass
+    def __call__(self, other_transform):
```
markvdw: Why `__call__` used here? Could this not just be a method like `Transform.chain()`?

st--: Because it's like function-chaining: y = t1(t2(x)) = g(x), so g = t1(t2) seemed to make sense

st--:
I'm now wondering if the order of chaining makes sense. At the moment, you would have to call
`Rescale(scale)(positive)` to get a transform for a positive parameter with a given scale. And `forward` goes from unconstrained x to constrained y.

`positive(Rescale(scale))` seems more intuitive, but that would only work by explaining the transform in terms of `x = t1(t2(y))`, so would have to swap which way is forward and backward?...

PR #551, opened 2017-11-13
Enabled nested AutoFlow types

rhaps0dy:
The type descriptions of AutoFlow inputs now can be any nested structure
that is accepted by `tensorflow.python.util.nest`.

This is perhaps an unwanted feature; it breaks the existing uses of AutoFlow and needs one extra class. On the other hand, if you really want a namedtuple input to your functions, it works wonderfully: the TensorType class is there solely to break the nesting at the desired depth.

### DIFF
--- a/gpflow/core/autoflow.py
+++ b/gpflow/core/autoflow.py
@@ -14,6 +14,9 @@
 
 
 from gpflow.misc import get_attribute
+from gpflow import settings
+import tensorflow as tf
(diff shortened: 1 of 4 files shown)


rhaps0dy:
Only I use this, it breaks existing code and can be worked around, so as markvdw said it's probably not a good idea to merge it.

markvdw:
It's good that this is here, if this issue ever pops up for more people.

PR #552, opened 2017-11-13
Explicit checking of None in set_parent

markvdw:
If `parent` (L115) has `__len__()` defined, this will be called in the `if` statement. This is not the intended check.

I think in this instance it's better to be explicit, as we do really want to check for `None`. Some objects may have a `__len__()` method defined. In my case, this relies on parameters which are set in the constructor, which are not initialised at the time that `if parent` is run.


jameshensman:
Explicit is better than implicit :) merged.

PR #556, opened 2017-11-15
adding value property

st--:
Now tested.

Question: is there (or should there be) a test to compare setting of the same parameter in different sessions with different values?

### DIFF
--- a/gpflow/params/parameter.py
+++ b/gpflow/params/parameter.py
@@ -201,6 +201,10 @@ class Parameter(Node):
     def fixed_shape(self):
         return self._fixed_shape
 
+    @property
+    def value(self):
+        return self.read_value()
+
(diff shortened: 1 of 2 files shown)


st--:
NB- I think the second assign (of self.m.p) should be a different number e.g. 3.0 and then there should be a check whether self.p is *also* 3.0 - if I'm understanding the concept correctly?

alexggmatthews:
I'm happy to merge this unless someone has a good reason for value having been removed in 1.0 . I personally think read_value is a too verbose. @awav @jameshensman ?

awav:
@alexggmatthews, @st-- I'm just curious if someone see my core review comments in PRs? :)

alexggmatthews:
@awav I'm sorry I don't know what you mean.

PR #565, opened 2017-11-19
Increase test coverage.

awav:
This PR major aim is an increasing the test coverage. Previously, `codecov` **was configured to calculate coverage score for both python folders**: `gpflow` and `tests`. This is a little bit unfair as we are really interested only in `gpflow` package. Because `tests` directory was included into `codecov` reports, the scores were skewed towards 100% a bit. When I turned off coverage for `tests` the score dropped from `97%` to `94.7%`, so I managed to push it back to `97.31%`.

Additionally, I removed `miniconda` installation, switched tensorflow version to 1.4 and turned on caching in travis builds - as a result installation takes twice less time then before.

@alexggmatthews, @jameshensman, @markvdw

### DIFF
--- a/gpflow/params/dataholders.py
+++ b/gpflow/params/dataholders.py
@@ -140,6 +140,14 @@ class Minibatch(DataHolder):
         self._seed = seed
 
     @property
+    def batch_size(self):
+        return self._batch_size
+
+    @batch_size.setter
+    def batch_size(self, value):
+        return self.set_batch_size(value)
+
+    @property
     def initializables(self):
         return [self._iterator_tensor]
 
@@ -164,7 +172,7 @@ class Minibatch(DataHolder):
         self._batch_size = size
         session = self.enquire_session(session)
         if session is not None:
-            self.initialize(session=session)
+            self.initialize(session=session, force=True)
 
     def _clear(self):
         self._reset_name()
--- a/gpflow/params/dataholders.py
+++ b/gpflow/params/dataholders.py
@@ -17,9 +17,6 @@
 # limitations under the License.
 
 
-from __future__ import absolute_import
-from pkg_resources import parse_version
-
 import tensorflow as tf
 
 from .. import misc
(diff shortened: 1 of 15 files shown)


markvdw:
Amazing, did not using miniconda fix the libgcc issue?

gpflow/params/dataholders.py
```diff
@@ -60,13 +57,7 @@ def trainable(self):
     def parameter_tensor(self):
         return self._dataholder_tensor
 
-    @property
-    def shape(self):
```
markvdw: Why is this property removed? I see that `batch_size` is added, but what if we want to find the input dimension?

tests/test_optimizers.py
```diff
@@ -58,19 +67,19 @@ def test_different_sessions(self):
         with self.test_context() as session1:
             gpflow.reset_default_session()
             opt = self.optimizer()
-            opt.minimize(demo, maxiter=10)
+            opt.minimize(demo, maxiter=1)
```
markvdw: Maybe `maxiter=2`? One is often a bit of a special case... I've had situations before where my optimiser would run once but not twice. I don't feel strongly though.

gpflow/params/dataholders.py
awav: it is inherited from `parameter` class

tests/test_optimizers.py
awav: Here we are just checking that optimizer will not crash when model was created subject to it was initialized properly. But, later it would be good to add mocking (or introspection mechanism) which will check that session was called once (or more) depending on optimizer.

awav:
@markvdw, miniconda was a source of issue, somehow it screws environment links, there is another reason why I removed miniconda - time overhead. Travis installs python3.6 in the beginning, so I think it is wasteful to download and install it again using miniconda. And one thing more, conda's commands cannot be cached by travis.

alexggmatthews:
@awav a clear improvement on the current coverage and should go in conditional on @jameshensman being happy. 

One general question for us all to think about is: "to what extent do we have numerical coverage of the algorithms?" It is very hard to measure this using automated metrics. We have had cases in the past where we have had line coverage but not spotted a bug. One practical suggestion would therefore be for you and others to raise an issue if they think there is an area of the code that lacks numerical checks.

alexggmatthews:
I suggest we squash this PR to reduce commit noise.

markvdw:
I'm happy with this, and would argue for it to be merged now. @jameshensman is away for a few days, so he may not see this.

I agree with the tendency to squash, unless @awav has particularly strong reasons.

awav:
I think that that squash unnecessary unless it is not a messy branch with not well organized commits. In my case, each commit is reasonable, it adds more and more tests - this helps with git bisect a lot. The `squash` must be applied to very dirty PRs, like bug fixing with multiple trials. Another thing is that I prefer keep history, again only if it is well organized and squash shouldn't be triggered by number of commits, but their quality :)

For now I'm not against squashing, green light! :)

PR #567, opened 2017-11-20
Explicit kwargs

st--:
The `conditional()` function has a lot of arguments; when specifying them using positional arguments it's easy to get them wrong, and in practice code specifies `full_cov`, `whiten` etc using keyword arguments anyway. Now that GPflow is based on python3, this PR changes the keyword arguments to keyword-only. This might break some code, but will a) make future code clearer and b) making it less likely to introduce errors by confusing order of arguments.


alexggmatthews:
Thank you for your pull request. I am willing to merge this. I will leave 24 hours for people to register any comments.

alexggmatthews:
@st-- As far as I'm aware you aren't on the contributor list? Your github id is a little cryptic. I'm happy to add you if you send me the details.

Issue #568, opened 2017-11-20
Predict call execution time degrades linearly with additional calls


NMRobert:
- **OS Platform and Distribution**: Linux Ubuntu 17.04
- **TensorFlow installed from (source or binary)**: binary
- **Tensorflow version**: tensorflow-gpu==1.4.0
- **GPflow version:**  gpflow-1.0.0 (from github master source)
- **CUDNN installed:** 6
- **GPU:**  GeForce GTX 1080 Ti
- **Python Version:** Python 3.5.3

Issue:
Repeated calls to .predict (either .predict_f or .predict_y) have degrading execution time when called successively. The initial call might take perhaps 10ms, but later calls might take as long as several seconds each. This poses a problem when many calls to *.predict are desired. I did not observe this behaviour in versions of gpflow prior to 1.0.0.  This occurs for at least the regression models (GPR, SGPR, SVGP), and possibly others. 

```
import numpy as np
import matplotlib.pyplot as plt
import gpflow
import time

X = np.linspace(0,10, 100).reshape(-1,1)
y = np.sin(X) + np.random.normal(0, 0.25, size=100).reshape(-1,1)

k = gpflow.kernels.RBF(input_dim=1)
m = gpflow.models.gpr.GPR(X, y, kern=k)
m.compile()
gpflow.train.ScipyOptimizer().minimize(m)

print('Done optimize.')
predict_duration_log = []
for i in range(250):
    start = time.time()
    m.predict_y(X) 
    end = time.time()
    predict_duration_log.append(end-start)

plt.plot(predict_duration_log)
plt.show()
```
[
![image](https://user-images.githubusercontent.com/9734331/33043675-278a22a4-ce3e-11e7-84d0-7ed818b313a6.png)
](url)

vdutor:
I replicated your experiment and came to the same conclusion. The code was running on CPU.

## GPflow 1.0
![gpflow_1_predict](https://user-images.githubusercontent.com/13780846/33067149-2d392af8-cea5-11e7-8213-8dcb910a30b0.png)

## GPflow 0.5
![gpflow_old_predict](https://user-images.githubusercontent.com/13780846/33067150-2d9d1626-cea5-11e7-9381-f32b00486570.png)

alexggmatthews:
@NMRobert @vdutor thank you for reporting and confirming.
@awav I'm away from my desk but is this possibly because the graph is growing each time we call predict?

awav:
[...]

Issue #561, opened 2017-11-16
Minibatch


hughsalimbeni:
Minibatching doesn't seem to be working for me. For example, 

```python
model = gpflow.models.svgp.SVGP(np.random.randn(1, 1), 
                                np.random.randn(1, 1), 
                                gpflow.kernels.RBF(1), 
                                gpflow.likelihoods.Gaussian(),
                                np.random.randn(1, 1),
                                minibatch_size=1)
model.compute_log_likelihood()
```
gives the following error

```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-3-e3c07a8fceb7> in <module>()
     10                                 np.random.randn(M, D),
     11                                 minibatch_size=2)
---> 12 model.compute_log_likelihood()

/[...]/GPflow/gpflow/decors.py in runnable(obj, *args, **kwargs)
    152                     _setup_storage(store, *af_args, **af_kwargs)
    153                     _build_method(method, obj, store)
--> 154             return _session_run(session, obj, store, *args, **kwargs)
    155         return runnable
    156     return autoflow_wrapper

/[...]/GPflow/gpflow/decors.py in _session_run(session, obj, store, *args, **kwargs)
    189         feed_dict.update(obj.feeds)
    190     initialize = kwargs.pop('initialize', False)
--> 191     obj.initialize(session=session, force=initialize)
    192     return session.run(store['result'], **kwargs)
    193 

/[...]/GPflow/gpflow/core/node.py in initialize(self, session, force)
     82                 session=session,
     83                 force=force,
---> 84                 feed_dict=self.initializable_feeds)
     85 
     86     def clear(self):

/[...]/GPflow/gpflow/misc.py in initialize_variables(variables, session, force, **run_kwargs)
     82             initializer = tf.variables_initializer(variables)
     83         else:
---> 84             uninitialized = tf.report_uninitialized_variables(var_list=variables)
     85             def uninitialized_names():
     86                 for uv in session.run(uninitialized):

/[...]/anaconda3/envs/py35cpu/lib/python3.5/site-packages/tensorflow/python/util/tf_should_use.py in wrapped(*args, **kwargs)
    105   """
    106   def wrapped(*args, **kwargs):
--> 107     return _add_should_use_warning(fn(*args, **kwargs))
    108   return tf_decorator.make_decorator(
    109       fn, wrapped, 'should_use_result',

[...]

PR #575, opened 2017-11-24
Address issues #568, #561

awav:
* Improves initialization checking mechanizm
* Iterator still cannot be checked if it were initialized for particular session or not, hence it will just always be initialized.


.travis.yml
```diff
@@ -13,7 +13,7 @@ install:
   - python setup.py install
 
 script:
-  - pytest -x --cov=./ -d --tx 3*popen//python=python3.6 --pyargs tests
```
markvdw: I would rather that the tests fail on the server asap. One can then run the tests locally for full coverage.

But I don't have a strong opinion.

markvdw:
This looks great. I'll review. We should try to get this in asap.

markvdw:
I'm happy to merge these changes as soon as the test coverage is what we require.

.travis.yml
awav: Disagree, travis and developer's environments are different and can produce unlike results, which *must* not happen, but still it is important to see whole report and all errors.

awav:
@markvdw you may want to squash, but in fact github show most of all of these commits for no reason, probably someone created a branch from mine when it hadn't merged into master yet, but squashed later (it is one of the drawbacks of squashing). This branch has only 4 commits, a couple of them have references to issues in github and were bounded.  I propose do not squash this branch.

markvdw:
Also, what does it mean for an "iterator to always be initialised"? Can this cause variables to be reset, like in earlier versions of the code?

awav:
@markvdw, This is a plug for now, until we find another proper solution. Everytime whenever you call autoflow, iterator will be re-initialized. It is a problem of TensorFlow - iterator doesn't provide an interface for checking if it were initialized for a session or not.

PR #576, opened 2017-11-24
Likelihood checks speed up.

awav:
```
=========================== slowest 5 test durations ===========================
417.60s call     tests/test_likelihoods.py::TestLikelihoodChecks::test_likelihood_checks
166.69s call     tests/test_ekerns.py::TestKernExpQuadrature::test_exKxz_pairwise
148.19s call     tests/test_gplvm.py::TestBayesianGPLVM::test_kernelsActiveDims
65.70s call     tests/test_hmc.py::CheckTrainingVariableState::test_multiple_runs
48.64s call     tests/test_coregion.py::TestEquivalence::test_predicts
```

Likelihood tests take extremely long time. This PR fixes that by using `pytest` parameterize feature and running each likelihood example in different context.

### DIFF
--- a/tests/test_likelihoods.py
+++ b/tests/test_likelihoods.py
@@ -22,6 +22,8 @@ import gpflow
 from gpflow import settings
 from gpflow.test_util import GPflowTestCase
 
+import pytest
+
 
 class LikelihoodSetup(object):
     def __init__(self, likelihood, Y, tolerance):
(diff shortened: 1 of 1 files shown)


markvdw:
This looks good, but it moves away from using `unittest` as the code, right? Is this the first part of the code where this happens? Can there be any negative consequences of this?

awav:
@markvdw, it is even better that we moves away from `unittests`. In this particular case `unittests` provide unnecessary level of abstraction, what's important that we have two distinct tests `pass` and `fail`, and data which we pass - done. Moreover `pytest` is much nicer and provides good features, which we can use, but some of them are not compatible with `unittests`.

markvdw:
Ok sure, as long as we're happy doing this, and the instructions on how to run the tests in the readme still works.

PR #577, opened 2017-11-25
Improve test code

st--:
Some of the test code has a lot of code duplication, which makes it harder to maintain/extend, or is otherwise inefficient. This PR aims to clean up the test code somewhat, making use of more advanced test framework features such as fixtures and parametrizations.


awav:
I would just re-write these tests with `@pytest.mark.parametrize`, because introducing additional package to the project is unnecessary. It provides same functionality. Yes, `@pytest.mark.parametrize` doesn't work with `unittests`, but the test itself does not require anything from`unittest` what cannot be provided by `pytest`.

st--:
Well, all the tests subclass GPflowTestCase, which in turn is based on tf.test.TestCase, which is a subclass of unittest.TestCase. So options are
a) duplicate code
b) introduce parameterized module
c) rewrite tests so they do not rely on GPflowTestcase (i.e. some other way of handling tensorflow teardown)

I think a) is the worst option, b) is the easiest. If you prefer c) over b) it'd be great if you could rewrite Gpflowtestcase so it doesn't inherit from unittest.TestCase.... at which point I'd be happy to move back to pytest.mark.parametrize...

awav:
Hello @st--, you dont' really need `unittest` at all, look at #579. Common and static data are packed whithin class, changing data passed through `@pytest.mark.parametrize`.

awav:
I could prepare `@pytest.mark.parametrize` changes, so that you could introduce your changes.

st--:
Sure, feel free to commit to my branch to show how you would deal with the tensorflow setup/teardown instead.

awav:
@st-- I'm going close this PR, looks like https://github.com/GPflow/GPflow/pull/583 covered your requests and does same job as you have got here. Is that okay?

awav:
@st--, Feel free to re-open this PR, if you still have other requests.

PR #579, opened 2017-11-26
Speed up some ekernels test on 40%.

awav:
* Created session_context monster (decorator / context manager) for pytests.
* Using distinct graphs and sessions per test case made quadrature tests a bit faster.
* Removed code duplication.


markvdw:
This is great.

PR #580, opened 2017-11-26
Exclusive behaviour for Minibatch

awav:
Until https://github.com/tensorflow/tensorflow/issues/14633 is resolved, Minibatch should be kept from `auto-initializing` feature. Simple cases are not affected. Only when you switch between sessions you will have to re-initialize Minibatch explicitly.

### DIFF
--- a/gpflow/misc.py
+++ b/gpflow/misc.py
@@ -185,9 +185,10 @@ def _find_initializable_tensors(intializables, session):
         if isinstance(v, (tuple, list)):
             status_tensors.append(v[0])
             boolean_tensors.append(v[1])
-        # TODO(@awav): Tensorflow Iterator must have init status.
+        # TODO(@awav): Tensorflow Iterator must have to be skipped at
+        # auto-intialization unless TensorFlow issue #14633 is resolved.
         elif isinstance(v, tf.data.Iterator):
-            yield v
+            continue
         else:
             for_reports.append(v)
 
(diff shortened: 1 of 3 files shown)


markvdw:
What is wrong with the behaviour that currently happens? Can you give an illustration?

awav:
@markvdw, 
My assumptions for Minibatch:

```python
mini = gpflow.Minibatch(np.random.randn(500, 1))
mini.read_value(session=gpflow.get_default_session())
Out[6]: array([[-1.21374478]])

# emulate auto-initialize VVV
mini.initialize(force=True)
mini.read_value(session=gpflow.get_default_session())
Out[9]: array([[ 0.47700428]])
```

But SVGP has minibatch non-None seed, therefore whenever auto-initialize is called for *minibatches*, it will return the same sequence of minibatches:
```
N, D = 5000, 1
X = np.linspace(-2, 2, N).reshape(N, 1)
Y = np.zeros((N, 1))
svgp = SVGP(X, Y, RBF(D), Gaussian(), X.copy(), minibatch_size=2)
svgp.X.read_value(session=gpflow.get_default_session())
Out[15]:
array([[ 1.78635727],
       [ 1.25665133]])

svgp.X.initialize(force=True)
svgp.X.read_value(session=gpflow.get_default_session())
Out[17]:
array([[ 1.78635727],
       [ 1.25665133]])
```

It means that whenever you call autoflow method it will read same minibatch element from stream.

markvdw:
Right, so to prevent the same minibatch from being returned by an autoflow method, we need to stop the minibatch from being re-initialised?

Just to check, is the autoflow case just an example, or can this behaviour show up anywhere other anywhere else?

I'll merge this soon.

awav:
> Right, so to prevent the same minibatch from being returned by an autoflow method, we need to stop the minibatch from being re-initialised?
Exactly.

> Just to check, is the autoflow case just an example, or can this behaviour show up anywhere other anywhere else?
Optimization.

PR #586, opened 2017-12-01
Use only np_float[int] in code.

awav:
Because TensorFlow accepts numpy dtypes, there is no need in `tf_float` and `tf_int`. For now, I marked `tf_like` setting properties as deprecated.


markvdw:
This looks good, but should we not just rename `np_type` to `dtype` or something? @jameshensman?

awav:
@markvdw do you mean float[int]_dtype?

Other options:
* inttype, floattype
* int_type, float_type
* dtype_float, dtype_int
* type_float, type_int

:)

st--:
I'd much prefer float_type and int_type. Whereas np_float seems ambiguous with np.float, which it is not, it's the specific precision type used throughout GPflow - whether tensorflow or numpy is used for any specific calculation, and I think calling the selected precision float_type better highlights that distinction. It's also consistent with settings.dtypes.float_type that has been around for ages (and for which settings.float_type is simply a shorthand).

awav:
Agree, `float[int]_type` is better.

markvdw:
Agreed. So we shouldn't see any more references to `np_float` in the code after this, right?

alexggmatthews:
@awav thank you for your pull request. I've marked as don't merge until we make the agreed changes.

alexggmatthews:
@awav I am happy to merge this once we have resolved CI. Is the codecov/patch failure I'm getting spurious?

PR #592, opened 2017-12-07
Pandas Model/Parameters printing.

awav:
Current printing of models and parameters is pretty naive and simple. As far as we use `pandas` in our project we can take advantage of using DataFrame. Here is an example of how model representation can look like:

screen shot 2017-12-05 at 16 23 43

`as_pandas_table()` method returns DataFrame itself and `__repr__` and `__str__` accordingly return string of that DataFrame.


gpflow/params/parameter.py
```diff
@@ -452,48 +460,6 @@ def _set_parameter_attribute(self, attr, value):
             raise GPflowError(msg.format(name, attr.interface))
         object.__setattr__(self, name, value)
```
alexggmatthews: Could you explain what  _html_table_rows was being used for? It seems to be mentioned in .covargerc only on main.

alexggmatthews:
@awav looks good - I just have the one question.

gpflow/params/parameter.py
awav: Guess, it was used for printing in notebooks. It looks like another dangled code, which we forgot to delete.

PR #595, opened 2017-12-08
Set gradient of sqrt(kernel) forced to 0 instead of inf

rhaps0dy:
As the added comment explains:

[In RBF.eKzxKxz] If an element of Kmms is 0, then its gradient is +inf and that gets backpropagated, but the overall gradient should instead be 0 (because the overall function is exp(-1/2 x)). We fix this with `tf.where`.

Also merged two variance power raising operations.

I've run the tests, and only one of them fails (`tests.test_autoflow.TestSVGP`) which doesn't seem related to this, so I'm doing this pull request anyways.

### DIFF
--- a/gpflow/ekernels.py
+++ b/gpflow/ekernels.py
@@ -138,7 +138,12 @@ class RBF(kernels.RBF):
         D = tf.shape(Xmu)[1]
         lengthscales = self.lengthscales if self.ARD else tf.zeros((D,), dtype=settings.float_type) + self.lengthscales
 
-        Kmms = tf.sqrt(self.K(Z, presliced=True)) / self.variance ** 0.5
+        # If an element of Kmms is 0, then its gradient is +inf and that gets
+        # backpropagated, but the overall gradient should instead be 0 (because
+        # the overall function is exp(-1/2 x). We fix this with `tf.where`.
+        Kmms_sq = self.K(Z, presliced=True)
+        Kmms = tf.sqrt(tf.where(tf.equal(Kmms_sq, tf.zeros((), dtype=settings.float_type)),
+                                tf.zeros_like(Kmms_sq), Kmms_sq))
         scalemat = tf.expand_dims(tf.eye(D, dtype=settings.float_type), 0) + 2 * Xcov * tf.reshape(lengthscales ** -2.0, [1, 1, -1])  # NxDxD
         det = tf.matrix_determinant(scalemat)
 
@@ -151,7 +156,7 @@ class RBF(kernels.RBF):
         smI_z = tf.reshape(ssmI_z, (N, D, M, M))  # NxDxMxM
         fs = tf.reduce_sum(tf.square(smI_z), [1])  # NxMxM
 
-        return self.variance ** 2.0 * tf.expand_dims(Kmms, 0) * tf.exp(-0.5 * fs) * tf.reshape(det ** -0.5, [N, 1, 1])
+        return self.variance ** 1.5 * tf.expand_dims(Kmms, 0) * tf.exp(-0.5 * fs) * tf.reshape(det ** -0.5, [N, 1, 1])
 
 
 class Linear(kernels.Linear):


hughsalimbeni:
What happens for very small values? I usually set variances to 1e-6 to avoid the nans, but perhaps there's some small values which will still give nans.

rhaps0dy:
Sorry I forgot the minimum working example.

## Minimum working example
```python
import numpy as np
import gpflow
import tensorflow as tf

kern = gpflow.ekernels.RBF(1, lengthscales=0.1)

tf_float = gpflow.settings.tf_float
ekz = kern.eKzxKxz(tf.constant([[-10], [10]], dtype=tf_float),
                   tf.constant([[10]], dtype=tf_float),
                   tf.constant([[0.1]], dtype=tf_float))

g, = tf.gradients(ekz, kern.lengthscales._unconstrained_tensor)

with tf.Session() as sess:
    kern.initialize(sess)
    print(sess.run([ekz, g]))
```

Without my patch, this outputs:
```
[array([[[ 0.        ,  0.        ],
        [ 0.        ,  0.21821789]]]), nan]
```

With my patch, this outputs:

```
[array([[[ 0.        ,  0.        ],
        [ 0.        ,  0.21821789]]]), 0.19777124203610966]
```

## Why does this happen?

Let's compute the gradient of the following expression, which is a simplification of what computes `Kmms` in `gpflow.ekernels.RBF.eKzxKxz`. _x_ here stands for the lengthscale-scaled distance.

d exp(-1/(2x)) / dx = exp(-1/(2x)) / (2x^2)

When evaluated for large _x_ with finite precision, this gives exp(-1/(2x)) = 0 .

Now let's compute the gradient of an equivalent expression, using the chain rule:

d sqrt(exp(-1/x)) / dx = 1/2 (exp(-1/x))^(-1/2) * exp(-1/x) / x^2

This expression is equivalent to the one above, but when evaluated for large _x_ with finite precision (such that exp(-1/x) = 0) we get:

1/2 ( 0 )^(-1/2) * 0 / x^2

which must calculate sqrt(1/0), which is of course `NaN`. And that's what causes problems.

@hughsalimbeni I don't quite understand what you mean, but perhaps this answered you. Sometimes the elements of the kernel matrix are zero merely because the points are far apart, and one can only influence it by setting the parameters.

st--:
@vdutor I guess this should be incorporated in #591 ?

gpflow/ekernels.py
```diff
@@ -138,7 +138,12 @@ def eKzxKxz(self, Z, Xmu, Xcov):
         D = tf.shape(Xmu)[1]
         lengthscales = self.lengthscales if self.ARD else tf.zeros((D,), dtype=settings.float_type) + self.lengthscales
 
-        Kmms = tf.sqrt(self.K(Z, presliced=True)) / self.variance ** 0.5
+        # If an element of Kmms is 0, then its gradient is +inf and that gets
+        # backpropagated, but the overall gradient should instead be 0 (because
+        # the overall function is exp(-1/2 x). We fix this with `tf.where`.
+        Kmms_sq = self.K(Z, presliced=True)
+        Kmms = tf.sqrt(tf.where(tf.equal(Kmms_sq, tf.zeros((), dtype=settings.float_type)),
```
awav: Can it be different from `settings.float_type` type? If it is even not, it still should be `Kmms_sq.dtype`.

rhaps0dy: I think you're correct, it should be `Kmms_sq.dtype`.

awav:
@rhaps0dy Could you add test cases? Thanks!

vdutor:
If #591 makes it into master, these changes will have to made in `expectations.py` instead of `ekernels.py`.

rhaps0dy:
I can add test cases, yes. Would just checking that the minimum working example I posted optimises and doesn't grow NaNs be alright?

If (hopefully "when") #591 hits master, propagating the changes to `expectations.py` should be very easy right?

markvdw:
Yes, @rhaps0dy, did you figure out the value of the derivatives as we discussed earlier?

rhaps0dy:
@markvdw I have! It's the same I wrote above but substituting "large x" with "small x".

x in that case represents the lengthscales. In the actual RBF kernel they're multiplied by the data and other things but we don't take the derivative of that.

vdutor:
@rhaps0dy  #591 is in master now, could you make your changes in the `expectations.py` file please?
Probably the easiest way to do this is creating a new PR by branching from master.

PR #601, opened 2017-12-21
Stef/expectations

stefanosele:
Hi guys here is the quadrature for the MarkovGaussian and the corresponding tests.

All test pass (locally at least). I still believe that we should not merge it yet to main since there is ability for the user to change the number of gauss-hermite points. In general the code doesn't account at all for the `kern.num_gauss_hermite_points`. By default it uses `H=40` while the kernels are constructed with `H=20`.


gpflow/expectations.py
```diff
@@ -492,9 +492,15 @@ def _expectation(p, kern, feat, none2, none3):
     return functools.reduce(tf.add, [_expectation_fn(k) for k in kern.kern_list])
 
 
-@dispatch((Gaussian, MarkovGaussian),
-            kernels.Sum, InducingPoints,
-            (mean_functions.Linear, mean_functions.Constant), type(None))
+@dispatch(Gaussian,
```
vdutor: What was the particular reason to split up these two?

vdutor:
Thanks for adding this to the expectations code, it is exactly what we needed.
I guess we could add `num_gauss_hermite_points` to gpflowrc and use that value as the default.

gpflow/expectations.py
stefanosele: I split it because there is no implementation for Linear and Constant mean functions with the MarkovGaussian distribution, and it wasn't falling back to the quadrature. Happy to revert it back if I misunderstood.

markvdw: This is correct.

Issue #332, opened 2017-02-06
Use tf distributions once they make it into tf main


alexggmatthews:
@dustinvtran suggests the following in #322
`
As another example, instead of coding up a separate densities.py and transforms.py, it would be nice to take advantage of the distributions and Bijectors class in tf.contrib.distributions. These are more comprehensive in robustness and vectorization. As tf.contrib.distributions is moving soon into core TensorFlow, it will continue to be expanded upon as the official distribution (and transformations) library.`

I think this would be worth considering once they are into the core tf code as we can then be sure of the test coverage.

goldingn:
Just a heads up that some of the darker corners ([one](https://github.com/tensorflow/tensorflow/issues/10131), [two](https://github.com/tensorflow/tensorflow/issues/10149)) of that module are a bit buggy at present. Hopefully they'll be fully tested when they move to main, but something to watch out for!

alexggmatthews:
As an update the tensorflow distributions library has made a lot of progress recently and now has a long description that can be found here:
https://t.co/S1PuBVC1kA

alexggmatthews:
It expect we would make gains in terms of composability with native tensorflow and speed if we incorporated it. 
@goldingn how has your experience been with this recently?

goldingn:
Awesome! I haven't taken a look in a while, but am thinking of switching over myself now

markvdw:
There are more recent discussions of this.

PR #607, opened 2018-01-02
Missing part in transform - backward_tensor

awav:
Hello everyone,

`transform.py` does have numpy _backward/forward_ methods, but only tensorflow forward method. Natural gradients implementation requires tensorflow backward transformation as well. This PR is intended to fix it.


hughsalimbeni:
Looks good. The fixed size of the lower triangular isn't a problem for the natural gradients as q_sqrt has known size.

alexggmatthews:
Thanks for the PR. Looks good. Long term we may consider whether such things can be simplified using tf.distributions as discussed in #332 .

PR #609, opened 2018-01-04
improvements and clarifications to the handling of num_latent

jameshensman:
Also fixes the coregion notebook


awav:
Hello James. Could you remove -k option for pytest in .travis file? It will turn on notebook tests checking.

vdutor:
@awav I removed the -k option and the coverage dropped 5%. Any idea how this happened?

PR #611, opened 2018-01-05
Minor clean-up: move example_profiler code to doc/source/notebooks (fixes bug)

jameshensman:


alexggmatthews:
hi @jameshensman could you add a bit more of a comment on the broken profiler code? At the moment it looks like the code is moved rather than removed.

awav:
@jameshensman, ping :)

jameshensman:
Hey @alexggmatthews , sorry to have dropped the ball on this one. 

The profiler code was broken because it was in the gpflow/notebooks dir, when everything else (including simple_regression, which it references) is in doc/source/notebooks. 

Moving this file to be with the others declutters by removing the (now empty) gpflow/notebooks dir.

PR #615, opened 2018-01-07
Gradient of `sqrt(RBF(Z, Z))` no longer NaN

rhaps0dy:
Before this commit, `Kmms = sqrt(exp(-1/x)) = exp(-1/(2x))`. These two expressions are equivalent mathematically. However, they are not equivalent when it comes to computing finite-precision gradients (dKmms/dx) via backpropagation when `Kmms**2 == 0.0`. Thus, we change sqrt(exp(-1/x)) to exp(-1/(2x)). This solves #595 in the expectations framework.

I'm not sure the test is up to your standards. I check that the gradient has a known-correct value (0.79108...) rather than checking that it's a finite number. This value was obtained using the `tf.where` patch of #595 . Is that reliable enough to be considered correct? Should I make more tests?

Also, I should note that the test `tests/test_notebooks.py::test_notebook[coreg_demo.ipynb]` doesn't pass on my machine, but neither does it pass in the master branch.


markvdw:
Hi @rhaps0dy, thanks for porting this. This is indeed a better way of doing it. I originally chose the previous implementation, as I thought it would allow more re-use of previously computed things (specifically the K matrix). I don't think this is as much of an issue as I thought.

I would request only two changes:
- I think you should be using `feat.Z` instead of `Z` when calling `kern.square_dist()`. `Z` has already been sliced, and `sqare_dist()` will slice it again. This may cause incorrect outputs.
- Having tests is a great thing, but in this case I think the benefit of the test is so marginal that I would argue it can even be left out. I don't think this is a problem that can actually creep back in easily. However, if you feel strongly that your test should be included, I'm happy to merge it. But I'm also happy to merge it without the test :)

rhaps0dy:
Hello! Sorry for the late reply. Regarding the first change: it looks to me like `square_dist` slices nothing, so the use of `Z` is correct.

https://github.com/GPflow/GPflow/blob/master/gpflow/kernels.py#L234-L248

In fact, the following code, which is the same as the test but with some `active_dims`, works just fine:
```py
import gpflow, tensorflow as tf, numpy as np
kern = gpflow.kernels.RBF(1, lengthscales=0.1, active_dims=slice(3, 4))
kern.variance = 2.
mean = np.zeros((1, 5), dtype=gpflow.settings.float_type)
mean[-1] = 10
cov = np.eye(5)[None, ...].astype(mean.dtype)
cov[0, 3, 3] = 0.1
p = gpflow.probability_distributions.Gaussian(mean, cov)
z = np.zeros((2, 5), dtype=mean.dtype)
z[0, 3] = -10
z[1, 3] = 10
z = gpflow.features.InducingPoints(z)
ekz = gpflow.expectations.expectation(p, (z, kern), (z, kern))
g, = tf.gradients(ekz, kern.lengthscales._unconstrained_tensor)
with tf.Session() as sess:
    z.initialize(sess)
    kern.initialize(sess)
    print(sess.run(g))
```

The second change: you're right that this particular problem is unlikely to creep in, even if we change the relevant kernel expectation function.

In my machine the test takes from 0.3 to 0.5 seconds to run (multiply that for every time it is ran while a developer waits), plus it might need to be changed if the interfaces to kernel and such change. On the other hand it provides a reference value for the gradient of eKzxKxz .

Perhaps we can just remove it, yeah.

markvdw:
You are completely right :) Should have checked rather than rely on my memory. I'll merge this asap. Thanks again!

PR #612, opened 2018-01-05
`q_sqrt` is now D x M x M.

markvdw:
As discussed in #610, this PR makes `q_sqrt` D x M x M. This was long overdue, and pulls `q_sqrt` in line with the TensorFlow convention of how to represent lists of matrices.

PR #617, opened 2018-01-08
VGP data update

javdrher:
Some minor fixes to perform the parameter updates of q_mu/q_sqrt for the VGP. This was broken after #612 , and when using the VGP as part of a Parameterized `compile()` was not called, so the code to update the parameters was moved to `_build()`


markvdw:
Ah yes, it looks like I missed that out. You're also adding a `LowerTriangular` transform, which wasn't there earlier. Was that also a bug?

javdrher:
On the cases I ran without the transform it worked, but it guess it should be there if it is part of original param? Ideally we'd take the current transform of `q_sqrt` before replacing it in case it was changed to something else by the user, but the transform might need to be updated for the new `num_latent` which differs for each transform.

markvdw:
Ok, so this feature was added specifically for your use case?

Could we recreate the transform if the existing transform is `LowerTriangular`, since we know what to do then? Alternatively, we could raise an error or do a sensible default?

javdrher:
This feature was added in 491b288a8636fb39ac1043b4a9648c5d6404f034 , long before GPflowOpt, our support for VGP relies on it

Right now it does recreate `LowerTriangular` regardless of the existing transform. Perhaps keep that, but issue a `UserWarning` if the transform was not a `LowerTriangular` ?
Alternatively, it seems only the LowerTriangular transform relies on `num_latent`, so we could just reuse the existing transform, and only intervene if its a `LowerTriangular`

markvdw:
Hi @javdrher, sorry for being a bit out of touch with these PRs lately. We're doing a push to get back on track. Is this still relevant?

PR #621, opened 2018-01-15
Expectation test clearance along with new test features.

awav:
Hello,

Expectation tests use one graph, it can cause memory issues. Luckily, I didn't see a lot of speedup improvement after tests graph separation, but I still feel that it is important to follow the "rule" one test one graph, therefore I publish this changes.

PS: Actually, the crucial part of this work lies in test_util new features - `session_tf` fixture and `cache_tensor` decorator.


markvdw:
I think this looks amazing @vdutor, can you ok this? If you're happy, I'll press the merge button.

vdutor:
I agree, very impressed by this clearance.

PR #627, opened 2018-01-20
fix vector handling and broadcasting in multivariate_normal density

### LINKED ISSUE #626: multivariate normal density cannot handle vectors as it claims
It seems that if `x` and `mu` are vectors, the call to `tf.matrix_triangular_solve` would fail (as `d = x - mu` is rank 1). Worse, if `x` and `mu` are of different dimensionality (e.g. one a row, one a column vector) broadcasting happens and fails silently.

ialong:
Fixes #626.

### DIFF
--- a/gpflow/densities.py
+++ b/gpflow/densities.py
@@ -71,18 +71,25 @@ def laplace(mu, sigma, y):
 
 def multivariate_normal(x, mu, L):
     """
-    L is the Cholesky decomposition of the covariance.
+    Computes the log-density of a multivariate normal.
+    :param x  : D or DxN sample(s) for which we want the density
+    :param mu : D or DxN mean(s) of the normal distribution
+    :param L  : DxD Cholesky decomposition of the covariance matrix
+    :return p : N vector of log densities for each of the N x's and/or mu's
 
-    x and mu are either vectors (ndim=1) or matrices. In the matrix case, we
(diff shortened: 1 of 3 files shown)


gpflow/densities.py
```diff
@@ -77,11 +77,11 @@ def multivariate_normal(x, mu, L):
     assume independence over the *columns*: the number of rows must match the
     size of L.
     """
-    d = x - mu
+    d = tf.cond(tf.rank(x) < 2, lambda: x[:, None], lambda: x) \
```
awav: Add some readability to the code:
```
x = tf.cond(tf.rank(x) < 2, lambda: x[:, None], lambda: x)
mu = tf.cond(tf.rank(mu) < 2, lambda: mu[:, None], lambda: mu)
d = x - mu
```
and num_col and num_dims types depend on L type
```
num_col = tf.cast(tf.shape(d)[1], L.dtype)
num_dims = tf.cast(tf.shape(d)[0], L.dtype)
```

ialong: Sounds good! (the use of settings.float_type was already there but your change sounds sensible)

awav:
It would be nice to have some tests covering 4 cases, *all rank 2*, *mu rank 2*, *x rank 2*, *none rank 2* :)

ialong:
@awav added the tests

PR #632, opened 2018-01-23
Vincent/pickling

vdutor:
A quick fix to allow pickling of models (Thanks to @awav).

### FYI: How to pickle a model 
```python
# write
model.anchor(gpflow.get_default_session())
model.clear()
with open("name", "wb") as fd:
	pickle.dump(model, fd)

# load
with open("name", "rb") as fd:
	model = pickle.load(fd)
	
model.compile()
```


vdutor:
Hi all,
I wanted to warn everybody about some weird behaviour we (@markvdw @awav) are seeing when clearing GPflow models. For some reason the minibatch's size is set to 1. A temporary fix is to manually reset the `seed` and `batchsize` before compiling the object:
```python
m.X.seed = 0
m.Y.seed = 0
m.X._batch_size = value
m.Y._batch_size = value
m.compile()
```

Issue #620, opened 2018-01-14
Docker image out of date


jwvdm:
Thanks for your great work on this library! I'm using GPflow for a programming assignment in a class that I'm teaching. I was hoping to point students towards your docker image. Unfortunately, it appears the docker image is still running 0.3.5. Is it indeed out of date, or might I have done something silly?

alexggmatthews:
Hi Jan-Willem. Yes it is certainly out of date. My apologies- I will add a ticket to regenerate it.
Alex

d-henderson:
Hello, looking to run the latest version of GPflow with docker. Is this PR ready to be finalised.

st--:
@awav  do we not have a new Docker image now?

st--:
It looks like the Docker image linked to from the documentation (https://gpflow.readthedocs.io/en/develop/intro.html, even for 2.0 version) is massively out of date (still 0.3.5). Should we remove the link, or update the docker image? (Who even has permission to update the docker image, and who knows what would be needed for that?)

PR #633, opened 2018-01-24
Update Dockerfile to support latest version

Canas:
Just had to add a `pip install pytest` to avoid import error and use the python 3 tensorflow tag.
This should help to address #620


d-henderson:
Hello, looking to run latest version of GPflow with docker. Is this PR ready to be finalised.

awav:
Now, GPflow has `pytest` as a dependency. I'm closing this ticket as it is not necessary anymore install pytest via pip.

PR #634, opened 2018-01-25
Cross-covariance between different RBF kernels

rhaps0dy:
All the `tests/test_expectations.py` pass, of course including the one I added to test this.

There are a couple of things I don't quite like about my code, and hopefully you know of a better way to do them:
  - Needing `nested_with`. That's because I needed to do `with params_as_tensors_for(a)` for every _different_ kernel and feature `a`, and you can't use `*map(params_as_tensors_for, unique_feat_kern)`.
  - Having to write another implementation for `square_dist` starting on line 350. This is because the desired lengthscale to use `(La + Lb)` doesn't belong to any kernel, and also to use the default kernel's implementation one would need to pass `tf.sqrt(La + Lb)`, only to have it be squared again.

### DIFF
--- a/gpflow/decors.py
+++ b/gpflow/decors.py
@@ -195,3 +195,17 @@ def _session_run(session, obj, store, *args, **kwargs):
 
 def _build_method(method, obj, store):
     store['result'] = method(obj, *store['arguments'])
+
+@contextlib.contextmanager
+def nested_with(generator):
+    "Implemented according to the Specification in PEP 343"
+    try:
+        mgr = next(generator)
+        mgr.__enter__()
+        try:
+            with nested_with(generator):
+                yield
+        finally:
+            mgr.__exit__(None, None, None)
+    except StopIteration:
+        yield
--- a/gpflow/expectations.py
+++ b/gpflow/expectations.py
@@ -15,12 +15,13 @@
 import functools
 import itertools as it
 import tensorflow as tf
+import numpy as np
 
 from . import kernels, mean_functions, settings
 from .probability_distributions import Gaussian, DiagonalGaussian, MarkovGaussian
 from .expectations_quadrature import dispatch, quadrature_fallback
 from .features import InducingFeature, InducingPoints
-from .decors import params_as_tensors_for
+from .decors import nested_with, params_as_tensors_for
 
 
 def expectation(p, obj1, obj2=None):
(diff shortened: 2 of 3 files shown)


rhaps0dy:
Well, the tests do work on my machine. Funny that even fixing the RNG it's not deterministic. Usually I've found that the test fails in kernels with very short lengthscales. But `np.testing.assert_allclose(analytic, quad, rtol=0.01, atol=1e-12)` passes even with very short lengthscales, though I'm not sure we want to relax requirements this way.

For that matter, [the documentation of `np.testing.assert_almost_equal`](https://docs.scipy.org/doc/numpy/reference/generated/numpy.testing.assert_almost_equal.html) states we should use other functions (for example `assert_allclose` ) to ensure consistent comparisons. Why are we using this one?

I've also made some changes that add an optional `H` argument to the `expectation_quadrature` functions and thus to the tests. Are you interested in them?

ialong:
@rhaps0dy hey man! Thanks for the PR! I am currently rewriting a lot of the expectations code to use my (hopefully) faster and more stable implementation. It should be merged in the next few days and then we can think of merging your PR as well. It might need some changing to use some of the new maths implementation as a template but shouldn't be too substantial.

st--:
@rhaps0dy #672 might help with your first point?

rhaps0dy:
Thanks for the heads up! I'll definitely use that, `nested_with` is no good. This isn't the only thing that needs to change, all the code that computes the expectation with two RBF kernels has changed, and I need to adapt it to that. That's quite a bit more work than just changing the `nested_with`, which is why I've been putting it off.

If I still worked on the project that used it I would have fixed it sooner, but it's a little dead... I'll get to it in the next couple of days, sorry for taking so long.

markvdw:
@vdutor / @ialong is this already supported, or is this still useful? Feel free to close the issue if it's no longer relevant.

ialong:
It’s not yet supported and it would still be useful, but the code should be updated to fit in with the new sped-up expectations

ialong:
@markvdw didn't realize you guys needed it urgently. @rhaps0dy and I discussed it and neither of us is likely to be able to find time before NIPS (at least not to do it properly - could probably produce a less thoroughly tested implementation). I think it would probably be easiest for me to do it rather than @rhaps0dy so I am happy to get this done first thing after NIPS

vdutor:
Hi @ialong , @rhaps0dy and @markvdw,
@sanket-kamthe just pinged me that he has an implementation that is ready to go.
I believe that Sanket and @dlqudwns need it more urgently (pre NIPS?).
I suggest that Sanket makes a separate PR and that we then sort things out?

sanket-kamthe:
I have an adapted version of @dlqudwns code. I am actually trying to cleanup (repeated code e.g.) other parts of the expectations as well. It might be better to merge this after multi-ouput PR [#724](https://github.com/GPflow/GPflow/pull/724) is finished as I was initially working on that branch ( However, I can quickly create a new PR from current master as well.)

ialong:
@sanket-kamthe would it be possible to have a look at the PR? Curious to have a look at the suggested changes to expectations as well as the cross eKzxKxz implementation

st--:
@ialong it's taken ages, but finally we're getting ready to merge #754 (which is up to date with HEAD) - can this PR be closed then or is there some functionality that goes beyond #754?

rhaps0dy:
It seems to me like the functionality in the new pull request is the same and we can close this one.

ialong:
Hey @st--, I agree with @rhaps0dy.
I am going to set some time aside to look at #754 over the weekend!

st--:
Thanks @ialong ! Closing this in favour of #754 then.

PR #635, opened 2018-01-27
Allow the epsilon parameter of the Robustmax likelihood to be trainable

john-bradshaw:
* Robustmax epsilon a parameter rather than a fixed float
* eps_K1 now a property so that changes with epsilon (with test)
* updated multiclass notebook to show how it could work

Superseding  #444, as brought up to date for GPflow 1.x

I set the epsilon  parameter by default to not-trainable so that behaves the same as previously.
I also added what I think is a sensible default prior for this parameter but don't know whether you'd rather it was added without one?


john-bradshaw:
Think it's failing due to notebook tests (the notebook worked fine on my machine though, and I don't think I changed one of the ones that it has failed on). 
How do these work exactly?

st--:
@john-bradshaw do you know what the `cdfs = cdfs * (1 - 2e-4) + 1e-4` line is for - just to ensure that (strictly) 0 < cdfs < 1?

john-bradshaw:
@st--  No sorry, I don't know, although I assume like you that is what it's for.

I don't think I wrote this line though. I think it was written by Mark or James.

PR #636, opened 2018-01-29
remove reindex_axis call in pretty_pandas_table

ialong:
`df.reindex_axis()` in `pretty_pandas_table` is now deprecated in pandas and can be avoided anyway.


ialong:
@awav I think it should be ready to be merged now

PR #641, opened 2018-02-02
Faster, More Stable Kernel Expectations

ialong:
Analytic kernel expectations (wrt Gaussian distributions) for the RBF kernel, Linear kernel and Sum kernels (when composed of RBF and Linear kernels) have been reimplemented in a more efficient and numerically stable way by better exploiting Cholesky factorisations and simplifications as well as avoiding superfluous transpositions and dimension augmentations.

**Main changes:**

- reimplemented RBF and Linear kernel expectations as well as the cross RBF Linear kernel expectation <K_lin_{Z, x} K_rbf_{x, Z}>_p(x)
- more efficient mean function expectation handling
- Streamlined quadrature expectations and unified the expectations files
- clearer distinction between quadrature_expectation and expectation which makes `@quadrature_fallback` superfluous
- "Private" _expectation methods are kept private
- Improved handling of Sum kernels (with 1 bug fixed on when cross expectations simplify)
- switched to assert_allclose instead of the less advisable assert_almost_equal
- switched around (feat, kern) to (kern, feat) in expectations for consistency
- improved test coverage and reduced error tolerance
- added, corrected and improved docstrings
- make covariance name consistent for all Gaussian `probability_distributions`: `p.cov`


markvdw:
So... beautiful...

One test fails though, but it looks like just a tolerance thing...

Also do you mind if we squash merge this, or would you prefer to have the full history in there?

ialong:
@markvdw haha thanks man. Yeah as I said in the comment above I am currently finishing work on the tests. Once I am done I'll let you know and we can definitely squash the commits, they're not that informative anyway

gpflow/expectations.py
```diff
@@ -13,75 +13,214 @@
...
+
+# By default multipledispatch uses a global namespace in multipledispatch.core.global_namespace.
+# We define our own GPflow namespace to avoid any conflict which may arise.
+gpflow_md_namespace = dict()
+dispatch = partial(dispatch, namespace=gpflow_md_namespace)
+
+
+# ========================== QUADRATURE EXPECTATIONS ==========================
+
+def quadrature_expectation(p, obj1, obj2=None, quad_points=None):
```
stefanosele: much better now that user can control the number of gauss_hermite points.
Should we rename it though to the standard name that we've been using so far? `num_gauss_hermite_points`

gpflow/expectations.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2017 the GPflow authors.
```
stefanosele: I kinda have the feeling that this file could benefit from splitting into different files as you have already done by splitting it into "sub-sections".
I'm just raising that comment here so it will trigger opinions from the other guys

ialong:
Should be ready to merge now

PR #642, opened 2018-02-03
<__str__> must implement pandas printing not <__repr__>

awav:
Hello everyone, 

As many of you know that default __str__ implemented using __repr__, and gpflow defines its own implementation of __repr__. But sometimes it causes troubles and misinterpretations when user does not print, but just returns object in `ipython`. Instead of short object description, you will get full table with all parameters:
```
In [10]: m
Out[10]:
                                                        class prior transform  \
SVGP-de20fbbe99/Sum/Matern32..  Parameter  None       +ve
SVGP-de20fbbe99/Sum/Matern32...  Parameter  None       +ve
SVGP-de20fbbe99/Sum/White...  Parameter  None       +ve
...
```
Ipython output for list of models looks terrible, you will not see elements or model objects, just many rows of parameters.

I resolved this issue:

```
In [2]: m
Out[2]: <SVGP name='SVGP' pathname='SVGP'>

In [3]: m.X
Out[3]: <DataHolder name='X' pathname='SVGP/X'>
```

### DIFF
--- a/gpflow/core/node.py
+++ b/gpflow/core/node.py
@@ -186,3 +186,9 @@ class Node(Parentable, ICompilable):
         compabilities instead of public `build`.
         """
         raise NotImplementedError('Private method `build` must be implemented by successor.')
+    
+    def __repr__(self):
+        type_name = self.__class__.__name__
+        name = self.name
+        msg = "<{type_name} '{name}'>"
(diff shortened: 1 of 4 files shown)


st--:
@awav why did you not merge this? the `__repr__` to Node class looked like a useful addition..

Issue #640, opened 2018-02-02
Issue with linear mean function.


svaibhavsingh:
When gpflow settings is changed with dtype=float32. 
A,b of mean function should be initialised with mentioning dtype too. (line 59,60) of mean_functions.py .
dtype=settings.float_type

markvdw:
Indeed, we've seen this as well. Originally, this was a design decision. The current fix is to pass numpy arrays of the desired type using `array.as_type('float32')`.

But I do think that this is cumbersome. We will make a PR to change this behaviour.

markvdw:
This issue is fixed with https://github.com/GPflow/GPflow/pull/647

PR #644, opened 2018-02-05
Update mean_functions.py

svaibhavsingh:
Pull request corresponding to the changes maentioned in issue #640


markvdw:
@svaibhavsingh, I think we're preferring a solution where we pass the `dtype` to the `Parameter` constructors for now. Thanks for the PR and for highlighting this issue. I'll commit this slight modification in a few minutes.

PR #645, opened 2018-02-05
Refactor index & naming rules.

awav:
Hello everyone,

Current tensorflow uniqness managing approach in GPflow has a couple of flaws: 1. hidden_*, full_* conceptions are a bit unclear and misleading - changed on build_[name or pathname]. 2. You may know that creating gpflow.Param(10) and then again gpflow.Param(1.) will not give an error, because of indexing of all built tensorflow's name scopes, but it does not generate index which will be guaranteed unique between your session and therefore it will cause issues with model saving (yes, after merging this one you are going to get a "Saver at GPflow" right after).
This PR re-considers implementation of the indexing and naming features, taking into account their current disadvantages.


markvdw:
@awav and went through the contribution and discussed this. I'm happy to merge as soon as the tests finished running.

PR #648, opened 2018-02-09
Deterministic ordering of variable list to enable reproducible results with ScipyOptimizer

### LINKED ISSUE #416: Reproducibility using seeds

jmxpearson:
- closes #416; similar in spirit to #417 
- Issue appears to affect only ScipyOptimizer. Does not affect, e.g.,
  AdamOptimizer on doc/source/notebook/classification
- Appears to be due to the fact that TensorFlow variables are not stably
  ordered across across replications, so var_list (and thus random
initial values for optimization) is not either.


jameshensman:
@awav doesn't this code originate somewhere in tf.contrib?

awav:
@jameshensman, nope. Fixed code from our repo.
@jmxpearson, I can see the issue: hashes for `set` elements changes depending on global python seed, so the order of var_list is different. The main question is how it affects ScipyOptimizer and why. Do you know answers? 
#416 is a bit different and not only about optimizers.

jmxpearson:
There are two key pieces:

1. `ScipyOptimizer` is responsible for intializing the vector of parameters using numpy. The random piece of this is, of course, repeatable if one sets the numpy random seed.

1. However, the function that packs `Parameter`s into a single vector relies on the order in which the relevant `tf.Variable`s come out of `var_list`. So if this ordering isn’t stable, the various pieces of the vector `x` in the optimization are reordered from one run to the next and so initialized differently. 

In the case of the TensorFlow optimizers, the `Variable` packing is handled by tf, and if one sets the random seed in TensorFlow, this is done correctly and repeatably. 

My understanding was that #416 was just about getting repeatable results from setting random seeds. I only ran a subset of models, so it’s quite possible there are other factors influencing the repeatability from other models.

awav:
Hello @jmxpearson. I apologize for not answering sooner. Can you show a case (mwe) when your change fixes the issue and w/o your fix breaks it. Thanks!

jmxpearson:
As noted in the original PR, [this notebook](
https://github.com/GPflow/GPflow/blob/master/doc/source/notebooks/classification.ipynb)
produces inconsistent numerical results without the change and exactly the
same results with.

PR #653, opened 2018-02-21
remove conditional shape changes from multivariate_normal

ialong:
tf.cond breaks things when the shapes are unknown, make `multivariate_normal` require column vectors or matrices (i.e. 2D input) and raise a warning when the shape is unknown or less than 2D


ialong:
@awav I think this is ready to be merged

markvdw:
@ialong densities.py:93 needs testing (or a nocover), then all tests should pass.

ialong:
@awav codecov is finally happy

PR #656, opened 2018-02-22
fix consistency of triangular transform

hughsalimbeni:
The free vector for the lower triangular transform wasn't being handled consistently. This PR makes the free state a vector of length `num_matrices * triangle_num`. A test ensures that the numpy and tensorflow transforms agree

### DIFF
--- a/gpflow/transforms.py
+++ b/gpflow/transforms.py
@@ -318,22 +318,21 @@ class LowerTriangular(Transform):
     """
     A transform of the form
 
-       tri_mat = vec_to_tri(x)
+       y = vec_to_tri(x)
 
-    x is of shape num_matrices x (N**2 + N)/2
-    y is of shape num_matrices x N x N.
+    x is the 'packed' version of shape num_matrices x (N**2 + N)/2
+    y is the 'unpacked' version of shape num_matrices x N x N.
+    
+    :param N: the size of the final lower triangular matrices.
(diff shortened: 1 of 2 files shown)


jameshensman:
@hughsalimbeni I edited your changes (quite a bit) to make th8ings simpler and to improve the docstrings.

The main upshot is that we no longer expect the 'free form' variable to have a flat shape, it's now "num_matrices x (N**2+N)/2". 

I also changed some of the code that was doing shape inference to use the shapes that are passed in the constructor. 

@awav might have some comments?

jameshensman:
@awav does this look okay?

gpflow/transforms.py
```diff
@@ -332,43 +332,28 @@ def __init__(self, N, num_matrices=1, squeeze=False):
...
-
-        Returns: Length of the vector with the lower triangular elements.
 
-        """
-        L = length / self.num_matrices
-        if int(((L * 8) + 1) ** 0.5) ** 2.0 != (L * 8 + 1):
-            raise ValueError("The free state must be a triangle number.")
-        return L
+        if self.squeeze and (num_matrices != 1):
+            raise ValueError("cannot squeeze matrices unless the first dim is 1.")
```
awav: it is worth mention about it in docstring:
```:raises ValueError: squeezing is impossible when dimension is 1.```

PR #664, opened 2018-03-01
Added simple multi-output kernels with various features for inference.

markvdw:


gpflow/conditionals.py
```diff
@@ -16,12 +16,133 @@
...
+# TODO: Ensure that R is handled correctly in all cases
+# TODO: Should there be consistentcy between fmean out, and f in?
+# Shapes to keep constant:
+#  - f      : M x L x R  or M x L  or  M x R
+#  - q_sqrt :
+
+
+@dispatch(InducingFeature, Kernel, object, object)
+@name_scope()
+def feature_conditional(feat, kern, Xnew, f, *, full_cov=False, full_cov_output=False, q_sqrt=None, white=False):
```
st--: How about a decorator that only takes the types for feat and kern, and combines the dispatch() and name_scope()? would make the rest of the code tidier.

Also, just call it conditional()! :-)

gpflow/models/svgp.py
```diff
@@ -101,7 +100,14 @@ def build_prior_KL(self):
             K = None
         else:
             K = self.feature.Kuu(self.kern, jitter=settings.numerics.jitter_level)
-        return kullback_leiblers.gauss_kl(self.q_mu, self.q_sqrt, K)
+        M = tf.shape(self.q_mu)[0]
+
+        # M x L x R -> M x RL
+        # TODO: Extra transpose gives axes same order for q_mu as q_sqrt. Not necessary for valid KL. Keep?
```
st--: transposes aren't free... what's the advantage of this then?

gpflow/multikernels.py
```diff
@@ -1,25 +1,40 @@
...
-class MultiKernel(Kernel):
+class IndependentMultiKernel(MultiKernel):
     pass
 
 
 class IndependentFeature(InducingFeature):
     pass
 
 
+class MultiInducingPoints(InducingPoints):
```
st--: Maybe I'm missing something but how is this different from the InducingPoints base class?

gpflow/multikernels.py
```diff
@@ -1,25 +1,40 @@
...
+
+    @params_as_tensors
+    def Kuf(self, kern, Xnew):
+        Kzx = kern.K(self.Z, Xnew)
+        return Kzx
+
+
 class Independent(Combination, IndependentMultiKernel):
-    def K(self, X, X2=None, presliced=False, full_cov_output=False):
+    def K(self, X, X2=None, presliced=False, full_cov_output=True):
```
st--: I just thought maybe I keep finding `full_cov_output` confusing because "full output-covariance" makes more sense than "full covariance output" (which is what I keep reading into this instead of "full covariance *between the* outputs"? Before we add it all to gpflow-master maybe we can spend a bit more time thinking about how to name things so they make sense...

gpflow/multikernels.py
```diff
@@ -35,26 +50,38 @@ def Kdiag(self, X, presliced=False, full_cov_output=False):
         return tf.matrix_diag(stacked) if full_cov_output else stacked  # N x P x P  or  N x P
 
 
-class IndependentFeature(InducingPoints):
+class IndependentSharedInducingPoints(InducingPoints, IndependentFeature):
+    @params_as_tensors
     def Kuf(self, kern, Xnew):
-        pass  # N x P x M
+        if type(kern) is Independent:
```
st--: Yep I think we should make Kuf() and Kuu() multiple-dispatch functions instead of inducing features methods. Will do that in a separate PR to the multipledispatch_featureconditionals branch.

gpflow/multikernels.py
```diff
@@ -1,25 +1,40 @@
...
 from .kernels import Kernel, Combination
 from .params import Parameter
 
 
-class IndependentMultiKernel(Kernel):
+class MultiKernel(Kernel):
     pass
 
 
-class MultiKernel(Kernel):
```
st--: I'd prefer being explicit and calling it MultiOutputKernel

gpflow/conditionals.py
```diff
@@ -122,6 +232,267 @@ def base_conditional(Kmn, Kmm, Knn, f, *, full_cov=False, q_sqrt=None, white=Fal
...
+    """
+    # TODO: Allow broadcasting over L if priors are shared?
+    # TODO: Change Kmn to be L x M x N x P? Saves a transpose...
+    R = tf.shape(f)[2]
+    M, L, N, P = [tf.shape(Kmn)[i] for i in range(Kmn.shape.ndims)]
+
+    Lm = tf.cholesky(Kmm)  # L x M x M
+
+    # Compute the projection matrix A
+    Kmn = tf.reshape(tf.transpose(Kmn, (1, 0, 2, 3)), (L, M, N * P))
```
st--: just checking- it's not possible to avoid the transpose by having Kmn computed differently ?

gpflow/conditionals.py
```diff
@@ -16,12 +16,133 @@
...
+from .kernels import Kernel
+from .multikernels import MultiKernel, IndependentMultiKernel, IndependentFeature, MultiInducingPoints
 from .probability_distributions import Gaussian
 
 
+# TODO: Make all outputs of conditionals equal
+# TODO: Add tensorflow assertions of shapes
+# TODO: Remove `conditional()`?
+# TODO: Ensure that R is handled correctly in all cases
+# TODO: Should there be consistentcy between fmean out, and f in?
```
st--: what does this comment mean by "consistency" ?

gpflow/models/svgp.py
markvdw: Code readability and that it's easier to see that the code is correct. Otherwise we're relying on slightly undefined behaviour of the KL divergence function. (i.e. that means and covariances don't interact in the whitened case). Actually, the more I think about it, the more I think that the extra transpose is necessary.

gpflow/conditionals.py
markvdw: Yes, there's a TODO for that up top.

gpflow/conditionals.py
```diff
@@ -122,6 +232,267 @@ def base_conditional(Kmn, Kmm, Knn, f, *, full_cov=False, q_sqrt=None, white=Fal
...
+            LTAr = tf.transpose(tf.reshape(LTA, (R, L * M, N, P)), [0, 3, 1, 2])  # R x P x LM x N
+            fvar = fvar[None, :, :, :] + tf.matmul(LTAr, LTAr, transpose_a=True)  # R x P x N x N
+        elif not full_cov and full_cov_output:
+            LTAr = tf.transpose(tf.reshape(LTA, (R, L * M, N, P)), [2, 0, 1, 3])  # N x R x LM x P
+            fvar = fvar[:, None, :, :] + tf.matmul(LTAr, LTAr, transpose_a=True)  # N x R x P x P
+        elif not full_cov and not full_cov_output:
+            # N x R x P
+            fvar = fvar[None, :, :] + tf.reshape(tf.reduce_sum(tf.square(LTA), (1, 2)), (R, N, P))
+            fvar = tf.transpose(fvar, (1, 0, 2))
+    return fmean[:, 0, :], fvar[:, 0, :]
```
st--: why only the first element? is this because it's length 1 along that dimension and that's supposed to be a squeeze?

gpflow/conditionals.py
```diff
@@ -122,6 +232,267 @@ def base_conditional(Kmn, Kmm, Knn, f, *, full_cov=False, q_sqrt=None, white=Fal
...
+        elif not full_cov and full_cov_output:
+            LTAr = tf.transpose(tf.reshape(LTA, (R, M, N, K)), [2, 0, 3, 1])  # N x R x K x M
+            fvar = fvar[:, None, :, :] + tf.matmul(LTAr, LTAr, transpose_b=True)  # N x R x K x K
+        elif not full_cov and not full_cov_output:
+            addvar = tf.reshape(tf.reduce_sum(tf.square(LTA), 1), (R, N, K))
+            fvar = fvar[:, None, :] + tf.transpose(addvar, (1, 0, 2))  # N x R x K
+    return fmean, fvar
+
+
+# def dependent_conditional(Kmn, Kmm, Knn, f, full_cov=False, full_cov_output=False, q_sqrt=None, white=False):
```
st--: What's the state of the commented-out code? Broken? Not tested? ...

st--:
There's a lot of duplicate code in the various types of conditionals ... e.g. they all do L = cholesky(Kmm), A = L^-1 Lmn ... I think it'd be much cleaner & easier to understand if we break things up from the bottom up, e.g. something like get_A_and_fvar with multiple dispatch for the different combinations of feature & kernel to return the appropriately shaped objects, and then a *single* "general" conditional that calls these helper functions instead of doing all the nitty-gritty reshaping by itself.

gpflow/models/svgp.py
st--: Do you mean some implicit broadcasting in the KL divergence implementation? or what?

gpflow/conditionals.py
```diff
@@ -36,55 +35,91 @@
...
     rmu, rvar = tf.map_fn(single_gp_conditional,
                           (Kmms, Kmns, Knns, fs, q_sqrts),
-                          (settings.float_type, settings.float_type))  # P x N x R  ,  P x N(x N)x R
+                          (settings.float_type, settings.float_type))  # P x N x 1  ,  P x N(x N) x 1
 
-    rmu = tf.transpose(rmu[:, :, 0])
-    rvar = tf.transpose(rvar[:, :, 0])
-    return rmu, rvar
+    if full_cov or full_cov_output:
+        raise NotImplementedError
```
vdutor: Why aren't we implementing this? 
We would like to use this case but with `full_cov` and/or `full_cov_output` equal to true.

markvdw: Just haven't got round to it. I stared a convenience function up top to do make the matrix diagonals etc (in this case all these output covariances are diagonal).

st--:
Already merged through #668

PR #665, opened 2018-03-02
Minor improvement to decorators and AutoBuild

st--:
- makes sure that the AutoBuild metaclass does not overwrite the `__init__`'s docstring anymore
- makes the naming of the functions inside the decorators more consistent
- improves the name_scope() docstring

### DIFF
--- a/gpflow/core/compilable.py
+++ b/gpflow/core/compilable.py
@@ -89,8 +89,7 @@ class AutoBuild(abc.ABCMeta):
             if autobuild_on and global_autobuild_on:
                 self.build()
                 self.initialize(force=True)
-        if origin_init.__doc__:
-            __init__.__doc__ = '\n'.join([origin_init.__doc__, __init__.__doc__])
+        __init__.__doc__ = origin_init.__doc__
         setattr(new_cls, '__init__', __init__)
         return new_cls
 
--- a/gpflow/core/compilable.py
+++ b/gpflow/core/compilable.py
@@ -89,6 +89,8 @@ class AutoBuild(abc.ABCMeta):
             if autobuild_on and global_autobuild_on:
                 self.build()
                 self.initialize(force=True)
+        if origin_init.__doc__:
+            __init__.__doc__ = '\n'.join([origin_init.__doc__, __init__.__doc__])
         setattr(new_cls, '__init__', __init__)
         return new_cls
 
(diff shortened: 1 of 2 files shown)


gpflow/decors.py
```diff
@@ -29,19 +29,24 @@
...
-    by either `name` option or `name` option is None then name of the
-    function will be used.
+    This decorator wraps a function so that it runs inside a TensorFlow
+    name scope. The name is given by the `name` option; if this is None,
+    then the name of the function will be used.
+    >>> @name_scope()
+    >>> def foo(...):
+    >>>     # now runs inside scope "foo"
+    >>> @name_scope('bar')
+    >>> def @baz(...):
```
jameshensman: is this @ meant to be here?

st--:
Can this be merged now please? Ta:)

gpflow/decors.py
```diff
@@ -29,19 +29,24 @@
...
 def name_scope(name=None):
     """
-    Name scope decorator does little trick with scope naming. The wrapped
-    function will be run inside TensorFlow name scope with name specified
-    by either `name` option or `name` option is None then name of the
-    function will be used.
+    This decorator wraps a function so that it runs inside a TensorFlow
+    name scope. The name is given by the `name` option; if this is None,
+    then the name of the function will be used.
+    >>> @name_scope()
```
awav: Doc generator might use markdown, could you wrap code with *``` ```*

gpflow/core/compilable.py
```diff
@@ -89,6 +89,8 @@ def __init__(self, *args, **kwargs):
             if autobuild_on and global_autobuild_on:
                 self.build()
                 self.initialize(force=True)
+        if origin_init.__doc__:
+            __init__.__doc__ = '\n'.join([origin_init.__doc__, __init__.__doc__])
```
awav: Guess, replacing will be just fine. You don't really need docstring of autobuild in descendant.

st--: Well, the autobuild parses the kwargs before passing them on, so might be useful to keep that in the docstring, no?

awav: autobuild doesn't have any docstrings

st--: It does: core/compilable.py:73: ```            """
            The `kwargs` may or may not contain 'autobuild' option. This option is
            inherited implicitly by all classes.
            """
```

st--: That's the thing that was overwriting the Model `__init__` docstrings in the first place.

PR #666, opened 2018-03-04
Adds rational-quadratic kernel and minor clean-up of notation

st--:
- renames square_dist and euclid_dist to scaled_square_dist and scaled_euclid_dist to be more consistent about what they actually return, and adds docstrings
- adds SquaredExponential as alias to RBF
- adds RationalQuadratic kernel
- includes a test for Rational Quadratic (\alpha->\infty) -> Squared Exponential

### DIFF
--- a/gpflow/kernels.py
+++ b/gpflow/kernels.py
@@ -231,6 +231,21 @@ class Stationary(Kernel):
             self.lengthscales = Parameter(lengthscales, transform=transforms.positive)
             self.ARD = False
 
+
+    def square_dist(self, X, X2):
+        warnings.warn('square_dist is deprecated and will be removed at '
+                      'GPflow version 1.2.0. Use scaled_square_dist.',
+                      DeprecationWarning)
+        return self.scaled_square_dist(X, X2)
+
+
+    def euclid_dist(self, X, X2):
+        warnings.warn('euclid_dist is deprecated and will be removed at '
+                      'GPflow version 1.2.0. Use scaled_euclid_dist.',
+                      DeprecationWarning)
+        return self.scaled_euclid_dist(X, X2)
+
+
     @params_as_tensors
     def scaled_square_dist(self, X, X2):
         """
--- a/gpflow/kernels.py
+++ b/gpflow/kernels.py
@@ -232,14 +232,14 @@ class Stationary(Kernel):
             self.ARD = False
 
 
-    def square_dist(self, X, X2):
+    def square_dist(self, X, X2):  # pragma: no cover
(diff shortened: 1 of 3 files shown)


gpflow/kernels.py
```diff
@@ -232,7 +232,11 @@ def __init__(self, input_dim, variance=1.0, lengthscales=None,
             self.ARD = False
 
     @params_as_tensors
-    def square_dist(self, X, X2):
+    def scaled_square_dist(self, X, X2):
+        """
+        Returns ((X - X2ᵀ)/lengthscales)².
+        Due to the implementation, the result may actually be negative.
+        """
```
markvdw: I'm against renaming this method. Second opinion of @jameshensman? If James agrees to the renaming, I would like to insist on deprecating `square_dist()` with a warning for the time being.

Also, with "the result may be actually be negative", you mean due to floating point precision, right? Could you clarify this in the docstring. Right now it just seems like the implementation is wrong.

gpflow/kernels.py
```diff
@@ -248,8 +252,11 @@ def square_dist(self, X, X2):
         return dist
 
 
-    def euclid_dist(self, X, X2):
-        r2 = self.square_dist(X, X2)
+    def scaled_euclid_dist(self, X, X2):
+        """
+        Returns |(X - X2ᵀ)/lengthscales| (L2-norm).
+        """
+        r2 = self.scaled_square_dist(X, X2)
```
markvdw: Idem above.

gpflow/kernels.py
```diff
@@ -266,7 +273,35 @@ class RBF(Stationary):
...
+    """
+    Rational Quadratic kernel,
+
+    k(r) = σ² (1 + r² / 2αℓ²)^(-α)
+
+    σ² : variance
+    ℓ  : lengthscales
+    α  : alpha, determines relative weighting of small-scale and large-scale fluctuations
+
+    For α→ ∞, the RQ kernel becomes equivalent to the squared exponential.
```
markvdw: Add space after alpha. Very nitpicky, but the look of the lack of space makes me sqirm :)

Also, I would do it myself, but then my review would be attached to an old commit.

st--: Funnily enough, there is no space *after* the \alpha either... it's just a quirk of the font :roll_eyes: but I'm happy to add a space beforehand if that makes you happier

gpflow/kernels.py
st--: Happy to add deprecated methods.

PR #669, opened 2018-03-07
Move to multiple dispatch for Kuu & Kuf, and some other cleaning-up

st--:


gpflow/conditionals.py
```diff
@@ -66,12 +76,13 @@
...
 
 def expand_independent_outputs(fvar, full_cov, full_cov_output):
+    # TODO point of this function?
     if not full_cov_output:
         # Output shape should be N x R or R x N x N, which it already is.
         return fvar
     elif full_cov:
         # Output shape should be ???
-        pass
+        raise NotImplementedError
```
markvdw: This function still needs to be implemented. In situations where we have independent outputs, we will always have a diagonal over P when full_cov_output is True. The intent for this function is that it reshapes from N x P to the appropriate shape.

gpflow/decors.py
```diff
@@ -123,14 +123,16 @@ def compute_something(self):  # self is parameterized object.
...
         feature on, `True`, or turning it off, `False`.
     """
-    prev_value = _params_as_tensors_enter(obj, convert)
+    prev_values = [_params_as_tensors_enter(o, convert) for o in objs]
     try:
         yield
     finally:
-        _params_as_tensors_exit(obj, prev_value)
+        for o, pv in zip(objs, prev_values):
+            _params_as_tensors_exit(o, pv)
```
markvdw: I think the changes in this file would best have been done in a separate PR. I think it's best if @awav signs off on these. I'm sure they're fine. I just don't know the code inside out.

st--: No problem, was thinking that as well (after the fact), but I can still make a PR just for that commit.

st--: See #672

gpflow/conditionals.py
st--: Yeah, but that's not something that's gonna happen in this branch...

st--:
The failed tests are due to the NotImplementedError inserted into the SVGP's build_KL ... not my fault :stuck_out_tongue:

vdutor:
I know, that's why I started to work on the KL-divergences :)

PR #672, opened 2018-03-08
allow params_as_tensors_for() to take several objects as varargs

st--:


st--:
@awav - how would you test this - create a param and check whether what we access is a tensor or an instance of param? in which test class in test_param would this belong?

st--:
@awav could you please clarify what you would like added before merging this?

PR #674, opened 2018-03-08
Natural Gradient Optimizer

hughsalimbeni:
Adds natural gradient optimization to gpflow models with variational distributions parameterized with `q_mu`, `q_sqrt`. Currently this only works on its own rather than in combination with other optimizers.


markvdw:
Can we have a notebook with an example for natural gradients?

gpflow/training/natgrad_optimizer.py
```diff
@@ -0,0 +1,322 @@
...
+#
+# Auxiliary gaussian parameter conversion functions.
+# 
+# The following functions expect their first and second inputs to have shape
+# DN1 and DNN, respectively. Return values are also of shapes DN1 and DNN.
+
+
+def swap_dimensions(method):
+    """
+    Converts between GPflow indexing and tensorflow indexing
```
st--: Is this still necessary after @markvdw's rolling of dimensions in #612  ?

hughsalimbeni: Yes, for the mean only now though (if the mean is KM1 and the var is KMM then all the matrix ops work out nicely). It could be written without, but at the expense of being slightly harder to read

tests/test_optimizers.py
```diff
@@ -215,3 +223,164 @@ class TestFtrlOptimizer(GPflowTestCase, OptimizerCase):
 # E               TypeError: Input 'lr' of 'ApplyProximalAdagrad' Op has type float32 that does not match type float64 of argument 'var'.
 #
 # ../../anaconda3/lib/python3.6/site-packages/tensorflow/python/framework/op_def_library.py:546: TypeError
+
+def test_VGP_vs_GPR(session_tf):
```
jameshensman: Just wanted to say that I absolutely love this test.

jameshensman: It could do with a docstring though. Obvious to me, bizarre to the unitiated

tests/test_optimizers.py
```diff
@@ -215,3 +223,164 @@ class TestFtrlOptimizer(GPflowTestCase, OptimizerCase):
...
+            feed_dict = self._gen_feed_dict(model, feed_dict)
+
+            for _i in range(maxiter):
+                for op in minimize_ops:
+                    session.run(op, feed_dict=feed_dict)
+
+        if anchor:
+            model.anchor(session)
+
+# # TODO(@hugh) this needs to be fixed
```
jameshensman: What's wrong with this code? We should fix it before merging.

gpflow/training/__init__.py
```diff
@@ -2,4 +2,8 @@
 
 from .scipy_optimizer import ScipyOptimizer
 from .hmc import HMC
+from .natgrad_optimizer import XiTransform
```
jameshensman: Do we really need all of these imports to run the natgrad code?

tests/test_optimizers.py
hughsalimbeni: This will be superseded by the Actions PR

hughsalimbeni: The problem is that if the q_mu, q_sqrt params are not trainable their change doesn't get noticed by the ordinary grad optimizer

gpflow/training/__init__.py
hughsalimbeni: Not necessary. Will remove

hughsalimbeni: Actually not sure. Leaving them in

awav: hm, someone will decide to create their transforms, this is just a shortcut

tests/test_optimizers.py
awav: Oh no!!! I've just realized what's wrong with this case!

gpflow/training/natgrad_optimizer.py
stefanosele: I don't see the purpose of this code anymore, since it is only called for the mean. Is it expected that two transposes and a matmul are more efficient than an einsum? If not then I would just remove that and use einsums

tests/test_optimizers.py
```diff
@@ -215,3 +227,213 @@ class TestFtrlOptimizer(GPflowTestCase, OptimizerCase):
...
+    m_svgp.set_trainable(False)
+    m_svgp.q_mu.set_trainable(True)
+    m_svgp.q_sqrt.set_trainable(True)
+    NatGradOptimizer(1.).minimize(m_svgp, [[m_svgp.q_mu, m_svgp.q_sqrt]], maxiter=1)
+
+    assert_allclose(m_sgpr.compute_log_likelihood(),
+                    m_svgp.compute_log_likelihood(), atol=1e-5)
+
+
+class CombinationOptimizer(Optimizer):
```
stefanosele: Shouldn't this class live in the `natgrad_optimizer` module?

hughsalimbeni: This is just for testing purposes and also serves as a demo for how the Actions might work. When Actions is done this won't be necessary

awav: @stefanosele, we will remove it as soon as Actions are merged into master.

tests/test_optimizers.py
awav: Agreed! @hughsalimbeni

gpflow/training/natgrad_optimizer.py
```diff
@@ -0,0 +1,330 @@
...
+import abc
+import functools
+
+import tensorflow as tf
+
+from gpflow.models import Model
+
+from . import optimizer
+from .. import settings
+from ..models import Model
```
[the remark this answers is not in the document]
awav: duplicate?

PR #678, opened 2018-03-11
GPflow Actions.

awav:
Hello everyone,

This PR introduces new GPflow abstraction for running complex optimization loops.
For e.g. example Natural Gradients optimizer can be used in combination with standard Tensorflow optimizers in non-trivial manner.
I created action followed abstractions, which are actually functors:
  - Loop
  - Condition
  - Optimization

How to use:

1.

```
m = <Any GPflow model>
opt1 = gpflow.train.AdamOptimizer(0.1).make_optimization(m)
opt2 = gpflow.train.MomentumOptimizer(0.0001).make_optimization(m)
cond = gpflow.actions.Condition(lambda ctx: ctx.iteration < 1e3, opt1, opt2)
loop = gpflow.actions.Loop(cond).with_settings(stop=int(1e6))
loop() # runs 1000 iterations with two optimizers.
```

2.

```
sudden_stop = lambda cxt: numpy.random.randn(1)[0] > 0.8)
exit_cond = lambda ctx: raise gpflow.actions.Loop.Break
cond = gpflow.actions.Condition(sudden_stop, exit_cond, opt1)
loop = loop.with_action(cond)()
```

Retrospective: in future we could add TensorFlow board, logging or extend an API of actions and make it even better.


st--:
Is your `Loop()` supposed to take the `cond` as argument?
What's the relation of this to gpflow-monitor ?

awav:
@st-- , thx, I have updated topic description. It is attempt to generalize what was done in gpflow-monitor, and make it less hacky.

gpflow/actions.py
```diff
@@ -0,0 +1,386 @@
...
+    
+    def with_run_kwargs(self, **kwargs: Dict[str, Any]) -> 'Optimization':
+        """
+        Replace Tensorflow session run kwargs.
+        Check Tensorflow session run [documentation](https://www.tensorflow.org/api_docs/python/tf/Session).
+
+            :param kwargs: Dictionary of tensors as keys and numpy arrays or
+                primitive python types as values.
+            :return: Optimization instance self reference.
+        """
```
lucabordo: I guess this should only be used through the public method run_kwargs(),
Currently there is no test that exercises this method?

gpflow/actions.py
```diff
@@ -0,0 +1,386 @@
...
+
+            :param context: Action context.
+        """
+        iterator = itertools.count(start=self.start, step=self.step)
+        for i in iterator:
+            self.with_iteration(i)
+            if self.stop is not None and i >= self.stop:
+                break
+            try:
+                self._action(context)
```
lucabordo: I am not a big fan of using exceptions for control flow - something that is rightly considered a bad idea in any other language. 
Because it's Pythonic (i.e. "Guido said we should do that" ;) ) it's not a big deal, and does not seem to be a source of possible confusion in this context...
Could we instead use enumerators? Could the actions just return an enum value saying whether "break" or "continue"?

lucabordo: I think this exception business gets nicer anyway if, rather than exposing publicly the Exception types, what gets exposed publicly is the 'continue_loop' and 'break_loop' functions.

gpflow/training/natgrad_optimizer.py
```diff
@@ -0,0 +1,388 @@
...
+            :param session: Tensorflow session.
+            :param var_list: List of tuples of variational parameters.
+            :return: Tensorflow natural gradient operation.
+        """
+        with session.as_default(), tf.name_scope(self.name):
+            # Create optimizer variables before initialization.
+            return self._build_natgrad_step_ops(model, *var_list)
+
+    def make_optimization(self, model, session=None, var_list=None, **kwargs):
+        """
```
lucabordo: Optimization (typo)

gpflow/actions.py
```diff
@@ -0,0 +1,386 @@
...
+        """
+        Elapsed time since owner started running dependent actions.
+
+            :return: Spent time [µs] by owner on running up to the call.
+        """
+        return self.owner.watcher.elapsed
+
+
+class Action(metaclass=abc.ABCMeta):
+    """
```
lucabordo: These actions seem to play the role of "OptimizationAction"s, specifically - or it is fine to keep this very general terminology?

awav: Later we can add TensorboardAction, LoggingAction and cetera.

gpflow/actions.py
awav: 1. Exceptions are inevitable, because action may return something meaningful in future to make monad-like execution
2. I like that, I will add break_loop & continue_loop methods

gpflow/actions.py
awav: in fact, run_kwargs does not run anything, it is run's kwargs for tensorflow session or did I misunderstand you?

gpflow/actions.py
```diff
@@ -0,0 +1,386 @@
...
+    @abc.abstractmethod
+    def run(self, context: ActionContext) -> None:
+        """
+        Run action's execution. Must be implemented by decendants.
+
+            :param context: Required action context.
+        """
+        pass
+    
+    @property
```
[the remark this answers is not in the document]
awav: We must have annotations. I hate personally python annotations, but they help for reading code.

gpflow/actions.py
lucabordo: But these are still actions done during the optimization?

gpflow/actions.py
lucabordo: I struggled to locate a place where the saved _run_args are used, but actually I now see it: it's obviously passed systematically to run

gpflow/training/scipy_optimizer.py
```diff
@@ -25,6 +25,32 @@ def __init__(self, **kwargs):
...
+        Make SciPy optimization tensor.
+        The `make_optimizer_tensor` method builds optimization tensor and initializes
+        all necessary variables created by optimizer.
+
+            :param model: GPflow model.
+            :param session: Tensorflow session.
+            :param var_list: List of variables for training.
+            :param kwargs: Scipy optional optimization parameters,
+                - `maxiter`, maximal number of iterations to perform.
+                - `disp`, if True, prints convergence messages.
```
lucabordo: It seems to return a tf.Tensor or a tf.Operation, and the same make_optimizer_operation would seem more natural to me (with the other method called make_optimizer_action). Just personal preferences, feel free to differ

gpflow/actions.py
awav: Or anything else :)
Optimization is only part of this process, action is more generalized action. E.g. loop is not optimization, condition too. Logger action is not part of optimization.

gpflow/training/scipy_optimizer.py
awav: `*_tensor` refers to tensorflow, not a tensor as a structure.

st--: Then call it make_tensorflow_optimizer ?

awav: @st-- latest :)

gpflow/actions.py
awav: The point is that action can be anything (including optimization). Technically, yes, loop created to help to simplify complex optimization steps, but still is not correct to call it "OptimiztionLoop", as you can run w/o actual optimization.

gpflow/training/natgrad_optimizer.py
awav: I don't see typo. Is that in description? Line below?

lucabordo: Optimizatin -> missing 'o'

awav: Ah, it is under the block with your comment. In review mode github does not show me anything below.

PR #682, opened 2018-03-18
Update gauss_kl to work with K matrices of shape L x M x M (previously M x M only)

vincentadam87:
* Updated gauss_kl
* added tests to check that KL(q(X)|p(X)) = \sum_l KL(q(x_l)|p(x_l))  ( for both full and diagonal q_sqrt )


gpflow/kullback_leiblers.py
```diff
@@ -80,26 +83,49 @@ def gauss_kl(q_mu, q_sqrt, K=None):
         trace = tf.reduce_sum(tf.square(Lq))
     else:
         if diag:
```
awav: Can you get rid of `if if else ifelse else if` :)?

tests/test_kldiv.py
```diff
@@ -172,6 +172,111 @@ def test_dense(self):
             self.assertTrue(np.allclose(res, 0.0))
 
 
+class EqualitySumKLBatchKL(GPflowTestCase):
```
awav: Please, use `pytest` instead of `GPflowTestCase`

tests/test_kldiv.py
```diff
@@ -172,6 +172,111 @@ def test_dense(self):
...
+            kl_batch = gpflow.kullback_leiblers.gauss_kl(self.mu_batch,self.sqrt_batch,self.K_batch)
+            res_batch = sess.run(kl_batch, feed_dict=self.feed_dict)
+            kl = gpflow.kullback_leiblers.gauss_kl(self.mu, self.sqrt, self.K)
+            res_sum = 0.
+            for l in range(self.L):
+                res_sum += sess.run(kl, feed_dict=self.feed_dicts[l])
+        self.assertTrue(np.allclose(res_sum, res_batch))
+
+
+class EqualitySumKLBatchKLqdiag(GPflowTestCase):
```
awav: Same is here: `pytest`

tests/test_kldiv.py
```diff
@@ -172,6 +172,111 @@ def test_dense(self):
...
+    p(X) has covariance L x M x M
+    Here, q(x_i) has covariance M x 1
+    p(x_i) has covariance M x M
+    """
+
+    def setUp(self):
+        M, L = 4, 3
+        self.M, self.L = M, L
+
+        self.mu_batch = tf.placeholder(settings.float_type, [M, L])
```
awav: Is it a duplication of previous setup? If so, you need `Datum` class with static `numpy` values. Convert them to tensors via fixtures of pytest.

gpflow/kullback_leiblers.py
```diff
@@ -80,26 +83,49 @@ def gauss_kl(q_mu, q_sqrt, K=None):
...
+
     twoKL = mahalanobis + constant - logdet_qcov + trace
 
     # Log-determinant of the covariance of p(x):
     if not white:
         log_sqdiag_Lp = tf.log(tf.square(tf.matrix_diag_part(Lp)))
         sum_log_sqdiag_Lp = tf.reduce_sum(log_sqdiag_Lp)
-        prior_logdet = tf.cast(num_latent, settings.float_type) * sum_log_sqdiag_Lp
+        # If K is L x M x M, num_latent is no longer implicit, no need to multiply the single kernel logdet
+        scale = tf.cast(L, settings.float_type) if K.get_shape().ndims == 2 else 1.
```
st--: `1.` -> `1.0` to be more explicit about the type (my first thought was "huh, you can python like a sentence?"...)

gpflow/kullback_leiblers.py
```diff
@@ -80,26 +83,49 @@ def gauss_kl(q_mu, q_sqrt, K=None):
...
         else:
-            Lp_tiled = tf.tile(tf.expand_dims(Lp, 0), [num_latent, 1, 1])
-            LpiLq = tf.matrix_triangular_solve(Lp_tiled, Lq, lower=True)
+
+            # If K is L x M x M, no need to tile the cholesky anymore
+            if K.get_shape().ndims == 3:
+                LpiLq = tf.matrix_triangular_solve(Lp, Lq, lower=True)
+            else:
+                Lp_tiled = tf.tile(tf.expand_dims(Lp, 0), [num_latent, 1, 1])
+                LpiLq = tf.matrix_triangular_solve(Lp_tiled, Lq, lower=True)
```
st--: to minimise code divergence, it'd be good to move the matrix_triangular_solve out of the if, only keep the definition of the argument inside the branching

gpflow/kullback_leiblers.py
```diff
@@ -80,26 +83,49 @@ def gauss_kl(q_mu, q_sqrt, K=None):
...
+                    ) # L x M
+                ) # M x L
+            else:
+                LpT = tf.transpose(Lp) # M x M
+                Lp_inv = tf.matrix_triangular_solve(Lp,
+                        tf.eye(M, dtype=settings.float_type), #  M x M
+                        lower=True) # M x M
+                K_inv = tf.matrix_diag_part(
+                    tf.matrix_triangular_solve(LpT, Lp_inv, lower=False)  # M x M
+                )[:,None] # M x 1 (implicit extension to broadcast)
```
st--: would be good to minimise code divergence and only have a single call to matrix_triangular_solve and matrix_diag_part, and only set arguments to these functions in the if: else:

gpflow/kullback_leiblers.py
vincentadam87: Well, there is no easy way to do that.
There are different treatment for the factors white x batch x diag...
I can do some inline ifs if that helps ,,,, like if x then y else z.

tests/test_kldiv.py
vincentadam87: not sure what that means $:

tests/test_kldiv.py
```diff
@@ -18,225 +18,128 @@
...
-    return np.sum(kls)
-
-
-class OneDTest(GPflowTestCase):
+    s = sqrt_diag if diag else sqrt
+    kl_batch = gpflow.kullback_leiblers.gauss_kl(mu,s,K_batch)
+
+    kl_sum = []
+    for n in range(Datum.N):
+        s = sqrt_diag[:,n][:,None] if diag else sqrt[n, :, :][None,:,:]
```
awav: Can you leave a comment why you added this line?

vincentadam87:
Thanks @st-- for the help!

tests/test_kldiv.py
```diff
@@ -18,225 +18,133 @@
...
-            kl_nonwhite = gpflow.kullback_leiblers.gauss_kl(self.mu, self.sqrt, self.I)
-
-            res_white = sess.run(kl_white, feed_dict=self.feed_dict)
-            res_nonwhite = sess.run(kl_nonwhite, feed_dict=self.feed_dict)
+    kl_white = gauss_kl(mu, s)
+    kl_nonwhite = gauss_kl(mu, s, I)
 
-            np.testing.assert_allclose(res_white, res_nonwhite)
+    np.testing.assert_allclose(kl_white.eval(),
+                               kl_nonwhite.eval())
```
awav: I apologize for the pickiness, but have lines longer than 80 chars and here you injected newline. Why is that?

PR #690, opened 2018-03-26
Added a comparison of manual implementation of KL with implementations based on tf.contrib.distributions

vincentadam87:
implementations by @vdutor 
tests by @vincentadam87


vdutor:
I would propose to change the upstream to `vdutor/multiple-output-gps` instead of directly merging into `master`. The functionality you added to the KLs is used by the multioutput framework and I need it to fix a currently broken unit test.

tests/test_kldiv.py
```diff
@@ -24,20 +25,62 @@
 from gpflow import settings
 from gpflow.test_util import session_tf
 
+
+def gauss_kl_tf_distributions(q_mu, q_sqrt, K=None):
+    """
+    Kullbach-Leiber divergence KL[q(U) || p(U)],
```
st--: Kullbach-Leiber -> Kullback-Leibler

tests/test_kldiv.py
```diff
@@ -24,20 +25,62 @@
...
+    with q(U) ~ N(q_mu, q_sqrt^2) the variational Gaussian posterior
+    and p(U) ~ N(0, K) the prior. If K is None we assume a whitened
+    prior, p(U) ~ N(0, I).
+
+    We assume L independent distributions, then
+
+    :param q_mu: L variational means, M x L
+    :param q_sqrt: L variational covariances,
+                - cholesky: L x M x M or
+                - diag elements: M x L
```
st--: just for clarification - is this still the sqrt of the diagonal elements?

st--:
@vdutor should we just merge this ?

vdutor:
Yes, this can be merged.

PR #693, opened 2018-03-27
added kwargs in _expectations to pass around gauss_hermite points. Al…

stefanosele:
Hi guys a simple pull request for fixing the following minor issues in `expectations.py`:

1) Added `kwargs` arguments in all `_expectation` signatures in order to correctly be able to pass around the `num_gauss_hermite_points`. In the old code if we are falling back to quadrature by a nested call of expectation, e.g. calling `expectation` inside one of the dispatched versions of `_expectation`, we loose track of the user defined `num_guass_hermite_points` and we revert to the default value. This is now handled

2) Computation of exKxz kernels was handled in the dispatcher by the general type `kernels.Kernel`. This type results in infinite loops (dispatcher chasing its tail) when calling it with an arbitrary kernel, e.g., Matern. I changed the type to `(kernels.Combination, kernels.RBF, kernels.Linear)`, so if the signature is not found it correctly falls back to quadrature.

### DIFF
--- a/gpflow/expectations.py
+++ b/gpflow/expectations.py
@@ -105,7 +105,7 @@ def _quadrature_expectation(p, obj1, feature1, obj2, feature2, num_gauss_hermite
     General handling of quadrature expectations for Gaussians and DiagonalGaussians
     Fallback method for missing analytic expectations
     """
-    num_gauss_hermite_points = 1 if num_gauss_hermite_points is None else num_gauss_hermite_points
+    num_gauss_hermite_points = 100 if num_gauss_hermite_points is None else num_gauss_hermite_points
 
     warnings.warn("Quadrature is used to calculate the expectation. This means that "
                   "an analytical implementations is not available for the given combination.")
--- a/gpflow/expectations.py
+++ b/gpflow/expectations.py
@@ -105,7 +105,7 @@ def _quadrature_expectation(p, obj1, feature1, obj2, feature2, num_gauss_hermite
     General handling of quadrature expectations for Gaussians and DiagonalGaussians
     Fallback method for missing analytic expectations
     """
-    num_gauss_hermite_points = 100 if num_gauss_hermite_points is None else num_gauss_hermite_points
+    num_gauss_hermite_points = 1 if num_gauss_hermite_points is None else num_gauss_hermite_points
 
     warnings.warn("Quadrature is used to calculate the expectation. This means that "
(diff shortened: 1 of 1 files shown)


vdutor:
Hi, 
Nice work - this needed to be done.
A minor issue I foresee with this implementation is that we will have to specify the number guass hermite points in models using `uncertain_conditional` or `expectations`, e.g. BayesianGPLVM.
These models can be configured with any kernel, so they will potentially fall back to quadrature.

gpflow/expectations.py
```diff
@@ -105,7 +105,7 @@ def _quadrature_expectation(p, obj1, feature1, obj2, feature2, num_gauss_hermite
     General handling of quadrature expectations for Gaussians and DiagonalGaussians
     Fallback method for missing analytic expectations
     """
-    num_gauss_hermite_points = 100 if num_gauss_hermite_points is None else num_gauss_hermite_points
```
stefanosele: that's definitely a mistake, I'm fixing it asap

stefanosele:
Hey @vdutor, I agree with you. We had a discussion earlier with @markvdw regarding that issue. But we are not sure whether we want to support quadrature expectations for "dummy" choice of dimensions and quadrature points.

st--:
Would it be an option to simply add num_gauss_hermite_points to gpflow.settings and use that instead of the hardcoded 100, rather than passing it around all the time ?

stefanosele:
@st-- You are not passing around the hardcoded 100 but the user-defined `num_gauss_hermite_points`. If none passed then it defaults to 100 (haven't touched that in this PR)

We also discussed with @vdutor and @markvdw of the gpflowrc option but the problem is that there is no default value for `num_gauss_hermite_points`. First even in the expectation module we have different hardcoded defaults, depending on the distribution. Also the `variational_expectation` method in `likelihoods` uses different default value. Apart from that a default value of 100 for a single dimension is prohibitive for 2-3 dims.

st--:
Yeah I didn't mean passing around hardcoded 100, I meant not passing around anything... but I see your point about different numbers needed for different distributions. Could it be a parameter of the distribution, though?

awav:
@stefanosele tests

awav:
:) So many kwargs, where exactly do you pass `**kwargs`? I can only see that kwargs wander around `_expectations` w/o going anywhere.

PR #694, opened 2018-03-27
add shape check for ARD hyperparameters to Kernel classes

st--:
Stationary kernels didn't check shape of lengthscales. If you have a non-ARD kernel, you should only be able to pass a single lengthscale. Shot myself in the foot because of this twice now. This PR adds checks (and tests for the checks).


hughsalimbeni:
Could we also have a cast to float so that lengthscales=1 doesn't break it?

gpflow/kernels.py
```diff
@@ -223,11 +223,16 @@ def __init__(self, input_dim, variance=1.0, lengthscales=None,
             else:
                 # accepts float or array:
                 lengthscales = lengthscales * np.ones(input_dim, dtype=settings.float_type)
+                if np.asarray(lengthscales).squeeze().shape != (input_dim,):
+                    raise ValueError("shape of lengthscales does not match input_dim")
```
awav: Uppercase "Shape ..."

gpflow/kernels.py
```diff
@@ -62,6 +62,37 @@ def __init__(self, input_dim, active_dims=None, name=None):
 
         self.num_gauss_hermite_points = 20
```
stefanosele: I know that this is irrelevant to this PR but just saw that.
This is deprecated and should be removed. Kernels don't use `num_gauss_hermite_points` as far as I can tell.

gpflow/kernels.py
```diff
@@ -60,7 +60,36 @@ def __init__(self, input_dim, active_dims=None, name=None):
...
+
+        Input:
+            value - scalar or array
+            ARD - None, False, or True. If None, infers ARD from shape of value
+        Output:
+            value - scalar if input_dim==1 or not ARD, array otherwise
+            ARD - False if input_dim==1 or not ARD, True otherwise
+        """
+        if ARD is None:
+            if np.asarray(value).squeeze().shape == ():
```
awav: `ARD = np.asarray(value).squeeze().shape != ()`

gpflow/kernels.py
```diff
@@ -60,7 +60,31 @@ def __init__(self, input_dim, active_dims=None, name=None):
             self.active_dims = np.array(active_dims, dtype=np.int32)
             assert len(active_dims) == input_dim
 
-        self.num_gauss_hermite_points = 20
+    def _validate_ard_shape(self, name, value, ARD=None):
```
awav: What does `name` mean?

st--: the name of the parameter, for sensible error messages

gpflow/kernels.py
```diff
@@ -60,7 +60,31 @@ def __init__(self, input_dim, active_dims=None, name=None):
...
+        """
+        Validates the shape of a potentially ARD hyperparameter
+
+        :param value: A scalar or an array.
+        :param ARD: None, False, or True. If None, infers ARD from shape of value.
+        :return: Tuple (value, ARG), where _value_ is a scalar if input_dim==1 or not ARD, array otherwise.
```
st--: ARG -> ARD

gpflow/kernels.py
awav: wouldn't it better to use kwargs then and restrict kwargs to one value? like:
```
def _validate_ard_shape(self, ARD: Optional[bool] = None, **kwargs: Dict[str, np.ndarray])
```

awav: in any case, `name` doesn't have docstring

st--:
ready to merge?

Issue #689, opened 2018-03-23
Model store to file


Bluejung:
Nice to meet you.

I want to save my VGPR models to a file. 

In order to do this, I searched related post. (https://github.com/GPflow/GPflow/issues/543)

However, I can't understand it.
Is there anyone who can explain or show an example using below code for me ?

Thank you :)
====================================================================
models = [] 
k = gpflow.kernels.Linear(input_dim=dimention, ARD=True) + gpflow.kernels.Polynomial(input_dim=dimention, ARD=True) + gpflow.kernels.ArcCosine(input_dim=dimention, ARD=True)

Z = kmeans(X, num_inducing)[0]
m = gpflow.models.SVGP(X, y, kern=k, likelihood=gpflow.likelihoods.Bernoulli(), Z=Z)
m.feature.set_trainable(True)
gpflow.train.ScipyOptimizer(options=dict(maxiter=200)).minimize(m)
models.append(m)

====================================================================

ambrishrawat:
As mentioned in [https://github.com/GPflow/GPflow/issues/543](#543 ), you can use TensorFlow saver and restore option as described [here](https://www.tensorflow.org/programmers_guide/saved_model)

* First, I would suggest adding a `name` string to your SVGP model constructor
* Then you can try something like this

```python
### some code
#
gpflow.models.SVGP(X, y, kern=k, likelihood=gpflow.likelihoods.Bernoulli(), Z=Z, name='svgp')
#
#
gpflow.train.ScipyOptimizer(options=dict(maxiter=200)).minimize(m)
#
#
saver = tf.train.Saver()
tf_graph = tf.Graph()
tf_session = tf.Session(graph=tf_graph)

with tf_graph.as_default():
        save_path = saver.restore(tf_session, "model.ckpt")
        print("Model saved in path: %s" % save_path)
```

For loading you could redefine the model

```python
### some code
#
gpflow.models.SVGP(X, y, kern=k, likelihood=gpflow.likelihoods.Bernoulli(), Z=Z, name='svgp')

saver = tf.train.Saver()
tf_graph = tf.Graph()
tf_session = tf.Session(graph=tf_graph)

with tf_graph.as_default():
        save_path = saver.restore(tf_session, "model.ckpt")
        print("Model loaded from path: %s" % save_path)
       ## fancy GP stuff
```

gdkrmr:
The following code does not work, it complains about uninitialized variables, if you run the last part for a second time, the error changes. I am on 
[...]

PR #695, opened 2018-03-28
name gaussian likelihoods

gdkrmr:
added name argument to `class Gaussian(Likelihood)`, this prevented storing and loading models to and from files. I am pretty sure that there are more of the same, but I am not familiar enough with the code base to point them out. See #689.

### DIFF
--- a/gpflow/likelihoods.py
+++ b/gpflow/likelihoods.py
@@ -141,8 +141,8 @@ class Likelihood(Parameterized):
 
 
 class Gaussian(Likelihood):
-    def __init__(self, var=1.0):
-        super().__init__()
+    def __init__(self, var=1.0, name=None):
+        super().__init__(name=name)
         self.variance = Parameter(
             var, transform=transforms.positive, dtype=settings.float_type)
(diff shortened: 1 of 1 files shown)


gdkrmr:
what still does not work is saving and loading the same model in a single Python session

awav:
Hi. `what still does not work is saving and loading the same model in a single Python session` can you elaborate what do you mean by _in single python session_?

gdkrmr:
This is because you cannot create two tensors with identical names in the same tf session. It is possible to solve this with `defer_build`, see #689.

awav:
What do you mean by "in single python session"? Is that your ipython notebook or terminal run? You could create models within different graph in that case. It doesn't make sense to store/restore models within same graph.
Also, I'm not sure what do you mean by saving models, because gpflow doesn't provide such feature at that moment. Yes, you can use tensorflow saver, but you need to be very careful about what exactly you are saving (graph can have more than just a model: optimizers and cetera) and for loading you have to use different graph.

PR #696, opened 2018-03-28
NatGrads notebook cleanup

awav:
Quick PR. ^^^ subj.


markvdw:
Hugh should look at this as well, but I have points:
- First mention of "Comparison with ordinary gradients in the non-conjugate case" should be "conjugate case" (since you're using a Gaussian likelihood)
- Likelihoods in cells 115 and 116 look wrong.

PR #700, opened 2018-04-03
Fixed regression to bug in #615

rhaps0dy:
Refactoring `expectations.py` re-introduced the bug fixed by #615, for which @markvdw and I decided to not introduce a test. This contains the bug fix and the test.

Lesson learned: always leave the tests! The bug might be re-introduced in the future.

### DIFF
--- a/tests/test_expectations.py
+++ b/tests/test_expectations.py
@@ -255,6 +255,30 @@ def test_eKzxKxz_no_uncertainty(session_tf, kernel, feature):
     assert_allclose(eKzxKxz, KzxKxz, rtol=RTOL)
 
 
+@test_util.session_context()
+def test_RBF_eKzxKxz_notNaN():
+    """
+    Ensure that <K_{Z, x} K_{x, Z}>_p(x) is not NaN and correct, when
+    K_{Z, Z} is zero with finite precision. See pull request #595.
+    """
+    kern = gpflow.kernels.RBF(1, lengthscales=0.1)
+    kern.variance = 2.
+
+    p = gpflow.probability_distributions.Gaussian(
+        tf.constant([[10]], dtype=gpflow.settings.tf_float),
+        tf.constant([[[0.1]]], dtype=gpflow.settings.tf_float))
+    z = gpflow.features.InducingPoints([[-10.], [10.]])
+
+    ekz = expectation(p, (kern, z), (kern, z))
+
+    g, = tf.gradients(ekz, kern.lengthscales._unconstrained_tensor)
+
+    sess = tf.get_default_session()
+    z.initialize(sess)
+    kern.initialize(sess)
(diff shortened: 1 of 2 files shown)


markvdw:
Lesson learned :)

markvdw:
Can you just confirm you're happy to merge @rhaps0dy?

tests/test_expectations.py
```diff
@@ -255,6 +255,30 @@ def test_eKzxKxz_no_uncertainty(session_tf, kernel, feature):
     assert_allclose(eKzxKxz, KzxKxz, rtol=RTOL)
 
 
+@test_util.session_context()
+def test_RBF_eKzxKxz_notNaN():
```
awav: `def test_RBF_eKzxKxz_notNaN():` =>
```
from gpflow.test_util import session_tf
def test_RBF_eKzxKxz_notNaN(session_tf):
```

rhaps0dy:
Yep, fine for me! There's not much to this commit, I just spotted that the bug was reintroduced.

tests/test_expectations.py
```diff
@@ -255,6 +255,30 @@ def test_eKzxKxz_no_uncertainty(session_tf, kernel, feature):
...
+        tf.constant([[10]], dtype=gpflow.settings.tf_float),
+        tf.constant([[[0.1]]], dtype=gpflow.settings.tf_float))
+    z = gpflow.features.InducingPoints([[-10.], [10.]])
+
+    ekz = expectation(p, (kern, z), (kern, z))
+
+    g, = tf.gradients(ekz, kern.lengthscales._unconstrained_tensor)
+
+    sess = tf.get_default_session()
+    z.initialize(sess)
```
awav: it has been initialized before

tests/test_expectations.py
```diff
@@ -255,6 +255,30 @@ def test_eKzxKxz_no_uncertainty(session_tf, kernel, feature):
...
+    kern.initialize(sess)
```
awav: same here

tests/test_expectations.py
```diff
@@ -255,6 +255,30 @@ def test_eKzxKxz_no_uncertainty(session_tf, kernel, feature):
...
+    np.testing.assert_almost_equal(sess.run(g), 0.79108496814443863)
```
awav: How did you get this magic `0.79108496814443863` number?

rhaps0dy: It's the value of the gradient obtained by running the code in #615 . One might as well just check that it's not `NaN` and leave some other test to catch whether the code that computes the kernel expectation is correct.

awav: I see, can you rename test name and change asserting on 
```
grad = sess.run(g)
assert grad is not None and not np.isnan(grad)
```

tests/test_expectations.py
awav: and remove `session_context()`

PR #701, opened 2018-04-03
expectations: pass num hermite points to _expectation() and minor cleaning

ialong:
This pull request fixes two issues:

- the signature `(Gaussian, identity_mean, None, kernel, features)` was calling itself whenever no specific implementation for `kernel` exists because `Identity` is a type of `Linear` mean function
- the default number of Hermite points (often too high) was used for nested calling of `expectation` as the `num_gauss_hermite_points` argument was not being passed beyond the first call

What is irking about the current fix for the second issue:

- `_expectation`'s should, in theory, have no knowledge of Hermite points and should not receive them as arguments
- the code is now slightly more verbose, though at least explicit
- to keep the code and signatures simple (and not changing 35 of them) the number of Hermite points argument (`nghp`) is not checked by multiple dispatch

The solution to fix all this would be to make `expectation` into a class but it is unclear at the moment whether this would have any other use cases. The Hermite points issue does not warrant, in my opinion, changing `expectation` to a class as that would imply a significant refactor and changes to code outside of the `expectations` submodule.

### DIFF
--- a/gpflow/expectations.py
+++ b/gpflow/expectations.py
@@ -51,9 +51,15 @@ dispatch = partial(dispatch, namespace=gpflow_md_namespace)
 
 def quadrature_expectation(p, obj1, obj2=None, num_gauss_hermite_points=None):
     """
-    Compute the expectation <obj1(x) obj2(x)>_p(x) using quadrature
-    p can be a (mu, cov) tuple or a probability_distribution
-    obj1 and obj2 can be kernels, mean functions, (kernel, features) tuples, or None
+    Compute the expectation <obj1(x) obj2(x)>_p(x)
+    Uses Gauss-Hermite quadrature for approximate integration.
+
+    :type p: (mu, cov) tuple or a `ProbabilityDistribution` object
+    :type obj1: kernel, mean function, (kernel, features), or None
+    :type obj2: kernel, mean function, (kernel, features), or None
+    :param int num_gauss_hermite_points: passed to `_quadrature_expectation` to set
+                                         the number of Gauss-Hermite points used
+    :return: a 1-D, 2-D, or 3-D tensor containing the expectation
     """
     if isinstance(p, tuple):
         assert len(p) == 2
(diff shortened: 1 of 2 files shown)


gpflow/expectations.py
```diff
@@ -171,14 +171,14 @@ def _quadrature_expectation(p, obj1, feature1, obj2, feature2, num_gauss_hermite
 
 # =========================== ANALYTIC EXPECTATIONS ===========================
 
-def expectation(p, obj1, obj2=None, num_gauss_hermite_points=None):
+def expectation(p, obj1, obj2=None, nghp=None):
```
awav: :param ...: description
:return: description

follow _reStructuredText_ (reST, sphinx) docstrings 
https://pythonhosted.org/an_example_pypi_project/sphinx.html

gpflow/expectations.py
```diff
@@ -571,37 +588,33 @@ def _expectation(p, linear_mean, none, kern, feat):
...
+    so _expectation(p, linear_mean, none, kern, feat) would call itself.
+    More specific signatures (e.g. (p, identity_mean, None, RBF, feat)) will
+    be found and used whenever available
     """
-    with params_as_tensors_for(constant_mean):
-        c = constant_mean(p.mu)  # NxQ
-        eKxz = expectation(p, (kern, feat))  # NxM
-
-        return c[..., None] * eKxz[:, None, :]
+    raise NotImplementedError
```
awav: So, other cases can't cause infinite loop? Did you check it? Can you prove it?

vdutor: I think it is definitely worth adding @stefanosele's case to the unit tests and show that this is remedied now.

ialong: Yes I have checked there are no other self-referential calls to `expectation`. Not sure a test is absolutely necessary

ialong: Ok I added a simple one by including a Matern kernel in exKxz expectation tests

gpflow/expectations.py
ialong: I have changed the docstrings, do they look ok now?

gpflow/expectations.py
stefanosele: I'm happy with that

stefanosele:
Good! Finally someone took the decision for  passing around kwargs vs. nghp.

Just close https://github.com/GPflow/GPflow/pull/693 and we're good to go

PR #704, opened 2018-04-05
changed Minibatch class so that it can change the shape of data freely

dlqudwns:
This fix was needed for me to change the amount of data that Minibatch holds.


awav:
Please, provide tests for your particular case.

dlqudwns:
Added test test_change_variable_size() in test_dataholders.py.

markvdw:
@awav can you confirm that this doesn't break anything you had in mind for autoflow and merge or close accordingly?

PR #706, opened 2018-04-05
a first mixedmok test

thevincentadam:
A first test for mixed mok. more to follow

### DIFF
--- a/tests/test_multioutput.py
+++ b/tests/test_multioutput.py
@@ -274,37 +274,6 @@ def test_seperate_independent_mof():
 # @pytest.mark.parametrize('shared_kern', [True, False])
 def test_mixed_mok_with_Id_vs_independent_mok():
     with session_context() as sess:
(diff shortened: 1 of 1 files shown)


tests/test_multioutput.py
```diff
@@ -85,6 +85,67 @@ class Data:
...
 
+
+def make_sqrt_data(rng, N, M):
+    return np.array([np.tril(rng.randn(M, M)) for _ in range(N)])  # N x M x M
+
+
+def expand_cov(G, W):
+    ''' G is L x M x M
+        W is L x L
+        Output is LM x LM
```
st--: what is this trying to compute? some kind of outer product?

vincentadam87: yes. the idea is to expand in the 'wasteful' LM * LM shape, covariances saved as L * M * M
(because this is what the dummy mok spits out)

PR #712, opened 2018-04-10
Notebook to demonstrate advanced usage such as combining GP with Neural Network

st--:


stefanosele:
the `KernelSpaceInducingPoints` all it does is to implement the deprecated `Kuf` method of the `Kernel` class. Preferably you should be registering a new dispatch version of the `Kuf` function.

The rest seem pretty good

st--:
@stefanosele Yeah, but that's not merged into GPflow-master yet so this PR can't use it.

jameshensman:
This needs testing, it's not picked up by the test_notebooks test.

st--:
The build fails because it runs out of time on travis and gets killed, not because the notebook is broken... @awav what to do about that ?

hughsalimbeni:
following #798, I think this PR should be fixed and merged. @st-- do you know what's breaking it?

st--:
It fails due to a really stupid thing where, in python 3.5, "{:.4}".format(np.array(1.23456)) works just fine, but in python 3.6 you need to convert the numpy scalar array (shape `()`) to a python float, which is really ugly ...

PR #721, opened 2018-04-16
Unify calling signature of (log)densities

st--:
densities.py contains tf calls to calculate log p(x | parameters). This PR renames the file to *log*densities, as that corresponds to the actual content, and also unifies the call signatures to have x first (as opposed to half of them x first and half of them x last). Just a GPflow-internal tidy-up (MWE not applicable).


st--:
Todo:
- Uniform: make log_heights a property
- Laplace: rename sigma to b

PR #723, opened 2018-04-17
Pytest used by test_util module

awav:
Hello everyone,

Quick fix for `pytest` import issue. Apparently, GPflow imports it in test_util.py module. Therefore it must go into requirements list.


vdutor:
Looks good

PR #724, opened 2018-04-17
Multi-output conditionals

vdutor:
Hi,

This PR implements a new framework for efficiently calculating (correlated) multi-output conditionals. Previously GPflow treated multi-dimensional outputs completely independent. This PR makes it now possible to deal with correlation between the outputs. We achieve this by using a "multiple-dispatch" `conditional` that uses different code paths depending on which type of correlation or structure the model encodes. A user can specify this type of correlation by using specific multi-output kernels and features. The multiple-dispatch `conditional` also allows a user to easily define his/her own `conditional` function. 

It is important to note that this PR is completely backward compatible. Even more, if a user is not interested in the new multi-output features he/she can safely ignore the entire `multioutput` module.
 
### Most important additions:

1) Introduction of MultiOutputFeatures (`Mof`) and MultiOutputKernels (`Mok`).
These are used to specify a particular setup of multi-output correlation.

2) Multiple-dispatch for `conditional`. This allows GPflow to select the most efficient conditional code depending on your choice of `Mof` and `Mok`.

3) Multiple-dispatch for `Kuu` and `Kuf`. Previously `Kuu(.)` and `Kuf(.)` were member functions of the `feature` class. This became cumbersome as the calculation of `Kuu` and `Kuf` also depends on the kernel used. In line with `conditional` we now also use multiple-dispatch to calculate `Kuu` and `Kuf` for a particular combination of `Mok` and `Mof`.  

4) The actual maths to efficiently calculate the output-correlated conditional (credits to @markvdw  ) 

5) `sample_conditional` Function that makes sure that the most efficient code is used to get a sample from the conditional distribution.

6) Minor: we updated a couple of models to use the new multi-output conditional.

## Minimal Example

- We created an extended notebook to illustrate the use of the new multi-output framework. Please have a look at `notebooks/multioutput.ipynb`, it contains a couple of informative UML-diagrams and tables.
- The unit tests `tests/test_multioutput.py` should also explain the concept.

This snippet shows how you can define a SVGP model that uses a different kernel for each output.
```python
import gpflow.multioutput.kernels as mk
import gpflow.multioutput.features as mf

M, D = 20, 1
X, Y = data()  # N x D, N x P

kern_list = [RBF(D) + Linear(D) for _ in range(P)]
kernel = mk.SeparateIndependentMok(kern_list)
feature = mf.SharedIndependentMof(gpf.features.InducingPoints(X[:M,...].copy()))
m = SVGP(X, Y, kernel, gpf.likelihoods.Gaussian(), feat=feature)
```
Note that nothing changes, except that we use a different kernel and feature.

This work is done in collaboration with @markvdw  @st-- -- @vincentadam87  @stefanosele


st--:
Thought: use names identifying the type of conditional for the name_scope() - e.g. "conditional_independentmok", "conditional_mixedmok", "conditional_feature" instead of just "conditional" -
 this might make debugging easier, to see which one actually got called?

tests/test_coregion.py
```diff
@@ -154,7 +154,7 @@ def test_predicts(self):
             self.assertTrue(np.allclose(pred_ydensity0, pred_ydensity_c0, atol=1e-2))
             pred_ydensity1 = self.vgp1.predict_density(self.Xtest, Ytest)
             pred_ydensity_c1 = self.cvgp.predict_density(X_augumented1, Y_augumented1)
-            assert_allclose(pred_ydensity1, pred_ydensity_c1, atol=1e-2)
+            # np.testing.assert_allclose(pred_ydensity1, pred_ydensity_c1, atol=1e-2)
```
st--: why did this get commented out ?

tests/test_method_equivalence.py
```diff
@@ -102,8 +102,11 @@ def test_all(self):
             assert_allclose(variances, variances[0], 1e-5)
             assert_allclose(lengthscales, lengthscales.mean(), 1e-4)
             mu0, var0 = models[0].predict_y(self.Xtest)
-            for m in models[1:]:
+            for i, m in enumerate(models[1:]):
                 mu, var = m.predict_y(self.Xtest)
+                print(i, type(m))
+                print(mu.shape)
+                print(var.shape)
```
st--: Tests should assert shapes, not print them?:)

gpflow/multioutput/features.py
```diff
@@ -0,0 +1,150 @@
...
+        self.feat = feat
+
+    def __len__(self):
+        return len(self.feat)
+
+
+class SeparateIndependentMof(Mof):
+    """
+    A different feature is used for each output.
+    Note: each feature should have the same number of points, M.
```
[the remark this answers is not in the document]
vdutor: we normally don't check user inputs, right?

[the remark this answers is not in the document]
st--: I would phrase it as "we don't check user inputs as much as we should"... but I think it's good to add checks, e.g. #694

gpflow/multioutput/conditionals.py
```diff
@@ -0,0 +1,451 @@
...
+
+    print("conditional: object, SharedIndependentMof, SeparateIndependentMok, object")
+    # Following are: P x M x M  -  P x M x N  -  P x N(x N)
+    Kmms = Kuu(feat, kern, jitter=settings.numerics.jitter_level)  # P x M x M
+    Kmns = Kuf(feat, kern, Xnew)  # P x M x N
+    kern_list = kern.kern_list if isinstance(kern, Combination) else [kern.kern] * len(feat.feat_list)
+    Knns = tf.stack([k.K(Xnew) if full_cov else k.Kdiag(Xnew) for k in kern_list], axis=0)
+    fs = tf.transpose(f)[:, :, None]  # P x M x 1
+    # P x 1 x M x M  or  P x M x 1
+    q_sqrts = tf.transpose(q_sqrt)[:, :, None] if q_sqrt.shape.ndims == 2 else q_sqrt[:, None, :, :]
```
dlqudwns: This part does not work with q_sqrt=None

gpflow/multioutput/features.py
```diff
@@ -0,0 +1,150 @@
...
+    M = tf.shape(Kmm)[0] * tf.shape(Kmm)[1]
+    jittermat = jitter * tf.reshape(tf.eye(M, dtype=float_type), tf.shape(Kmm))
+    return Kmm + jittermat
+
+
+@dispatch(SharedIndependentMof, SharedIndependentMok)
+def Kuu(feat, kern, *, jitter=0.0):
+    Kmm = Kuu(feat.feat, kern.kern)  # M x M
+    jittermat = tf.eye(len(feat), dtype=float_type) * jitter
+    return Kmm + jittermat
```
dlqudwns: This seems to adding jitter twice since Kuu already add a jitter? I am not sure about all cases but at least for some cases it adds jitter twice.

vdutor: Maybe I'm missing something, but I don't see where/why the jitter is added twice? The first time `Kuu` is called without `jitter` arguments, which means that the default (i.e. `0`) is used. We then add `jittermat` once. The same goes for the other cases.

tests/test_coregion.py
vdutor: Good catch, I'll uncomment it.

gpflow/multioutput/conditionals.py
vdutor: Correct, I'll fix that. Thanks

gpflow/multioutput/features.py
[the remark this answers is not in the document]
markvdw: My opinion is that we should check in cases where an error is common, and we get a cryptic error message if the error is made.

For example, usually, I think TensorFlow's error message points you to a reasonable place to figure out what you did wrong. However, sometimes, it sends you down 10 levels before some arcane shape error is thrown based on a tensor a few layers away from what you actually defined. In this case, a check would be useful.

gpflow/multioutput/features.py
dlqudwns: You're right. I think I was confused: sorry!

gpflow/dispatch.py
```diff
@@ -0,0 +1,10 @@
+from multipledispatch import dispatch, Dispatcher
+from functools import partial
+
+# By default multipledispatch uses a global namespace in multipledispatch.core.global_namespace
+# We define our own GPflow namespace to avoid any conflict which may arise
+gpflow_md_namespace = dict()
+dispatch = partial(dispatch, namespace=gpflow_md_namespace)
+
+conditional = Dispatcher('conditional')
```
st--: Should these go into conditionals.py as well ?

st--:
@all: *Should* Kuu() add a jitter?

gpflow/multioutput/features.py
```diff
@@ -0,0 +1,150 @@
...
+
+
+@dispatch(SeparateIndependentMof, SharedIndependentMok, object)
+def Kuf(feat, kern, Xnew):
+    return tf.stack([Kuf(f, kern.kern, Xnew) for f in feat.feat_list], axis=0)  # L x M x N
+
+
+@dispatch(SharedIndependentMof, SeparateIndependentMok, object)
+def Kuf(feat, kern, Xnew):
+    return tf.stack([Kuf(feat.feat, k, Xnew) for k in kern.kern_list], axis=0)  # L x M x N
```
dlqudwns: It seems that in master branch Combination kernel is no longer using kern_list and this makes error.

gpflow/conditionals.py
```diff
@@ -17,13 +17,95 @@
...
 from . import settings, mean_functions
 from .decors import name_scope
-from .features import InducingPoints
+from .dispatch import conditional, sample_conditional
 from .expectations import expectation
+from .features import Kuu, Kuf, InducingPoints, InducingFeature
+from .kernels import Kernel, Combination
 from .probability_distributions import Gaussian
 
+float_type = settings.float_type
```
st--: This means the functions in here will only use the value from gpflowrc at startup, and not support run-time changes to float_type or jitter_level - would be better to use settings.float_type etc.

Joshuaalbert:
Having just pulled the new version of gpflow, I'm now quite happy to see this multioutput work. I'm just curious, the fully correlated outputs only refers to LMC right? Or does it also allow convolutional models (which I'd be quite interested in using, and therefore coding up if it's not possible yet). Also, nice work @markvdw

Issue #685, opened 2018-03-22
Remove Matern's numerical issues


nrontsis:
The current Matern implementation has numerical issues with its gradient, as reported in #490. A numerical fix was suggested, which can occasionally be inefficient due to performance or stability reasons.

I believe that the issue is caused by tensorflow's autodiff trying to apply the chain rule to get the derivatives of the kernel at points where some distances between points are zero. At those points the derivative of distance is not defined (its limit goes to infinity). However, the kernel function as a whole is differentiable with finite derivatives.

The issue can be completely removed by defining the matern function as a new tensorflow operation with predefined gradients. It will take as an input the squared distance and provide the kernel value and gradient at that point. The expressions for both are [very simple](https://github.com/SheffieldML/GPy/blob/a6c154753d65c4d0a3d7af0e32803b72038cb68b/GPy/kern/src/stationary.py#L439-L440).

Would you be interested in such a PR?

jhoareau:
Also experiencing quite a bit of problems with Matern, that would be nice. +1

st--:
Sounds like a good idea. We'll have to get GPflow working with tf 1.7 first though...!

nrontsis:
Great, let me know when this is done.

nrontsis:
Also, @st-- couldn't #421 be removed by using a `scaled_euclid_dist` that is guaranteed to return positive results? See e.g. [here](https://stackoverflow.com/a/40053188/6738265).

(I believe that the performance/memory issues mentioned in the stack overflow answer should not be a concern since the dominant operation is the Cholesky decomposition)

nrontsis:
One more question: why tensorflow 1.7 is needed for the first suggestion?

Issue #490, opened 2017-08-21
Issues in GP regression with the Matern Kernel. Is RBF more numerically stable than the Matern Kernel?


DiegoRomeres:
I am doing GP regression with some real data dim(X)=[N,3] and dim[Y] = [N,1]. When I use the RBF Kernel it works fine when I use the Matern kernel it gives me some errors, probably in computing the cholesky during the optimazation. Here an idea of my code:

```
import GPflow

m = GPflow.gpr.GPR(X, Y, kern=GPflow.kernels.RBF(3))
m.optimize()


m2 = GPflow.gpr.GPR(X, Y, kern=GPflow.kernels.Matern52(3)+ GPflow.kernels.White(input_dim=3))
m2.optimize()
```
This is the error that I obtain

```
2017-08-21 17:49:05.142629: W tensorflow/core/platform/cpu_feature_guard.cc:45] The TensorFlow library wasn't compiled to use SSE4.1 instructions, but these are available on your machine and could speed up CPU computations.
2017-08-21 17:49:05.142639: W tensorflow/core/platform/cpu_feature_guard.cc:45] The TensorFlow library wasn't compiled to use SSE4.2 instructions, but these are available on your machine and could speed up CPU computations.
2017-08-21 17:49:05.142641: W tensorflow/core/platform/cpu_feature_guard.cc:45] The TensorFlow library wasn't compiled to use AVX instructions, but these are available on your machine and could speed up CPU computations.
2017-08-21 17:49:05.142643: W tensorflow/core/platform/cpu_feature_guard.cc:45] The TensorFlow library wasn't compiled to use AVX2 instructions, but these are available on your machine and could speed up CPU computations.
2017-08-21 17:49:05.142645: W tensorflow/core/platform/cpu_feature_guard.cc:45] The TensorFlow library wasn't compiled to use FMA instructions, but these are available on your machine and could speed up CPU computations.
2017-08-21 17:49:05.246520: I tensorflow/stream_executor/cuda/cuda_gpu_executor.cc:893] successful NUMA node read from SysFS had negative value (-1), but there must be at least one NUMA node, so returning NUMA node zero
2017-08-21 17:49:05.246789: I tensorflow/core/common_runtime/gpu/gpu_device.cc:940] Found device 0 with properties: 
name: GeForce GTX 1080
major: 6 minor: 1 memoryClockRate (GHz) 1.86
pciBusID 0000:01:00.0
Total memory: 7.92GiB
[...]

Issue #421, opened 2017-05-30
NaNs in predictions with Matern52 and Matern32 kernels


rm4216:
Dear all

I have been using the GPR regression model to make posterior predictions with Gaussian likelihood using Matern52, and Matern32 kernels. However, with a particular set of 3-dimensional pints and a particular choice of lengthscales I obtained all NaNs on posterior mean and variance, both for isotropic and ARD version of each kernel. Here is the code that should reproduce the issue. (I also modified the Matern52 kernel to be able to obtain numerical values of the covariance matrix "npK52_iso" and "npK52_ARD", I did not apply any modification to Matern32)
# Dataset
```pythonscript
import numpy as np
import tensorflow as tf
import GPflow

full_dim = 3
# Five data points in 3 dimensions
D_X = np.array([
 [31.55437035,  31.55437038,  32.36345678],
 [31.55437037,  31.55437037,  32.36345679],
 [31.55437037,  31.55437037,  32.36345679],
 [28.31802469,  21.84533348, -29.12711068],
 [28.31802469,  21.84533348, -29.12711111],
])
# Trying with different dataset does not show any problem
# D_X = np.random.uniform(-32, 32, 5*full_dim).reshape(-1, full_dim)

Y = np.random.uniform(2, 7, D_X.shape[0]).reshape(-1, 1)    # any random choice
X_test = np.random.uniform(-32, 32, 10*full_dim).reshape(-1, full_dim)  # random test set
```
# Matern52 tests
```pythonscript
# Matern52 test isotropic
kernel52_iso = GPflow.kernels.Matern52(input_dim=3, ARD=False)
kernel52_iso.lengthscales._array = np.array([0.01095283])     # critical lengthscale

GPmodel = GPflow.gpr.GPR(D_X, Y, kernel52_iso)
mean52_iso, vars52_iso = GPmodel.predict_f(X_test)  # posterior mean is all nan

K52_iso = kernel52_iso.K(D_X)
gg = K52_iso.graph
sess = tf.InteractiveSession(graph=gg)
npK52_iso = K52_iso.eval(session=sess)  # Covariance matrix contains nan

# Matern52 test ARD
kernel52_ARD = GPflow.kernels.Matern52(input_dim=3, ARD=True)
kernel52_ARD.lengthscales._array = np.array([0.01095283, 0.01095283, 0.01095283])     # critical lengthscales

GPmodel = GPflow.gpr.GPR(D_X, Y, kernel52_ARD)
[...]

PR #727, opened 2018-04-21
Numerically stable Matern32

nrontsis:
tldr: A numerically stable implementation of Matern32. This issue aims to address #685. A minimal example is not provided as the issues in the current implementation are rare, affecting e.g. ill-conditioned datasets, and iterative frameworks such as GPflowOpt.

The suggested implementation avoids the two issues existing in the current implementation:
1.  NaNs in the kernel outputs (see #421)
2.  Inf in the kernel's gradient (see #490)

Regarding (1): this is because the current implementation of `scaled_square_dist` can give negative results. ~~A new implementation of `scaled_square_dist` is introduced, taken from [here](https://stackoverflow.com/questions/37009647/compute-pairwise-distance-in-a-batch-without-replicating-tensor-in-tensorflow/40053188#40053188).~~ We use `tf.abs` (only in the forward pass) to get around this.

Regarding (2) this is due to the chain rule applied to `tf.sqrt`, which at zero has an infinite derivative. I avoid this by feeding analytically the gradient of `(1 + tf.sqrt(d2))*tf.exp(-tf.sqrt(d2)`, which has always bounded derivatives.

Tests are included that compare against the checking implementation.


nrontsis:
Just to mention that the same idea could be applied to the Matern 1/2 or 5/2, Exponential and (with a small adaptation) Cosine Kernels.

gpflow/kernels.py
```diff
@@ -250,22 +250,19 @@ def euclid_dist(self, X, X2):  # pragma: no cover
...
-        dist = -2 * tf.matmul(X, X2, transpose_b=True)
-        dist += tf.reshape(Xs, (-1, 1)) + tf.reshape(X2s, (1, -1))
+            X2 = X
+        else:
+            X2 = X2 / self.lengthscales
+
+        dist = tf.reduce_sum(tf.squared_difference(
+            tf.expand_dims(X, 1),
+            tf.expand_dims(X2, 0)), 2)
+
```
hughsalimbeni: In my experience this is very much slower than the code it replaces

nrontsis: Is this a concern? I thought that other operations e.g. Cholesky decomposition is the bottleneck.

hughsalimbeni: It can be, especially when the input dimension is large. I think the gradient override is compatible with the existing squared distance computation, though.

nrontsis: Okay. I this case we can get around the negative distances by adding a tf.abs which we will ignore in the gradients via a gradient override.

nrontsis:
Now that I think it more,  I believe that a simpler hack would be to replace the euclidean distance calculation to:
```
r2 = self.scaled_square_dist(X, X2)
return tf.sqrt(tf.maximum(r2, 1e-12))
```
It shouldn't make such a difference, since `scaled_square_dist` is not so precise (i.e. it can negative values).

hughsalimbeni:
Seems like a good way to go. Do you have a minimal example which exhibits the numerical instability?

nrontsis:
No, I've experienced these issues while doing Bayesian Optimization and I don't have something simple at hand.

gpflow/kernels.py
```diff
@@ -424,13 +424,28 @@ class Matern32(Stationary):
...
     @params_as_tensors
     def K(self, X, X2=None, presliced=False):
         if not presliced:
             X, X2 = self._slice(X, X2)
-        r = self.scaled_euclid_dist(X, X2)
-        return self.variance * (1. + np.sqrt(3.) * r) * \
-               tf.exp(-np.sqrt(3.) * r)
+        d = self.square_dist(X, X2)
+
+        with self.graph.gradient_override_map(
```
awav: A bit more about *gradient_override_map*
From docs: https://www.tensorflow.org/api_docs/python/tf/Graph

gradient_override_map(op_type_map)
**EXPERIMENTAL**: A context manager for overriding gradient functions.

awav: Does it mean there is no guarantee that this will work or supported in future by TensorFlow?

nrontsis: Yeah, I guess this might not be future proof.

awav:
Fantastic! Now you can convert it to the test. Make it this way (w.r.t [contributing](https://github.com/GPflow/GPflow/blob/master/contributing.md#code-quality-requirements)):

```
import pytest
from gpflow.test_utils import session_tf
from gpflow.kernels import Matern12, Matern32, Matern52, Exponential

class Datum:
    lengthscales = ...
    # cetera...

@pytest.mark.parametrize('kernel', [Matern12, Matern32, Matern52, Exponential]):
def test_error(session_tf, kernel):
    k = kernel(..., lengthscales=Datum.lengthscale, variance=Datum.variance)
    ... 
```
Check this [test](https://github.com/GPflow/GPflow/blob/master/tests/test_expectations.py) as an example.

gpflow/kernels.py
```diff
@@ -424,11 +424,17 @@ class Matern32(Stationary):
     The Matern 3/2 kernel
     """
 
+    # Used to bypass gradients of tf.sqrt that can lead to infs.
+    tf.RegisterGradient("Matern32Grad")(
```
awav: You can remove this.

tests/test_kerns.py
```diff
@@ -43,6 +43,60 @@ def test_1d(self):
             self.assertTrue(np.allclose(gram_matrix, reference_gram_matrix))
 
 
+class Matern32Old(gpflow.kernels.Stationary):
```
awav: No need in old implementation, once you cover your case with new test.

tests/test_kerns.py
```diff
@@ -43,6 +43,60 @@ def test_1d(self):
...
+    @gpflow.decors.params_as_tensors
+    def K(self, X, X2=None, presliced=False):
+        if not presliced:
+            X, X2 = self._slice(X, X2)
+        r = self.scaled_euclid_dist(X, X2)
+        return self.variance * (1. + np.sqrt(3.) * r) * \
+               tf.exp(-np.sqrt(3.) * r)
+
+
+class TestMatern32(GPflowTestCase):
```
awav: Same here, you can remove it. And it is not advised to use GPflowTestCase. Use pytest framework.

nrontsis:
Thanks, all done.

tests/test_scaled_euclid_dist.py
```diff
@@ -1,34 +1,46 @@
...
-    kernels = {'Matern12': Matern12, 'Matern32': Matern32,
-               'Matern52': Matern52, 'Exponential': Exponential,
-               'Cosine': Cosine}
-
-    k = kernels[kernel](Datum.D)
-
+    
+    k = kernel(Datum.D)
     K = k.compute_K_symm(Datum.X)
     assert not np.isnan(K).any(), 'There are NaNs in the output of the ' + kernel + ' kernel.'
```
nrontsis: kernel -> str(kernel) to avoid a TypeError.

tests/test_scaled_euclid_dist.py
```diff
@@ -1,34 +1,46 @@
...
 
-    X = tf.placeholder(float_type)
+    X = tf.placeholder(settings.float_type)
     dK = session_tf.run(tf.gradients(k.K(X, X), X)[0], feed_dict={X: Datum.X})
     assert not np.isnan(dK).any(), 'There are NaNs in the gradient of the ' + kernel + ' kernel.'
```
nrontsis: Same here

nrontsis:
@awav thanks alot for cleaning up the code. Is there anything else you would like to do? (apart from pushing the minor change in the assert message)

awav:
Once tests pass, I will merge it.

nrontsis:
Thanks! Last thing: added asserts for Infs as well, and added a forgotten __name__ in the first assert message.

nrontsis:
However, a side effect of clipping at non-zero is getting zero gradient at zero distances for non-differentiable kernels like the Exponential (check its 1d gradient [here](http://www.wolframalpha.com/input/?i=derivative+exp(abs(x)))).

A more concise treatment would be to clip at zero for Matern12, Exponential and Cosine kernels but not for the Matern32 and Matern52 which are differentiable. I don't know if we ever prefer NaNs over zeros though.

PR #738, opened 2018-05-01
GPflow ScipyOptimizer passes options to actual ScipyOptimizer incorrectly

awav:
Hello all,

There is a bug in current ScipyOptimizer. It doesn't pass options dictionary to actual scipy optimizer. You can check new test for a failure case.

w/o this fix, master will ignore `gtol` option:
```
o1 = gpflow.train.ScipyOptimizer(options={'gtol': gtol_value})
```


stefanosele:
That's great. I can confirm that with this fix ScipyOptimizer is working as intended

markvdw:
If @stefanosele says so... :)

Issue #734, opened 2018-04-29
Problem with KL_divergence.


svaibhavsingh:
I faced an error here in the below given line:
constant = - tf.size(q_mu, out_type=settings.float_type)
making out_type to be default (i.e int32 or int64) and then type casting the constant worked for me.

Full traceback of the error is given below:

File "/home/vaibhav/natgrad/lib/python3.5/site-packages/gpflow/decors.py", line 49, in name_scope_wrapper
    return method(*args, **kwargs)
  File "/home/vaibhav/natgrad/lib/python3.5/site-packages/gpflow/kullback_leiblers.py", line 73, in gauss_kl
    constant = - tf.size(q_mu, out_type=settings.float_type)
  File "/home/vaibhav/natgrad/lib/python3.5/site-packages/tensorflow/python/ops/array_ops.py", line 371, in size
    return size_internal(input, name, optimize=True, out_type=out_type)
  File "/home/vaibhav/natgrad/lib/python3.5/site-packages/tensorflow/python/ops/array_ops.py", line 404, in size_internal
    return gen_array_ops.size(input, name=name, out_type=out_type)
  File "/home/vaibhav/natgrad/lib/python3.5/site-packages/tensorflow/python/ops/gen_array_ops.py", line 6688, in size
    "Size", input=input, out_type=out_type, name=name)
  File "/home/vaibhav/natgrad/lib/python3.5/site-packages/tensorflow/python/framework/op_def_library.py", line 720, in _apply_op_helper
    attr_value.type = _MakeType(value, attr_def)
  File "/home/vaibhav/natgrad/lib/python3.5/site-packages/tensorflow/python/framework/op_def_library.py", line 179, in _MakeType
    _SatisfiesTypeConstraint(i, attr_def, param_name=attr_def.name)
  File "/home/vaibhav/natgrad/lib/python3.5/site-packages/tensorflow/python/framework/op_def_library.py", line 60, in _SatisfiesTypeConstraint
    ", ".join(dtypes.as_dtype(x).name for x in allowed_list)))
TypeError: Value passed to parameter 'out_type' has DataType float64 not in list of allowed values: int32, int64

hughsalimbeni:
I've had this issue in the past too, and got around it by using `.set_shape` so the size is known at graph build time. It would be good to fix this as it is quite annoying

awav:
@svaibhavsingh, resolved.

PS: Markdown works in comments. Could you please use code blocks next time, like:

[...]

Issue #725, opened 2018-04-18
`gauss_kl` error when mean has non-static shape


rhaps0dy:
When running `gpflow.kullback_leiblers.gauss_kl` on something with shape undefined at graph-building time, such as a `tf.placeholder`, the following happens:
```
TypeError: Value passed to parameter 'out_type' has DataType float64 not in list of allowed values: int32, int64
```
And [this GPflow line](https://github.com/GPflow/GPflow/blob/master/gpflow/kullback_leiblers.py#L73) is the one that causes the error. It seems that tensorflow has different code paths for fully defined and partially defined shapes, see [this Tensorflow snippet](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/ops/array_ops.py#L399-L404).

Minimum failing example:
```python
import gpflow
import tensorflow as tf
import numpy as np

mu = tf.ones([1, 4], dtype=gpflow.settings.float_type)
sig = tf.ones([4, 1, 1], dtype=gpflow.settings.float_type)
mu_ph = tf.placeholder(gpflow.settings.float_type, [1, None])

with tf.Session() as sess:
    print(sess.run(gpflow.kullback_leiblers.gauss_kl(mu, sig)))  # works
    print(sess.run(gpflow.kullback_leiblers.gauss_kl(mu_ph, sig),
                   feed_dict={mu_ph: np.ones([1, 4])}))  # doesn't work
```

Suggested fix:

```python
tf.cast(-tf.size(q_mu, out_type=settings.float_type), dtype=gpflow.settings.float_type)
```

Should I make a pull request?

markvdw:
@hughsalimbeni, you looked at #734, do you think this would solve it?

hughsalimbeni:
Looks like a good fix to me. @rhaps0dy are you happy to do this?

PR #742, opened 2018-05-02
Fixed `gauss_kl` when the shape of the Gaussian parameters is unknown

rhaps0dy:
Also added a test to prevent future regressions.

This addresses the bug in both #734 and #725  .

### DIFF
--- a/tests/test_kldiv.py
+++ b/tests/test_kldiv.py
@@ -145,5 +145,15 @@ def test_oned(session_tf, white, mu, sqrt, K_batch):
     np.testing.assert_allclose(kl.eval(), kl_tf.eval())
 
 
+def test_unknown_size_inputs(session_tf):
+    mu_ph = tf.placeholder(settings.float_type, [None, None])
+    sqrt_ph = tf.placeholder(settings.float_type, [None, None, None])
+    mu = np.ones([1, 4], dtype=settings.float_type)
+    sqrt = np.ones([4, 1, 1], dtype=settings.float_type)
+    np.testing.assert_allclose(
+        session_tf.run(gauss_kl(*map(tf.constant, [mu, sqrt]))),
+        session_tf.run(gauss_kl(mu_ph, sqrt_ph), {mu_ph: mu, sqrt_ph: sqrt}))
+
+
 if __name__ == "__main__":
     tf.test.main()
--- a/tests/test_kldiv.py
+++ b/tests/test_kldiv.py
@@ -146,13 +146,19 @@ def test_oned(session_tf, white, mu, sqrt, K_batch):
 
 
 def test_unknown_size_inputs(session_tf):
+    """
+    Test for #725 and #734. When the shape of the Gaussian's mean had at least
+    one unknown parameter, `gauss_kl` would blow up. This happened because
+    `tf.size` can only output types `tf.int32` or `tf.int64`.
+    """
     mu_ph = tf.placeholder(settings.float_type, [None, None])
     sqrt_ph = tf.placeholder(settings.float_type, [None, None, None])
     mu = np.ones([1, 4], dtype=settings.float_type)
     sqrt = np.ones([4, 1, 1], dtype=settings.float_type)
+    feed_dict = {mu_ph: mu, sqrt_ph: sqrt}
     np.testing.assert_allclose(
         session_tf.run(gauss_kl(*map(tf.constant, [mu, sqrt]))),
-        session_tf.run(gauss_kl(mu_ph, sqrt_ph), {mu_ph: mu, sqrt_ph: sqrt}))
+        session_tf.run(gauss_kl(mu_ph, sqrt_ph), feed_dict=feed_dict))
 
 
 if __name__ == "__main__":
--- a/tests/test_kldiv.py
+++ b/tests/test_kldiv.py
@@ -155,10 +155,15 @@ def test_unknown_size_inputs(session_tf):
     sqrt_ph = tf.placeholder(settings.float_type, [None, None, None])
     mu = np.ones([1, 4], dtype=settings.float_type)
     sqrt = np.ones([4, 1, 1], dtype=settings.float_type)
+    
     feed_dict = {mu_ph: mu, sqrt_ph: sqrt}
-    np.testing.assert_allclose(
-        session_tf.run(gauss_kl(*map(tf.constant, [mu, sqrt]))),
-        session_tf.run(gauss_kl(mu_ph, sqrt_ph), feed_dict=feed_dict))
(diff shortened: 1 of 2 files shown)


tests/test_kldiv.py
```diff
@@ -145,5 +145,15 @@ def test_oned(session_tf, white, mu, sqrt, K_batch):
     np.testing.assert_allclose(kl.eval(), kl_tf.eval())
 
 
+def test_unknown_size_inputs(session_tf):
+    mu_ph = tf.placeholder(settings.float_type, [None, None])
+    sqrt_ph = tf.placeholder(settings.float_type, [None, None, None])
+    mu = np.ones([1, 4], dtype=settings.float_type)
```
awav: np.ones => tf.ones

tests/test_kldiv.py
```diff
@@ -145,5 +145,15 @@ def test_oned(session_tf, white, mu, sqrt, K_batch):
...
+    sqrt = np.ones([4, 1, 1], dtype=settings.float_type)
+    np.testing.assert_allclose(
+        session_tf.run(gauss_kl(*map(tf.constant, [mu, sqrt]))),
```
awav: no need in `map` anymore after using tf.ones

tests/test_kldiv.py
```diff
@@ -145,5 +145,15 @@ def test_oned(session_tf, white, mu, sqrt, K_batch):
     np.testing.assert_allclose(kl.eval(), kl_tf.eval())
 
 
+def test_unknown_size_inputs(session_tf):
```
awav: it would be nice to have docstring describing the issue and reference to github issue

tests/test_kldiv.py
```diff
@@ -145,5 +145,15 @@ def test_oned(session_tf, white, mu, sqrt, K_batch):
...
+    mu_ph = tf.placeholder(settings.float_type, [None, None])
+    sqrt_ph = tf.placeholder(settings.float_type, [None, None, None])
+    mu = np.ones([1, 4], dtype=settings.float_type)
+    sqrt = np.ones([4, 1, 1], dtype=settings.float_type)
+    np.testing.assert_allclose(
+        session_tf.run(gauss_kl(*map(tf.constant, [mu, sqrt]))),
+        session_tf.run(gauss_kl(mu_ph, sqrt_ph), {mu_ph: mu, sqrt_ph: sqrt}))
```
awav: > Signature: sess.run(fetches, feed_dict=None, options=None, run_metadata=None)

Therefore your call should be
```
feed_dict = {mu_ph: mu, sqrt_ph: sqrt}
session_tf.run(gauss_kl(mu_ph, sqrt_ph), feed_dict=feed_dict))
```

awav:
Hello, @rhaps0dy! I would appreciate if you could summarize why this issue happened and how you solved it.

PS: from pull request template,
> Description
> Describe your contribution by writing down the context and reasons. Make it clear whether it is a bug > fix, feature or enhancement. Do not make it very wordy, keep it short and simple.
>
> Bad example: This is a bug-fix found at #111.
> Bad example: Add new SuperMagic optimizer.
> Good example: Working on feature #111, we found that parameters of VGP are not instantiated > properly. It causes arbitrary failures at training. The bug is fixed by adding necessary parameters to the list of trainables.

tests/test_kldiv.py
rhaps0dy: Does that work, even for passing them to the placeholders? (I tested and it seems that it doesn't). The point of having np.ones is to be able to have unknown-size tensors later.

rhaps0dy:
Hi @awav. Sorry, I thought it would be fine having done the bug description carefully in #725. I'll just copy the issue, it contains all the information required by the pull request template.

Point taken about the docstring though. I've also run pylint, and the only warnings I introduced are these:

```
W:148,29: Redefining name 'session_tf' from outer scope (line 25) (redefined-outer-name)
W:156, 4: Redefining name 'mu' from outer scope (line 52) (redefined-outer-name)
W:157, 4: Redefining name 'sqrt' from outer scope (line 68) (redefined-outer-name)
```

but the rest of the file is littered with them as well.

# `gauss_kl` error when mean has non-static shape

When running `gpflow.kullback_leiblers.gauss_kl` on something with shape undefined at graph-building time, such as a `tf.placeholder`, the following happens:
```
TypeError: Value passed to parameter 'out_type' has DataType float64 not in list of allowed values: int32, int64
```
And [this GPflow line](https://github.com/GPflow/GPflow/blob/master/gpflow/kullback_leiblers.py#L73) is the one that causes the error. It seems that tensorflow has different code paths for fully defined and partially defined shapes, see [this Tensorflow snippet](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/ops/array_ops.py#L399-L404).

Minimum failing example:
```python
import gpflow
import tensorflow as tf
import numpy as np

mu = tf.ones([1, 4], dtype=gpflow.settings.float_type)
sig = tf.ones([4, 1, 1], dtype=gpflow.settings.float_type)
mu_ph = tf.placeholder(gpflow.settings.float_type, [1, None])

with tf.Session() as sess:
    print(sess.run(gpflow.kullback_leiblers.gauss_kl(mu, sig)))  # works
    print(sess.run(gpflow.kullback_leiblers.gauss_kl(mu_ph, sig),
                   feed_dict={mu_ph: np.ones([1, 4])}))  # doesn't work
```

tests/test_kldiv.py
```diff
@@ -145,5 +145,21 @@ def test_oned(session_tf, white, mu, sqrt, K_batch):
...
+    one unknown parameter, `gauss_kl` would blow up. This happened because
+    `tf.size` can only output types `tf.int32` or `tf.int64`.
+    """
+    mu_ph = tf.placeholder(settings.float_type, [None, None])
+    sqrt_ph = tf.placeholder(settings.float_type, [None, None, None])
+    mu = np.ones([1, 4], dtype=settings.float_type)
+    sqrt = np.ones([4, 1, 1], dtype=settings.float_type)
+    feed_dict = {mu_ph: mu, sqrt_ph: sqrt}
+    np.testing.assert_allclose(
+        session_tf.run(gauss_kl(*map(tf.constant, [mu, sqrt]))),
```
awav: *One remark:* it is always preferable to be more explicit in tests and more wordy.

```
known_shape_tf = gauss_kl(*map(tf.constant, [mu, sqrt]))
unknown_shape_tf = gauss_kl(mu_ph, sqrt_ph)
known_shape = session_tf.run(known_shape_tf)
unknown_shape = session_tf.run(unknown_shape_tf, , feed_dict=feed_dict)
np.testing.assert_allclose(known_shape, unknown_shape)
```

rhaps0dy:
Thank you very much for finishing this!

PR #747, opened 2018-05-05
ndiagquad logspace feature for improved numerical stability

st--:
Better numerical stability for predict_density().
More tests.

### DIFF
--- a/gpflow/quadrature.py
+++ b/gpflow/quadrature.py
@@ -80,15 +80,18 @@ def mvnquad(func, means, covs, H: int, Din: int=None, Dout=None):
     return tf.reduce_sum(fX * wr, 0)
 
 
-def ndiagquad(funcs, H: int, Fmu, Fvar, **Ys):
+def ndiagquad(funcs, H: int, Fmu, Fvar, logspace: bool=False, **Ys):
     """
     Computes N Gaussian expectation integrals of one or more functions
     using Gauss-Hermite quadrature. The Gaussians must be independent.
 
-    :param funcs: Callable or Iterable of Callables that operates elementwise
+    :param funcs: the integrand(s):
+        Callable or Iterable of Callables that operates elementwise
     :param H: number of Gauss-Hermite quadrature points
     :param Fmu: array/tensor or `Din`-tuple/list thereof
     :param Fvar: array/tensor or `Din`-tuple/list thereof
+    :param logspace: if True, funcs are the log-integrands and this calculates
+        the log-expectation of exp(funcs)
     :param **Ys: arrays/tensors; deterministic arguments to be passed by name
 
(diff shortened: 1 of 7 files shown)


gpflow/quadrature.py
```diff
@@ -80,15 +80,18 @@ def mvnquad(func, means, covs, H: int, Din: int=None, Dout=None):
     return tf.reduce_sum(fX * wr, 0)
 
 
-def ndiagquad(funcs, H: int, Fmu, Fvar, **Ys):
+def ndiagquad(funcs, H: int, Fmu, Fvar, logspace: bool=False, **Ys):
```
awav: Could you fill all type annotations? https://docs.python.org/3/library/typing.html

gpflow/quadrature.py
```diff
@@ -80,15 +80,18 @@ def mvnquad(func, means, covs, H: int, Din: int=None, Dout=None):
...
     """
     Computes N Gaussian expectation integrals of one or more functions
     using Gauss-Hermite quadrature. The Gaussians must be independent.
 
-    :param funcs: Callable or Iterable of Callables that operates elementwise
+    :param funcs: the integrand(s):
```
awav: Type information should be in annotations, in docstring descriptions what it does. This `...: ...:` format is a bit odd.

awav:
How did you improve numerical stability?

st--:
@awav I wasn't using unittest, just pytest. I used session_context() instead of session_tf() because from its docstring it wasn't clear that you could use it to run() tensors as well - I've made that clearer.

st--:
@awav numerical stability: tf.reduce_logsumexp(logp) instead of tf.log(tf.reduce_sum(tf.exp(logp)))

gpflow/quadrature.py
st--: no, because I don't know how to describe the types - it could be tf.Tensor, tf.Variable, ... - maybe even a numpy array - and it doesn't make sense to have a complicated Union[] all over the place: if you add an appropriate type in gpflow I'm happy to add it here.

gpflow/quadrature.py
st--: I know - how would I write that it can be a function, or an iterable of functions, each of which takes some number of elements (not defined, but same as length of Fmu / Fvar)? Similar problem to your other comment...

st--:
NB- if anyone has a better suggestion for the name of the flag other than `logspace` I'd be happy to hear!

Joshuaalbert:
Using `reduce_logsumexp` will imporve stability, especially for gradient calculations however I don't see why you should have the `logspace` option. If it improves stability, then it should always be used. Am I missing a possible use case for the `logspace` option?

st--:
Can this be merged please?

PR #754, opened 2018-05-10
RBF Cross-Covariance

sanket-kamthe:
### Description
New  **feature**

- Adds RBF cross covariance in expectations. 
Previous implementation defaulted to quadrature expectations for cross co-variance when kernel and feature were not the same. 

- Should supersede previous pull request for the same feature  [#634](https://github.com/GPflow/GPflow/pull/634)
- Added test cases to cover all possible scenarios

- Cleaned up code based on [#672`](https://github.com/GPflow/GPflow/pull/672) which allowed 
```params_as_tensors_for()``` to take multiple arguments.

For e.g.,  long winding context manager
```python

 with params_as_tensors_for(kern1), params_as_tensors_for(feat1), \
        params_as_tensors_for(kern2), params_as_tensors_for(feat2):
````
is replaced by cleaner
```python
 with params_as_tensors_for(kern1, feat1, kern2, feat2):
````

### DIFF
--- a/doc/source/notebooks/multioutput.ipynb
+++ b/doc/source/notebooks/multioutput.ipynb
@@ -85,7 +85,8 @@
     "ExecuteTime": {
      "end_time": "2018-06-19T10:58:45.195606Z",
      "start_time": "2018-06-19T10:58:44.172232Z"
-    }
+    },
+    "collapsed": true
    },
    "outputs": [],
    "source": [
@@ -109,7 +110,8 @@
     "ExecuteTime": {
      "end_time": "2018-06-19T10:58:46.500202Z",
      "start_time": "2018-06-19T10:58:46.486870Z"
-    }
+    },
+    "collapsed": true
    },
    "outputs": [],
    "source": [
@@ -136,7 +138,8 @@
     "ExecuteTime": {
      "end_time": "2018-06-19T10:58:48.106040Z",
      "start_time": "2018-06-19T10:58:48.092379Z"
-    }
+    },
+    "collapsed": true
    },
    "outputs": [],
    "source": [
@@ -172,7 +175,8 @@
     "ExecuteTime": {
      "end_time": "2018-06-19T10:58:48.863635Z",
      "start_time": "2018-06-19T10:58:48.766655Z"
-    }
+    },
+    "collapsed": true
    },
    "outputs": [],
    "source": [
@@ -287,7 +291,8 @@
     "ExecuteTime": {
      "end_time": "2018-06-19T10:58:53.293025Z",
      "start_time": "2018-06-19T10:58:53.290410Z"
-    }
+    },
+    "collapsed": true
    },
    "outputs": [],
    "source": [
@@ -349,7 +354,8 @@
     "ExecuteTime": {
      "end_time": "2018-06-19T10:58:58.849062Z",
      "start_time": "2018-06-19T10:58:58.840498Z"
-    }
+    },
+    "collapsed": true
    },
    "outputs": [],
    "source": [
@@ -363,7 +369,8 @@
     "ExecuteTime": {
      "end_time": "2018-06-19T10:58:59.599039Z",
      "start_time": "2018-06-19T10:58:59.546311Z"
-    }
+    },
+    "collapsed": true
    },
    "outputs": [],
    "source": [
@@ -489,7 +496,8 @@
     "ExecuteTime": {
      "end_time": "2018-06-19T10:59:09.467172Z",
      "start_time": "2018-06-19T10:59:09.459302Z"
-    }
+    },
+    "collapsed": true
    },
    "outputs": [],
    "source": [
@@ -503,7 +511,8 @@
     "ExecuteTime": {
      "end_time": "2018-06-19T10:59:11.113435Z",
      "start_time": "2018-06-19T10:59:11.001429Z"
-    }
+    },
+    "collapsed": true
    },
    "outputs": [],
    "source": [
@@ -661,7 +670,8 @@
     "ExecuteTime": {
      "end_time": "2018-06-19T10:59:21.702858Z",
      "start_time": "2018-06-19T10:59:21.673859Z"
-    }
+    },
+    "collapsed": true
    },
    "outputs": [],
    "source": [
@@ -675,7 +685,8 @@
     "ExecuteTime": {
      "end_time": "2018-06-19T10:59:22.210502Z",
      "start_time": "2018-06-19T10:59:22.089982Z"
-    }
+    },
+    "collapsed": true
    },
    "outputs": [],
    "source": [
@@ -820,7 +831,8 @@
     "ExecuteTime": {
      "end_time": "2018-06-19T10:59:34.470299Z",
      "start_time": "2018-06-19T10:59:34.439405Z"
-    }
+    },
+    "collapsed": true
    },
    "outputs": [],
    "source": [
@@ -834,7 +846,8 @@
     "ExecuteTime": {
      "end_time": "2018-06-19T10:59:35.106186Z",
      "start_time": "2018-06-19T10:59:34.971615Z"
-    }
+    },
+    "collapsed": true
    },
    "outputs": [],
    "source": [
@@ -1008,7 +1021,8 @@
     "ExecuteTime": {
      "end_time": "2018-06-19T10:59:48.838797Z",
      "start_time": "2018-06-19T10:59:48.808721Z"
-    }
+    },
+    "collapsed": true
    },
    "outputs": [],
    "source": [
@@ -1022,7 +1036,8 @@
     "ExecuteTime": {
      "end_time": "2018-06-19T11:00:14.728947Z",
      "start_time": "2018-06-19T11:00:14.543249Z"
-    }
+    },
+    "collapsed": true
    },
    "outputs": [],
    "source": [
@@ -1191,7 +1206,8 @@
     "ExecuteTime": {
      "end_time": "2018-06-19T11:00:24.912718Z",
      "start_time": "2018-06-19T11:00:24.896649Z"
-    }
+    },
+    "collapsed": true
    },
    "outputs": [],
    "source": [
@@ -1205,7 +1221,8 @@
     "ExecuteTime": {
      "end_time": "2018-06-19T11:00:25.585881Z",
      "start_time": "2018-06-19T11:00:25.489128Z"
-    }
+    },
+    "collapsed": true
    },
    "outputs": [],
    "source": [
@@ -1378,7 +1395,8 @@
     "ExecuteTime": {
      "end_time": "2018-06-19T11:00:35.438985Z",
      "start_time": "2018-06-19T11:00:35.422006Z"
-    }
+    },
+    "collapsed": true
    },
    "outputs": [],
    "source": [
@@ -1392,7 +1410,8 @@
     "ExecuteTime": {
      "end_time": "2018-06-19T11:00:35.866898Z",
      "start_time": "2018-06-19T11:00:35.769898Z"
-    }
+    },
+    "collapsed": true
    },
    "outputs": [],
    "source": [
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: tests/test_expectations.py [hunk 1]
```diff
@@ -101,6 +106,9 @@ def dirac_markov_gauss():
 def rbf_kern():
     return kernels.RBF(Data.D_in, variance=rng.rand(), lengthscales=rng.rand() + 1.)
 
+@cache_tensor
+def rbf_kern_2():
```
st--: knowing what this PR is about I understand why you've got essentially the same code twice, but it'd be good to have a short comment explaining this duplication!

st--:
@sanket-kamthe Looks like this is passing - anything else needed before we merge it?

sanket-kamthe:
@st--  I will add the comment for the second rbf_kern() and then yes everything is ready I think for merge.

awav:
@sanket-kamthe any updates on it?

gpflow/expectations.py
```diff
@@ -230,7 +230,15 @@ def expectation(p, obj1, obj2=None, nghp=None):
     try:
         return _expectation(p, obj1, feat1, obj2, feat2, nghp=nghp)
     except NotImplementedError as e:
-        print(str(e))
+
+        def custom_warning_format(msg, *args, **kwargs):
```
awav: I'm not sure about this lines. Do you still need this?

sanket-kamthe: Yes, I would keep it. Otherwise the old msg was warning which was not informative at all. These two lines are the only way that I know of for format warning msg.

Also, unlike other print statements we eliminated, this is not a debug logger, we really want the user to know that specific expectation is solved by quadrature, if that is expected behaviour  then the user can choose to ignore the warning. 

I am not a fan of printing on stdout and logger is not appropriate for this case, as we are explicitly catching an error.

awav: a lot of things have changed since you created this PR, we added logger and converted some warning to switchable logs

gpflow/expectations.py
```diff
@@ -364,57 +372,90 @@ def _expectation(p, kern1, feat1, kern2, feat2, nghp=None):
     Compute the expectation:
     expectation[n] = <Ka_{Z1, x_n} Kb_{x_n, Z2}>_p(x_n)
         - Ka_{.,.}, Kb_{.,.} :: RBF kernels
-    Ka and Kb as well as Z1 and Z2 can differ from each other, but this is supported
-    only if the Gaussian p is Diagonal (p.cov NxD) and Ka, Kb have disjoint active_dims
-    in which case the joint expectations simplify into a product of expectations
+    Ka and Kb as well as Z1 and Z2 can differ from each other.
 
-    :return: NxMxM
+    :return: Nxdim(Z1)xdim(Z2)
```
awav: N x dim(Z1) x dim(Z2)

gpflow/expectations.py
```diff
@@ -364,57 +372,90 @@ def _expectation(p, kern1, feat1, kern2, feat2, nghp=None):
...
+            La, Lb = map(get_squared_length_scales, (Ka, Kb))
+            LaLb = La * Lb / (La + Lb)  # average length scale
 
-        sqrt_det_L = tf.reduce_prod(0.5 * squared_lengthscales) ** 0.5
-        C = tf.cholesky(0.5 * tf.matrix_diag(squared_lengthscales) + Xcov)  # NxDxD
+        sqrt_det_L = tf.reduce_prod(LaLb) ** 0.5
+        C = tf.cholesky(tf.matrix_diag(LaLb) + Xcov)  # NxDxD
         dets = sqrt_det_L / tf.exp(tf.reduce_sum(tf.log(tf.matrix_diag_part(C)), axis=1))  # N
 
+        def get_C_and_z_terms(Z, C=C):
```
awav: docstrings: what are these C and z terms?

gpflow/expectations.py
```diff
@@ -364,57 +372,90 @@ def _expectation(p, kern1, feat1, kern2, feat2, nghp=None):
...
+            z2_CC_inv_mu = z1_CC_inv_mu  # NxM
+            Z2 = Z1
+        else:
+            Z2, _ = Kb._slice(feat2.Z, p.mu)
+            C_inv_z2, z2_CC_inv_z2 = get_C_and_z_terms(Z2 / Lb * LaLb)
+            z2_CC_inv_mu = 2 * tf.matmul(C_inv_z2, C_inv_mu, transpose_a=True)[:, :, 0]  # NxM2
+
+        z1_CC_inv_z2 = tf.matmul(C_inv_z1, C_inv_z2, transpose_a=True)  # NxM1xM2
+
+        exponent_mahalanobis = mu_CC_inv_mu + tf.expand_dims(z2_CC_inv_z2, 1) + \
```
awav: z2_CC_inv_z2 = tf.expand_dims(z2_CC_inv_z2, 1)
z1_CC_inv_z1 = tf.expand_dims(z1_CC_inv_z1, 2)
z1_CC_inv_mu = tf.expand_dims(z1_CC_inv_mu, 2) ...

gpflow/expectations.py
```diff
@@ -364,57 +372,90 @@ def _expectation(p, kern1, feat1, kern2, feat2, nghp=None):
...
+        exponent_mahalanobis = tf.exp(-0.5 * exponent_mahalanobis)  # NxM1xM2
+
+        if Z1 == Z2:
+            # Compute sqrt(self.K(Z)) explicitly to prevent automatic gradient from
+            # being NaN sometimes, see pull request #615
+            dist_term = tf.exp(-0.25 * Ka.square_dist(Z1, None))
+        else:
+            # Compute exp( -.5 (Z-Z')^top (L_1+L_2)^{-1} (Z-Z') )
+            lengthscales_rms = tf.sqrt(La + Lb)
+            Z1 = Z1 / lengthscales_rms
```
awav: `Z1 /= lengthscales_rms` and others

gpflow/expectations.py
awav: tbh, I think you need add some math clarification of what's going on here, _not only for this function_

tests/test_expectations.py
awav: @sanket-kamthe agreed, add docstring to this test describing what's going on

gpflow/expectations.py
sanket-kamthe: I know, but this is still valid case, as you can see original statement is still print.

awav: point is that it is not really a scope of this task, It is global problem which is solvable by changing logger level or turning off warning messages

awav: btw, `expectations.py` don't use warnings anymore, only logger

gpflow/expectations.py
```diff
@@ -230,7 +230,15 @@ def expectation(p, obj1, obj2=None, nghp=None):
     try:
         return _expectation(p, obj1, feat1, obj2, feat2, nghp=nghp)
     except NotImplementedError as e:
-        print(str(e))
+
+        def custom_warning_format(msg, *args, **kwargs):
+            # ignore everything except the message
+            return str(msg) + '\n'
+
+        warnings.formatwarning = custom_warning_format
```
st--: This needs to be reset to the previous version afterwards.

gpflow/expectations.py
```diff
@@ -364,57 +372,95 @@ def _expectation(p, kern1, feat1, kern2, feat2, nghp=None):
...
-        squared_lengthscales = kern.lengthscales ** 2. if kern.ARD \
-            else tf.zeros((D,), dtype=settings.tf_float) + kern.lengthscales ** 2.
+        def get_squared_length_scales(kern):
+            squared_lengthscales = kern.lengthscales ** 2. if kern.ARD \
+                        else tf.zeros((D,), dtype=settings.tf_float) + kern.lengthscales ** 2.
+            return squared_lengthscales
 
-        sqrt_det_L = tf.reduce_prod(0.5 * squared_lengthscales) ** 0.5
-        C = tf.cholesky(0.5 * tf.matrix_diag(squared_lengthscales) + Xcov)  # NxDxD
+        if Ka == Kb:
```
st--: For understanding correctly: does this mean "equal lengthscale parameters"?

gpflow/expectations.py
```diff
@@ -364,57 +372,95 @@ def _expectation(p, kern1, feat1, kern2, feat2, nghp=None):
...
+            La = get_squared_length_scales(Ka)
+            Lb = La
+            LaLb = La * 0.5
+        else:
+            La, Lb = map(get_squared_length_scales, (Ka, Kb))
+            LaLb = La * Lb / (La + Lb)  # average length scale
+
+        sqrt_det_L = tf.reduce_prod(LaLb) ** 0.5
```
st--: notation is slightly confusing/inconsistent with the maths - I understand you don't want to multiply and divide by 2, but then maybe better to call it half_geom_mean_scale (or something along those lines)?

st--:
This PR needs to undo the deprecation of Stationary.scaled_square_dist() in kernels.py

gpflow/expectations.py
sanket-kamthe: Yes, equal length scales, we use variances at the end only.

gpflow/expectations.py
sanket-kamthe: calling it  `half_mean_L`

gpflow/expectations.py
```diff
@@ -110,9 +110,6 @@ def _quadrature_expectation(p, obj1, feature1, obj2, feature2, num_gauss_hermite
     """
     num_gauss_hermite_points = 100 if num_gauss_hermite_points is None else num_gauss_hermite_points
 
-    logger.warn("Quadrature is used to calculate the expectation. This means that "
-                "an analytical implementations is not available for the given combination.")
```
st--: implementations -> implementation

gpflow/expectations.py
```diff
@@ -364,57 +361,108 @@ def _expectation(p, kern1, feat1, kern2, feat2, nghp=None):
...
+        # along M2
+        z1_CC_inv_mu = tf.expand_dims(z1_CC_inv_mu, 2)
+        z1_CC_inv_z1 = tf.expand_dims(z1_CC_inv_z1, 2)
+
+        # expanded version of ((Z1 + Z2)-mu) (CCT)-1 ((Z1 + Z2)-mu)
+        exponent_mahalanobis = mu_CC_inv_mu + z2_CC_inv_z2 + \
+                               z1_CC_inv_z1 + 2 * z1_CC_inv_z2 - \
+                               z1_CC_inv_mu- z2_CC_inv_mu  # NxM1xM2
+
+        exponent_mahalanobis = tf.exp(-0.5 * exponent_mahalanobis)  # NxM1xM2
```
rhaps0dy: I think it would make sense to name this left-hand-side `mahalanobis`

gpflow/expectations.py
```diff
@@ -364,57 +361,108 @@ def _expectation(p, kern1, feat1, kern2, feat2, nghp=None):
...
+
+        if Z1 == Z2:
+            # CAVEAT : Compute sqrt(self.K(Z)) explicitly
+            # to prevent automatic gradient from
+            # being NaN sometimes, see pull request #615
+            dist_term = tf.exp(-0.25 * Ka.scaled_square_dist(Z1, None))
```
rhaps0dy: Even if `Z1 == Z2`, shouldn't the lengthscales here be `lengthscales_rms` like in the `else` branch? Hold on, I'm going to build a failing test case.

sanket-kamthe: I don't think so, as this Z1 and Z2 are not same as feat1.Z and feat2.Z they are already scaled. Moreover, unless the objects are not same this branch is never visited. This condition was added to make sure we do not lose performance for both rbfs being exactly same.

ialong: I agree with @sanket-kamthe here. `Ka.scaled_square_dist` takes care of the square rooting and scaling.

st--:
@sanket-kamthe Hi Sanket, would you be up for porting this PR to gpflow 2 ?

Issue #756, opened 2018-05-14
Optimization slows down when redone many times


tomaskrehlik:
1. When you rerun optimisation over the same model many times it becomes slower and slower.
2. MWE:
```
opt = gpflow.train.ScipyOptimizer()
mean = gpflow.mean_functions.Constant()
k = gpflow.kernels.Matern52(1) 

x = np.atleast_2d(np.random.rand(100))
y = np.atleast_2d(np.random.rand(100))

m = gpflow.models.GPR(x, y, kern=k, mean_function=mean)

for i in range(100):
    %time opt.minimize(m)
```
3.
    * GPflow 1.1.1
    * Tensorflow version tensorflow==1.7.0
    * No GPU
    * Python 3.6.4
    * macOS

Is there a way how to clean the tf graph easily, it seems to be growing with every iteration.

markvdw:
That is strange. I don't immediately see why the graph should grow with this code.

hughsalimbeni:
I think this is inevitable as the scipy optimizer build a graph each time (https://github.com/GPflow/GPflow/blob/bac97289f9c3194f13729eb6b8531b4c6afcaec8/gpflow/training/external_optimizer.py#L221). If you use the actions framework you can call the optimize operation directly, and not have to rebuild the graph each time

Brunarspunar:
I had a quite similar problem. It was fixed by resetting the graph in each iteration, see the last comment in #719. In your case you will have to reassign the trained parameters from the previous iteration, but I guess that is possible too.

markvdw:
I like @hughsalimbeni's solution. Did you have a one-liner example in mind?

awav:
@tomaskrehlik , Hello! Are you sure that you are using scipy optimizer correctly, maybe you need: `opt.minimize(m, **maxiter=1000**)`? In any case,  you can use `session.run(optimizer.optimize_tensor)`.

sanket-kamthe:
@awav  is the actions framework that @hughsalimbeni  is mentioning documented somewhere ?  I need to actually do the optimisation in the issue where `opt.minimize` needs to be called within a loop (RL loop)

kozisek:
@hughsalimbeni Can the Action framework be used with ScipyOptimizer?

Using it with other optimizers (like AdamOptimizer) does work as a workaround for this problem.

awav:
@kozisek, 
> Can the Action framework be used with ScipyOptimizer?
No.

[...]

PR #761, opened 2018-05-17
Mitigate ScipyOptimizer graph expansion by initializing optimize_tensor only once

kozisek:
# Mitigate graph expansion by initializing optimize_tensor only once

## Description
While investigating the cause of #756 , we found that the primary suspect of the slow down is expanding computation graph which grows by ~380 operations on every run of `optimizer.minimize()`. The cause of the expansion is a repeated instantiation of the optimize_tensor on every `minimize()` call even if the `ScipyOptimizer` instance is not recreated.
By creating a new optimize_tensor instance only once (and then on demand) as in the actions framework examples, the graph grows only by 11 operations on each run. While this does not fix the problem described in #756 completely, it mitigates it substantially.

## Minimal example
Using a modified example from #756 with a seed for reproducibility
```
import gpflow
import numpy as np

opt = gpflow.train.ScipyOptimizer()
mean = gpflow.mean_functions.Constant()
k = gpflow.kernels.Matern52(1) 

np.random.seed(42)
x = np.atleast_2d(np.random.rand(100))
y = np.atleast_2d(np.random.rand(100))

m = gpflow.models.GPR(x, y, kern=k, mean_function=mean)

def optimize(iterations):
    for i in range(iterations):
        opt.minimize(m)
    return

%time optimize(100)
```

Before the change, the computation takes ~4 minutes. After the change, it takes ~10 seconds to finish while reaching the same objective.

### Disclaimer
The loop above is meant only to showcase the slow down of repeated computations, not as the primary use case of repeated calls of the `optimizer.minimize()` function. The use case of running the `minimize()` function repeatedly can range from reusing the same model with different data to loading a saved model and running further optimization.

### DIFF
--- a/gpflow/training/scipy_optimizer.py
+++ b/gpflow/training/scipy_optimizer.py
@@ -80,11 +80,11 @@ class ScipyOptimizer(optimizer.Optimizer):
 
         session = model.enquire_session(session)
         self._model = model
-        optimizer = self.make_optimize_tensor(model, session,
-            var_list=var_list, maxiter=maxiter, disp=disp)
-        self._optimizer = optimizer
+        if initialize or self._optimizer is None:
+            self._optimizer = self.make_optimize_tensor(model, session,
+                      var_list=var_list, maxiter=maxiter, disp=disp)
         feed_dict = self._gen_feed_dict(model, feed_dict)
-        optimizer.minimize(session=session, feed_dict=feed_dict, **kwargs)
+        self._optimizer.minimize(session=session, feed_dict=feed_dict, **kwargs)
         if anchor:
             model.anchor(session)
 


awav:
> The loop above is meant only to showcase the slow down of repeated computations

* Actually, all optimizers have this affect. 
* `minimize` function has loop inside and default maxiter value is 1000, therefore you are running 100 x 1000.
* It is even more tricky with scipyoptimizers. https://docs.scipy.org/doc/scipy/reference/optimize.html. It doesn't allow you to run optimization step by step, scipy uses callbacks and controls optimization loop by itself.

awav:
@kozisek Hello Jakub, any thoughts on comments above?

PR #764, opened 2018-05-22
Add logger to settings

vdutor:
# Logging capabilities

Recently, we needed to inform a user about specific functions being called inside GPflow.
For example, in the expectations framework when quadrature is used and in the multi-output framework to indicate which conditional is used.
We did this using simple print-statements, but this isn't a good practice.
This PR adds a standard Python `logger` to GPflow which can be used for these use-cases.

### Minimal working example

```python
from gpflow import settings
logger = settings.logger()
...
logger.debug("Quadrature is used to calculate the kernel expectation for Matern-1/2")
...
```

The debugging level `(CRITICAL, WARNING, INFO, etc.)` is set in the RC-file and can be changed to a one's needs.

### DIFF
--- a/gpflow/_settings.py
+++ b/gpflow/_settings.py
@@ -92,11 +92,12 @@ class _SettingsManager(object):
         return self.logging.level
 
     def logger(self):
-        frm = inspect.stack()[1]
-        mod = inspect.getmodule(frm[0])  # calling module
-        level = getattr(logging, self.logging.level)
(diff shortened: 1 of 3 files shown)


markvdw:
This looks great. Do you want it merged now, or only once the code actually contains some things that use the logger?

awav:
@markvdw, I'll merge it soon ;)

PR #771, opened 2018-05-25
Small adjustment to let ndiagquad pass non-Tensor keywords

Joshuaalbert:
### Title
Small modifcation to ndiagquad to allow arbitrary object keyword passing

### Description

This allows non-Tenor parameters to be passed to likelihood methods by conditionally skipping the `reshape` and `tile` operations.

### Example
For example, suppose I've observed phase variations across a field of view for a radio antenna.
These variations are due to the integrated line of sight electron content (TEC). There is physical grounding to suppose TEC is Gaussian correlated, but not phase (because it is a wrapped quantity).
Therefore, my dataspace is phase, but my model space is TEC and the relation between them is related to the frequency of observation. Furthermore, while eval_freq below is a scalar, a tensor heteroscedastic_variance could be used.

``` python
# in the model class
    def predict_y(self, Xnew, num_samples, eval_freq=None):
        """
        Draws the predictive mean and variance at the points `X`
        num_samples times.
        X should be [N,D] and this returns [S,N,num_latent], [S,N,num_latent]
        """
        Xnew = tf.tile(Xnew[None,:,:],[num_samples,1,1])
        Fmean, Fvar = self._build_predict(Xnew, full_cov=False, S=None)
        hetero_variance = tf.square(self.likelihood.hetero_noise(Xnew))
        return self.likelihood.predict_mean_and_var(Fmean, Fvar, eval_freq=eval_freq, hetero_variance=hetero_variance)

# in the likelihood
    @params_as_tensors
    def conditional_mean(self, F, eval_freq=None,hetero_variance=None, **unused_kwargs):  
        """The mean of the likelihood conditioned on latent."""
        eval_freq = self.freq if eval_freq is None else eval_freq
        tec2phase = self.tec_conversion/eval_freq
        phase = F*tec2phase
        return phase

    @params_as_tensors
    def conditional_variance(self, F,hetero_variance=None, **unused_kwargs):
        return hetero_variance 
```

## Code quality requirements
No major changes made

### DIFF
--- a/gpflow/quadrature.py
+++ b/gpflow/quadrature.py
@@ -133,7 +133,7 @@ def ndiagquad(funcs, H: int, Fmu, Fvar, **Ys):
     gh_w = wn.reshape(-1, 1) * np.pi ** (-0.5 * Din)  # H**Din x 1
 
     for name, Y in Ys.items():
-        if not isinstance(Y,(np.array,tf.Tensor)):
+        if not isinstance(Y,(np.ndarray,tf.Tensor)):
             # some things that one might wish to pass the likelihood are not 
             # Tensors (like when the latent and likelihood are in different
             # spaces, and you want to transform between them for the 


st--:
Hi, your example does not actually use the added feature of ndiagquad, how would your `conditional_mean()` get called with keyword arguments - could you please extend the MWE?

Also, it would be better to have a separate function to check whether an argument can be used as a tensor or not (this is not just np.ndarrays and tf.Tensors, but also tf.Variables - see https://www.tensorflow.org/api_docs/python/tf/contrib/framework/is_tensor though we would need to add the np.ndarray check on top and may not want to depend on contrib so perhaps better to roll our own?).

PR #789, opened 2018-06-13
Speed up notebooks

st--:
This PR adds a switch to use a much smaller number of iterations for minimize(), sample() and loops in continuous integration tests compared to running locally, which significantly speeds up the tests. It also adds a sample_feed_dict() method to Model which is needed to speed up the MCMC predict_f calls on the samples.


st--:
Because of the checks for CI environment, the notebooks are now slightly messier - I hope this is made up for by the fact that now only one of the five slowest tests is a notebook test, and the overall travis runtime is reduced by about one third.

vdutor:
Went through the code - ready to be merge. @markvdw @awav

gpflow/test_util.py
```diff
@@ -107,3 +108,23 @@ def test_context(self, graph=None):
         graph = self.test_graph if graph is None else graph
         with graph.as_default(), self.test_session(graph=graph) as session:
             yield session
+
+
+def is_continuous_integration():  # pragma: no cover
```
sanket-kamthe: Isn't it easier and cleaner to write 
```python
is_continous_integration = os.environ.get('CI', False)
```

gpflow/test_util.py
```diff
@@ -107,3 +108,23 @@ def test_context(self, graph=None):
...
+    ci = os.environ.get('CI')
+    return (ci == 'true') or (ci == '1')
+
+def nb_niter(n, test_n=2):  # pragma: no cover
```
sanket-kamthe: Same here just a shorter code and easier to read 

```python
return n if is_continous_integration else test_n
```

gpflow/test_util.py
vdutor: @sanket-kamthe -- I've made your proposed changes

st--: No, it's not, just the environment variable existing does not mean that we're in CI - it could be set to "false" or "0". I think explictly checking for true or 1 is safer.

st--:
we could change is_continuous_integration to a module-level global variable, and rename the function to _is_continuous_integration() ... ? would then have to change all the notebooks as well, but might be cleaner...

gpflow/test_util.py
sanket-kamthe: I agree but this code already does that if it is set to zero or False it will be zero or False. 
If 'CI' set to False then it should return False. Travis and other environments always have this flag set to `True`.
Check her for details https://docs.travis-ci.com/user/environment-variables/#Default-Environment-Variables

st--: No, it doesn't; environment variables are always strings, and "0" and "False" both evaluate to `True`.

st--: Anyway, changed it back already :P

st--:
Ready to merge, please!
@awav if you prefer notebook_niter instead of nb_niter I'm happy to change that, it just needs to be consistent..

PR #792, opened 2018-06-15
New monitor initial commit

ahsimb:
### Description

This is close to Mark's original implementation of the monitor. The main principal change is that it works as a callback function provided to an optimiser (or whatever code that wants to use it).

The callback is added to all existing optimisers except HMC.

Also the proper unit testing is provided.

### Minimal working example

In the monitor module docstring.


gpflow/training/monitor/__init__.py
```diff
@@ -1,4 +1 @@
-from .actions import *
-
-print("Warning: `monitor` is still in beta. It is slightly buggy, will change, "
-      "and is not tested properly in the unittests.")
+from .monitor import *
```
awav: Super bad! :) https://stackoverflow.com/questions/2386714/why-is-import-bad

gpflow/training/monitor/monitor.py
```diff
@@ -0,0 +1,790 @@
...
+
+
+def get_hr_time() -> float:
+    """
+    Gets high resolution time. Mainly defined here for the convenience of unit testing.
+    """
+    return timer()
+
+
+def create_global_step(tf_session: tf.Session) -> tf.Variable:
```
awav: Minor: we don't have another abstraction for session, `tf_session` can be renamed to `session`.

gpflow/training/monitor/monitor.py
```diff
@@ -0,0 +1,790 @@
...
+    :param checkpoint_dir: checkpoint files directory.
+    :param saver: The saver object, if not provided a default saver object will be created.
+    """
+    checkpoint_path = tf.train.latest_checkpoint(checkpoint_dir)
+    print("Restoring session from `%s`." % checkpoint_path)
+    saver = saver or get_default_saver()
+    saver.restore(tf_session, checkpoint_path)
+
+
+def get_default_saver() -> tf.train.Saver:
```
awav: Recommended: Add default argument for `max_to_keep`?

gpflow/training/monitor/monitor.py
```diff
@@ -0,0 +1,790 @@
...
+def restore_session(tf_session: tf.Session, checkpoint_dir: str,
+                    saver: Optional[tf.train.Saver] = None) -> None:
+    """
+    Restores Tensorflow session from the latest checkpoint.
+    :param tf_session: The TF session
+    :param checkpoint_dir: checkpoint files directory.
+    :param saver: The saver object, if not provided a default saver object will be created.
+    """
+    checkpoint_path = tf.train.latest_checkpoint(checkpoint_dir)
+    print("Restoring session from `%s`." % checkpoint_path)
```
awav: Major: print => logger. Check gpflow logger.

gpflow/training/monitor/monitor.py
```diff
@@ -0,0 +1,790 @@
...
+
+
+def get_default_saver() -> tf.train.Saver:
+    """
+    Creates Tensorflow Saver object with 3 recent checkpoints to keep.
+    """
+    return tf.train.Saver(max_to_keep=3)
+
+
+class MonitorContext(object):
```
awav: Minor: Can you move comments for public properties to docstring?

gpflow/training/monitor/monitor.py
```diff
@@ -0,0 +1,790 @@
...
+
+    def print_summary(self, monitor_tasks: Optional[Iterator['MonitorTask']]=None) -> None:
+        """
+        Prints the tasks' timing summary.
+        :param monitor_tasks: List of monitoring tasks to print the summary for, defaults to
+        all tasks.
+        """
+        print("Tasks execution time summary:")
+        for mon_task in (monitor_tasks or self._monitor_tasks):
+            print("%s:\t%.4f (sec)" % (mon_task.task_name, mon_task.total_time))
```
awav: Major: print => logging

gpflow/training/monitor/monitor.py
```diff
@@ -0,0 +1,790 @@
...
+
+        with params_as_tensors_for(self._model):
+            tf_x, tf_y = self._model.X, self._model.Y
+
+        if self._display_progress:  # pragma: no cover
+            try:
+                import tqdm
+                wrapper = tqdm.tqdm
+            except ImportError:
+                print("LML monitor task: to display progress install `tqdm`.")
```
awav: Major: print => proper logging

gpflow/training/monitor/monitor.py
```diff
@@ -0,0 +1,790 @@
...
+        self._full_lml = tf.placeholder(settings.tf_float, shape=())
+        self._summary = tf.summary.scalar(model.name + '/full_lml', self._full_lml)
+        self._display_progress = display_progress
+
+    def run(self, context: MonitorContext, *args, **kwargs) -> None:
+
+        with params_as_tensors_for(self._model):
+            tf_x, tf_y = self._model.X, self._model.Y
+
+        if self._display_progress:  # pragma: no cover
```
awav: Minor. For avoiding assigning 2x times wrapper with `lambda x: x` you can do this:
```
wrapper = None
if self._display_progress:
  ... pass ...
if wrapper is None:
   wrapper = lambda x: x
```

gpflow/training/monitor/monitor.py
```diff
@@ -0,0 +1,790 @@
...
+                import tqdm
+                wrapper = tqdm.tqdm
+            except ImportError:
+                print("LML monitor task: to display progress install `tqdm`.")
+                wrapper = lambda x: x
+        else:   # pragma: no cover
+            wrapper = lambda x: x
+
+        lml = 0.0
+        num_batches = -(-len(self._model.X._value) // self._minibatch_size)  # round up
```
awav: Recommended: There is `math.ceil`.
```
num_values = len(self._model.X._value)
num_batches = math.ceil(num_values / self._minibatch_size)
```

gpflow/training/monitor/monitor.py
```diff
@@ -0,0 +1,790 @@
...
+        :param parameters: List of model parameters to send to TensorBoard. If not
+        provided all parameters will be sent to TensorBoard.
+        :param additional_summaries: List of additional summary objects to send to TensorBoard.
+        """
+        super().__init__(model, event_path)
+        all_summaries = additional_summaries or []
+        parameters = parameters or list(model.parameters)
+
+        # Add scalar parameters
+        all_summaries += [tf.summary.scalar(p.full_name, tf.reshape(p.constrained_tensor, []))
```
awav: From `full_name` docstring: "WARNING: WILL BE DEPRICATED SOON."

gpflow/training/natgrad_optimizer.py
```diff
@@ -76,10 +78,11 @@ def minimize(self, model, var_list=None, session=None, feed_dict=None,
         opt = self.make_optimize_action(model, session=session, var_list=var_list, **kwargs)
         with session.as_default():
             for _i in range(maxiter):
-                opt()
+                result = opt()
```
awav: What does `opt()` return? Is that something useful and usable for callback?

gpflow/training/scipy_optimizer.py
```diff
@@ -54,15 +53,18 @@ def make_optimize_tensor(self, model, session=None, var_list=None, **kwargs):
             model.initialize(session=session)
             return optimizer
 
-    def minimize(self, model, session=None, var_list=None, feed_dict=None,
-                 maxiter=1000, disp=False, initialize=False, anchor=True, **kwargs):
+    def minimize(self, model, session=None, var_list=None, feed_dict=None, maxiter=1000,
+                 disp=False, initialize=False, anchor=True, step_callback=None, **kwargs):
```
awav: I know, it will sound strange, but could you add type annotation for step_callback?

gpflow/training/tensorflow_optimizer.py
```diff
@@ -95,6 +95,8 @@ def minimize(self, model, session=None, var_list=None, feed_dict=None,
             initialized before for gotten session.
         :param anchor: If `True` trained variable values computed during optimization at
             particular session will be synchronized with internal parameter values.
+        :param step_callback: A callback function to execute at each optimization step,
+        supplied with the current value of the packed variable vector.
```
awav: Minor: you forgot indentation after newline.

gpflow/training/tensorflow_optimizer.py
```diff
@@ -112,7 +114,9 @@ def minimize(self, model, session=None, var_list=None, feed_dict=None,
         session = model.enquire_session(session)
         with session.as_default():
             for _ in range(maxiter):
-                opt()
+                result = opt()
+                if step_callback is not None:
+                    step_callback(result)
```
awav: Same here, is `result` useful for step_callback?

gpflow/training/monitor/monitor.py
ahsimb: OK, changed

gpflow/training/monitor/monitor.py
ahsimb: OK, makes sense, done.

gpflow/training/monitor/monitor.py
ahsimb: The intention was to print into stdout, not log.

gpflow/training/monitor/monitor.py
ahsimb: Changed

gpflow/training/monitor/monitor.py
ahsimb: Changed.

gpflow/training/monitor/monitor.py
markvdw: I agree with @ahsimb. It's not really debug, it's really about user interaction.

gpflow/training/monitor/monitor.py
markvdw: Yeah I kind of agree with this.

markvdw:
@awav @ahsimb What do you think about using the GPflow logger instead of print? I'm mostly in favour of print, since we want to actually use this for user-interaction, so I think we should always print. Also, currently monitor doesn't actually depend on anything in GPflow too much. It would be nice to have the option to spin it off as a separate tool.

gpflow/training/monitor/monitor.py
ahsimb: OK, good idea, done.

ahsimb:
Re: logging. I think logger is a wrong thing for printing timings, but right thing for printing task execution info and errors. I also agree that it might be a good idea to take monitor out of GPFlow. There are two tasks - tensorboards that are specific to GPFlow. Everything else is generic. We could make GPFlow extension of the monitor and put those two tasks in it. Going back to logging, There are only two insignificant places where messages are printed - when the graph is restored from a checkpoint and when we don't find `tqdm` installed. It's probably not worth keeping a reference to GPFlow for the sake of these two things.

gpflow/training/monitor/monitor.py
ahsimb: Changed.

gpflow/training/monitor/monitor.py
ahsimb: I guess // is more efficient - no conversion from int to float and back. Obviously it's just a question of code beauty.

gpflow/training/monitor/monitor.py
ahsimb: Should I use `pathname` instead?

gpflow/training/monitor/monitor.py
st--: math.ceil returns a float so needs casting back to int but I think this is not going to be time-critical and in this case I would favour being explicit about what you mean.

gpflow/training/natgrad_optimizer.py
ahsimb: I just realised I didn't change the interface in the optimizer.py. That's where I should have documented the step_callback.

Basically we pass back whatever the optimiser returns. If anything, it will most likely be the values of trainable variables. So one can write a specialised task that having a specific optimiser in mind will do something with this returned value.

gpflow/training/monitor/__init__.py
ahsimb: I deleted the __init__.py altogether.

gpflow/training/monitor/monitor.py
ahsimb: Changed.

gpflow/training/tensorflow_optimizer.py
ahsimb: Fixed

doc/source/notebooks/monitor-tensorboard.ipynb
```diff
@@ -88,425 +53,423 @@
...
+    "\n",
+    "sleep_task = mon.SleepTask(0.01).with_name('sleep').with_name('sleep')\n",
+    "\n",
+    "saver_task = mon.CheckpointTask('./monitor-saves').with_name('saver')\\\n",
+    "    .with_condition(mon.PeriodicIterationCondition(10))\\\n",
+    "    .with_exit_condition(True)\n",
+    "\n",
+    "std_tboard_task = mon.StandardTensorBoardTask(m, './model-tensorboard').with_name('std_tboard')\\\n",
+    "    .with_condition(mon.PeriodicIterationCondition(10))\\\n",
+    "    .with_exit_condition(True)\n",
```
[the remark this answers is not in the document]
ahsimb: No, we can scrap with_global_step_tensor. I added it because I thought we need global_step it in the summary. Since we don't actually need it there it makes things easier.

gpflow/training/monitor/monitor.py
```diff
@@ -0,0 +1,801 @@
...
+        """
+        self._flush_immediately = flush_immediately
+        return self
+
+    def with_global_step_tensor(self, global_step_tensor: tf.Variable) -> 'BaseTensorBoardTask':
+        """
+        Adds optimisation global step to the summary.
+        """
+        if self._summary is not None:
+            self._summary = tf.summary.merge(
```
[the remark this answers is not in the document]
ahsimb: This is exactly what I've done.

gpflow/training/monitor/monitor.py
```diff
@@ -0,0 +1,801 @@
...
+            raise RuntimeError('To run a TensorBoard monitor task the TF session object'
+                               ' must be provided when creating an instance of the Monitor')
+
+        summary = context.session.run(self._summary, feed_dict=feed_dict)
+        self._file_writer.add_summary(summary)
+        if self._flush_immediately:
+            self.flush()
+
+
+class StandardTensorBoardTask(BaseTensorBoardTask):
```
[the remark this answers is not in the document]
ahsimb: OK, I renamed all tensorboard tasks except the abstract to xxxToTensorBoardTask.

ahsimb:
Sorry about the junk commit comment.

vdutor:
I've been using the new monitor framework and it works perfectly. I also really like the new syntax. @markvdw @awav , can we merge this, please?

st--:
One more thing before merging: why does `with_exit_condition()` not have a default argument? "with_exit_condition(False)" is a no-op anyways, and "with_exit_condition(True)" seems a bit redundant...

ahsimb:
I think it's better to be explicit about the `with_exit_condition`. We may decide to change the default to True in the future. You either don't care whether it runs or not in which case the default decides it for you, or you do care and set it. In the latter case the behavior of your code doesn't change when the default changes.

hughsalimbeni:
Hi any chance of merging this soon? I'd really like to use it!

awav:
@ahsimb @st-- , I agree with Ti point. Moreover, `with_exit_condition` stands for exit condition will be applied and default value `True` is reasonable, but you need ability to change it back whenever you want.

gpflow/training/optimizer.py
```diff
@@ -37,7 +37,24 @@ def make_optimize_tensor(self, model, session=None, var_list=None, **kwargs):
 
     @abc.abstractmethod
     def minimize(self, model, session=None, var_list=None, feed_dict=None,
-                 maxiter=1000, initialize=True, anchor=True, **kwargs):
+                 maxiter=1000, initialize=True, anchor=True, step_callback=None, **kwargs):
```
awav: Minor: Docs for `NotImplemented` method is bad idea, it does nothing. Descendants can re-define what each of the arguments means.

gpflow/training/scipy_optimizer.py
awav: @ahsimb highlight the callback signature

gpflow/training/scipy_optimizer.py
```diff
@@ -70,6 +69,9 @@ def minimize(self, model, session=None, var_list=None, feed_dict=None,
             initialized before for gotten session.
         :param anchor: If `True` trained parameters computed during optimization at
             particular session will be synchronized with internal parameter values.
+        :param step_callback: A function to be called at each optimization step;
```
awav: @ahsimb Same here about signature of callback, are these values tensors or numpy arrays? ;)

gpflow/training/tensorflow_optimizer.py
awav: The reason why I'm asking is that should you pass any other information or not. @vdutor, @st-- , what do you think?

setup.py
```diff
@@ -22,6 +22,7 @@
     'multipledispatch>=0.4.9',
     'pytest>=3.5.0',
     'h5py>=2.7.0',
+    'matplotlib>=2.0.2'
```
awav: why did we change it?

markvdw:
@awav @ahsimb Looks like I was a bit trigger happy on the merge. Sorry for that. Can these comments be addressed in a separate PR?

PR #794, opened 2018-06-15
Use a single index-tensor in vec_to_tri instead of map_fn

mrksr:
Hello everybody!

# Why this PR?
For the last few days I have been playing around with using [distributed tensorflow](https://www.tensorflow.org/deploy/distributed) for gpflow training. Most things are fairly straightforward, but there seems to be one exeception: The `vec_to_tri`-transform. Specifically, the forward-implementation using `map_fn` seems to prevent tensorflow from placing parameters such as `q_sqrt` on a parameter server instead of a worker. I am not quite sure why this is the case, but I can say that if the code is reimplemented without `map_fn`, the problem goes away.

**Update**: `tensorflow 1.10` seems to solve the issue as well.

# Example Code
Here is a small program that showcases the problem:
```python
import sys

import numpy as np
import tensorflow as tf
import gpflow
from gpflow import settings

cluster = tf.train.ClusterSpec({
    'ps': ['localhost:2222'],
    'worker': ['localhost:2223'],
})

if '--ps' in sys.argv:
    job_name = 'ps'
else:
    job_name = 'worker'

server = tf.train.Server(
    cluster,
    job_name=job_name,
    task_index=0,
    config=tf.ConfigProto(**settings.session),
)

if job_name == 'ps':
    server.join()
else:
    with tf.Session(server.target).as_default() as sess:
        device_setter = tf.train.replica_device_setter(
            ps_device='/job:ps',
            worker_device='/job:worker',
            cluster=cluster,
        )
        with tf.device(device_setter):
            q_sqrt = np.eye(5, dtype=settings.float_type)[None, ...]
            q_sqrt = gpflow.params.Parameter(q_sqrt, transform=gpflow.transforms.LowerTriangular(5, 1))

            q_sqrt.compile()
            print('q_sqrt', sess.run(q_sqrt.parameter_tensor))
```

If you save the program and start two instances `python example.py --ps` and `python example.py`, you will be greeted with an exception looking something like this:
```
InvalidArgumentError (see above for traceback): Cannot colocate nodes 'Parameter-feb74d67-0/IsVariableInitialized' and 'Parameter-feb74d67-0/unconstrained: Cannot merge devices with incompatible jobs: '/job:ps/task:0' and '/job:worker/replica:0/task:0/device:CPU:0'
         [[Node: Parameter-feb74d67-0/IsVariableInitialized = IsVariableInitialized[_class=["loc:@Parameter-feb74d67-0/unconstrained"], dtype=DT_DOUBLE, _device="/job:ps/task:0"](Parameter-feb74d67-0/unconstrained)]]
```
If you change  the transform of `q_sqrt` to something different, `q_sqrt/unconstrained` will be placed on `/job:ps` and everything works as expected.

# Review
I did not benchmark the new implementation against the old, so it is possible that the old implementation is faster.


gpflow/misc.py
```diff
@@ -153,13 +153,13 @@ def int_shape(x):
     # Check M is a valid triangle number
     assert((matrix * (N + 1)) == (2 * M))
     """
-    indices = list(zip(*np.tril_indices(N)))
-    indices = tf.constant([list(i) for i in indices], dtype=tf.int64)
+    D = int(vectors.shape[0])
```
hughsalimbeni: This will break if vectors has unknown shape. Maybe this is fine for most uses, but I think the motivation for the existing code is to deal with dynamic shapes. If this code is adopted then D should probably also be an input.

mrksr:
@hughsalimbeni commented that in the first implementation I assumed that the shape of `vectors` will always be known, which was not necessary in the current implementation (thanks for taking a look!).
As a basis for discussion I reimplemented the tiling-operations in tensorflow to support unknown shapes. Depending on how well tensorflow does constant propagation, this might be just as fast as the numpy-implementation or slower if the tilings have to be repeatet. Does anybody have an idea if the tiling operations will be fused into a constant? Any ideas about the performance of `map_fn`?

I guess another option would be to add another parameter to the function (comparable to `N`). But I fear this would mean that we do not really support unknown shapes again.

hughsalimbeni:
In all practical settings I've come across the `q_sqrt` matrix has known shape anyway, so I personally wouldn't be fussed about making the shape static. It would be really useful to have a benchmark comparison of the three methods. Actually, while we're at it, I'd be interested in whether the transform actually saves time over using a full matrix (so that's 4 methods: with `map_fn`, with static shape, with tile and with no transform). Certainly looking at traceplots for large M and large D in the SVGP model I can see the `scatter_nd` ops taking up a negligible compute, but perhaps nearly halving the number of parameters saves time overall.

jameshensman:
@mrksr this looks interesting. @lucabordo might have a comment?

lucabordo:
It looks like this needs a bit of investigation. Thanks @mrksr for sharing this example which I can indeed repro

mrksr:
I will run some timing comparisons once I get the chance. I can say that distributed training of GPs is really quite worth it from a performance side.

mrksr:
Here are the results of a few performance tests. I implemented four version discussed above:

* **_map**: The current implementation
* **_numpy**: My first implementation using numpy to construct the indexer
* **_tensorflow_unknown_D**: My second implementation using tensorflow to construct the indexer
* **_tensorflow_known_D**: An alternative tensorflow-implementation where the `D`-dimension is resolved at tensor construction time 

To test the timings I sample a large number of random triangular matrices, forward-transform them to vectors and construct a backward-transforming tensor. I then evaluate this tensor multiple times using `timeit`. I think the overhead introduced by the slowly growing graph should be negligable.

Here is what I think is the most relevant result: I set `D=2` and vary N between 1 and 1000. This corresponds to a SVGP with 2 outputs and up to 1000 pseudo-inputs.
timings_n_small_d
The three new implementations are all very comparable, but at some point map will be a bit faster. This becomes more pronounced with larger `D=200`:
timings_n

Lastly, I set `N=200` and vary `D` which shows very similar results:
timings_n_small_d

I think the areas where `map_fn` starts to be faster are not very relevant in practice and for "normal" examples the new implementation is considerably faster. I would therefore propose to use the **_tensorflow_unknown_D** implementation since it is quite general. What do you think?

Here is the Notebook: [vec_to_tri_benchmark.zip](https://github.com/GPflow/GPflow/files/2127633/vec_to_tri_benchmark.zip)
I did not vary session parameters. The benchmarks were done on a machine with 12 CPUs and more than enough RAM.

mrksr:
The **_tensorflow_unknown_D** implementation is the one currently implemented in this PR. What are the next steps to merge this? @hughsalimbeni @jameshensman

hughsalimbeni:
I ran the notebook on a K40 GPU and this is what the plots look like:

download1

download2


download3

hughsalimbeni:
But if we're in the M=O(1000) regime the current way is better

mrksr:
The tiling seems to dominate at some point, since if `D=1`, `map_fn` remains slower:
2_replicas
1_replica

hughsalimbeni:
Note new [tensorflow 1.10](https://github.com/tensorflow/tensorflow/releases/tag/v1.10.0) includes 

> Add FillTriangular Bijector to create triangular matrices

which may be worth considering in the future

alexggmatthews:
@mrksr @hughsalimbeni  thanks for all this work. I think that distributions is intended to stay in the main tf repo but maybe it has changed since I lasts talked to them. Personally, I don't see an issue depending on tensorflow probability in principal though we would need to check the details. The triangular transform has been a constant pain to maintain and is not part of our core purpose, so I would happily use a decent implementation from outside.

alexggmatthews:
@mrksr Ok the warning method is pretty informative. Note also the news about TensorFlow 2.0 #824

mrksr:
I think we can close this considering #871. If GPflow is based on tensorflow-probability, using the `FillTriangular` transform seems like the obvious choice. @awav @alexggmatthews @hughsalimbeni

PR #799, opened 2018-06-20
Monte-Carlo likelihoods

st--:
This PR provides Monte-Carlo likelihoods. First version of the code written by @markvdw. It adds a soft-max likelihood and, for comparison purposes, a stochastic Gaussian likelihood.


st--:
@markvdw currently, predict_mean_and_var() always returns zero - is this intentional?

gpflow/likelihoods.py
```diff
@@ -116,10 +116,10 @@ def variational_expectations(self, Fmu, Fvar, Y):
 
 
 class Gaussian(Likelihood):
-    def __init__(self, var=1.0, name=None):
+    def __init__(self, variance=1.0, name=None):
```
markvdw: I totally agree with this renaming.

gpflow/likelihoods.py
```diff
@@ -583,3 +583,136 @@ def conditional_variance(self, F):
         E_y = tf.matmul(phi, Ys)
         E_y2 = tf.matmul(phi, tf.square(Ys))
         return tf.reshape(E_y2 - tf.square(E_y), tf.shape(F))
+
+
+class MonteCarloLikelihood(Likelihood):
+    def __init__(self, name=None):
+        super().__init__(name)
+        self.num_monte_carlo_points = 100
+        del self.num_gauss_hermite_points
```
markvdw: Maybe at some points in the future we should change this to a single property? E.g. `num_numerical_estimation_points` or something? This should be done after consulting with other people who have looked at quadrature before.

gpflow/likelihoods.py
```diff
@@ -583,3 +583,136 @@ def conditional_variance(self, F):
...
+
+    def _mc_evals(self, func, Fmu, Fvar, Y=None, epsilon=None):
+        N, D = tf.shape(Fmu)[0], tf.shape(Fvar)[1]
+        if epsilon is None:
+            epsilon = tf.random_normal((self.num_monte_carlo_points, N, D), dtype=settings.float_type)
+        mc_x = Fmu[None, :, :] + tf.sqrt(Fvar[None, :, :]) * epsilon
+        mc_Xr = tf.reshape(mc_x, (self.num_monte_carlo_points * N, D))
+
+        if Y is not None:
```
markvdw: Is this to support likelihoods which don't have Y?

gpflow/likelihoods.py
```diff
@@ -583,3 +583,136 @@ def conditional_variance(self, F):
...
         return tf.reshape(E_y2 - tf.square(E_y), tf.shape(F))
+
+
+class MonteCarloLikelihood(Likelihood):
+    def __init__(self, name=None):
+        super().__init__(name)
+        self.num_monte_carlo_points = 100
+        del self.num_gauss_hermite_points
+
+    def _mc_evals(self, func, Fmu, Fvar, Y=None, epsilon=None):
```
markvdw: Should this not be a static method or just a function somewhere? I don't see it referring to `self` anywhere.

gpflow/likelihoods.py
```diff
@@ -583,3 +583,136 @@ def conditional_variance(self, F):
...
+        and the predictive variance
+
+           \int\int y^2 p(y|f)q(f) df dy  - [ \int\int y^2 p(y|f)q(f) df dy ]^2
+
+        Here, we implement a default Monte Carlo routine.
+        """
+        mc_cm = self._mc_evals(self.conditional_mean, Fmu, Fvar, epsilon=epsilon)
+
+        mean, var = tf.nn.moments(mc_cm, [0])
+        unbiased_var = var * self.num_monte_carlo_points / (self.num_monte_carlo_points - 1)
```
markvdw: What's the reasoning behind this? Could you add a small comment? All it needs is something that explains what `tf.nn.moments()` returns, and then why the factor makes it unbiased.

gpflow/likelihoods.py
st--: No, predict_mean_and_var() doesn't have Y

gpflow/likelihoods.py
st--: self.num_monte_carlo_points. Yeah, was thinking it could be moved to quadrature.py, but I think that might be scope for a separate PR.

st--:
We probably want a test that compares GaussianMC() and Gaussian() and checks they give reasonably close answers

gpflow/likelihoods.py
st--: done

gpflow/likelihoods.py
```diff
@@ -583,3 +583,132 @@ def conditional_variance(self, F):
...
+        return tf.reduce_mean(mc_logp, axis=0)  # N x 1
+
+
+class GaussianMC(MonteCarloLikelihood, Gaussian):
+    """
+    Stochastic version of Gaussian likelihood for comparison.
+    """
+    def __init__(self, variance=1.0, name=None):
+        super().__init__(name=name)
+        self.variance = variance
```
markvdw: Not that I really care about this, but this looks like a perfect example of thinking about cooperative multiple inheritance. Should the constructor of `MonteCarloLikelihood` not also call `super()` with kwargs?

If you're interested, check out https://www.youtube.com/watch?v=EiOglTERPEo. It's a good watch anyway.

st--: Like in 69f4c59 ? :)

gpflow/likelihoods.py
```diff
@@ -583,3 +582,105 @@ def conditional_variance(self, F):
...
+    """
+    The soft-max multi-class likelihood.
+    """
+
+    def __init__(self, num_classes, **kwargs):
+        super().__init__(**kwargs)
+        self.num_classes = num_classes
+
+    def logp(self, F, Y):
+        return -tf.nn.sparse_softmax_cross_entropy_with_logits(logits=F, labels=Y[:, 0])[:, None]
```
st--: Is `Y[:, 0]` the correct thing to do here? shouldn't it be tf.squeeze(axis=1), maybe an assert that it only has one dimension?

gpflow/likelihoods.py
```diff
@@ -203,8 +202,8 @@ def variational_expectations(self, Fmu, Fvar, Y):
 
 
 class StudentT(Likelihood):
-    def __init__(self, deg_free=3.0):
-        Likelihood.__init__(self)
+    def __init__(self, deg_free=3.0, **kwargs):
```
st--: "deg_free" sounds pretty terrible... any better idea for how to call it ?

hughsalimbeni: degrees_of_freedom, d_o_f, dof??

gpflow/likelihoods.py
hughsalimbeni: Yes I put this in a while ago when we decided not to support multiple columns of multiclass classification. An assert would be good

gpflow/likelihoods.py
st--: Scipy & numpy just call it `df`... would you prefer `dof`?

st--:
Codecov/patch fails because of implementation of new SoftMax likelihood is untested. Can't think of a good way of testing that. Please merge!

st--:
@markvdw SoftMax sets self.num_classes but this is not used anywhere...

markvdw:
@st-- Shouldn't `testclass.py` be under the directory `tests` or something?

st--:
Woops, testclass shouldn't've gotten added.

PR #807, opened 2018-06-29
Fixed discovered problem with the monitor, in particular with TensorBoard tasks

ahsimb:
### Description

Addressed remaining comments in the initial PR

Fixed the problem with the TensorBoard tasks writing to the same location. Since the tf.summary.FileWriter doesn't allow shared access only one task was able to write events into the location. Made the tasks sharing the FileWriter in case they use the same location.

Fixed a bug with ImageToTensorBoardTask. It's now closing the Figure objects. 

### Minimal working example

See monitor-tensorboard notebook. Now TensorBoard should have all created elements.


gpflow/training/monitor.py
```diff
@@ -631,16 +632,27 @@ class BaseTensorBoardTask(MonitorTask):
     providing these values as the input values dictionary.
     """
 
+    _all_file_writers = {}  # type: Dict[str, tf.summary.FileWriter]
```
awav: Can we avoid to use global variable here? Most of the time this will not be an issue, but nevertheless this dictionary can grow fast. Simple example: using loop and create different directories for tensorboard, like 'path_1', 'path_2', ... 'path_n', and, btw,  they will never be closed - that's another issue. Options:
* pass writer explicitly (it is easy to do) and create one only when 'str' passed.
* create abstraction for TensorBoard and pass this one instead of the file writer.

ahsimb: I agree that using a global variable is not nice in general and keeping not closed file writers is bad (although in practice it's very unlikely cause problems). So I made some improvements.
I didn't want to pass it over to a user to deal with the FileWriter for two reasons. One would expect the framework to do things like opening and closing files. More importantly, a user is likely to run into the same problem as us - trying to open multiple writers (not everyone reads every line of documentation). TF kindly doesn't raise any exception, it just silently doesn't do anything. So the user will jump to immediate conclusion that the Monitor doesn't work.
Handling writers at the monitor level is also problematic because the user can create more than  one monitor.
So I left the FileWriter cache in a global variable but now I make sure that all writers are closed when the monitor stops running. Unless they are used by another monitor, in which case the last one will close all remaining writers.
I also made the monitor running in a context which slightly simplifies the syntax and ensures that the tasks close their writers.

gpflow/training/monitor.py
```diff
@@ -630,6 +729,11 @@ class BaseTensorBoardTask(MonitorTask):
...
     method where it can calculate the correspondent values. It will then call the _eval_summary
     providing these values as the input values dictionary.
+
+    A TensorBoard task requests access to the Tensorflow summary FileWrite object providing the
+    location of the event file. The FileWriter object will be created if it doesn't exist. When
+    the task is no longer needed the `close` method should be called. This will release the
+    FileWriter object.
     """
 
     _all_file_writers = {}  # type: Dict[str, tf.summary.FileWriter]
```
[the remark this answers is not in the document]
ahsimb: Oops, sorry, well spotted!

awav:
@ahsimb, okay, the code became much more complex, this is the very last what we want (or just me, depending on level of disagreement). It is very brave to invent your own GC inside monitor problem with GCed language for standing problem ;)
Here is my solution, I use it time to time when I don't want to bother myself with tensorflow writers. This is a thin wrapper for `tf.summary.Writer` with nice feature which allows you to log not only tensorflow tensors, but any values (`np.ndarray`s, standard types - `int`, `float` and cetera)

```python
class TensorboardWriter():
    def __init__(self, logdir, suffix=None):
        suffix = '' if suffix is None else '_' + suffix
        self.writer = tf.summary.Writer(logdir=logdir, filename_suffix=prefix)

    def log_scalar(self, name, value, step):
        summary = tf.Summary(value=[tf.Summary.Value(tag=name, simple_value=value)])
        self.writer.add_summary(summary, step)
    
    def log_histogram(self, name, values, step):
        pass

    def __enter__(self, ...):
        pass

    def __exit__(self, ...):
        pass
```

@vdutor , @st-- , @markvdw 
How to use this, several options:
1. Almost, nothing changes for user except you need to create `TensorboardWriter('path/to/logdir')` before you actually create tasks. User passes wrapper object instead of  logdir `str`. I don't mind to allow to pass string as well and create writer inside task, when user doesn't have many tasks to manage and doesn't want to bother him/herself. 
  + User can choose different names of files inside logdir (using suffixes)
  + No problem with deallocating and closing writers within loop
  + Nice features, like logging standard python types
  + `with` context can be used as well

So our usecase becames:
```
tfboard1 = TensorboardWriter('./monitor-saves')
print_taskprint_t  = mon.PrintTimingsTask(tfboard1)\
    .with_name('print')\
    .with_condition(mon.PeriodicIterationCondition(10))\
    .with_exit_condition(True)

sleep_task = mon.SleepTask(0.01).with_name('sleep').with_name('sleep')

saver_task = mon.CheckpointTask(tfboard1)\
    .with_name('saver')\
    .with_condition(mon.PeriodicIterationCondition(10))\
    .with_exit_condition(True)

tfboard2 = TensorboardWriter('./monitor-saves', suffix='red')
model_tboard_task = mon.ModelToTensorBoardTask(tfboard2, m)\
    .with_name('model_tboard')\
    .with_condition(mon.PeriodicIterationCondition(10))\
    .with_exit_condition(True)

lml_tboard_task = mon.LmlToTensorBoardTask(tfboard2, m)\
    .with_name('lml_tboard')\
    .with_condition(mon.PeriodicIterationCondition(100))\
    .with_exit_condition(True)
```

2. Get rid of logdir arguments in Tensorboard tasks and Monitor object will pass writer wrapper inside tasks with context.

Thanks

ahsimb:
I am not sure I understood how this wrapper class addresses the problem. As a user I can writer it like this, and the second task will never write anything.

`model_tboard_task = mon.ModelToTensorBoardTask(TensorboardWriter(my_dir), m)\
    .with_name('model_tboard')\
    .with_condition(mon.PeriodicIterationCondition(10))\
    .with_exit_condition(True)

lml_tboard_task = mon.LmlToTensorBoardTask(TensorboardWriter(my_dir), m)\
    .with_name('lml_tboard')\
    .with_condition(mon.PeriodicIterationCondition(100))\
    .with_exit_condition(True)
`
What's actually wrong with my FileWriter cache?

awav:
User has choice how to use `TensorboardWriter(my_dir)`, when in original case user couldn't control creation of writers at all. Plus, user can explicitly create different writers with suffixes (`TensorboardWriter('.', suffix='task_1')`) to write into the same logdir, which of course must be documented in notebooks and examples.

1. This approach gives more control over the problem with writers.
2. It doesn't blows up our codebase much.
3. Wrapper around the writer can have potentially nice methods which is useful for other things apart from monitor stuff.

Issue with tensorflow writers is not a monitor problem. Providing full solution requires too much unnecessary effort, but we can provide flexible interface to reduce the headache.

awav:
@markvdw , @vdutor , @st-- your opinions?

ahsimb:
OK, let me clarify what the issue is about.

The fundamental problem is that a). TensorFlow doesn't allow shared access to the event file. b) It doesn't show any error when such access is attempted. Only the first FileWriter object writes data, all others fail silently.

There are two possible approaches to the problem.

A. Try to make the user aware of this problem, let him/her handle it and hope that he/she will do it right.

B. Provide a workaround so that the user doesn't need to be concerned about this issue.

I favour B even if it takes writing some ugly looking code, so long as it works. Not all users thoroughly study the doc. Let the one who always does this throw the first stone into me. People usually consult the doc when they get an error (preferably with some meaningful message). In this particular case there will be no error. Personally, in this situation I would just assume that the tool is flaky and can't be relied on.

What @awav is suggesting seems to fall under A. We ask the user to create either a FileWriter or some wrapper around it and explain that multiple objects cannot be created with the same event file.

awav:
@ahsimb, there is no silver bullet here, solution B. provides other difficulties: our own GC inside monitoring or global variables, neither of this solutions are good. 

> let him/her handle it and hope that he/she will do it right.

_There is no hope, there is a trade-off_. Solution, which I proposed is to force a user use flexible way, so that user could choose. The example which I showed is so simple, that I do not really understand how someone can't miss that point if it is in _documentation/notebook_ - forewarned IS forearmed. 
 I don't buy the point that people don't read documentation (Something wrong with me then ;)). 

Another, even simpler solution __is just use suffixes for writer__, by default. Like it is done for all GPflow objects. Introduce name for monitor tasks and when you create writer do this:

```
name = self.__class__.__name__ if name is None else name
writer = tf.summary.FileWriter(logdir=logdir, filename_suffix=name)
```

ahsimb:
I think there is a risk of this discussion sliding into non-productive zone. Let me try to put it into the format of a code review as I understand it.

This PR solves the problem I have described earlier (or so the PR claims). Could interested parties please put their thoughts in roughly one of the two areas:

1. The problem is not worth solving. This is what @awav seems to be saying. If others agree I will scrap the FileWriters cache and let the user handle the writers. It's an easy change.

2. The solution has such and such technical issues. Please be specific. Comments like "your code looks like GC" are hard to address.

Thanks

ahsimb:
Ditto. Do we want the user to deal with this problem or we deal with it ourselves? 

>  more freedom
I don't understand how the "freedom" is limited. The user can choose to write data from whatever task to whatever location, same or different. The only thing, I don't use suffixes, which I can easily add if needed.

st--:
How about a global state, but for the *pathnames only* -- and then generating a warning (but not erroring) when a user tries to instantiate a TensorboardWriter() with a path that's already been opened, telling the user that this *may* otherwise silently fail (because another tf.FileWriter may or may no longer be open already). Means we don't need to worry about GC, but there's still at least some kind of indication why things might be going wrong for unsuspecting users.

ahsimb:
Sounds like a good compromise to me. I would just always throw an error. If the user wants to reuse the location he/she must close the TensorboardWriter (or better yet use it in `with`). In the `close` method we will remove the path from the collection.

We don't actually worry about GC (real one), it's just the @awav allegoric comparison, since both GC and the writer cache I made do reference counting.

awav:
@ahsimb , @st-- , in original solution there was a problem with tensorflow graphs, because writer can keep as many as user want and when graph(s) is released by user, it is still kept by writer. Python GC will never collect this object.

@st--, with your case there is an issue of growing dictionary/list in global variable, once path is added - it is never removed. You can resolve it by @ahsimb solution, but again we will implement this: `Try to make the user aware of this problem, let him/her handle it and hope that he/she will do it right.`. I conclude there is no difference between my approach and this one.

There I see this choices:
1. Make wrapper for file writer with extra features for writing to the board any types (not only values from tensors)
2. Make `Monitor` responsible for creating file writer and pass it with context. Tasks will use another writer to other places (logdirs) if user passed explicitly writer to the task. I guess this is good alternative.
3. Make global variable work, taking @st-- and upgrading @ahsimb ideas. So, we can leave cache for logdirs + suffixes, it is a set of tuples (logdir, suffix). It can be a thin wrapper. And we can get rid of `close` function as well!

```python
class _TensorboardDirs(UserList):
    def __contains__(self, item: Tuple[str, Optional[str]]):
        if item in self.__dict__['data']:
            raise RuntimeError('Tensorboard directory {dir} is already in use'. Even file suffix {suffix}.format(dir=item[0], suffix=item[1]))
        return False

_tensorboard_dirs = __TensorboardDirs([])

class AnyTask():
    def __del__(self):
        _tensorboard_dirs.remove(self.event_file_tuple)
        super.__del__()
```

vdutor:
Hi all,

Thank you all for the discussion - this has definitely improved the end result.
I would vouch for solution (4): it gives all control to the user, but prevents the user of creating `FileWriter`s writing to the same TensorBoard event file.

gpflow/training/monitor.py
```diff
@@ -621,6 +640,90 @@ def run(self, context: MonitorContext, *args, **kwargs) -> None:
...
+        """
+
+        self._location = (str(PurePath(logdir)), filename_suffix)
+        self._is_locked = False
+        self.__lock_location()
+        super().__init__(logdir, graph, max_queue, flush_secs, filename_suffix=filename_suffix)
+
+    def __del__(self):
+        self.__release_location()
+        if hasattr(super(), '__del__'):
```
awav: All objects have __del__ method. It is not necessary to check it.

gpflow/training/monitor.py
```diff
@@ -621,6 +640,90 @@ def run(self, context: MonitorContext, *args, **kwargs) -> None:
...
+
+        :param logdir: Directory where event file will be written.
+        :param graph: A `Graph` object
+        :param max_queue: Size of the queue for pending events and summaries.
+        :param flush_secs: How often, in seconds, to flush the added summaries and events to disk.
+        :param filename_suffix: Optional suffix of the event file's name.
+        """
+
+        self._location = (str(PurePath(logdir)), filename_suffix)
+        self._is_locked = False
```
awav: there is no need in extra `self._is_locked` variable, you can make a function which checks that location is present in locations set.

gpflow/training/monitor.py
```diff
@@ -621,6 +640,90 @@ def run(self, context: MonitorContext, *args, **kwargs) -> None:
...
+    def reopen(self) -> None:
+        """
+        Reopens the summary.FileWriter that has been previously closed. Attempts to lock the event
+        file location. Will raise an error if the location  has been taken by another writer since
+        it was closed by this writer.
+        """
+        self.__lock_location()
+        super().reopen()
+
+    def __lock_location(self) -> None:
```
awav: jooc, why double `_ _`?

gpflow/training/monitor.py
```diff
@@ -629,18 +732,21 @@ class BaseTensorBoardTask(MonitorTask):
...
-    def __init__(self, event_path: str, model: Optional[Model]=None) -> None:
+    def __init__(self, file_writer: LogdirWriter, model: Optional[Model]=None) -> None:
         """
-        :param event_path: Path of the event file where the summary protocol buffers will
-        be written to.
+        :param file_writer: Event file writer object.
         :param model: Model object
         """
         super().__init__()
+        self._file_writer = file_writer
```
awav: Raise error if it is not `LogdirWriter`

gpflow/training/tensorflow_optimizer.py
```diff
@@ -96,8 +96,9 @@ def minimize(self, model, session=None, var_list=None, feed_dict=None,
         :param anchor: If `True` trained variable values computed during optimization at
             particular session will be synchronized with internal parameter values.
         :param step_callback: A callback function to execute at each optimization step.
-            If provided, the function will be called with the current value of the packed variable
-            vector as an argument.
+            Currently the function is called with no arguments.
+            TODO: consider passing the optimisation state, e.g. values of trainable variables.
```
awav: Can you remove `TODO` from docstrings? Maybe create an github issue instead?

gpflow/training/monitor.py
ahsimb: I don't think it's true. Try this
`hasattr(object(), '__del__')`

gpflow/training/monitor.py
ahsimb: Yes, but why is it better?

gpflow/training/monitor.py
ahsimb: I want it to be a private attribute.

gpflow/training/monitor.py
ahsimb: OK, will do

gpflow/training/monitor.py
```diff
@@ -69,43 +69,39 @@
...
 checkpoint_task = mon.CheckpointTask(checkpoint_dir="./model-saves")\
-    .with_name('checkpoint')\
-    .with_condition(mon.PeriodicIterationCondition(15))\
-
-monitor_tasks = [print_task, tensorboard_task, checkpoint_task]
+        .with_name('checkpoint')\
+        .with_condition(mon.PeriodicIterationCondition(15))\
 
-# Create and start the monitor.
+# This task will create a Tensoreflow summary of the model for Tensorboard. It will run at
```
vdutor: Minor: typo in "Tensoreflow"

gpflow/training/monitor.py
```diff
@@ -629,18 +732,21 @@ class BaseTensorBoardTask(MonitorTask):
     If the summary object contains one or more placeholders it may also override the `run`
     method where it can calculate the correspondent values. It will then call the _eval_summary
     providing these values as the input values dictionary.
+
+    A TensorBoard task requests access to the Tensorflow summary FileWrite object providing the
```
vdutor: Minor: typo in "FileWrite"

gpflow/training/monitor.py
awav: no need in variables which actually do nothing extra on top of `in` operation for set

vdutor: I agree with Artem here, if we would rename `used_locations` to `locked_locations` we achieve the same functionality with less additional logic.

gpflow/training/monitor.py
awav: Yep, remove `hasattr(object(), '__del__')` with super().__del__()

awav: @ahsimb Now, I see why `_is_locked` is needed, the name is a bit confusing. I would rename it to `_active` - like you are activating this writer, when you close it you deactivate it and __del__ will do nothing, but in the same time another writer could be created and place same logdir in the set.

ahsimb: `del` is not a method of an object. It's a Python statement that deletes (garbage collects) an object. A class may implement a destructor `__del__`. The destructor, if defined, will be called when the object is removed either by `del` or GC. The parent class could have defined its own destructor which should also be called to clear the parent's resources. I believe the way I've done this is correct.

awav: I know that `__del__` is not part of the object. What I meant is that `__del__` is not overwritten by FileWriter and therefore you don't need to call `hasattr(object(), '__del__')` and `super().__de__()` respectively.

gpflow/training/monitor.py
ahsimb: My comment "Yes, but why is it better?" was misleading. The variable is needed. But I think we are now in agreement about this. I changed the names as per the suggestions.

awav:
Thank a lot! I'm going to merge it.

PR #811, opened 2018-07-10
Optimizer's callback accept step number.

awav:
Hello everyone,

Reasons for this PR:

- `step_callback` docstring states that optimizer doesn't pass arguments to the function and function can have as many arguments as user wants including nil arguments, but an optimizer pass results of `session.run` operation to the `step_callback`, therefore causes failures. [update] Function must be with signature `fun(*args, **kwargs)`, but it is still confusing that nothing is passed into callback.
- Optimizer minimize returns tensorflow operation, which after running this tf.Operation though `session.run` returns `None`.
- It would be nice to expose step number to callback explicitly.

Thanks!

### DIFF
--- a/gpflow/training/natgrad_optimizer.py
+++ b/gpflow/training/natgrad_optimizer.py
@@ -66,10 +66,8 @@ class NatGradOptimizer(optimizer.Optimizer):
             :param anchor: Synchronize updated parameters for a session with internal
                 parameter's values.
             :param step_callback: A callback function to execute at each optimization step.
-                The function should accept variable argument list. Currently the function is called
-                with no arguments but this may change in the future. For instance the values of
-                trainable variables may be provided.
-            :type step_callback: Callable[[], None]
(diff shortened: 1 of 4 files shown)


ahsimb:
There is a so called Liskov substitution principle. 
[https://en.wikipedia.org/wiki/Liskov_substitution_principle]

Basically, with regards to arguments, the derived class can have more generic IN parameters and more specific OUT parameters (including return). So that the caller who doesn't know that it deals with a derived class can be happy.

If the interface (Optimizer) declares the input parameter as Callback with arbitrary list of arguments the implementation e.g. Tensorflow_optimizer cannot redefine it in a more specific way.

The advantage of declaring a callback with arbitrary list of parameters is that it preserves compatibility when the list of parameters changes. If I am not interested in any parameters (which I guess is most common case) my code will work whatever the parameters are.

awav:
As discussed offline, I made changes to docstrings.

PR #814, opened 2018-07-19
Parameter.trainable was not properly a property and Parameterized.trainable did not special-case DataHolders

st--:
```python
sm = gpflow.models.SVGP(np.array([[1.0]]), np.array([[1.0]]), gpflow.kernels.RBF(1),
    gpflow.likelihoods.Gaussian(), np.array([[1.0]]),
    q_diag=False, whiten=False, minibatch_size=None)
sm.trainable = False  # does not work
sm.set_trainable(False)  # does work
```

The reason was that while in Parameterized, trainable was a property, this was not the case in Parameter. This is fixed by this PR.
Also, there was code duplication in Parameterized between trainable.setter and set_trainable(); in this PR, trainable.setter simply calls set_trainable.

### DIFF
--- a/gpflow/params/parameter.py
+++ b/gpflow/params/parameter.py
@@ -246,6 +246,14 @@ class Parameter(Node):
             return Build.NO
         return Build.YES
 
+    @property
+    def trainable(self):
+        return self._trainable
+
+    @trainable.setter
+    def trainable(self, value):
+        self.set_trainable(value)
+
     def set_trainable(self, value):
         if not isinstance(value, bool):
             raise ValueError('Fixed property value must be boolean.')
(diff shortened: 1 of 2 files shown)


st--:
Turns out I kinda accidentally fixed two separate bugs in this PR... by failing to run the code and instead spotting a different bug and thinking that had been the issue, and wanting to clean up some "duplicated" code along the way that turned out to get rid of the actual bug Mark noticed ... :rofl:

st--:
Bug 1 (the one @markvdw encountered):
Parameterized.trainable.setter called self.param.trainable=value even for DataHolders, and DataHolders.trainable was a property without a setter, so that raised an exception. set_trainable() had a check for that.
Fix: Parameterized.trainable.setter now calls set_trainable() instead of redundant code.

Bug 2 (the one I spotted and _thought_ was the one Mark was talking about):
Parameter.trainable was NOT a property, just an attribute; set_trainable() did the actual changes to whether tensorflow trains the parameter, and changed the value of the `trainable` attribute accordingly. Assigning to a Parameter.trainable would simply change the value of the flag, _without_ changing whether the parameter would actually be trained!
Fix: Internal flag is now called _trainable, and Parameter.trainable.setter calls set_trainable().

PR #825, opened 2018-08-17
Adding widedeepnetworks repository link

alexggmatthews:
I implemented the deep arcosine kernel in the code that accompanies this paper using GPflow:

https://arxiv.org/abs/1804.11271

I've added a link to the README of the project. 

Particular effort has gone into numerical stability, which allowed us to backpropagate. 

The code uses an old version of GPflow, has only basic unit test framework and would probably benefit from an option to use multiple parameters. That said it is still useful to link it in my opinion.


alexggmatthews:
Given that is a one liner I think the build failures are a false negative.

PR #827, opened 2018-08-17
Cleaning up stationary kernel implementations

st--:
Stationary kernels only depend on the distance r between x and x'; hence instead of duplicating the distance calculation in each child class of `Stationary` it would make more sense to simply define their implementation through k(r).


st--:
@vdutor one thing I'm not quite sure about yet is the name of `Kr()`. I think it'd be good to call it something slightly further away from `K()` to make the distinction clear. `k_r()` maybe? or even `k_radial`?

st--:
Please merge #828 in here first

PR #829, opened 2018-08-19
support broadcasting

hughsalimbeni:
# Broadcasting over leading dimensions 
This PR modifies the kernels and conditionals to support broadcasting over (arbitrarily many) extra leading dimensions, similar to most of the tensorflow ops. So e.g. `kern.K(X)` now supports `X` being shape `[S, ..., T, N, D]`, and returns  `[S, ..., T, N, N]`

The purpose of this PR is to facilitate models that need multiple, potentially correlated, samples. 

## Summary of new supported operations:

`kern.K(X)` supports `X` of shape  `[..., N, D]` and returns `[..., N, N]`
`kern.K(X, X2)` supports `X`, `X2` of shape  `[..., T, N, D]`, `[..., T, M, D]` and returns `[..., N, M]`, where the leading dimensions must match 
`kern.Kdiag(X)` supports `X` of shape  `[..., N, D]` and returns `[..., N]`

<s>`conditionals.conditional(Xnew, X, kern, f, full_cov=False, q_sqrt=None)` supports `Xnew` of shape [..., N, D] and returns `[..., N, Dy]`, `[..., Dy, N, N]` </s> 

EDIT leaving out the conditionals for now, as needs further discussion 

## Why not use multioutput?

Suppose a model has `Dy` independent GPs and we want to compute the full covariance between `S` different sets of inputs, each of size `N`. To use the multioutput framework for this we would need to have a `Dy * S`-output kernel, but `S` is not a parameter of our model so this is unsatisfactory. Also, if we actually have dependent GPs then the correlation structure is very tricky in the flattened version, but is straightforward using broadcasting over the leading dimensions.

### DIFF
--- a/gpflow/conditionals.py
+++ b/gpflow/conditionals.py
@@ -337,7 +337,7 @@ def _sample_mvn(mean, cov, cov_structure):
     if cov_structure == "diag":
         sample = mean + tf.sqrt(cov) * eps  # N x P
     elif cov_structure == "full":
-        cov = cov + (tf.eye(tf.shape(mean)[1], dtype=settings.float_type) * settings.numerics.jitter_level)[None, ...]  # N x P x P
+        cov = cov + (tf.eye(tf.shape(mean)[1], dtype=settings.float_type) * settings.numerics.jitter_level)  # N x P x P
         chol = tf.cholesky(cov)  # N x P x P
         return mean + (tf.matmul(chol, eps[..., None])[..., 0])  # N x P
     else:
--- a/gpflow/conditionals.py
+++ b/gpflow/conditionals.py
@@ -337,20 +337,20 @@ def uncertain_conditional(Xnew_mu, Xnew_var, feat, kern, q_mu, q_sqrt, *,
 def _sample_mvn(mean, cov, cov_structure):
     """
     Returns a sample from a D-dimensional Multivariate Normal distribution
-    :param mean: [..., D]
-    :param cov: [..., D] or [..., D, D]
+    :param mean: N x D
+    :param cov: N x D or N x D x D
     :param cov_structure: "diag" or "full"
     - "diag": cov holds the diagonal elements of the covariance matrix
     - "full": cov holds the full covariance matrix (without jitter)
-    :return: sample from the MVN of shape [..., D]
+    :return: sample from the MVN of shape N x D
     """
-    eps = tf.random_normal(tf.shape(mean), dtype=settings.float_type)  # [..., D]
+    eps = tf.random_normal(tf.shape(mean), dtype=settings.float_type)  # N x P
     if cov_structure == "diag":
-        sample = mean + tf.sqrt(cov) * eps  # [..., D]
+        sample = mean + tf.sqrt(cov) * eps  # N x P
     elif cov_structure == "full":
-        cov = cov + (tf.eye(tf.shape(mean)[-1], dtype=settings.float_type) * settings.numerics.jitter_level)  # [..., D, D]
-        chol = tf.cholesky(cov)  # [..., D, D]
-        return mean + (tf.matmul(chol, eps[..., None])[..., 0])  # [..., D]
+        cov = cov + (tf.eye(tf.shape(mean)[1], dtype=settings.float_type) * settings.numerics.jitter_level)[None, ...]  # N x P x P
+        chol = tf.cholesky(cov)  # N x P x P
+        return mean + (tf.matmul(chol, eps[..., None])[..., 0])  # N x P
     else:
         raise NotImplementedError  # pragma: no cover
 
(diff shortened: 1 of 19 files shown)


gpflow/conditionals.py
```diff
@@ -22,7 +22,7 @@
 from .features import Kuu, Kuf, InducingPoints, InducingFeature
 from .kernels import Kernel, Combination
 from .probability_distributions import Gaussian
-
+from .misc import tile_over_new_axis, swap_final_dims
```
awav: Proposal to rename functions:
* `tile_over_new_axis` => `repeat`
* `swap_final_dims` => `transpose`, so that if people hasn't specified axes function permutes last two dims.

gpflow/conditionals.py
```diff
@@ -171,55 +171,62 @@ def base_conditional(Kmn, Kmm, Knn, f, *, full_cov=False, q_sqrt=None, white=Fal
...
-    :param Knn: N x N  or  N
+    :param Kmn: ... x M x N
+    :param Kmm: ... x M x M
+    :param Knn: ... x N x N  or ... x N
     :param f: M x R
     :param full_cov: bool
     :param q_sqrt: None or R x M x M (lower triangular)
     :param white: bool
-    :return: N x R  or R x N x N
+    :return: ... x N x R  or ... x R x N x N
```
awav: It became hard to read when spaces are used. What do you think about `... x N x R` => `(...xNxR)`?

tests/test_broadcasting.py
```diff
@@ -0,0 +1,125 @@
...
+#
+#     def test_rbf_indices_active_dims(self):
+#         self._test_indices_active_dims(gpflow.kernels.RBF)
+#
+#     def test_all_no_active_dims(self):
+#         # TODO: make parallel
+#         for Kern in Kerns:
+#             self._test_no_active_dims(Kern)
+
+class ConditionalTest(GPflowTestCase):
```
awav: please use `pytests`, hint: test becomes just a function:
```
from gpflow.test_utils import session_tf
def test_conditional(session_tf):
    pass
```

gpflow/misc.py
```diff
@@ -251,5 +251,57 @@ def _get_tensor_safe(name, index, graph):
...
+        size = tf.concat([tf.shape(X)[:-1], tf.reshape(sl.stop - start, [1])], 0)
+
+        return tf.slice(X, begin, size)
+
+
+def tile_over_new_axis(X : tf.Tensor, repeats : tf.Tensor, axis : int):
+    """
+    Inserts a new axis and tiles k times
+    e.g. if X.shape = [S, N, D], and axis=-2 then this is equivalent to
+     tf.tile(X[:, :, None, :], [1, 1, repeats, 1])
```
awav: * Formatting issues - space before "tf.tile" and "supports negative ..."
* Would be nice to have docstrings for parameters

gpflow/misc.py
```diff
@@ -251,5 +251,57 @@ def _get_tensor_safe(name, index, graph):
...
+    return tf.transpose(X, perm)
+
+
+def slice_final_dim(X: tf.Tensor, sl: slice, dim: int):
+    """
+    Takes the slice sl from the final dimension of X
+    If the slice is the whole of the final dimension (i.e. sl=slice(0, dim)) then there is nothing to do
+    NB tensorflow does not support slicing with step!=1, so raise error if this is attempted
+    """
+    if (sl.start is None) and (sl.step is None) and (sl.stop == dim):  # nothing to do
```
awav: Simply: `sl == slice(None, dim, None)`

gpflow/misc.py
```diff
@@ -251,5 +251,57 @@ def _get_tensor_safe(name, index, graph):
...
+        return X
+    else:
```
awav: There is no necessity for `else` here, you can remove it.

gpflow/misc.py
```diff
@@ -251,5 +251,57 @@ def _get_tensor_safe(name, index, graph):
     except KeyError:
         return None
 
+
+def swap_final_dims(X: tf.Tensor):
+    """
+    equivalent to np.swapaxes(X, -2, -1)
```
awav: Could you replace

> equivalent to np.swapaxes(X, -2, -1)

with

> Equivalent `np.swapaxes(X, -2, -1)`. Check `https://docs.scipy.org/doc/numpy-1.14.0/reference/generated/numpy.swapaxes.html` for details.

gpflow/misc.py
```diff
@@ -251,5 +251,57 @@ def _get_tensor_safe(name, index, graph):
...
+    If the slice is the whole of the final dimension (i.e. sl=slice(0, dim)) then there is nothing to do
+    NB tensorflow does not support slicing with step!=1, so raise error if this is attempted
+    """
+    if (sl.start is None) and (sl.step is None) and (sl.stop == dim):  # nothing to do
+        return X
+    else:
+        if sl.step and (sl.step != 1):
+            raise NotImplementedError('only slices with step=1 supported by tensorflow')
+
+        start = 1 if sl.start is None else sl.start
```
awav: start = sl.start or 1

gpflow/conditionals.py
st--: @awav With "x" the letter separating the dimension sizes, I personally find it harder to read when spaces are NOT used ... maybe use the multiplication symbol "×" instead? As in N×N×R ?

gpflow/misc.py
hughsalimbeni: done

gpflow/misc.py
hughsalimbeni: done

gpflow/misc.py
hughsalimbeni: done

gpflow/misc.py
hughsalimbeni: actually it was this line that was breaking some things. It seems there are situations where the slice is as big as input_dim, but it still needs to do something. Removing so now more tests pass, though it seems we're doing a bit of unnecessary slicing here (and in the current master, too)

tests/test_broadcasting.py
hughsalimbeni: Will refactor for py_test now

gpflow/conditionals.py
hughsalimbeni: happy to make these changes, but doesn't that overload the meanings, as tensorflow already implements `repeat` and `transpose`, which have different behaviour?

gpflow/conditionals.py
hughsalimbeni: I personally find [N, D] easier to read, but happy with whatever

gpflow/conditionals.py
hughsalimbeni: annoyingly the behaviour of the tensorflow functions isn't really consistent with the rest of the broadcasting conventions, but there you go...

hughsalimbeni:
@vdutor thanks for that. I'll incorporate in the next commit.

hughsalimbeni:
@awav any further comments?

gpflow/kernels.py
```diff
@@ -122,14 +124,15 @@ def _slice(self, X, X2):
...
         """
         if isinstance(self.active_dims, slice):
-            X = X[:, self.active_dims]
+            X = slice_final_dim(X, self.active_dims, self.input_dim)
             if X2 is not None:
-                X2 = X2[:, self.active_dims]
+                X2 = slice_final_dim(X2, self.active_dims, self.input_dim)
         else:
-            X = tf.transpose(tf.gather(tf.transpose(X), self.active_dims))
+            X = tf.gather(X, self.active_dims, axis=-1)
```
awav: `X = X[..., self.active_dims]` ?

gpflow/conditionals.py
```diff
@@ -36,18 +36,15 @@
 def _conditional(Xnew, feat, kern, f, *, full_cov=False, full_output_cov=False, q_sqrt=None, white=False):
     """
     Single-output GP conditional.
-
```
awav: Why did you remove `\n` ?

gpflow/kernels.py
```diff
@@ -122,14 +124,15 @@ def _slice(self, X, X2):
...
             if X2 is not None:
-                X2 = tf.transpose(tf.gather(tf.transpose(X2), self.active_dims))
-        input_dim_shape = tf.shape(X)[1]
+                X2 = tf.gather(X2, self.active_dims, axis=-1)
```
awav: Here the same, `X = X[..., self.active_dims]`

gpflow/misc.py
```diff
@@ -251,5 +251,22 @@ def _get_tensor_safe(name, index, graph):
     except KeyError:
         return None
 
+
+def slice_final_dim(X: tf.Tensor, sl: slice, dim: int):
```
awav: what is `dim` for? it is not used in function

gpflow/kernels.py
hughsalimbeni: I've looked again and it does work for slice but not for the gather

gpflow/misc.py
hughsalimbeni: Turns out this whole method can be removed

gpflow/kernels.py
awav: as far as I know numpy indexing is translated to tf.gather and etc.

gpflow/misc.py
```diff
@@ -251,5 +251,55 @@ def _get_tensor_safe(name, index, graph):
...
+    Equivalent to np.swapaxes(X, -2, -1).
+    Check https://docs.scipy.org/doc/numpy-1.14.0/reference/generated/numpy.swapaxes.html for details.
+    """
+    K = tf.rank(X)
+    num_broadcasting_dims = tf.reduce_max([K - 2, 0])
+    perm = tf.concat([tf.range(num_broadcasting_dims), [K - 1, K - 2]], 0)
+    return tf.transpose(X, perm)
+
+
+def slice_final_dim(X: tf.Tensor, sl: slice, dim: int):
```
awav: Previous question disappeared: what dim is for, because it is not used in function.

hughsalimbeni:
any thoughts about what's going on with codecov? It's complaining about files that haven't been touched

PR #836, opened 2018-08-29
sample_conditional to return mean and var as well as samples, and generate more than one sample

st--:
## `sample_conditional`: signature change

In many cases, it is useful to have access to mean and (co)variance as well as the samples from `sample_conditional`. This PR will change the interface of sample_conditional.

Specifically, `sample_conditional()` will now return a 3-tuple of samples as well as mean and covariance (as these have to be computed anyways, and will allow reusing downstream). Code using sample_conditional() will have to change, e.g. `samples, _, _ = sample_conditional(...)` if you don't care about mean/cov.

`sample_conditional()` now takes the same full_cov and full_output_cov keyword arguments as conditional(). Requesting both full_cov _and_ full_output_cov is not possible. [Some other combinations may not be implemented (yet) by all implementations.]

It is now also possible to pass a `num_samples` keyword argument: if this is None (default), behaviour is as before, returning an N x P sample. If num_samples is an integer S, then S samples will be generated in parallel, and the return shape of samples will be S x N x P.

### DIFF
--- a/gpflow/conditionals.py
+++ b/gpflow/conditionals.py
@@ -154,9 +154,9 @@ def _sample_conditional(Xnew, feat, kern, f, *, full_cov=False, full_output_cov=
     if full_cov:
         # mean: N x P
         # cov: P x N x N
-        mean = tf.transpose(mean, [0, 1])  # now P x N
+        mean = tf.matrix_transpose(mean)  # now P x N
         samples = _sample_mvn(mean, cov, 'full', num_samples=num_samples)  # (S x) P x N
-        samples = tf.transpose(samples, [-1, -2])  # now (S x) N x P
+        samples = tf.matrix_transpose(samples)  # now (S x) N x P
 
     else:
         cov_structure = "full" if full_output_cov else "diag"
--- a/gpflow/conditionals.py
+++ b/gpflow/conditionals.py
@@ -368,12 +368,12 @@ def _sample_mvn(mean, cov, cov_structure=None, num_samples=None):
     S = num_samples if num_samples is not None else 1
 
     if cov_structure == "diag":
-        assert rank(cov) == rank(mean)
+        assert rank_cov == rank_mean
         eps = tf.random_normal([S, N, D], dtype=settings.float_type)  # S x N x D
-        sample = mean + tf.sqrt(cov) * eps  # S x N x D
+        samples = mean + tf.sqrt(cov) * eps  # S x N x D
     elif cov_structure == "full":
-        assert rank(cov) == rank(mean) + 1
-        jittermat = settings.numerics.jitter_level * tf.eye(D, batch_shape=[N], dtype=settings.float_type)[None, ...]  # N x D x D
+        assert rank_cov == rank_mean + 1
+        jittermat = settings.numerics.jitter_level * tf.eye(D, batch_shape=[N], dtype=settings.float_type)  # N x D x D
         eps = tf.random_normal([N, D, S], dtype=settings.float_type)  # N x D x S
         chol = tf.cholesky(cov + jittermat)  # N x D x D
         samples = mean[..., None] + tf.matmul(chol, eps)  # N x D x S
@@ -382,7 +382,7 @@ def _sample_mvn(mean, cov, cov_structure=None, num_samples=None):
         raise NotImplementedError  # pragma: no cover
 
     if num_samples is None:
-        return samples[..., 0]  # [N, D] (full_cov) or [N, P] (not full_cov)
+        return samples[0]  # N x D
     return samples  # S x N x D
 
 def _expand_independent_outputs(fvar, full_cov, full_output_cov):
--- a/gpflow/conditionals.py
+++ b/gpflow/conditionals.py
@@ -374,12 +374,14 @@ def _sample_mvn(mean, cov, cov_structure=None, num_samples=None):
     - "full": cov holds the full covariance matrix (without jitter)
     :return: sample from the MVN of shape N x D
     """
-    N, D = tf.shape(mean)[0], tf.shape(mean)[1]
-    # assert shape(cov) == (N, D) or (N, D, D)
-    S = num_samples if num_samples is not None else 1
+    mean_shape = tf.shape(mean)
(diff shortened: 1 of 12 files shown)


gpflow/conditionals.py
```diff
@@ -142,18 +143,37 @@ def _sample_conditional(Xnew, feat, kern, f, *, full_output_cov=False, q_sqrt=No
...
-    return _sample_mvn(mean, var, cov_structure)
+    if full_cov:
+        if full_output_cov:
+            raise NotImplementedError("The combination of both full_cov and full_output_cov is not "
+                                      "implemented for sample_conditional.")
+		N, D = tf.shape(mean)[0], tf.shape(mean)[1]
+		S = num_samples
+		jittermat = settings.numerics.jitter_level * tf.eye(N, batch_shape=[D], dtype=settings.float_type)  # [D, N, N]
+		L = tf.cholesky(var + jittermat)  # [D, N, N]
+		nu = tf.random_normal([D, N, S], dtype=float_type)  # [D, N, S]
```
hughsalimbeni: you can reuse the `_mvn_sample` code here with two transposes (on the mean). Also, it would be great if all this code was negatively indexed so supported arbitrary leading dim

st--: Yeah, this is by no means meant to be done! I just put together the minimal amount of code to support what we need in gpflux... it'd also be good for the sample_mvn to handle num_samples, etc. Feel free to push to this branch!

markvdw:
Beautiful.

markvdw:
But I do think a few lines in the diff need testing (take a look at codecov/patch).

gpflow/conditionals.py
```diff
@@ -140,30 +140,25 @@ def _sample_conditional(Xnew, feat, kern, f, *, full_cov=False, full_output_cov=
...
-        S = num_samples
-        jittermat = settings.numerics.jitter_level * tf.eye(N, batch_shape=[D], dtype=settings.float_type)  # [D, N, N]
-        L = tf.cholesky(var + jittermat)  # [D, N, N]
-        nu = tf.random_normal([D, N, S], dtype=float_type)  # [D, N, S]
-        samples = mean[..., None] + tf.transpose(tf.matmul(L, nu), [1, 0, 2])  # [N, D, S]
+        # mean: N x P
+        # var: P x N x N
+        mean = tf.transpose(mean, [0, 1])  # now P x N
+        samples = _sample_mvn(mean, var, 'full', num_samples=num_samples)  # (S x) P x N
+        samples = tf.transpose(samples, [-1, -2])  # now (S x) N x P
```
hughsalimbeni: `matrix_transpose` here will do

st--: thanks... the first one (two lines above) was also wrong :/ fixed now.

gpflow/conditionals.py
```diff
@@ -364,17 +359,29 @@ def _sample_mvn(mean, cov, cov_structure):
     - "full": cov holds the full covariance matrix (without jitter)
     :return: sample from the MVN of shape N x D
     """
-    eps = tf.random_normal(tf.shape(mean), dtype=settings.float_type)  # N x P
+    rank_mean = len(mean.get_shape())
+    rank_cov = len(cov.get_shape())
+    N, D = tf.shape(mean)[0], tf.shape(mean)[1]
```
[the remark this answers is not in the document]
st--: We could have a broadcasting version for num_samples=None and an [S, N, P] version for num_samples=S being an integer...? could be a separate PR if you actually need that functionality though.

st--:
@hughsalimbeni @markvdw I've added the full_cov and num_samples arguments to the sample_conditional for tensor instead of feature instance, and made the tests cover everything. Hopefully now all tests should pass; please merge if you're happy now!

st--:
@hughsalimbeni in case you didn't see the comment, _sample_mvn / sample_conditional could be made to broadcast over the first dimensions for num_samples=None if you need that anywhere.

gpflow/conditionals.py
```diff
@@ -344,17 +374,29 @@ def _sample_mvn(mean, cov, cov_structure):
     - "full": cov holds the full covariance matrix (without jitter)
     :return: sample from the MVN of shape N x D
     """
-    eps = tf.random_normal(tf.shape(mean), dtype=settings.float_type)  # N x P
+    rank_mean = len(mean.get_shape())
```
awav: These two lines will not work for unknown shapes. Requires changes to use `tf.equal_assert(..)`.

st--: Isn't tensor.get_shape().ndim a property of the static graph? Is the rank of a tensor (the length of the shape) not known at graph-build time?

gpflow/conditionals.py
```diff
@@ -129,31 +130,60 @@ def _conditional(Xnew, X, kern, f, *, full_cov=False, q_sqrt=None, white=False):
 # ----------------------------------------------------------------------------
 
 
-@sample_conditional.register(object, InducingFeature, Kernel, object)
+@sample_conditional.register(TensorArray, InducingFeature, Kernel, TensorArray)
 @name_scope("sample_conditional")
-def _sample_conditional(Xnew, feat, kern, f, *, full_output_cov=False, q_sqrt=None, white=False):
+def _sample_conditional(Xnew, feat, kern, f, *, full_cov=False, full_output_cov=False, q_sqrt=None, white=False, num_samples=None):
```
awav: Wouldn't it better to introduce `cov_structure` with different options, `full_cov, full_output_cov`, instead of introducing more and more parameters?

gpflow/conditionals.py
```diff
@@ -22,6 +22,7 @@
 from .features import Kuu, Kuf, InducingPoints, InducingFeature
 from .kernels import Kernel, Combination
 from .probability_distributions import Gaussian
+from .types import TensorArray
```
awav: This should be done in single PR everywhere, not only this place.

gpflow/conditionals.py
st--: that should be a separate PR :stuck_out_tongue:

Issue #839, opened 2018-08-31
slice active_dims


Joshuaalbert:
In kernels, in the case of active_dims is a slice, the following happens:

``` bash
elif isinstance(active_dims, slice):
            self.active_dims = active_dims
            if active_dims.start is not None and active_dims.stop is not None and active_dims.step is not None:
>               assert len(range(*active_dims)) == input_dim  # pragma: no cover
E               TypeError: type object argument after * must be an iterable, not slice
```

I'll submit a PR for it I guess.

## MVCE
``` python 
import gpflow
gpflow.kernels.RBF(3,active_dims=slice(0,3,1))
```

PR #840, opened 2018-08-31
fix active_dims=slice

Joshuaalbert:
This is a fix for #839 which is a bug when active dims is a slice and the start, stop and step are not None.


st--:
@Joshuaalbert I'm happy to merge this (into develop, not master, as master should now point to the latest actual release whereas develop is now the place to add new ... developments:)), but would you be able to add a regression test? I.e., a test that would have failed before merging this PR, and that would no longer fail after merging this PR. Let me know if you've got any questions about that. Thanks!

Joshuaalbert:
@st-- done

PR #856, opened 2018-09-21
Fixing the problem with monitoring ScipyOptimizer

ahsimb:
The ScipyOptimizer works with its own copies of optimised variables. To see original variables changing in TensorBoard one has to explicitly write the current optimiser's state into them. The Monitor now does that. The user just have to provide the optimiser's object in the monitor initialization.

Also added an optional context parameter in case user wants to share the context between multiple optimisation sessions.


st--:
Awesome, thank you!! One note- we've now got a separate "develop" branch into which PRs should get merged ("master" should be equivalent to the latest actual release). Could you please change the target of this PR (and make sure there are no merge conflicts with develop)?

markvdw:
I'm happy with this to be merged. @ahsimb, did you manage to look at the issue of keeping track of what iteration a previous run of ScipyOptimizer left off at? Maybe an example in a notebook would be helpful as well.

ahsimb:
Yes, very simple solution. The user can create the MonitorContext and provide it to the Monitor instead of letting the monitor creating it. I should add some comments about this into the notebook.

st--:
Really nice, much cleaner than the hacks I had! Thank you :)

PR #871, opened 2018-10-20
GPflow 2.0

awav:
Hello Everyone!

TensorFlow 2.0 is coming and will bring important changes. In short, TensorFlow will have _function_ style execution by making Eager mode default, details are [here](https://github.com/tensorflow/community). This is very exciting news, as with "dynamic-like" graph creation users will have more flexibility and be able to do more (e.g. proper `for` loops and conditions). I started `awav/gpflow-2.0` branch as an experimental solution and eager mode playground, but eventually `awav/gpflow-2.0` will be merged into the GPflow `master`. GPflow 2.0 branch is expected to be unstable for now.

*Any feedback and suggestions are welcome!*

Plans of what we would like to do:

* At first, integrate GPflow with `tensorflow_probabilities`.
* We are considering to support both current TensorFlow static graph execution and Eager modes. It is reasonable, but implies difficulties in implementation.
* Unify saver interface with possibility to use TensorFlow (Keras) saver.
* Support Keras monitoring as well.

What `awav/gpflow-2.0` provides now:

* Project layout is cleaned and "optimized".
* Autoflow methods are removed, there is no `build_*` functions.
* `@params_as_tensors` are removed. There is no more such things as `tf.Tensor` which you can run via `session.run` and Parameter class is actually inherited from Eager Variable.
* TensorFlow Bijectors are used as Parameter transforms.
* TensorFlow distributions are used as Parameter priors.
* `Parameterized` is renamed to `Module`

There is a prototype [example](https://github.com/GPflow/GPflow/blob/awav/gpflow-2.0/prototype_example.py). For now only GPR and SVGP models are ported. In that file you will find rough example how to use SVGP and what you can do with GPflow 2.0.

Thank you!

### DIFF
--- a/tests/test_coregion.py
+++ b/tests/test_coregion.py
@@ -21,9 +21,10 @@ rng = np.random.RandomState(0)
 
 
 class Datum:
-    X = [rng.rand(10, 2) * 10, rng.rand(20, 2) * 10]
-    Y = [np.sin(x) + 0.9 * np.cos(x * 1.6) + rng.randn(*x.shape) * 0.8 for x in X]
-    label = [np.zeros((10, 1)), np.ones((20, 1))]
+    N1, N2 = 6, 12
+    X = [rng.rand(N1, 2) * 1, rng.rand(N2, 2) * 1]
+    Y = [np.sin(x[:, :1]) + 0.9 * np.cos(x[:, 1:2] * 1.6) + rng.randn(x.shape[0], 1) * 0.8 for x in X]
+    label = [np.zeros((N1, 1)), np.ones((N2, 1))]
     perm = list(range(30))
     rng.shuffle(perm)
     Xtest = rng.rand(10, 2) * 10
@@ -50,14 +51,15 @@ def _prepare_models():
     vgp0 = gpflow.models.VGP((Datum.X[0], Datum.Y[0]),
                              kernel=k0,
                              mean_function=Constant(),
-                             likelihood=gpflow.likelihoods.Gaussian(), num_latent=2)
+                             likelihood=gpflow.likelihoods.Gaussian(), num_latent=1)
     vgp1 = gpflow.models.VGP((Datum.X[1], Datum.Y[1]),
                              kernel=k1,
                              mean_function=Constant(),
-                             likelihood=gpflow.likelihoods.Gaussian(), num_latent=2)
+                             likelihood=gpflow.likelihoods.Gaussian(), num_latent=1)
     # 2. Coregionalized GPR
-    kc = gpflow.kernels.SquaredExponential()
-    kc.trainable = False  # lengthscale and variance is fixed.
+    kc = gpflow.kernels.SquaredExponential(active_dims=[0, 1])
+    kc.lengthscale.trainable = False
+    kc.variance.trainable = False  # variance is handles by the coregion kernel
     coreg = gpflow.kernels.Coregion(output_dim=2, rank=1, active_dims=[2])
     coreg.W.trainable = False
     lik = gpflow.likelihoods.SwitchedLikelihood([gpflow.likelihoods.Gaussian(),
@@ -69,7 +71,7 @@ def _prepare_models():
                              kernel=kc * coreg,
                              mean_function=mean_c,
                              likelihood=lik,
-                             num_latent=2
+                             num_latent=1
                              )
 
     # Train them for a small number of iterations
@@ -137,28 +139,29 @@ def test_predict_f():
     assert_allclose(pred_f1, pred_fc1, atol=1.0e-2)
 
     # check predict_f_full_cov
-    vgp0.predict_f_full_cov(Datum.Xtest)
-    cvgp.predict_f_full_cov(Datum.X_augumented0)
-    vgp1.predict_f_full_cov(Datum.Xtest)
-    cvgp.predict_f_full_cov(Datum.X_augumented1)
+    vgp0.predict_f(Datum.Xtest, full_cov=True)
+    cvgp.predict_f(Datum.X_augumented0, full_cov=True)
+    vgp1.predict_f(Datum.Xtest, full_cov=True)
+    cvgp.predict_f(Datum.X_augumented1, full_cov=True)
 
 
 def test_predict_y():
     vgp0, vgp1, cvgp = _prepare_models()
-    pred_y0 = vgp0.predict_y(Datum.Xtest)
-    pred_yc0 = cvgp.predict_y(
+    mu1, var1 = vgp0.predict_y(Datum.Xtest)
+    c_mu1, c_var1 = cvgp.predict_y(
         np.hstack([Datum.Xtest, np.zeros((Datum.Xtest.shape[0], 1))]))
 
-    # predict_y returns results for all the likelihodds in multi_likelihood
-    assert_allclose(pred_y0[0], pred_yc0[0][:, :np.array(Datum.Ytest).shape[1]], atol=1.0e-2)
-    assert_allclose(pred_y0[1], pred_yc0[1][:, :np.array(Datum.Ytest).shape[1]], atol=1.0e-2)
-    pred_y1 = vgp1.predict_y(Datum.Xtest)
-    pred_yc1 = cvgp.predict_y(
+    # predict_y returns results for all the likelihoods in multi_likelihood
+    assert_allclose(mu1, c_mu1[:, :1], atol=1.0e-5)
+    assert_allclose(var1, c_var1[:, :1], atol=1.0e-5)
+
+    mu2, var2 = vgp1.predict_y(Datum.Xtest)
+    c_mu2, c_var2 = cvgp.predict_y(
         np.hstack([Datum.Xtest, np.ones((Datum.Xtest.shape[0], 1))]))
 
-    # predict_y returns results for all the likelihodds in multi_likelihood
-    assert_allclose(pred_y1[0], pred_yc1[0][:, np.array(Datum.Ytest).shape[1]:], atol=1.0e-2)
-    assert_allclose(pred_y1[1], pred_yc1[1][:, np.array(Datum.Ytest).shape[1]:], atol=1.0e-2)
+    # predict_y returns results for all the likelihoods in multi_likelihood
+    assert_allclose(mu2, c_mu2[:, 1:2], atol=1.0e-5)
+    assert_allclose(var2, c_var2[:, 1:2], atol=1.0e-5)
 
 
 def test_predict_log_density():


alexggmatthews:
Hello @awav . Let's organize a chat over hangouts next week.

alexggmatthews:
Hi @awav , my two main thoughts after talking are:
1) Have we done any speed benchmarking / profiling? If we go for an eager heavy incarnation what slow down are we expecting across tasks we care about?
2) Relatedly. We should think about how easy it is to have code that is largely agnostic to eager/ vs graph mode. What code complexity do we pay for gains in being agnostic in this way? Is incorporating Keras a good approach for this? 
I'll follow up with you and James towards the end of this week or the beginning of next week.

mrksr:
Hello everyone!
The last few days I have been playing around with this branch and the newish `tf.estimator` API, trying to see how hard it would be to use GPflow in that setting. I am happy to say that things work fairly straightforwardly and almost all the work around these APIs can be used to spawn graphs, train and evaluate models and save checkpoints. At least on my end this makes a considerable amount of code obsolte which is always a good thing.

Now for the almost part: I love the way gpflow is moving in removing most of the magic from the library and relying more on tensorflow primitives. I think the fact that models no longer handle their own data is a major step in the right direction. However, one major magic remains: `Parameters`. I think they a re super convenient and should stay. However, the way they are implemented right now as inheriting from `tfe.Variable` adds a major drawback: It is no longer straightforward to use gpflow in a distributed tensorflow setting. I do not really know what the concrete problem is, I can only say that the `tf.estimator` API and specifically eager variables seem to not play nice together, meaning that clusters are never used even if specified. This also happens if I remove the `tf.enable_eager_execution()` call in `__init__.py` (which I am sure is only there for debugging purposes anyway).

To get things to work, I wrote a small replacement implementation for `Parameter` which is a bit more agnostic. The code is not very good but it should give you an idea:
```python
class Parameter:
    def __init__(
            self,
            data=None,
            unconstrained_tensor=None,
            dtype=None,
            name=None,
            prior=None,
            transform=None,
            trainable=True,
    ):
        if data is None and unconstrained_tensor is None:
            raise ValueError('You need to pass either initial data or an unconstrained tensor.')

        if tf.contrib.framework.is_tensor(data):
            unconstrained_tensor = data
            transform = None
            trainable = False

        if name is None:
            name = 'unnamed'

        if transform is None:
            transform = tfp.bijectors.Identity()
        self.transform = transform

        self.prior = prior

        if unconstrained_tensor is None:
            initial_value = self.transform.inverse(tf.constant(data, dtype=dtype))

            unconstrained_tensor = tf.get_variable(
                name,
                dtype=dtype,
                initializer=initial_value,
            )
        self.unconstrained_tensor = unconstrained_tensor
        self.constrained_tensor = self.transform.forward(unconstrained_tensor, name=f'{name}_constrained')

        self.trainable = trainable

    @property
    def constrained(self):
        return self.constrained_tensor

    def __call__(self):
        return self.constrained

    def log_prior(self):
        if self.prior is None:
            return tf.constant(0., dtype=self.unconstrained_tensor.dtype)

        log_prob = self.prior.log_prob(self.constrained_tensor)
        log_det_jacobian = bijector.forward_log_det_jacobian(
            self.unconstrained_tensor,
            self.unconstrained_tensor.shape.ndims,
        )
        return log_prob + log_det_jacobian
```
The main difference to what is currently done is to use `tf.get_variable` to initialize an unconstrained tensor if neccesary and not to tuch the data very much otherwise. I use `tf.get_variable` as it is the recommended way to create variables in the documentation for `estimator` and `distributed`. There is the plus that it is quite agnostic about the current context, but there is the drawback that you need a sensible way to name variables. So I added a `name` parameter and instanciate the model using `variable_scopes` to mimic the behaviour of current `master` which generates "subpaths" for variables. I think it should be quite possible to find an elegant solution for that.

After a few fixups to `Module` (and adding `name` and `dtype` to the parameters where needed) I can now use gpflow with `tf.estimator` in a distributed setting. I am not sure if this breaks eager mode (probably?), but I hope it should not be too hard to make this compatible. I think this implementation is fairly close (albeit much simpler) to how other libraries in the tensorflow universe handle parameters, so it might be a bit less surprising than the current way of inheriting from `tfe.Variable`.

I think all of this is very close to the discussion in #897, where @bodin-e makes very good points about compatibility of `gpflow` to the rest of the tensorflow ecosystem. With this branch I think we are almost there and if we manage to handle `Parameter` the right way, hopefully eager mode, the library usecase and the distributed usecase are all easy.

What do you think?

awav:
Hello @mrksr,

Thanks for testing GPflow-2.0 and your valuable feedback!

Before I address your questions and issues which you are having with gpflow-2.0, I would like to clarify the purpose of gpflow2.0 branch.

TensorFlow 2.0 (TF2.0) will replace 1.0 version very soon. The gpflow-2.0 branch is a transparent way of showing to GPflow community what will change and spot what must be changed on early stages. You can look at [gpflow 2.0 preview notebook](https://github.com/GPflow/GPflow/blob/awav/gpflow-2.0/doc/source/notebooks/gpflow-2.0-preview.ipynb) to get an idea how it will look like when TF2.0 will be released.

* >  This also happens if I remove the tf.enable_eager_execution() call in __init__.py (which I am sure is only there for debugging purposes anyway).

  The `tf.enable_eager_execution()` appears in `__init__.py` not with testing intentions, but to emulate TF2.0 behaviour right from the start. In short, TF2.0 will work with eager by default, [here are details](https://medium.com/tensorflow/whats-coming-in-tensorflow-2-0-d3663832e9b8).

* > I use tf.get_variable as it is the recommended way to create variables in the documentation for estimator and distributed.

  > So I added a name parameter and instantiate the model using variable_scopes...

  The [RFC](https://github.com/tensorflow/community/blob/master/rfcs/20180817-variables-20.md) describes what the variable does mean in TF2.0. There will not be `tf.get_variable(...)` function anymore. The same fate awaits `variable_scope` and that's great!

* >  I can only say that the tf.estimator API and specifically eager variables seem to not play nice together. 

  > After a few fixups to Module (and adding name and dtype to the parameters where needed) ...

  I'm not sure _where_ exactly the problem is. My hypothesis is that the current version of estimator (assuming that you used TF1.12) doesn't recognise parameter as a subclass of variable. I can't tell more without actual error output. Could you share the `diff` with your changes made in the gpflow's module? Regardless of that issue, the estimator stays in TF2.0 and it is supposed to work with future variables. Also, subclassing from [TF2.0 variable](https://github.com/tensorflow/community/blob/master/rfcs/20180817-variables-20.md) is a natural way for the parameter class implementation.
  Moreover, the module implementation will be provided by [TF2.0](https://github.com/tensorflow/community/pull/56) as well, which resembles GPflow parametrised class.

You can find TF2.0 preview pip package [here](https://pypi.org/project/tf-nightly-2.0-preview/). Full details of upcoming changes are in [RFCs](https://github.com/tensorflow/community). 

Kind regards,
Artem Artemev

mrksr:
Thanks for the pointers to the tensorflow documents and some more insight into your reasoning, @awav! It is very possible that the problems I ran into just stem from the fact that `tf.estimator` and eager mode are note quite compatible yet in tensorflow 1.12. I tried to write my code following the documentation for the estimators and there, they use `get_variable` and `variable_scope` everywhere. Eager is (as far as I can see) not mentioned there in any way yet, so I had to guess quite a bit along the way.

I will read through the documents and rerun my code with the preview packages, let's see what happens :). It was hard for me to really pinpoint the problem with eager variables since depending on the small implementation details problems would arise in either non-distributed mode or distributed mode. And sometimes distributed mode would just silently "fail" and fall back non-distributed training. So I decided against spending too much time there and did the parameter implementation I showed above. If the problem persists with tensorflow 2 (preview) I will try and provide a minimal example as it will be much more interesting to debug.

mrksr:
I read through the documents you linked and some more of the tensorflow source code and I definitely agree that it is no use to think about incompatibilities in tensorflow 1.12 code. Things will change much too drastically.

That being said, I think my previous feedback boils down to one question: Given that gpflow parameters will inherit from `tf.Variable`, can a parameter still only be a proxy for an arbitrary tensor augmented with a prior and maybe a transform? In other words, will there be a way to use gpflow in a mode which does not create new variables at all or which does so using a user-supplied function?

I think that's a feature worth thinking about. In my experience, most tensorflow libraries are quite opinionated when it comes to constructing graphs, generating variables and deriving things like loss functions or optimizers. That's fine as long as you do not mix libraries, then things tend to become complicated. And I think one of the major strenghts of `gpflow-2.0` when compared to current master is a massive increase in interoperability. I love that direction gpflow is taking.

---

Another point: Seeing as gpflow-2.0 is quite compatible to `tf.estimator`, have you considered getting rid of `gpflow.training.optimize`?

awav:
@mrksr,

> Given that gpflow parameters will inherit from `tf.Variable`, can a parameter still only be a proxy for an arbitrary tensor augmented with a prior and maybe a transform?

Parameter will be not be a "proxy", it will be extension of the variable class with prior and transformation capabilities. In turn, the variable is simply a tensor with `trainable` property. 

> In other words, will there be a way to use gpflow in a mode which does not create new variables at all or which does so using a user-supplied function?

What do you mean by user-supplied function? Passing a function result as initial value to a parameter or something else? Can you provide a use case? E.g. in TF2.0 you can do this (same thing is valid for parameter class):

```python
In [3]: v = tf.Variable(tf.random.normal((3, 3)))
```
> have you considered getting rid of `gpflow.training.optimize`?

It is already removed in gpflow2.0 branch

mrksr:
> Parameter will be not be a "proxy", it will be an extension of the variable class with prior and transformation capabilities. In turn, the variable is simply a tensor with `trainable` property.

I understand that the new implementation will not be "proxy-based", which is why I am wondering if it can still be used that way. A concrete example is that I want to use, say, an instance of `RBF` to calculate some kernel matrices. But for some reason, the `lengthscales` I would like to use should not be a variable but rather the result of some calculation, i.e. an arbitrary tensor. Importantly, that would not just be an initial value but rather the way the `lengthscales` should be calculated, say inside of a `tf.function`.

To me, the implementation of `RBF` consists of two distinct parts: The functional part which is side-effect free and calculates (say) kernel matrices given inputs and hyper-parameters. And the stateful part which keeps track of the current values of the hyper-parameters (i.e. the `Parameter` instances). In current master, these parts are heavily conflated as the parameters are accessed directly in the calculations (via `self.lengthscales`). Thus, to access the functional implementation, one has to "bypass" the `Parameters` by passing tensors at construction time and triggering special behavior in `Parameter` which disables the statefulness. I do not think this is possible with in awav/gpflow-2.0 right now, as the lengthscales are always variables. I would agree that this proxy-approach is not very elegant and should not be needed. However, I think the functional implementation of `RBF` should still be exposed and be usable without the stateful part.

A minor note on transforms and priors: Since the output of my neural network might be unconstrained and gpflow keeps track of proper transforms and priors for parameters anyway, it would be good to be able to rely on that part of the code as well. But I guess having to specify those manually is not very problematic.

---

> What do you mean by user-supplied function? Passing a function result as initial value to a parameter or something else? Can you provide a use case?

A user-supplied `add_variable_fn` is how this problem is handled in other libraries, for example in [tensorflow_probability](https://www.tensorflow.org/probability/api_docs/python/tfp/layers/default_multivariate_normal_fn). This allows the user to specify how new variables should be created if they are needed and can be used to add glue-code between different libraries. Of cause, this is a Tensorflow 1 implementation and variable creation behavior will change significantly, so this approach might get deprecated along with `get_variable`. Do you know how tensorflow probability will change for Tensorflow 2 in that regard?

---

> It is already removed in gpflow2.0 branch

Hm, is the branch on here out of date maybe? The module seems to [exist here](https://github.com/GPflow/GPflow/blob/5c6b240df8c395198358ca47fb1f3bc654be75c2/gpflow/training/optimize.py). It is a bit simpler though.

bodin-e:
Great work with this branch, like of plenty of things I see in here. 

On the topic of Kernels and user-provided kernel parameters: use-cases would be approximate hyperparameter marginalisation via MCMC (e.g. the variance and lengthscale of an RBF being samples), the kernel parameters being the output of another model or as Markus mentioned, calculate matrices using fixed parameters without the need to initialize variables. 

All this would be very easy (just using standard tensorflow constructs) if Kernels were tensor in - tensor out. My suggestion is to simply allow the user to pass in tensors as parameters:
```python
RBF(
  variance=variance,  # tensor (or Parameter)
  lengthscale=lengthscale # tensor (or Parameter)
``` 
If automatic Parameter instantiation is needed, for less verbosity for a common use-case, a flag could be used to indicate if to instantiate Parameters from the provided arguments:    
```python
RBF(
  variance=variance,  # tensor (or Parameter)
  lengthscale=lengthscale # tensor (or Parameter),
  instantiate_params=False
``` 
where if instantiate_params is true it wraps the arguments in a Parameter (if not already a Parameter) and the default being True, i.e.
RBF(
  variance=1,
  lengthscale=1
``` 
gives the current behavior.

**Comment by awav** [comment]
@bodin-e , @mrksr 

Just to make sure that we are on the same page, here is a minor clarification about TF2.0.
In TF2.0 you can't postpone execution of tensor. It is still part of the graph, but you are getting result immediately after calling an operation. E.g. when you call `tf.random.normal((2, 2))`, the sample will be available right after. It means there is no intermediate graph construction stage.

```python
In [3]: tf.random.normal((2, 2))
Out[3]:
<tf.Tensor: id=12, shape=(2, 2), dtype=float32, numpy=
array([[ 0.29597756,  0.88601315],
       [ 0.73969084, -0.22102852]], dtype=float32)>
```

TensorFlow variable is always root node in the computation graph. For the reason that TF1.0 doesn't hide graph building from the user and you actually manipulate operations (functions) though standard pythonic objects (tensors), it is possible to interchange variables and functions in TF1.0. @bodin-e, It means that you can pass the computation tensor (the result of any function) as an input to the GPflow 1.0 parameter, considering that there are other hacks, like `params_as_tensor`, helping to convert a proxy parameter object to a tensor when it is needed.

TF2.0 replaces heavy session/graph things with lightweight well familiar pythonic functions. The variable class becomes true root node which is not replaceable by a function or a result of it.

Inheriting from tensorflow variables gives huge advantage: the user would be able to pass gpflow parameters to standard tensorflow functions without any hassle. In turn, it means that gpflow can get rid of the `param_as_tensor` decorator and other magic functions:

```python
p = gpflow.Parameter(tf.random.uniform((5, 5)))
p_2 = tf.pow(p, 2)  # Done without converting parameter into a tensor
```

I would prefer to keep parameter class simple, coincise and follow [single responsibility principle](https://en.wikipedia.org/wiki/Single_responsibility_principle). The "beast" classes (which can do everything or many things) are prone to bugs, hardly maintainable and scale confusion.

It does make sense what you are proposing, but can be implemented in slightly different way.

* Let's say you want to change RBF kernel replacing lengthscales with a result of a function. You could create an CustomRBF object, inheriting from RBF and overload the variable property, the behavior of which you want to change:

```python
class CustomRBF(gpflow.kernels.RBF):
    def __init__(self, lengthscales, *args, **kwargs):
        super().__init__(*args, **kwargs)
        self._lengthscales_fun = lengthscales

    @property
    def lengthscales(self):
        return self._lengthscales_fun()
```


* I believe that it is doable (not 100% sure) to make another object which will wrap function call and do what you want while pretending to be a "variable", so it would look like this:

```python
def function():
    return tf.random.normal((2, 2)) + 10.

f = gpflow.FunctionWrapper(function)
f_2 = tf.pow(fun, 2)  # f calls `function()` somewhere inside somehow...
kernel = gpflow.kernels.RBF(variance=f_2)
kernel.lengthscales = f
```

st--:
Reworking gpflow to work with tf2-style eager mode means we'll have to break a lot of things anyways - so this is a good opportunity to think about what else we might want to improve in terms of how gpflow works/how we use it, e.g. separate the model from the data instead of passing in X and Y at model construction as can be seen in this PR - use tf probability distributions for q(u) (no more manual passing around of `q_mu` and `q_sqrt`)? separate approximation and inference (`GP()` objects to pass around)? - if you've got any thoughts/ideas along those lines it'd be great to hear them! in that case, could you perhaps prepare a brief (single page) slide/document that describes how you would envision a "better gpflow interface" (e.g. contrast old/current vs new/gpflow2 behaviour/usage) and link that from here? The majority of GPflow maintainers will meet up to discuss Thursday next week, so ideally we'd be able to see your thoughts by then (sorry for the somewhat short notice!), though we'd appreciate your thoughts one way or another (as always, no guarantees of implementing them straight away or at all, but constructive feedback is always useful and appreciated!) :-)

bodin-e:
Nice work with this branch.

Had a chat with @mrksr and we are thinking something on the lines of the following concerning kernels:
https://www.overleaf.com/read/jvsbztcfvjvb

Our main point is to separate the kernel instance parameters from the kernel functionality, in order to:
- make kernels easy to use also when the parameters are outputs from a different model, are sampled from a distribution etc (tensors in - tensors out).
- make kernel functionality usable also without kernel instance creation, for when that might make more sense, such as for one-off usage.

The separation between functionality that can be pure (staticmethod / function) and the functionality relying on object instance members has clearly increased in gpflow in the last year or so all over the library (with great benefit). We argue that similar benefits can be had in the kernel package.

mrksr:
> Let's say you want to change RBF kernel replacing lengthscales with a result of a function. You could create an CustomRBF object, inheriting from RBF and overload the variable property, the behavior of which you want to change:
> 
> ```python
> class CustomRBF(gpflow.kernels.RBF):
>     def __init__(self, lengthscales, *args, **kwargs):
>         super().__init__(*args, **kwargs)
>         self._lengthscales_fun = lengthscales
> 
>     @property
>     def lengthscales(self):
>         return self._lengthscales_fun()
> ```

One small additional clarification: I agree that this `CustomRBF`-construction works in principle. However, this implementation still implies that a variable will be created by the original initializer (and suspiciously be called `<some_kernel_prefix>.lengthscales`) although it will not be used. To me this is a code smell and can lead to confusion and problems (for example with checkpointing) later on. However, what we are proposing I think is super close to the `CustomRBF`-proposition, we just would like to see a proper separation of the static and stateful code-parts.

bodin-e:
> > Let's say you want to change RBF kernel replacing lengthscales with a result of a function. You could create an CustomRBF object, inheriting from RBF and overload the variable property, the behavior of which you want to change:
> > ```python
> > class CustomRBF(gpflow.kernels.RBF):
> >     def __init__(self, lengthscales, *args, **kwargs):
> >         super().__init__(*args, **kwargs)
> >         self._lengthscales_fun = lengthscales
> > 
> >     @property
> >     def lengthscales(self):
> >         return self._lengthscales_fun()
> > ```
> 
> One small additional clarification: I agree that this `CustomRBF`-construction works in principle. However, this implementation still implies that a variable will be created by the original initializer (and significantly be called `<some_kernel_prefix>.lengthscales`) although it will not be used. To me this is a code smell and can lead to confusion and problems (for example with checkpointing) later on. However, what we are proposing I think is super close to the `CustomRBF`-proposition, we just would like to see a proper separation of the static and stateful code-parts.

Not to mention that you then need to make a custom kernel class for every kernel you want to use.

awav:
@bodin-e , @mrksr this solution will work. No need in creating custom kernels, just replace parameters with your function wrapped within another object.

```
def function():
    return tf.random.normal((2, 2)) + 10.

f = gpflow.FunctionWrapper(function)
f_2 = tf.pow(f, 2)  # f calls `function()` somewhere inside somehow...
kernel = gpflow.kernels.RBF(variance=f_2)
kernel.lengthscales = f
```

Proposed solution (in overleaf) breaks integrity of the kernel class and calls from other parts of the code (`kuf`, `kuu`) to kernels will require knowledge about implementation details, as parameters vary from kernel to kernel.

bodin-e:
> @bodin-e , @mrksr this solution will work. No need in creating custom kernels, just replace parameters with your function wrapped within another object.
> 
> ```
> def function():
>     return tf.random.normal((2, 2)) + 10.
> 
> f = gpflow.FunctionWrapper(function)
> f_2 = tf.pow(f, 2)  # f calls `function()` somewhere inside somehow...
> kernel = gpflow.kernels.RBF(variance=f_2)
> kernel.lengthscales = f
> ```
> Proposed solution (in overleaf) breaks integrity of the kernel class and calls from other parts of the code (`kuf`, `kuu`) to kernels will require knowledge about implementation details, as parameters vary from kernel to kernel.

That looks similar to my current solution when I'm using gpflow kernels, where I do things like that inside a function I send e.g. gpflow kernels inside a lambda. However, ideally things like that wouldn't be needed.

In the proposed idea, the intention is that K and K_diag would not be used anymore other than for usage of the pure kernel functionality. The invariance to implementation details of each kernel you get by using __call__  instead (which would be the one to call within for e.g. the conditionals etc).

bodin-e:
As a bonus, if conditionals etc use the __call__ interface rather than K and K_diag, such functionality would also be usable with a function of the same call signature in place of a kernel instance.   
```python
def custom_kernel(X, Y=None, presliced=False, full=True):
  ...
  if full:
    return ...
  return ...
```
some_utility_like_condtionals(kernel=custom_kernel, ...)

mrksr:
> @bodin-e , @mrksr this solution will work. No need in creating custom kernels, just replace parameters with your function wrapped within another object.
> 
> ```
> def function():
>     return tf.random.normal((2, 2)) + 10.
> 
> f = gpflow.FunctionWrapper(function)
> f_2 = tf.pow(f, 2)  # f calls `function()` somewhere inside somehow...
> kernel = gpflow.kernels.RBF(variance=f_2)
> kernel.lengthscales = f
> ```
> 
> Proposed solution (in overleaf) breaks integrity of the kernel class and calls from other parts of the code (`kuf`, `kuu`) to kernels will require knowledge about implementation details, as parameters vary from kernel to kernel.

I do not think the proposesed solution is a final one, it just showcases the separation in static and stateful code. I think we might be talking past each other a bit.

Concerning the code-piece you showed above: Will `kerel.lengthscale` and `kernel.variance` be instances of `Variable` in this case, even if they are not needed? Could I calculte a kernel matrix for an `RBF`-Kernel without ever creating any variable? Would you agree that this is desireable? Also, can I use `Kernel`-instances without any variables inside the GPflow library?

bodin-e:
Thank you for the hangouts discussion @awav @mrksr last week concerning kernels.

Markus and I have written down the more elaborated suggestion together with some code examples, see the attached pdf.

Happy to discuss things further and/or answer any questions relating to the suggestions.

[gpflow_kernel_proposal.pdf](https://github.com/GPflow/GPflow/files/2988391/gpflow_kernel_proposal.pdf)

easysoft2k15:
Hi,
I'm trying to use *gpflow 2.0* with *Tensorflow 2.0*.
I'm using the *prototype_example.py* as a guide line but looks like the following method is not available:

`gpflow.optimize(loss_cb, adam, m.trainable_variables, 10)`

1.  What's the correct way to optimize hyper parameters right now?
2.  What is the timing to bring gpflow 2.0  into master?

Thank You

st--:
@awav how does this PR relate to the `develop-2.0` branch we've got now ? I expect there won't be any further commits to this branch? Should we close this PR and re-open one from develop-2.0 -> develop? Or what do you suggest?

st--:
(Urgh, the github "let's just interleave all commits and notifications and actual comments in one place" means it's really hard to actually read through the design discussion. Maybe this would've been better off as an issue rather than a pull request!)

PR #877, opened 2018-11-09
Support TensorFlow 1.12

vdutor:
Hi all,

This PR makes GPflow stable and tested against the latest TF release 1.12.

## Changes:
- removed unused, old types: `np_float, tf_float, np_int, tf_int`
- resolved some of the warnings that were showing when running GPflow
- allowed for broadcasting in mean functions `Constant` and `Linear`


.travis.yml
```diff
@@ -8,8 +8,8 @@ before_install:
   - sudo apt-get update -qq
 
 install:
-  - pip install numpy scipy pandas pytest nbformat nbconvert jupyter_client jupyter matplotlib pytest-xdist pytest-cov codecov multipledispatch
-  - pip install https://storage.googleapis.com/tensorflow/linux/cpu/tensorflow-1.8.0-cp36-cp36m-linux_x86_64.whl
+  - pip install numpy scipy pandas pytest nbformat nbconvert jupyter_client jupyter matplotlib pytest-xdist pytest-cov codecov multipledispatch mock
```
awav: Why do you need mock?

gpflow/mean_functions.py
```diff
@@ -102,19 +111,21 @@ def A(self, A):
...
 
+
 class Constant(MeanFunction):
     """
     y_i = c,,
     """
     def __init__(self, c=None):
         MeanFunction.__init__(self)
         c = np.zeros(1) if c is None else c
+        c = np.reshape(c, (1, -1))
```
st--: Can you explain what is expected to happen if you pass in a vector-valued `c`?

gpflow/params/parameter.py
```diff
@@ -364,7 +364,9 @@ def _build(self):
         constrained = self._build_constrained(unconstrained)
         prior = self._build_prior(unconstrained, constrained)
 
-        self._is_initialized_tensor = tf.is_variable_initialized(unconstrained)
+        self._is_initialized_tensor = True
+        if not isinstance(unconstrained, tf.Tensor):
```
st--: could we add a comment explaining this change please?

gpflow/mean_functions.py
vdutor: it will get reshaped to a vector having shape `[1, D]`. We need this first dimension for correct broadcasting.

vdutor:
I reverted the number of cores to be used on Travis to `1`, so the tests pass and we can merge this branch and deal with the slow tests in a different PR.

tests/test_likelihoods.py
```diff
@@ -247,80 +249,76 @@ def test_pred_mean_and_var(self):
...
+    F, Y, feed = _prepare(dimF=2, dimY=1)
+    Fvar = tf.exp(tf.stack([F[:, 1], -10.0 + tf.zeros(tf.shape(F)[0], dtype=F.dtype)], axis=1))
+    F = tf.stack([F[:, 0], tf.zeros(tf.shape(F)[0], dtype=F.dtype)], axis=1)
+    Ylabel = 1 - Y  # We need the 1 - Y, as we need to pass the *label* to SoftMax
+
+    def logistic_link(x):
+        return 1.0 / (1.0 + tf.exp(-x))
+
+    ls = gpflow.likelihoods.SoftMax(2)
+    ls.num_monte_carlo_points = int(5e6)
```
awav:
```suggestion
    ls.num_monte_carlo_points = int(1e7)
```

PR #882, opened 2018-11-21
broadcasting conditionals

hughsalimbeni:
Currently conditionals can only accept shape `N, D` for `Xnew`. The PR allows conditional to accept `...,N,D`, bringing it in line with other tensorflow functions

### DIFF
--- a/gpflow/conditionals.py
+++ b/gpflow/conditionals.py
@@ -74,6 +74,25 @@ def _conditional(Xnew, feat, kern, f, *, full_cov=False, full_output_cov=False, 
     return fmean, _expand_independent_outputs(fvar, full_cov, full_output_cov)
 
 
+# def broadcast_rowwise_over_leading_dims(fn, M, output=-1):
+#     """
+#     Performs reshaping and transposing to get a function to broadcast over leading dimensions
+#
+#     :param fn: a function which takes [A, B] and returns [a,...,b]
+#     :param M: an array of shape [..., A, B]
+#     :return: an array of shape [..., a...b]  where the operation is [.x.x.A, B]
+#     """
+#     N, D = tf.shape(M)[-2], tf.shape(M)[-1]
+#     leading_dims = tf.shape(M)[:-2]
+#     leading_size = tf.reduce_prod(leading_dims)
+#     M_flattened = tf.reshape(M, [leading_size * N, D])
+#     ret_flattened = fn(M_flattened)
+#     flat_shape = tf.shape(ret_flattened)
+#     new_shape = tf.concat([leading_dims, flat_shape[:-1], [N]], 0)
+#     ret = tf.reshape(ret_flattened, new_shape)
+#     return ret
+
+
 @conditional.register(object, object, Kernel, object)
 @name_scope("conditional")
 def _conditional(Xnew, X, kern, f, *, full_cov=False, q_sqrt=None, white=False):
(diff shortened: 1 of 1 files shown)


gpflow/conditionals.py
```diff
@@ -153,8 +185,8 @@ def _sample_conditional(Xnew, feat, kern, f, *, full_cov=False, full_output_cov=
     if full_cov:
         # mean: N x P
         # cov: P x N x N
-        mean = tf.matrix_transpose(mean)  # now P x N
-        samples = _sample_mvn(mean, cov, 'full', num_samples=num_samples)  # (S x) P x N
+        mean_PN = tf.matrix_transpose(mean)  # now P x N
```
hughsalimbeni: This isn't really part of the PR, but it needed fixing

gpflow/conditionals.py
```diff
@@ -74,6 +74,25 @@ def _conditional(Xnew, feat, kern, f, *, full_cov=False, full_output_cov=False,
     return fmean, _expand_independent_outputs(fvar, full_cov, full_output_cov)
 
 
+# def broadcast_rowwise_over_leading_dims(fn, M, output=-1):
```
st--: don't add commented-out code ...

gpflow/conditionals.py
```diff
@@ -112,13 +131,26 @@ def _conditional(Xnew, X, kern, f, *, full_cov=False, q_sqrt=None, white=False):
...
-    num_data = tf.shape(X)[0]  # M
-    Kmm = kern.K(X) + tf.eye(num_data, dtype=settings.float_type) * settings.numerics.jitter_level
-    Kmn = kern.K(X, Xnew)
+    num_data = tf.shape(X)[-2]  # M
+    Kmm = kern.K(X) + tf.eye(num_data, dtype=settings.float_type) * settings.numerics.jitter_level  #  [M, M]
+
+    Ns, D = tf.shape(Xnew)[-2], tf.shape(Xnew)[-1]
+    leading_dims = tf.shape(Xnew)[:-2]  # [...]
+    leading_size = tf.reduce_prod(tf.concat([[1,], leading_dims], 0))
+    Xnew_flattened = tf.reshape(Xnew, [leading_size * Ns, D])  #[prod(...)*N,D]
```
st--:
```suggestion
    Xnew_flattened = tf.reshape(Xnew, [leading_size * Ns, D])  #[prod(...)*N, D]
```

gpflow/conditionals.py
```diff
@@ -112,13 +131,26 @@ def _conditional(Xnew, X, kern, f, *, full_cov=False, q_sqrt=None, white=False):
...
+    Knm_flattened = kern.K(Xnew_flattened, X)  # [prod(...)*N, M]
+    # flat_shape = tf.shape(Knm_flattened)
```
st--: remove?

gpflow/conditionals.py
```diff
@@ -182,13 +214,15 @@ def _sample_conditional(Xnew, X, kern, f, *, q_sqrt=None, white=False, full_cov=
     samples = _sample_mvn(mean, cov, cov_structure, num_samples=num_samples)
     if full_cov:
         samples = tf.matrix_transpose(samples)
+        mean = tf.matrix_transpose(mean)
```
st--: this adds an unnecessary matrix_transpose ... would it be worth generating mean_PN instead? (could be mean_PN = mean in one branch and mean_PN = tf.matrix_transpose(mean) in the other)

gpflow/conditionals.py
```diff
@@ -212,47 +246,62 @@ def base_conditional(Kmn, Kmm, Knn, f, *, full_cov=False, q_sqrt=None, white=Fal
...
-    Lm = tf.cholesky(Kmm)
+    N = tf.shape(Kmn)[-1]
+    M = tf.shape(f)[0]
+    leading_dims = tf.shape(Kmn)[:-2]
+    Lm = tf.cholesky(Kmm)  # [M,M]
 
     # Compute the projection matrix A
-    A = tf.matrix_triangular_solve(Lm, Kmn, lower=True)
-
+    # A = broadcast_rowwise_over_leading_dims(lambda x: tf.matrix_triangular_solve(Lm, x, lower=True), Kmn)
```
st--: remove?

tests/test_param.py
```diff
@@ -354,30 +354,30 @@ def test_floating_assign(self):
...
-            self.m.p = param
-
-            assert self.m.p is param
-            assert p.name.startswith('Parameter')
-            assert p.root is p
-
-            self.m.d = new_param
-            assert self.m.d is new_param
-            assert self.m.d.pathname == '{name}/d'.format(name=self.m.name)
+    # def test_create_and_replace(self):
```
st--: why did all these get commented out?

gpflow/conditionals.py
hughsalimbeni: good point. Will do

tests/test_param.py
hughsalimbeni: tensorflow1.12 breaks this test. I've commented out to see if the build passes for the moment. @awav might have an idea why this test is broken. I suspect it's connected with the new horrible error messages tensorflow gives about unused tensors (a gpflow specific problem it seems, as vanilla tensorflow doesn't give the errors)

gpflow/conditionals.py
hughsalimbeni: Yes I agree

hughsalimbeni: Though a transpose is quite cheap

gpflow/conditionals.py
hughsalimbeni: It thought I'd got rid of this already but I'll do so again if not

hughsalimbeni:
The only test now that fails is the Bernoulli equivalence which crashes travis. @st--  any ideas?

tests/test_param.py
vdutor: Yes, we've identified this problem. We resolved it in `vincent/update_tf112` so you can pull in that branch if you wish.

awav:
Can't merge it, unless you move this changes to `awav/gpflow-2.0`.

gpflow/conditionals.py
awav: I wouldn't be so confident! Transpose is cheap in numpy, not in TensorFlow. Requires confirmation though.

st--: I've seen cases where transposes were actually one of the bottlenecks - can be a lot of memory to copy around ...

awav: As far as I remember transpose op in tensorflow is simple permutation of elements, therefore memory allocations are involved.

gpflow/conditionals.py
awav: Require change

gpflow/conditionals.py
awav: Requires change

gpflow/conditionals.py
```diff
@@ -212,47 +225,56 @@ def base_conditional(Kmn, Kmm, Knn, f, *, full_cov=False, q_sqrt=None, white=Fal
...
-    Lm = tf.cholesky(Kmm)
+    N = tf.shape(Kmn)[-1]
+    M = tf.shape(f)[0]
+    leading_dims = tf.shape(Kmn)[:-2]
+    Lm = tf.cholesky(Kmm)  # [M,M]
 
     # Compute the projection matrix A
-    A = tf.matrix_triangular_solve(Lm, Kmn, lower=True)
-
+    Lm = tf.broadcast_to(Lm, tf.concat([leading_dims, tf.shape(Lm)], 0))  # [...,M,M]
```
awav:
```suggestion
    Lm = tf.broadcast_to(Lm, tf.concat([leading_dims, tf.shape(Lm)], 0))  # [..., M, M]
```

tests/test_param.py
awav: did you forget to uncomment this test?

PR #884, opened 2018-11-23
Travis script

awav:
Our tests run for >20 mins in Travis and it hits the limit, the consequence: Travis kills testing randomly. We can run long running tests for master and develop branches only. This script detects in which branch tests were run and decides either include notebooks in the testing list or not.


st--:
Would this still run notebook tests before a PR can get merged, or does this mean develop could be broken after merging a PR that looked "green" ? I think running the notebooks is an important part of the test suite (both because that's a good kind of integration test, and to make sure they don't end up out of date)...

PR #885, opened 2018-11-24
Update transforms.py

ionut199418:
# Standard GPflow pull request

At first, thank you very much for spending time on contributing to GPflow.\
This template exists to help us to grasp your work without asking very common questions. The template helps you to get thorough feedback as well.

## PR content:

* [Title](#title)
* [Description](#description)

### Title

Squeeze it. Be creative. There is no necessity for any marks like [BUG], {feature} and cetera. We have github labels for that.

### Description

Describe your contribution by writing down the context and reasons. Make it clear whether it is a **bug fix**, **feature** or **enhancement**. Do not make it very wordy, keep it short and simple.

* Bad example: `This is a bug-fix found at #111.`
* Bad example: `Add new SuperMagic optimizer.`
* Good example: `Working on feature #111, we found that parameters of VGP are not instantiated properly. It causes arbitrary failures at training. The bug is fixed by adding necessary parameters to the list of trainables.`

### Minimal working example

No matter how good the description is, the best thing you can do for making a reviewer's life much easier is to add a minimal working example (MWE). Ideally, it should be a short code snippet with relevant comments. If a MWE is big (> 50 lines), you should consider [gist](https://gist.github.com) instead.

* If your PR is a *bug* related fix, then your MWE must disclose the bug explicitly showing what happened before and after.
* When you propose a new *feature*, the MWE must show different use cases and demonstrate its benefits.

## Code quality requirements

For details, look at the code quality requirements section in [contributing.md](../contributing.md). The *main* points are repeated here:

* _New feature's_ code must be accompanied with [detailed](../contributing.md) docstrings.
* _New features_ must be covered with tests.
* _Fixed bug_ must be proved by test suites.
* Use type annotations.
* Use `.pylintrc` for code formatting.

### DIFF
--- a/gpflow/transforms.py
+++ b/gpflow/transforms.py
@@ -126,7 +126,7 @@ class Log1pe(Transform):
     A transform of the form
     .. math::
 
-       y = \log(1 + \exp(x))
+       y = \log(x)
 
     x is a free variable, y is always positive.
 


gpflow/transforms.py
```diff
@@ -126,7 +126,7 @@ class Log1pe(Transform):
     A transform of the form
     .. math::
 
-       y = \log(1 + \exp(x))
+       y = \log(x)
```
hughsalimbeni: Why this change? This transform is softplus, not log

ionut199418: I needed the log transform for my HMC sampler and I didn't know at that time how to change the file.

st--:
Hi @ionut199418, we do appreciate contributions to GPflow, but I'm closing this PR because it doesn't make sense as it is - it's only changing a comment, making it inconsistent with the code. I believe the "log" transform that you want is already implemented - it's the "Exp" class in gpflow.transforms (the latent variable is in logspace; the transform goes from the unconstrained latent variable to the constrained parameter in "real" space and hence is the exponential). If you want to open a new PR, please make sure it's self-contained, code and documentation are consistent, and to actually fill in the template (instead of leaving it there as is as you did here). I hope that makes sense, and please don't hesitate to ask if you've got any further questions on GPflow and/or how to contribute!

PR #892, opened 2018-12-13
CircleCI instead of TravisCI

awav:
^^^ subj.


.circleci/images/Dockerfile
```diff
@@ -0,0 +1,36 @@
+# Copyright 2016 The GPflow authors
```
awav:
```suggestion
# Copyright 2018 The GPflow authors
```

vdutor:
LGTM. You only need to add the codecov reports back in.

PR #895, opened 2018-12-18
Broadcasting of conditionals

vdutor:
Hi all,

This PR includes and supersedes PR #882. The PR generalises gpflow's conditional code to work with tensors that have an arbitrary number of leading dimension. Before we assumed that `Xnew` passed in `gpflow.conditional` was a two-dimensional tensor, typically referred to as `[N, D]`. This PR makes it possible to pass in `Xnew` tensors of shape `[..., N, D]`, where `[...]` can be any leading dimensions. The changes required for this are minor:
1) A couple of `tf.matmul`'s are replaced with `tf.tensordot`'s.
2) `kernels.K(X1, X2)` returns the kernel evaluated on every pair in `X1` and `X2`. This is the same behaviour as before, but now tensors with leading dims are supported. 
E.g. if `X1` is `[N1, S1, D]` and `X2: [N2, S2, D]`, `kernel.K(X1, X2)` will return a tensor of shape `[N1, S1, N2, S2]`.  `kernel.K(X1)` returns `[N1, S1, S1]`.
3) `base_conditional` is changed to support this generalisation (done by @hughsalimbeni )
4) `_mix_latent_gps` is changed to support this generalisation.


tests/test_multioutput.py
```diff
@@ -47,7 +47,7 @@ def assert_all_array_elements_almost_equal(arr, decimal):
         np.testing.assert_almost_equal(arr[i], arr[i+1], decimal=decimal)
 
 
-def check_equality_predictions(sess, models, decimal=4):
+def check_equality_predictions(sess, models, decimal=3):
```
st--: why did you have to reduce the accuracy?

tests/test_conditionals.py
```diff
@@ -160,5 +163,48 @@ def test_whiten(self):
             assert_allclose(var_difference, 0, atol=4)
 
 
+@pytest.mark.parametrize("full_cov", [True, False])
+@pytest.mark.parametrize("features_inducing_points", [False, True])
+def test_base_conditional_vs_ref(session_tf, full_cov, features_inducing_points):
+    """
+    Test that conditionals agree with a slow-but-clear numpy implementation
```
st--: I think this is great, would be even better to have the slow-but-clear numpy implementation in its own function (just to make the test a bit easier to follow and to decouple setup from calling conditionals).

gpflow/kernels.py
```diff
@@ -761,6 +768,18 @@ def Kdiag(self, X, presliced=False):
         return reduce(tf.multiply, [k.Kdiag(X) for k in self.kernels])
 
 
+def _broadcasting_elementwise_op(op, a, b):
+    r"""
+    Apply binary operation `op` to every pair of in input tensors `a` and `b`.
```
vdutor:
```suggestion
    Apply binary operation `op` to every pair in tensors `a` and `b`.
```

gpflow/conditionals.py
```diff
@@ -73,6 +73,7 @@ def _conditional(Xnew, feat, kern, f, *, full_cov=False, full_output_cov=False,
     return fmean, _expand_independent_outputs(fvar, full_cov, full_output_cov)
 
 
+
```
vdutor:
```suggestion
```

gpflow/multioutput/conditionals.py
```diff
@@ -463,3 +444,72 @@ def fully_correlated_conditional_repeat(Kmn, Kmm, Knn, f, *, full_cov=False, ful
             addvar = tf.reshape(tf.reduce_sum(tf.square(LTA), axis=1), (R, N, K))  # R x N x K
             fvar = fvar[None, ...] + addvar  # R x N x K
     return fmean, fvar
+
+
+def _mix_latent_gp(W, g_mu, g_var, full_cov, full_output_cov):
+    r"""
+    Takes the mean and variance of a uncorrelated L-dimensional latent GP
```
vdutor:
```suggestion
    Takes the mean and variance of an uncorrelated L-dimensional latent GP
```

gpflow/multioutput/conditionals.py
```diff
@@ -463,3 +444,72 @@ def fully_correlated_conditional_repeat(Kmn, Kmm, Knn, f, *, full_cov=False, ful
...
+    and returns the mean and the variance of the mixed GP, `f = W \times g`,
+    where both f and g are GPs.
+
+    :param W: [P, L]
+    :param g_mu: [..., N, L]
+    :param g_var: [..., N, L] or [L, ..., N, N]
```
vdutor:
```suggestion
    :param g_var: [..., N, L] (full_cov = False) or [L, ..., N, N] (full_cov=True)
```

tests/test_broadcasting.py
```diff
@@ -41,14 +41,14 @@
 
     kernels.Linear,
     kernels.Polynomial,
-    kernels.ArcCosine,
-    kernels.Periodic,
+    # kernels.ArcCosine,
+    # kernels.Periodic,
```
st--: They don't broadcast ?

tests/test_broadcasting.py
```diff
@@ -41,14 +41,14 @@
...
 ]
 
 def _test_no_active_dims(Kern, sess):
     S, N, M, D = 5, 4, 3, 2
     X1 = tf.identity(np.random.randn(S, N, D))
-    X2 = tf.identity(np.random.randn(S, M, D))
+    X2 = tf.identity(np.random.randn(M, D))
```
st--: why not add this as an extra test instead of replacing the previous one (which tests slightly different behaviour)?

gpflow/conditionals.py
```diff
@@ -181,13 +183,15 @@ def _sample_conditional(Xnew, X, kern, f, *, q_sqrt=None, white=False, full_cov=
     samples = _sample_mvn(mean, cov, cov_structure, num_samples=num_samples)
     if full_cov:
         samples = tf.matrix_transpose(samples)
+        mean = tf.matrix_transpose(mean)
```
st--: Hm, was this code broken before (in the full_cov=True case) ? What shape is mean supposed to be - could we sort that out in this PR? Also, would be good to bring this in line with the other _sample_conditional: introduce a local variable mean_PN that is the potentially transposed one so we don't have to transpose twice for the return value.

gpflow/conditionals.py
```diff
@@ -199,9 +203,9 @@ def base_conditional(Kmn, Kmm, Knn, f, *, full_cov=False, q_sqrt=None, white=Fal
       q(g2) = N(g2;f,q_sqrt*q_sqrt^T)
     This method computes the mean and (co)variance of
       q(g1) = \int q(g2) p(g1|g2)
-    :param Kmn: M x N
+    :param Kmn: M x [...] x N
```
st--: should this not be [...] x M x N ?

tests/test_conditionals.py
vdutor: yes, you are right! These tests can use a bit of restructuring

gpflow/kernels.py
```diff
@@ -761,6 +768,18 @@ def Kdiag(self, X, presliced=False):
...
+def _broadcasting_elementwise_op(op, a, b):
+    r"""
+    Apply binary operation `op` to every pair in tensors `a` and `b`.
+    :param op: binary operator on tensors, e.g. tf.add, tf.substract
+    :param a: tf.Tensor, shape [n_1, ..., n_a]
+    :param b: tf.Tensor, shape [m_1, ..., m_b]
+    :return: tf.Tensor, shape [n_1, ..., n_a, m_1, ..., m_b]
+    """
+    flatres = op(tf.reshape(a, [-1, 1]), tf.reshape(b, [1, -1]))
+    return tf.reshape(flatres, tf.concat([tf.shape(a), tf.shape(b)], 0))
```
st--: the compile-time version with get_shape() didn't work ?

gpflow/mean_functions.py
```diff
@@ -64,7 +64,7 @@ def __init__(self, A=None, b=None):
 
     @params_as_tensors
     def __call__(self, X):
-        return tf.matmul(X, self.A) + self.b
+        return tf.tensordot(X, self.A, [[-1], [0]]) + self.b
```
st--: is this needed because tf.matmul() does not broadcast properly ? (i.e., once tf implements properly-broadcasting matmul, would we be able to revert this change?)

gpflow/multioutput/conditionals.py
```diff
@@ -120,12 +120,12 @@ def single_gp_conditional(t):
                           (Kmms, Kmns, Knns, fs, q_sqrts),
                           (settings.float_type, settings.float_type))  # P x N x 1, P x 1 x N x N or P x N x 1
 
-    fmu = tf.matrix_transpose(rmu[:, :, 0])  # N x P
+    fmu = _rollaxis(rmu[..., 0], 1)  # N x P
 
     if full_cov:
-        fvar = rvar[:, 0, :, :]  # P x N x N
+        fvar = rvar[..., 0, :, :]  # P x N x N
```
st--: why does it have a dimension of length 1 in it anyways ?

gpflow/multioutput/conditionals.py
```diff
@@ -463,3 +444,72 @@ def fully_correlated_conditional_repeat(Kmn, Kmm, Knn, f, *, full_cov=False, ful
             addvar = tf.reshape(tf.reduce_sum(tf.square(LTA), axis=1), (R, N, K))  # R x N x K
             fvar = fvar[None, ...] + addvar  # R x N x K
     return fmean, fvar
+
+
+def _mix_latent_gp(W, g_mu, g_var, full_cov, full_output_cov):
+    r"""
+    Takes the mean and variance of an uncorrelated L-dimensional latent GP
+    and returns the mean and the variance of the mixed GP, `f = W \times g`,
```
st--: I would just write `f = W g` but also point out that W is a matrix with shape [...]

gpflow/kernels.py
vdutor: Tried both, if I remember correctly 79 tests failed with the `get_shape()` approach.

gpflow/mean_functions.py
vdutor: Indeed, once tensorflow/tensorflow#216 is resolved we can use matmul here. The behaviour we want works out-of-the-box with numpy's matmul:
```
In [1]: X = np.random.randn(10, 5, 4)                                           
In [2]: A = np.random.randn(4, 3)                                               
In [3]: np.matmul(X, A).shape                                                   
Out[3]: (10, 5, 3)
```

gpflow/multioutput/conditionals.py
vdutor: This is due to the `tf.map` used above

gpflow/multioutput/conditionals.py
vdutor: Alright, I'll remove the `\times`. The `W` matrix doesn't have any leading dimensions though.

tests/test_broadcasting.py
vdutor: Not yet, I can have look if I can make them broadcast-able as well.

gpflow/kernels.py
awav: `get_shape()` doesn't return tensor, it is compile time shape (which is sometimes unknown), not run-time.

tests/test_broadcasting.py
hughsalimbeni: Yes this way is better. X1 and X2 shouldn't have to match leading dims - that case doesn't come up in gpflow at the moment (it would be for multiple datasets/inducing points and multiple observations, i.e. multiple parallel models, which we never do)

gpflow/multioutput/conditionals.py
```diff
@@ -463,3 +444,72 @@ def fully_correlated_conditional_repeat(Kmn, Kmm, Knn, f, *, full_cov=False, ful
...
+    where both f and g are GPs.
+
+    :param W: [P, L]
+    :param g_mu: [..., N, L]
+    :param g_var: [..., N, L] (full_cov = False) or [L, ..., N, N] (full_cov = True)
+    :return: f_mu and f_var, shape depends on `full_cov` and `full_output_cov`
```
st--: Could you add what shape they have in each case, please?

tests/test_broadcasting.py
vdutor: Yes, the shape now would be `[S, N, S, M]`. You can think of `kern.k(., .)` as a tensordot operation on the last dimension, i.e. reducing the last dimension (here `D`) and then concatenating both tensors from left to right.

gpflow/conditionals.py
```diff
@@ -111,13 +111,15 @@ def _conditional(Xnew, X, kern, f, *, full_cov=False, q_sqrt=None, white=False):
         - variance: N x R (full_cov = False), R x N x N (full_cov = True)
     """
     logger.debug("Conditional: Kernel")
-    num_data = tf.shape(X)[0]  # M
-    Kmm = kern.K(X) + tf.eye(num_data, dtype=settings.float_type) * settings.numerics.jitter_level
-    Kmn = kern.K(X, Xnew)
+    num_data = tf.shape(X)[-2]  # M
+    Kmm = kern.K(X) + tf.eye(num_data, dtype=settings.float_type) * settings.numerics.jitter_level  #  [..., M, M]
```
awav: 1. `settings.float_type` => `X.dtype`
2. `settings.numerics.jitter_level` => `settings.jitter`

gpflow/conditionals.py
```diff
@@ -111,13 +111,15 @@ def _conditional(Xnew, X, kern, f, *, full_cov=False, q_sqrt=None, white=False):
...
+    Kmn = kern.K(X, Xnew)  # [M, ..., N]
+
     if full_cov:
-        Knn = kern.K(Xnew)
+        Knn = kern.K(Xnew)  # [...,N,N]
```
awav: [discuss] These if else are ugly, just a note for next PRs: we need `kern.K(X, diag=not full_cov)` or `kern.K(X, full=full_cov)`

gpflow/conditionals.py
```diff
@@ -211,47 +221,65 @@ def base_conditional(Kmn, Kmm, Knn, f, *, full_cov=False, q_sqrt=None, white=Fal
...
     num_func = tf.shape(f)[1]  # R
-    Lm = tf.cholesky(Kmm)
+    N = tf.shape(Kmn)[-1]
+    M = tf.shape(f)[0]
 
-    # Compute the projection matrix A
-    A = tf.matrix_triangular_solve(Lm, Kmn, lower=True)
+    # get the leadings dims in Kmn to the front of the tensor
+    # if Kmn has rank two, i.e. [M, N], this is the identity op.
+    K = tf.rank(Kmn)
```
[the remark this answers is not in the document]
vdutor: yes, good point - i'll rename it

gpflow/conditionals.py
vdutor: yes, I agree. We should add this to our todo list

gpflow/conditionals.py
vdutor: There are plenty of places where we do this. We should probably have a separate PR that deals with all of them at once.

hughsalimbeni:
I made a couple of small changes to make it work with q_mu of shape [..., M, D]. This is only checked  for the q_sqrt=None case. Perhaps it would be possible to have q_sqrt of shape [..., D, M, M] but I haven't tried this.

hughsalimbeni:
Here is an alternative way of doing the multisample: https://github.com/GPflow/GPflow/tree/hugh/broadcasting_matched_multi_sample. 

I'm **not** suggesting at this stage to change this PR, but leaving this here for reference. 

To clarify, this PR has

`cov([A,...,B,N,D], [C,...,D,M,D]) = [A,...,B,N,C,...,D,M]`
which seems sensible and follows the tensordot convension, whereas the alternative https://github.com/GPflow/GPflow/tree/hugh/broadcasting_matched_multi_sample has 

`cov([A,...,B,N,D], [A,...,B,M,D]) = [A,...,B,N,M]`
where the A,...,B must match, and 
`cov([A,...,B,N,D], [M,D]) = [A,...,B,N,M]` 
and 
`cov([N,D], [A,...,B,M,D]) = [A,...,B,N,M]`

PR #898, opened 2018-12-22
Ask people to use SO in ISSUE_TEMPLATE

vdutor:
Updated  GitHub issue template to ask people to use stack overflow for questions.


awav:
I've replaced single issue template with sub-templates.

PR #902, opened 2019-01-24
Modified Saver to work with file-like objects as well as path names.

Joel-Jennings:
This required updating h5py>=2.9.0 (http://docs.h5py.org/en/stable/high/file.html).
I have a use case where I'd like to save the GPFlow model to a cloud datastore

Updated tests to test this.

Also removed some unused imports.


st--:
Replaced by #912 to make CircleCI run it

PR #903, opened 2019-01-24
Typo "NOSET" -> "NOTSET" in gpflowrc

rhaps0dy:
The available Python logging levels are as written now in the file:
https://docs.python.org/3/library/logging.html#levels

I don't think it's worth making a test for this. You can check the bug by doing:
```python
import gpflow
gpflow.settings.logging.level = 'NOSET'
gpflow.settings.logger()
```
and
```python
import gpflow
gpflow.settings.logging.level = 'NOTSET'
gpflow.settings.logger()
```


st--:
Closed in favour of #911

PR #907, opened 2019-01-28
Fix AutoBuild metaclass to not screw up method signatures

st--:


gpflow/core/compilable.py
```diff
@@ -35,6 +36,65 @@ class AutoBuildStatus(enum.Enum):
...
+    The only difference to decorator.decorate() is to use our custom AutobuildFunctionMaker instead.
+    """
+    evaldict = dict(_call_=caller, _func_=func)
+    fun = AutobuildFunctionMaker.create(
+        func, "return _call_(_func_, %(shortsignature)s)",
+        evaldict, __wrapped__=func)
+    if hasattr(func, '__qualname__'):
+        fun.__qualname__ = func.__qualname__
+    return fun
+
```
awav:
```suggestion


```

PR #910, opened 2019-01-28
Remove unnecessary warnings, add DataHolder shape check

st--:
This PR deals with a couple of warnings that confused users new to GPflow:

* Older versions of the multipledispatch package raised an ambiguity warning for the conditionals module. This warning does not come up when using a version >= 0.5. Accordingly, we bump the required version in setup.py.

* logdensities.multivariate_normal() raised a warning whenever shapes weren't known at graph construction time - this is only used in the GPR class and particularly confuses new users. This PR removes the warnings altogether.

* We also add an optional shape check: extra argument `check_ndim_is` to the DataHolder class that will raise a ValueError when the shape of the passed-in value (i.e. data array) doesn't have the right ndims. This is activated for all the models in the models/ submodule, enforcing X and Y to be two-dimensional (N x D, even if D=1).


gpflow/params/dataholders.py
```diff
@@ -40,13 +41,20 @@ class DataHolder(Parameter):
     :param fix_shape: Default value is `False` and indicates that shape
         of internal tensor does not have specific shape, in other words,
         it is None.
+    :param allow_1d: do not check shape of data input value.
```
awav: This is very weird argument! What is so special about 1D case, that it must go with all other fixes?

gpflow/logdensities.py
```diff
@@ -85,15 +82,6 @@ def multivariate_normal(x, mu, L):
     p[n] = log pdf of:
     x[n] ~ N(mu, LL^T) or x ~ N(mu[n], LL^T) or x[n] ~ N(mu[n], LL^T)
     """
-    if x.shape.ndims is None:
```
awav: This is already fixed in develop.

setup.py
```diff
@@ -19,7 +19,7 @@
     'numpy>=1.10.0',
     'scipy>=0.18.0',
     'pandas>=0.18.1',
-    'multipledispatch>=0.4.9',
+    'multipledispatch>=0.5',
```
awav: Shouldn't we check first what was the reason for that warning before getting rid of it?

tests/test_logdensities.py
```diff
@@ -26,32 +26,25 @@
...
-            x_tf, mu_tf, tf.convert_to_tensor(L))
-
-        gp_result = session_tf.run(gp_result, feed_dict={x_tf: x, mu_tf: mu})
-
-        if mu.shape[1] > 1:
-            if x.shape[1] > 1:
-                sp_result = [mvn.logpdf(x[:,i], mu[:,i], cov) for i in range(mu.shape[1])]
-            else:
-                sp_result = [mvn.logpdf(x.ravel(), mu[:, i], cov) for i in range(mu.shape[1])]
+    x_tf = tf.placeholder(settings.float_type)
```
awav: Fixed in develop.

gpflow/logdensities.py
```diff
@@ -85,13 +82,6 @@ def multivariate_normal(x, mu, L):
     p[n] = log pdf of:
     x[n] ~ N(mu, LL^T) or x ~ N(mu[n], LL^T) or x[n] ~ N(mu[n], LL^T)
     """
-    mu_ndims = mu.shape.ndims
-    x_ndims = x.shape.ndims
-    if x_ndims is not None and x_ndims != 2:
```
awav: Please, leave this file without change. Warnings removed in develop. Exceptions should stay.

st--: If you want an exception it should be a tf.Assert, no?

awav: No, tf.Assert is a run-time, I want it on graph building stage.

gpflow/logdensities.py
```diff
@@ -85,12 +82,11 @@ def multivariate_normal(x, mu, L):
...
-        raise ValueError('Shape of x must be 2D.')
-    if mu_ndims is not None and mu_ndims != 2:
-        raise ValueError('Shape of mu must be 2D.')
+
+    if x.shape.ndims is not None and x.shape.ndims != 2:
```
awav: Just checkout this file from your branch :)

gpflow/models/gpmc.py
```diff
@@ -47,8 +47,8 @@ def __init__(self, X, Y, kern, likelihood,
             L L^T = K
 
         """
-        X = DataHolder(X)
-        Y = DataHolder(Y)
+        X = DataHolder(X, check_ndim_is=2)
```
awav: We've already discussed it and decided that this change shouldn't be in any Parameter, but in the model itself.

gpflow/logdensities.py
st--: why do you even care? Alessandro added the check, and it only gets used in GPR anyways ...

gpflow/models/gpmc.py
st--: Yeah... that's what I did? I changed it in the model?

gpflow/logdensities.py
```diff
@@ -85,17 +83,16 @@ def multivariate_normal(x, mu, L):
...
-    p = - 0.5 * tf.reduce_sum(tf.square(alpha), 0)
-    p -= 0.5 * num_dims * np.log(2 * np.pi)
-    p -= tf.reduce_sum(tf.log(tf.matrix_diag_part(L)))
-    return p
+
+    assert_x_shape = tf.Assert(tensor_ndim_equal(x, 2), [
+        'multivariate_normal requires the shape of x to be (N, D)'])
+    assert_mu_shape = tf.Assert(tensor_ndim_equal(mu, 2), [
+        'multivariate_normal requires the shape of mu to be (N, D)'])
+    with tf.control_dependencies([assert_x_shape, assert_mu_shape]):
```
st--: For some reason the asserts don't actually get triggered...

st--: because I tested with constants, not with placeholders fed in through the feed_dict...

thevincentadam:
Looks good to me!

awav:
Hello @st-- ,

1. PR assumes 2 changes at the same time. As far as remember, we agreed on policy 1 PR - 1 change. If it is not reflected in `contributions.md`, then we should add it.
2. If main contribution is a warnings removings, then it was solved and already in develop.
3. I disagree with DataHolder changes.

Details:

The `check_ndim_is` functionality better to be done by a model, not by a dataholder. The dataholder instance can contain a data of _any type and any shape, without any restrictions_. It is unintuitive and misleading why so much attention allocated to the single case in that class. In turn, current models, indeed, handle 1D case incorrectly. There are two ways of changing this situation. You can begin with making models work with 1D case (are only kernels changes required for that?). And, secondly, you can change model arguments verification method to test shape correctness even before passing it to the dataholder. These methods would exactly solve the problem in the place where it originates.

The TensorFlow's asserts are a good way of detecting errors on early stage. But here, the information on `ndims` exists even before running `session.run(...)` method. It pretty good solution and more than enough, as we know where it is ran and what tensors it uses. Introducing +100 lines of code for handling this assert is a bit of over-engineering. And:

> why do you even care? Alessandro added the check, and it only gets used in GPR anyways ...

I'm sorry, but as a reviewer, I do not understand this question. We can discuss it offline.

Thanks!

st--:
Split into #964 and #965.

PR #915, opened 2019-01-28
Cleaned up SVGP Notebook

aboustati:
Clean-up of SVGP notebook


awav:
@aboustati , don't forget we run this notebooks in CI, and usually it takes a lot of time. Consider this notebook as integration test. `1000` datapoints and `1000` iteration steps are time consuming.

st--:
1000 datapoints and 1000 iterations is fine, but wrap them in one of test_util.notebook_niter, notebook_range, notebook_list to make sure that the CI only does 2 instead of 1000 steps. :-)

st--:
Yes, it still works on CircleCI (otherwise all the notebook tests would've been taking an hour instead of under ten minutes!).

PR #916, opened 2019-01-28
Tips & Tricks notebooks changes

awav:
subj.

### DIFF
--- a/doc/source/notebooks/tips_and_tricks.ipynb
+++ b/doc/source/notebooks/tips_and_tricks.ipynb
@@ -35,7 +35,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "_Bad example:_"
+    "This example below illustrates typical situation when a computation time increases proportionally to the number of created GPflow objects."
    ]
   },
   {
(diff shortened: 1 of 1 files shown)


doc/source/notebooks/tips_and_tricks.ipynb
```diff
@@ -418,7 +418,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "Or you can `anchor()` your model to the session after the optimisation step. After that your parameters' cache have been updated."
+    "Or you can `anchor()` your model to the session after the optimisation step. The `anchor()` updates parameters' cache. **The `anchor()` is significantly more time consuming than `read_value()`, do not call it too often without necessity.**"
```
st--: is anchor(session) also slower than read_value(session)? or does it just seem like that?

awav:
```suggestion
    "Or you can `anchor(...)` your model to the session after the optimisation step. The `anchor()` updates parameters' cache. **The `anchor(...)` is significantly more time consuming than `read_value(...)`, do not call it too often without necessity.**"
```

awav:
```suggestion
    "Or you can `anchor(session)` your model to the session after the optimisation step. The `anchor()` updates parameters' cache. **The `anchor(session)` is significantly more time consuming than `read_value(session)`, do not call it too often without necessity.**"
```

PR #923, opened 2019-02-04
GPflow glossary document

awav:
This glossary should ease reading GPflow's documentation and the code, explaining some non-trivial definitions and abbreviations. For example, we use some uppercase letters like N, L, M, D and others implying that reader familiar with their meaning, of course it is not true. To avoid cumbersome and tedious repetitions in the docs, I created a single definition list markdown file where all well established definitions must take place.


GLOSSARY.md
```diff
@@ -0,0 +1,31 @@
+<dl>
```
st--: Would be good to have a header & brief explanation. E.g.:

```
Glossary
========

GPflow does not always follow standard Python naming conventions,
and instead tries to follow the notation in the relevant GP papers.
The following is the convention we aim to follow in the code:
```

GLOSSARY.md
```diff
@@ -0,0 +1,31 @@
...
+  <dt>GPR</dt>
+  <dd>gaussian process regression</dd>
```
st--:
```suggestion
  <dd>Gaussian process regression</dd>
```

GLOSSARY.md
```diff
@@ -0,0 +1,31 @@
...
+
+  <dt>SVGP</dt>
+  <dd>stochastic variational inference for Gaussian process models</dd>
+
+  <dt>[] shape constructions</dt>
```
st--:
```suggestion
  <dt>`[...]` shape annotations</dt>
```

GLOSSARY.md
```diff
@@ -0,0 +1,31 @@
...
+  <dd>the way of describing tensor shape in docstrings and comments. Example: [..., N, D, D], this is tensor with unknown number of leading dimensions and last two dimensions are equal</dd>
```
st--:
```suggestion
  <dd>The way of describing tensor shapes in docstrings and comments. Example: [..., N, D, D], this is a tensor with an arbitrary number of leading dimensions, and the last two dimensions are equal</dd>
```

GLOSSARY.md
```diff
@@ -0,0 +1,31 @@
...
+
+  <dt>X</dt>
+  <dd>and its variations like Xnew in GPflow code mean input data (numpy or tensorflow)</dd>
```
st--:
```suggestion
  <dd>(and variations like Xnew) refers to input points; always of rank 2, e.g. shape [N, D], even when D=1</dd>
```
(I think a lot of the entries can describe both numpy arrays or tensorflow tensors so I think it's confusing if this is the only place where it's mentioned explicitly)

GLOSSARY.md
```diff
@@ -0,0 +1,31 @@
...
+
+  <dt>Y</dt>
+  <dd>throughout is used to denote observed data, potentially with multiple output dimensions</dd>
```
st--:
```suggestion
  <dd>refers to observed output values, potentially with multiple output dimensions; always of rank 2, e.g. shape [N,
 P], even when P=1</dd>
```
Would it be useful to point towards the notebooks here (Coregionalization still has "P=1", the outputs are distinguished by one of the columns in the input points)?

GLOSSARY.md
```diff
@@ -0,0 +1,31 @@
...
+  <dd>stands for data or minibatch size in docstrings and shape constructions</dd>
+
+  <dt>P</dt>
+  <dd>stands for the number of output dimensions in docstrings and shape constructions</dd>
+
+  <dt>D</dt>
+  <dd>stands for the number of input dimensions in docstrings and shape constructions</dd>
+
+  <dt>Z</dt>
+  <dd>inducing points tensor</dd>
```
st--:
```suggestion
  <dd>inducing points</dd>
```

GLOSSARY.md
```diff
@@ -0,0 +1,31 @@
...
+
+  <dt>...</dt>
+  <dd>leading dimensions in docstrings and in the code it is ellipsis.</dd>
```
st--:
```suggestion
  <dd>arbitrary leading dimensions</dd>
```
(this is true both for docstrings and code)

st--: maybe move this right underneath the entry for shapes ?

GLOSSARY.md
awav: @st-- , it is slightly misleading because later we say that the ellipsis sign `...` denotes leading dimensions.

awav: @st-- , but I agree that empty brackets don't tell much as well.

GLOSSARY.md
```diff
@@ -0,0 +1,36 @@
...
+
+  <dt>X</dt>
+  <dd>(and variations like Xnew) refers to input points; always of rank 2, e.g. shape <i>[N, D]</i>, even when <i>D=1</i></dd>
+
+  <dt>Y</dt>
+  <dd>(and variations like Ynew) refers to observed output values, potentially with multiple output dimensions; always of rank 2, e.g. shape <i>[N, P]</i>, even when <i>P=1</i></dd>
+
+  <dt>Z</dt>
+  <dd>refers to inducing points</dd>
+
```
st--:
```suggestion

  <dt>M</dt>
  <dd>stands for the number of inducing features (e.g. length of Z)</dd>

```

GLOSSARY.md
```diff
@@ -0,0 +1,39 @@
...
+  <dt>M</dt>
+  <dd>stands for the number of inducing features (e.g. length of Z)</dd>
+
+  <dt>N</dt>
+  <dd>stands for data or minibatch size in docstrings and shape constructions</dd>
```
awav:
```suggestion
  <dd>stands for the number of data or minibatch size in docstrings and shape constructions</dd>
```

README.md
```diff
@@ -6,7 +6,8 @@
 
 [Website](https://gpflow.org) |
 [Documentation](https://gpflow.readthedocs.io/en/master/) |
-[Tips & Tricks](https://github.com/GPflow/GPflow/blob/develop/doc/source/notebooks/tips_and_tricks.ipynb)
+[Glossary](GLOSSARY.md) |
```
awav:
```suggestion
[Glossary](GLOSSARY.md) |```

PR #929, opened 2019-02-04
Nicolas/notebook classification

NicolasDurrande:
Nothing super fancy here: just a new notebook on classif + minor edits to the regression one so that they have a similar structure


doc/source/notebooks/manual/basics/classification.ipynb
```diff
@@ -4,11 +4,299 @@
...
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "m = gpflow.models.VGP(X, Y,\n",
+    "                      likelihood=gpflow.likelihoods.Bernoulli(),\n",
+    "                      kern=gpflow.kernels.Matern52(1))\n",
+    "\n",
+    "o = gpflow.train.ScipyOptimizer()\n",
+    "o.minimize(m);"
```
vdutor: can you add a `maxiter` arg to the optimiser please
```
from gpflow.test_util import notebook_niter
o.minimize(m, maxiter=notebook_niter(10e3))
```

doc/source/notebooks/manual/basics/classification.ipynb
```diff
@@ -4,11 +4,299 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import numpy as np\n",
+    "import tensorflow as tf\n",
```
vdutor: you are not using `tf`, can be removed :)

doc/source/notebooks/manual/basics/classification.ipynb
NicolasDurrande: I'd like the code to be as neat as possible so I would be in favour of keeping it the way it is. If it happens to be too slow on the servers, I am happy to reduce the number of observation points, and to stop this optim before convergence.

doc/source/notebooks/manual/basics/classification.ipynb
NicolasDurrande: done

PR #933, opened 2019-02-15
Minor performance improvement: pass Cholesky of K to gauss_kl

vdutor:
This PR adds the ability to pass the Cholesky of K to `gauss_kl(.)` for the unwhitened case, instead of `K`. It prevents the double cost of inverting `K` twice, in both conditionals and gauss_kl.

## changes:
- added `K_cholesky` kwarg to `kullback_leiblers.gauss_kl`
- updated file to the new way of documenting array shapes. From `A x B` to `[A, B]`
- added unit test

### DIFF
--- a/gpflow/kullback_leiblers.py
+++ b/gpflow/kullback_leiblers.py
@@ -42,13 +42,17 @@ def gauss_kl(q_mu, q_sqrt, K=None, *, K_cholesky=None):
 
     K is the covariance of p, [M, M] or [L, M, M]
     K_cholesky is the cholesky of the covariance of p, [M, M] or [L, M, M]
-    Either `K` or `K_cholesky` have to be None.
-    If `K` and `K_cholesky` are None, compute the KL divergence to p(x) = N(0, I) instead.
+    
+    Note: if no K matrix is given (both `K` and `K_cholesky` are None),
+    `gauss_kl` computes the KL divergence from p(x) = N(0, I) instead.
+    The K matrix can be passed either directly as `K`, or as its Cholesky factor, `K_cholesky`. 
+    In either case, it can be a single matrix [M, M], in which case the sum of the L KL divergences 
+    is computed by broadcasting, or L different covariances [L, M, M].
     """
 
     if (K is not None) and (K_cholesky is not None):
-        raise ValueError("Ambigious args are passed to `gauss_kl`, "
-                         "Either `K` or `K_cholesky` need to be `None`")
+        raise ValueError("Ambiguous arguments: gauss_kl() must only "
+                         "be passed one of `K` or `K_cholesky`.")
 
     white = (K is None) and (K_cholesky is None)
     diag = q_sqrt.get_shape().ndims == 2
--- a/gpflow/kullback_leiblers.py
+++ b/gpflow/kullback_leiblers.py
@@ -22,7 +22,7 @@ from .decors import name_scope
 
 
 @name_scope()
-def gauss_kl(q_mu, q_sqrt, K=None):
+def gauss_kl(q_mu, q_sqrt, K=None, *, K_cholesky=None):
     """
     Compute the KL divergence KL[q || p] between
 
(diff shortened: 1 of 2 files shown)


gpflow/kullback_leiblers.py
```diff
@@ -22,7 +22,7 @@
 
 
 @name_scope()
-def gauss_kl(q_mu, q_sqrt, K=None):
+def gauss_kl(q_mu, q_sqrt, K=None, *, K_cholesky=None):
```
st--: Huh, does this work?? I had no idea! so you can pass K both by position or keyword, but K_cholesky must be passed as a keyword argument?

gpflow/kullback_leiblers.py
```diff
@@ -33,38 +33,46 @@ def gauss_kl(q_mu, q_sqrt, K=None):
...
-    If K is None, compute the KL divergence to p(x) = N(0, I) instead.
+    K is the covariance of p, [M, M] or [L, M, M]
+    K_cholesky is the cholesky of the covariance of p, [M, M] or [L, M, M]
+    Either `K` or `K_cholesky` have to be None.
+    If `K` and `K_cholesky` are None, compute the KL divergence to p(x) = N(0, I) instead.
     """
 
-    white = K is None
+    if (K is not None) and (K_cholesky is not None):
+        raise ValueError("Ambigious args are passed to `gauss_kl`, "
```
st--:
```suggestion
        raise ValueError("Ambiguous arguments: gauss_kl() must only be passed one of `K` or `K_cholesky`.")
```
(and delete following line)

gpflow/kullback_leiblers.py
```diff
@@ -33,38 +33,46 @@ def gauss_kl(q_mu, q_sqrt, K=None):
...
+    q_sqrt can be a matrix [M, L], each column represents the diagonal of a
         square-root matrix of the covariance of q.
 
-    K is the covariance of p.
-    It is a positive definite matrix (M x M) or a tensor of stacked such matrices (L x M x M)
-    If K is None, compute the KL divergence to p(x) = N(0, I) instead.
+    K is the covariance of p, [M, M] or [L, M, M]
+    K_cholesky is the cholesky of the covariance of p, [M, M] or [L, M, M]
+    Either `K` or `K_cholesky` have to be None.
+    If `K` and `K_cholesky` are None, compute the KL divergence to p(x) = N(0, I) instead.
```
st--: It would be good to make a bit clearer what happens. How about:

"If no K matrix is given (both `K` and `K_cholesky` are None), computes the KL divergence from p(x) = N(0, I) instead.
The K matrix can be passed either directly as `K`, or as its cholesky factor, `K_cholesky`. In either case, it can be a single matrix [M, M], in which case the sum of the L KL divergences is computed by broadcasting, or L different covariances [L, M, M]."

(note that it is the KL[q || p] *from* p *to* q)

gpflow/kullback_leiblers.py
vdutor: Yeah, it works as you would expect. 
```
In [6]: def foo(a=None, *, b=None):
   ...:     print(a, b)

In [7]: foo(a=1, b=2)
1 2

In [8]: foo(a=1)
1 None

In [9]: foo(b=2)
None 2

In [10]: foo(1, b=2)
1 2

In [11]: foo(2, a=1)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-11-00078d28f012> in <module>
----> 1 foo(2, a=1)
TypeError: foo() got multiple values for argument 'a'

In [12]: foo(1, 2)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-12-d3151bdab612> in <module>
----> 1 foo(1, 2)
TypeError: foo() takes from 0 to 1 positional arguments but 2 were given
```

I think it's the best thing to do as it is backwards compatible and "forces" people to explicitly specify if they passing the cholesky

st--: Yeah :) I think it's great!

st--: (I just had no idea it would work)

PR #936, opened 2019-02-18
Nicolas/playing with kernels

NicolasDurrande:
Notebook *manipulating kernels* in the new format


markvdw:
This is great. I guess it could be improved with visualisations of the higher dimensional examples, but we're not in this to write an intro book I guess :)

PR #945, opened 2019-02-19
fix fully correlated conditional repeat shapes

marctom:
Fix shapes returned by `fully_correlated_conditional_repeat`

There's inconsistency between shapes of `fmean` and `fvar` returned by `fully_correlated_conditional_repeat`. This causes problems with downstream functions for sampling.

Minimal working example:

This test catches the bug:

```
@pytest.mark.parametrize("func", [fully_correlated_conditional_repeat,
                                  fully_correlated_conditional])
def test_fully_correlated_conditional_repeat_shapes(func):
    L, M, N, P = Data.L, Data.M, Data.N, Data.P
    R = 1

    Kmm = tf.ones((L * M, L * M))
    Kmn = tf.ones((L * M, N, P))
    Knn = tf.ones((N, P))
    f = tf.ones((L * M, R))
    q_sqrt = None
    white = True

    m, v = func(Kmn, Kmm, Knn, f,
                full_cov=False,
                full_output_cov=False,
                q_sqrt=q_sqrt,
                white=white)

    assert v.shape == m.shape
```

Output before fix:

```
func = <function fully_correlated_conditional at 0x7ff689444268>

    @pytest.mark.parametrize("func", [fully_correlated_conditional_repeat,
                                      fully_correlated_conditional])
    def test_fully_correlated_conditional_repeat_shapes(func):
        L, M, N, P = Data.L, Data.M, Data.N, Data.P
        R = 1
    
        Kmm = tf.ones((L * M, L * M))
        Kmn = tf.ones((L * M, N, P))
        Knn = tf.ones((N, P))
        f = tf.ones((L * M, R))
        q_sqrt = None
        white = True
    
        m, v = func(Kmn, Kmm, Knn, f,
                    full_cov=False,
                    full_output_cov=False,
                    q_sqrt=q_sqrt,
                    white=white)
    
>       assert v.shape.as_list() == m.shape.as_list()
E       assert [3] == [20, 3]
E         At index 0 diff: 3 != 20
E         Right contains more items, first extra item: 3
E         Use -v to get the full diff

tests/test_multioutput.py:314: AssertionError
```


st--:
@marctom thanks for spotting this and fixing it ! Looks good, I'm only wondering if we might ever end up with an fvar that has an undefined shape with `fvar.shape.ndim` being None (which would break the code you added)?

marctom:
Many thanks for looking into this.

Dynamic shapes are indeed painful - fortunately they go away eager mode!

But in this case I'm not sure if I fully understand; `fvar.shape.ndims` returns the number of dimensions, regardless of what they are specified or not as here:
```
In [1]: import tensorflow as tf                                                                                                                                                          

In [2]: a = tf.placeholder(shape=(None,10), dtype=tf.float32)                                                                                                                            

In [3]: a.shape                                                                                                                                                                          
Out[3]: TensorShape([Dimension(None), Dimension(10)])

In [4]: a.shape.ndims                                                                                                                                                                    
Out[4]: 2
```

The code could indeed break if we used a dynamic shape of one tensor to reshape another, but we don't do this.

st--:
If you just do `a = tf.placeholder(dtype=tf.float32)`, then `a.shape.ndims` is None... and that is what actually happens in the DataHolders! Can you replace it with tf graph code that builds the shape dynamically?

PR #952, opened 2019-03-05
Fixing bad docstrings in conditionals

thevincentadam:
quick fix to update dosctrings


gpflow/conditionals.py
```diff
@@ -101,7 +103,7 @@ def _conditional(Xnew, X, kern, f, *, full_cov=False, q_sqrt=None, white=False,
     :param X: data points, size M x D.
     :param kern: GPflow kernel.
     :param f: data matrix, M x R, representing the function values at X,
-        for K functions.
+        for R functions.
```
awav: Where is R definition? Do we have it in glossary?

awav: Where is R definition? Do we have it in glossary?

thevincentadam: Just above : 
`:param f: data matrix, M x R,...`
definitely no K!

thevincentadam: It is sort of the implicit letter for the usual `num_latent`

PR #964, opened 2019-04-03
proper shape checks for multivariate_normal

st--:
#914 (f8fc156ed41d5489ca61978a4e7bed43b0d22419) simply removed the shape checks of logdensities.multivariate_normal() that issued a warning, this puts them back properly as tensorflow graph assertions.

Without this, the following results in a segfault:
```python
import numpy as np
import tensorflow as tf
from gpflow.logdensities import multivariate_normal
A = np.random.randn(5)
B = np.random.randn(5)
L = np.tril(np.random.randn(5, 5))
vA = tf.placeholder(tf.float64)
vB = tf.placeholder(tf.float64)
vL = tf.placeholder(tf.float64)
res = multivariate_normal(vA, vB, vL)
sess = tf.Session()
sess.run(res, {vA: A, vB: B, vL: L})
```

With this PR, tensorflow instead raises an InvalidArgumentError exception with a helpful error message.
(Note that all DataHolders etc correspond to a tf.placeholder() with static shape None as above!)


gpflow/logdensities.py
```diff
@@ -85,17 +83,15 @@ def multivariate_normal(x, mu, L):
...
-    alpha = tf.matrix_triangular_solve(L, d, lower=True)
-    num_dims = tf.cast(tf.shape(d)[0], L.dtype)
-    p = - 0.5 * tf.reduce_sum(tf.square(alpha), 0)
-    p -= 0.5 * num_dims * np.log(2 * np.pi)
-    p -= tf.reduce_sum(tf.log(tf.matrix_diag_part(L)))
-    return p
+    with tf.control_dependencies([
+            assert_tensor_ndim(x, 2, 'multivariate_normal requires the shape of x to be (N, D)'),
+            assert_tensor_ndim(mu, 2, 'multivariate_normal requires the shape of mu to be (N, D)'),
+            tf.assert_equal(tf.shape(L), [tf.shape(x)[0], tf.shape(x)[0]]),
```
vdutor: It's a bit confusing that you write in the two lines above that `x` and `mu` should be `(N, D)`, which implies that you check in this assert whether `L` is `(N, N)`, but in the docstrings it is stated that `L` should be `(D, D)`. Can you resolve this mismatch?

st--: The docstring says x and mu should be (D, N), so I think it's all correct as is?

vdutor: then change the `x to be (N, D)`?

st--: Sorry, you're absolutely right - it was confusing. I've changed it!

gpflow/misc.py
```diff
@@ -285,5 +285,26 @@ def _get_tensor_safe(name, index, graph):
...
+    """
+    tensor_shape = tf.shape(tensor)
+    tensor_ndim = tf.shape(tensor_shape)
+    # The return value of tf.equal() is a 1-vector; turn it into a scalar using tf.reduce_all():
+    return tf.reduce_all(tf.equal(tensor_ndim, ndim))
+
+
+def assert_tensor_ndim(tensor: tf.Tensor, ndim: int, message: Optional[str]=None):
+    if message is None:
+        message = "assert_tensor_ndim: {}'s shape does not have ndim {}".format(tensor, ndim)
```
awav: The trace will point to the error line, there is no need in specifying the tensor name. Concise error message is preferable, too much information isn't good as well.

```suggestion
        message = "Tensor shape does not have ndim {}".format(ndim)
```

gpflow/misc.py
```diff
@@ -285,5 +285,26 @@ def _get_tensor_safe(name, index, graph):
...
+
+    if tensor.shape.ndims is not None:
+        if tensor.shape.ndims != ndim:
+            raise ValueError(message)
+    return tf.Assert(tensor_ndim_equal(tensor, ndim), [message])
```
awav: I would leave this assert without a message as it is self descriptive. (Btw, it is strange to have the same message as in raise)

```suggestion
    return tf.Assert(tensor_ndim_equal(tensor, ndim))
```

gpflow/logdensities.py
```diff
@@ -73,29 +71,27 @@ def laplace(x, mu, sigma):
 def multivariate_normal(x, mu, L):
     """
     Computes the log-density of a multivariate normal.
-    :param x  : Dx1 or DxN sample(s) for which we want the density
-    :param mu : Dx1 or DxN mean(s) of the normal distribution
-    :param L  : DxD Cholesky decomposition of the covariance matrix
-    :return p : (1,) or (N,) vector of log densities for each of the N x's and/or mu's
+    :param x  : N×1 or N×R sample(s) for which we want the density
```
awav:
```suggestion
    :param x  : [N, 1] or [N, R] sample(s) for which we want the density
```

gpflow/logdensities.py
```diff
@@ -73,29 +71,27 @@ def laplace(x, mu, sigma):
...
+    :param mu : N×1 or N×R mean(s) of the normal distribution
```
awav:
```suggestion
    :param mu : [N, 1] or [N, R] mean(s) of the normal distribution
```

gpflow/logdensities.py
```diff
@@ -73,29 +71,27 @@ def laplace(x, mu, sigma):
...
+    :param L  : N×N Cholesky decomposition of the covariance matrix
```
awav:
```suggestion
    :param L  : [N, N] Cholesky decomposition of the covariance matrix
```

gpflow/logdensities.py
```diff
@@ -73,29 +71,27 @@ def laplace(x, mu, sigma):
...
+    :return p : (1,) or (R,) vector of log densities for each of the columns of x and/or mu
 
-    x and mu are either vectors or matrices. If both are vectors (N,1):
+    x and mu are either vectors or matrices. If both are vectors (N×1):
```
awav:
```suggestion
    x and mu are either vectors or matrices. If both are vectors [N, 1]:
```

gpflow/logdensities.py
```diff
@@ -73,29 +71,27 @@ def laplace(x, mu, sigma):
...
-
-    d = x - mu
-    alpha = tf.matrix_triangular_solve(L, d, lower=True)
-    num_dims = tf.cast(tf.shape(d)[0], L.dtype)
-    p = - 0.5 * tf.reduce_sum(tf.square(alpha), 0)
-    p -= 0.5 * num_dims * np.log(2 * np.pi)
-    p -= tf.reduce_sum(tf.log(tf.matrix_diag_part(L)))
-    return p
+    with tf.control_dependencies([
+            assert_tensor_ndim(x, 2, 'multivariate_normal requires the shape of x to be (N, R)'),
```
awav:
```suggestion
            assert_tensor_ndim(x, 2, 'multivariate_normal requires the shape of x to be [N, R]'),
```

gpflow/logdensities.py
```diff
@@ -73,29 +71,27 @@ def laplace(x, mu, sigma):
...
+            assert_tensor_ndim(mu, 2, 'multivariate_normal requires the shape of mu to be (N, R)'),
```
awav:
```suggestion
            assert_tensor_ndim(mu, 2, 'multivariate_normal requires the shape of mu to be [N, R]'),
```

awav:
Overall is good and ready for merge. I left some "suggestions" to commit.

gpflow/misc.py
st--: It's often useful to know _which_ tensor is actually responsible for an error, no?

PR #965, opened 2019-04-03
DataHolder shape check

st--:
GPflow always requires the data tensors such as `X` and `Y` to be rank-2/ndim=2, e.g. input N x D, even when D=1. This is a common source of bugs for people new to GPflow, who might simply pass in a flat array when input/output are one-dimensional. This then leads to shape mismatches somewhere in the middle of the code, not with any helpful error messages giving an indication *why* it's not working. This PR adds a `check_ndim_is` flag to DataHolder which gives a more helpful error message when passing wrongly-shaped arrays to such a DataHolder.


st--:
@awav can you please approve this PR or state what you would like changed, and in which way, or why you think this should not be merged?

condnsdmatters:
Can this PR be closed? It seems to be a gpflow 1 PR - are improvements still underway?

PR #968, opened 2019-04-22
URGENT: FIX TensorFlow api compatibility problem

TITAN-PyCompat:
Dear developers,

I found some compatibility problems in your use of TensorFlow APIs, which might lead to crashes in a low version of TensorFlow. And I fix it for you as below:

`tf.variables_initializer` is the updated version of tf.initialize_variables, however, it appears until tf 0.12.0-rc0. So for version before that, it's better to use `tf.initialize_variables` to be safe.

`tf.global_variables_initializer` is the updated version of `tf.initialize_all_variables`, however, it also appears until tf 0.12.0-rc0. So for version before that, it's better to use `tf.initialize_all_variables` to be safe.

@davidsheldon @jameshensman @icouckuy @vals @javdrher 
Hope you could take my advice and look forward to your reply! 😄

### DIFF
--- a/gpflow/misc.py
+++ b/gpflow/misc.py
@@ -199,7 +199,11 @@ def vec_to_tri(vectors, N):
 def initialize_variables(variables=None, session=None, force=False, **run_kwargs):
     session = tf.get_default_session() if session is None else session
     if variables is None:
-        initializer = tf.global_variables_initializer()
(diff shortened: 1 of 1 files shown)


awav:
Hello @TITAN-PyCompat, thanks for your PR. Unfortunately, we don't have plans to support old versions of TF. We already have a lot of features incompatible with < TF1.12 (e.g. broadcasting)

PR #990, opened 2019-06-18
Make ScipyOptimizer reusable (without growing tf graph)

st--:
Currently, the ScipyOptimizer only exposes a `minimize()` method that adds ops to the default graph, thereby slowing down performance when run repeatedly (e.g. in online learning or backtesting). This is an updated version of awav/reuse-scipyoptimizer.


st--:
I know this is a bit hacky, but it's something that I actually need and it'd be much easier if I wouldn't have to copy these files manually/work off a feature branch ... and I know other people need this feature as well. Could we please get this merged? @awav if you've got any ideas for how to clean it up before merging please say!

awav:
Changes made in original branch

Issue #995, opened 2019-07-05
Cannot use TensorFlow optimizer (e.g. AdamOptimizer) with TF v1.14


wiseodd:
System information:

* Python version: 3.7.3
* TensorFlow installed from: binary
* TensorFlow version: 1.14
* GPflow installed from: binary
* GPflow version: 1.4.1

Just do `gpflow.train.AdamOptimizer()`, it will throw an error saying that it's not found. Downgrading to TF version 1.13.1 fixes this.

awav:
https://stackoverflow.com/questions/56741160/importerror-cannot-import-name-adamoptimizer-in-gpflow

PR #997, opened 2019-07-09
GPflow TF 1.14 compatible. Fixes optimizers importing issue.

vdutor:
This fixes issue https://github.com/GPflow/GPflow/issues/995 and makes GPflow compatible with the latest TF version 1.14.

Details:
TensorFlow wrapped `tf.train` in a module deprecation-wrapper, so we simply go and get the items from inside the wrapper.

### DIFF
--- a/gpflow/training/tensorflow_optimizer.py
+++ b/gpflow/training/tensorflow_optimizer.py
@@ -13,8 +13,8 @@
 # limitations under the License.
 
 import sys
-
 import tensorflow as tf
+from pkg_resources import parse_version
 
 from . import optimizer
 from .. import misc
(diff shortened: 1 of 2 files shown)


gpflow/training/tensorflow_optimizer.py
```diff
@@ -164,7 +164,13 @@ def _register_optimizer(name, optimizer_type):
 
 
 # Create GPflow optimizer classes with same names as TensorFlow optimizers
-for key, train_type in tf.train.__dict__.items():
+
+if parse_version(tf.VERSION) >= parse_version("1.14"):
+    train_items = tf.train._dw_wrapped_module.__dict__.items()
```
st--:
```suggestion
    # tensorflow 1.14 deprecated the `train` module and wraps it in a DeprecationWrapper
    # which requires this hack to get access to the list of objects in the namespace
    train_items = tf.train._dw_wrapped_module.__dict__.items()
```

PR #1004, opened 2019-07-17
Small updates for convolutional kernel

markvdw:


gpflow/kernels.py
```diff
@@ -841,7 +841,19 @@ def on_separate_dimensions(self):
 class Convolutional(Kernel):
     """
     Conv
```
st--:
```suggestion
```

gpflow/kernels.py
```diff
@@ -906,6 +919,19 @@ def compute_patches(self, X):
 
 
 class WeightedConvolutional(Convolutional):
+    """
+    WeightedConvolutional
```
st--:
```suggestion
```

gpflow/features.py
```diff
@@ -150,14 +150,26 @@ def Kuu(feat, kern, *, jitter=0.0):
 
 
 class InducingPatch(InducingPointsBase):
+    """
+    InducingPatch
```
st--:
```suggestion
```
Having the class name in the docstring is superfluous..

doc/source/train_mnist.py
```diff
@@ -0,0 +1,419 @@
+# Copyright 2019 Mark van der Wilk, Vincent Dutordoir
```
st--: Do you expect this file to be included in the sphinx documentation? If not, would this file fit better in a dedicated `examples` directory or so?

markvdw: Yes, I was not expecting this to be in the Sphinx docs. Do you have any suggestions for a dir?

st--: GPflow.git/examples ?

st--: Though tbh I think a notebook would work better (it would make it more visible, could include some visualisations, and it would be included in the CI build to ensure it won't break unnoticedly at some point in the future).

requirements_tests.txt
```diff
@@ -0,0 +1 @@
+sacred==0.7.4
```
awav: Don't think it is necessary

markvdw: Why not?

awav: ^

examples/train_mnist.py
```diff
@@ -0,0 +1,425 @@
+# Copyright 2019 Mark van der Wilk, Vincent Dutordoir
```
awav: Why do you need sacred to test the functionality?

PR #1019, opened 2019-08-12
GPflow docs copyedit: Manipulating models

akhayes:
Check for:

mistakes/typos
house style/formatting
consistency
sufficient (or missing) cross-linkage between documents
clarify unclear sections/explanations
(See https://prowlerio.atlassian.net/browse/PTKB-6960)


doc/source/notebooks/understanding/models.ipynb
```diff
@@ -848,11 +848,11 @@
...
     "\n",
-    "In this very simple demo, we'll implement linear multiclass classification. There will be two parameters: a weight matrix and a 'bias' (offset). The key thing to implement is the private `_build_likelihood` method, which should return a tensorflow scalar representing the (log) likelihood. By decorating the function with `@gpflow.params_as_tensors`, Param objects can be used inside `_build_likelihood`: they will appear as appropriate (constrained) tensors."
+    "In this very simple demo, we'll implement linear multiclass classification. There are two parameters: a weight matrix and a 'bias' (offset). The key thing to implement is the private `_build_likelihood` method, which returns a TensorFlow scalar that represents the (log) likelihood. If you decorate the function with `@gpflow.params_as_tensors`, you can use `Param` objects inside `_build_likelihood`; they appear as appropriate (constrained) tensors."
```
st--: I think the last sentence can be given a bit more meaning - verbose explanation: you can always use `Param` objects anywhere you want, but normally you would have to type, e.g., `self.kernel.variance.constrained_tensor` to get at the tf tensor to use when building a tf Graph - when a method is decorated by `@gpflow.params_as_tensors`, you only need to type `self.kernel.variance` instead, which is now the tf tensor object (and you can't access any of the other features of a Param object) - this is particularly useful when writing for example the model likelihood as there'll be lots of references to constrained_tensor attributes of various params.

doc/source/notebooks/understanding/models.ipynb
```diff
@@ -36,7 +36,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "Create a very simple GPR model without building it in TensorFlow graph."
+    "Create a very simple GP regression model without building it in the TensorFlow graph:"
```
st--: maybe worth highlighting this a bit more - we *need* to wrap it in the `defer_build` because some of the things demonstrated in the following, such as adding transforms and priors, only works for *non*-compiled models, not necessarily obvious to people new to gpflow..

PR #1034, opened 2019-08-22
Added convolutional code

markvdw:
I'm still going to compare the speed of this vs GPflow 1.0.


markvdw:
The speed is roughly equivalent, as you expected.

gpflow/inducing_variables/convolutional_patch.py
```diff
@@ -0,0 +1,5 @@
+from .inducing_variables import InducingPoints
```
vdutor: would rename this file to `inducing_patch.py`, makes a bit more sense imo.

gpflow/kernels/convolutional.py
```diff
@@ -0,0 +1,120 @@
...
+    """
+
+    def __init__(self, basekern, img_size, patch_size, colour_channels=1):
+        super().__init__()
+        self.img_size = img_size
+        self.patch_size = patch_size
+        self.basekern = basekern
+        self.colour_channels = colour_channels
+
+        # NOTE: basekern now is input-dim independent, so no longer need to check this?
```
vdutor: I would just remove this note

gpflow/kernels/convolutional.py
```diff
@@ -0,0 +1,120 @@
...
+    @property
+    def patch_len(self):
+        return np.prod(self.patch_size)
+
+    @property
+    def num_patches(self):
+        return (self.img_size[0] - self.patch_size[0] + 1) * (
+                self.img_size[1] - self.patch_size[1] + 1) * self.colour_channels
+
+    # @autoflow((settings.float_type,))
```
vdutor: best to delete this. No more autoflow in gpflow2

gpflow/kernels/base.py
```diff
@@ -119,19 +119,19 @@ def slice_cov(self, cov: tf.Tensor) -> tf.Tensor:
...
-        pass
+        raise NotImplementedError
 
     def __call__(self, X, Y=None, full=True, presliced=False):
         if not full and Y is not None:
             raise ValueError(
                 "Ambiguous inputs: `diagonal` and `y` are not compatible.")
         if not full:
-            return self.K_diag(X, presliced=presliced)
-        return self.K(X, Y, presliced=presliced)
```
st--: What's the intent behind no longer passing through the `presliced` argument ?

gpflow/kernels/convolutional.py
```diff
@@ -0,0 +1,120 @@
+import numpy as np
+import tensorflow as tf
+
+from ..base import Parameter, positive
+from ..utilities.ops import square_distance
+from .base import Kernel
+from ..config import default_float
+
+
+class Convolutional(Kernel):
```
vdutor: I think it would make sense to go for a single `Convolutional` class which has `weights=np.ones()` or `weights=Parameter(np.ones())` depending on an `init` parameter. Quite easy to do and introduces less code duplication.

gpflow/kernels/base.py
markvdw: `presliced` was only used for certain bits in the kernel expectation quadrature. In correspondence with @awav and @vincent we decided to simplify the kernel expectations. A consequence of this is that we will remove certain quadrature fallback behaviour that required the `presliced` behaviour. This simplifies kernels as well.

I will remove all `presliced` references in a separate PR.

gpflow/kernels/convolutional.py
markvdw: @awav said this as well. Will do.

PR #1035, opened 2019-08-23
Mark/gpflow 2.0/kl dispatch

markvdw:
Simple multiple dispatch for a KL function. This is necessary to support faster inference with more structured KL divergences. E.g. VFF.


st--:
The prior distribution p(0, Kuu) is defined by kernel and inducing variables, which the changes in this PR dispatch on, but the KL[q(u)||p(u)] also needs to dispatch on the q(u) for more structured variational approximations (as in the VFF example you mention). Would probably need some kind of class hierarchy for distributions as well -- at which point the function that we define is "just" a KL (not specifically for the prior), and it dispatches on the types of p(u) and q(u)  (and p(u) can then be constructed out of Kuu, but doesn't need to know about kernel/inducing variables directly).

PR #1041, opened 2019-09-06
Fix ImageToTensorboardTask graph pollution

st--:
`gpflow.training.monitor.ImageToTensorboardTask` creates new tensorflow operations each time its `run()` method is called. This PR fixes that behaviour, by separating op construction (now moved into `__init__`) from running them in `run()`.


st--:
Yes, not adding *any* ops to the graph throughout optimisation actually gave me a 10% speedup.
It will probably look quite different in gpflow2, but I still think it would be good to get this PR merged in as soon as possible.

gpflow/training/monitor.py
```diff
@@ -1050,8 +1050,15 @@ def __init__(self, file_writer: LogdirWriter, func: Callable[[], Figure], func_n
 
         super().__init__(file_writer)
         self.func = func
-        self.placeholder = tf.placeholder(tf.float64, [1, None, None, None])
-        self._summary = tf.summary.image(func_name, self.placeholder)
+
+        self.buf_placeholder = tf.placeholder(tf.string)
+        # Create TF image and load its content from the buffer.
+        image_tensor = tf.image.decode_png(self.buf_placeholder)
```
vdutor:
```suggestion
        self.image_tensor = tf.image.decode_png(self.buf_placeholder)[None]
```

vdutor: Could potentially cleanup the code a bit. If you accepts this change, you can delete the next two lines.

st--: Cleaned it up even more, feeding the image_tensor directly to tf.summary.image() instead of having an intermediate tf.placeholder and two separate session.run() calls

PR #1042, opened 2019-09-09
Add setters to SettingsManager aliases such as float_type

st--:
The `SettingsManager` (singleton accessed as `gpflow.settings`) provides a number of aliases as shortcuts, defined as a property, e.g. `gpflow.settings.float_type` is equivalent to `gpflow.settings.dtypes.float_type`. However, changing the float_type in the same way did not work, as the property did not define a setter. This PR adds the missing setter definitions.


st--:
Alright, let's drop gpflow1!

PR #1048, opened 2019-09-13
GPflow 2.0 notebook update: GPs for big data

st--:


st--:
NB- there seems to be a bug, calling `run_adam()` multiple times fails.

st--:
Requires #1063 to be merged first

PR #1052, opened 2019-09-17
Periodic++

jahall:
This is a proposed **enhancement** which introduces periodic versions of the Matern family of kernels, to supplement the standard SE-based one. I have found these to be useful to have in my toolkit since the ubiquitous SE one is often too darn smooth!

### DIFF
--- a/gpflow/kernels.py
+++ b/gpflow/kernels.py
@@ -786,9 +786,9 @@ class PeriodicMatern32(Periodic):
 
 class PeriodicMatern52(Periodic):
     """
-    A periodic version of the Matern 3/2 kernel.
+    A periodic version of the Matern 5/2 kernel.
 
-    Derived using the Matern 3/2 kernel once mapped the original inputs through
+    Derived using the Matern 5/2 kernel once mapped the original inputs through
     the mapping u=(cos(x), sin(x)).
 
     The resulting kernel can be expressed as:
--- a/gpflow/kernels.py
+++ b/gpflow/kernels.py
@@ -693,7 +693,7 @@ class Periodic(Kernel):
     The derivation can be achieved by mapping the original inputs through the
     transformation u = (cos(x), sin(x)).
 
-    The resulting RBF-based kernel can be expressed as:
+    For the SquaredExponential base kernel, the result can be expressed as:
         k(r) = σ² exp{ -0.5 sin²(π r / γ) / ℓ² }
 
     where:
--- a/gpflow/kernels.py
+++ b/gpflow/kernels.py
@@ -706,52 +706,49 @@ class Periodic(Kernel):
     is absorbed into lengthscale hyperparameter).
     """
     def __init__(self, input_dim=None, period=1.0, variance=1.0,
-                 lengthscales=1.0, basekern=None, active_dims=None, name=None):
+                 lengthscales=1.0, base=None, active_dims=None, name=None):
         """
         The Periodic kernel supports the specification of any Stationary kernel
-        as the base through the `basekern` parameter. If this is provided then
-        `input_dim`, `variance`, `lengthscales` and `active_dims` parameters are
+        as the base through the "base" parameter. If this is provided then
+        input_dim, variance, lengthscales and active_dims parameters are
         taken from there.
 
-        If `basekern` is not provided, then a SquaredExponential is constructed
-        implicitly using the relevant arguments:
-
-            basekern = SquaredExponential(
-                input_dim,
-                period=period,
-                lengthscales=lengthscales,
-                active_dims=active_dims,
-            )
+        If a "base" kernel is not provided, then a SquaredExponential is
+        constructed implicitly using the input_dim, variance, lengthscales
+        and active_dims parameters.
         """
-        if basekern is None:
+        if (base is not None) and (input_dim is not None):
+            raise ValueError("input_dim should be defined through the base kernel, not explicitly")
+
+        if base is None:
             warnings.warn(
                 'Implicit specification of the base kernel for Periodic is '
-                'deprecated, use Periodic(basekern=SquaredExponential({i}, '
+                'deprecated, use Periodic(base=SquaredExponential({i}, '
                 'variance={v}, lengthscales={l}, active_dims={a}), period={p}) '
                 'instead.'.format(i=input_dim, v=variance, l=lengthscales,
                                   a=active_dims, p=period),
                 DeprecationWarning)
-            basekern = SquaredExponential(
+            base = SquaredExponential(
                 input_dim,
                 variance=variance,
                 lengthscales=lengthscales,
                 active_dims=active_dims)
 
-        if not isinstance(basekern, Stationary):
-            raise TypeError("Periodic requires a Stationary kernel as the `basekern`")
+        if not isinstance(base, Stationary):
+            raise TypeError("Periodic requires a Stationary kernel as the `base`")
 
-        super().__init__(basekern.input_dim, basekern.active_dims, name=name)
-        self.basekern = basekern
+        super().__init__(base.input_dim, base.active_dims, name=name)
+        self.base = base
         self.period = Parameter(  # No ARD support for period yet
             period, transform=transforms.positive, dtype=settings.float_type)
 
     @property
     def ARD(self):
-        return self.basekern.ARD
+        return self.base.ARD
 
     @params_as_tensors
     def Kdiag(self, X, presliced=False):
-        return self.basekern.Kdiag(X)
+        return self.base.Kdiag(X)
 
     @params_as_tensors
     def K(self, X, X2=None, presliced=False):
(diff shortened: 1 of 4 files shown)


st--:
Hi @joehall87, this is a great addition to GPflow. However, you rewrote all the implementations of K(r) for the various Matern kernels that are already in GPflow: all subclasses of `Stationary` (such as `Matern12`, `Matern32`, `Matern52`, but also `Cosine` kernel) implement a `K_r(r)` method which is doing essentially the same job as your `_K_from_sin_r`. Could you rework your PR to simply wrap *any* stationary kernel (i.e. one that defines `K_r`) and feed it the appropriate symmetrised abs(sin(πr / γ) / ℓ)? E.g. the interface could look like
```python
k = Periodic(Matern32(1, lengthscales=ℓ), period=γ))  # passing in an object to be wrapped
```
or
```python
k = Periodic(Matern32, lengthscales=ℓ, period=γ)  # passing in a class constructor
```
and we could specify a default behaviour to be backwards-compatible... thoughts, anyone ?

joehall87:
This is a great idea! Being able to extend to arbitrary stationary kernels would be awesome. I think the first interface is cleanest. Let me rework my PR accordingly - and ensure that current behaviour of `Periodic` is backward compatible.

joehall87:
Ok, in order to retain backward compatibility I went with the second option suggested by @st--. The new version will now allow the specification of an arbitrary Stationary kernel as the base class.

jahall:
Hi there - just wondering if this looks good to folks, or whether it is out of scope for GPflow?

gpflow/kernels.py
```diff
@@ -683,42 +683,42 @@ def Kdiag(self, X, presliced=False):
 
 class Periodic(Kernel):
     """
-    The periodic kernel. Defined in  Equation (47) of
+    The periodic family of kernels. Can be used to wrap any Stationary kernel
+    to transform it into a periodic version. The canonical form (based on the
+    RBF kernel) can be found in Equation (47) of
```
st--:
```suggestion
    SquaredExponential kernel) can be found in Equation (47) of
```

gpflow/kernels.py
```diff
@@ -683,42 +683,42 @@ def Kdiag(self, X, presliced=False):
...
     Neural Networks and Machine Learning, pages 133--165. Springer, 1998.
 
-    Derived using an RBF kernel once mapped the original inputs through
-    the mapping u=(cos(x), sin(x)).
+    The derivation can be achieved by mapping the original inputs through the
+    transformation u = (cos(x), sin(x)).
 
-    The resulting kernel can be expressed as:
-        k(r) =  σ² exp{ -0.5 sin²(π r / γ) / ℓ²}
+    The resulting RBF-based kernel can be expressed as:
```
st--:
```suggestion
    For the SquaredExponential base kernel, the result can be expressed as:
```

gpflow/kernels.py
```diff
@@ -683,42 +683,42 @@ def Kdiag(self, X, presliced=False):
...
+        k(r) = σ² exp{ -0.5 sin²(π r / γ) / ℓ² }
 
     where:
-    r  is the Euclidean distance between the input points
+    r is the Euclidean distance between the input points
```
st--:
```suggestion
    r is the Euclidean distance between the input points,
```

gpflow/kernels.py
```diff
@@ -683,42 +683,42 @@ def Kdiag(self, X, presliced=False):
...
     σ² is the variance parameter,
     γ is the period parameter.
 
     (note that usually we have a factor of 4 instead of 0.5 in front but this is absorbed into lengthscale
     hyperparameter).
     """
-
-    def __init__(self, input_dim, period=1.0, variance=1.0,
-                 lengthscales=1.0, active_dims=None, name=None):
-        # No ARD support for lengthscale or period yet
```
st--: Is this comment still applicable - it still sets `self.ARD = False` below? Should this just be a pass-through property that returns `self.base.ARD`?

gpflow/kernels.py
```diff
@@ -683,42 +683,42 @@ def Kdiag(self, X, presliced=False):
...
+    def __init__(self, input_dim, period=1.0, variance=1.0, lengthscales=1.0,
+                 base_class=SquaredExponential, active_dims=None, name=None, **kw):
+        if not issubclass(base_class, Stationary):
+            raise TypeError("the base class for Periodic must be a stationary kernel")
```
st--:
```suggestion
            raise TypeError("Periodic requires a Stationary kernel as the base_class")
```

st--:
@jahall I left a few minor PR comments - can't easily fix it myself as I don't have edit permissions on your branch, but once those are sorted out I'd be happy to merge this into develop.

gpflow/kernels.py
jahall: I guess the comment is still applicable in that there is still no ARD support for period.

However, I will update self.ARD to be a pass-through and maybe in a follow-up PR we can add support for different periods per dimension...in which case we might want to rename period -> periods?

jahall:
Thanks for this @st--! I have merged your comments.

Also, I altered the code to adhere to your first suggestion of passing in an _instantiated_ base class i.e.
```python
k = Periodic(base=Matern32(1, lengthscales=ℓ), period=γ)
```
as I think this is cleaner, and making it backward compatible is possible if we just put the new `base` kwarg down the list of existing kwargs.

And I expanded the test suite to include a 2d with lengthscale ARD test.

jahall:
@st-- Do these changes look good from your perspective?

gpflow/kernels.py
```diff
@@ -683,42 +683,59 @@ def Kdiag(self, X, presliced=False):
...
+    is absorbed into lengthscale hyperparameter).
+    """
+    def __init__(self, input_dim=None, period=1.0, variance=1.0,
+                 lengthscales=1.0, base=None, active_dims=None, name=None):
+
+        if base is None:
+            warnings.warn(
+                'Indirect specification of the base kernel for Periodic is '
+                'deprecated, use Periodic(base=SquaredExponential({i}, variance={v}, '
+                'lengthscales={l}), period={p}) instead.'.format(
```
st--: What about `active_dims`?

gpflow/kernels.py
```diff
@@ -683,42 +683,59 @@ def Kdiag(self, X, presliced=False):
...
     γ is the period parameter.
 
-    (note that usually we have a factor of 4 instead of 0.5 in front but this is absorbed into lengthscale
-    hyperparameter).
-    """
+    (note that usually we have a factor of 4 instead of 0.5 in front but this 
+    is absorbed into lengthscale hyperparameter).
+    """
+    def __init__(self, input_dim=None, period=1.0, variance=1.0,
+                 lengthscales=1.0, base=None, active_dims=None, name=None):
```
st--: Might be good to have an `__init__` docstring that points out you should *either* pass in `base=...` *or* the arguments as before (in which case it will implicitly construct a SquaredExponential base kernel)?

st--:
Thanks for the prod! I think it looks good, if you could just fix the docstring/warning to account for active_dims and make sure it's clear what is happening, and if no-one else has any comments, I'll go and merge it!

jahall:
Yep, I'll go ahead and do that. One other question: I see that the `Convolutional` kernel uses the nomenclature `basekern`. Do you think its worth renaming `base` -> `basekern` for consistency?

gpflow/kernels.py
jahall: Added docstring in latest commit.

gpflow/kernels.py
jahall: Added active_dims in latest commit.

st--:
No, I think `base` is much more sensible and I'd rather also change the Convolutional kernel to use `base` as well.

gpflow/kernels.py
```diff
@@ -683,42 +683,75 @@ def Kdiag(self, X, presliced=False):
...
+        """
+        The Periodic kernel supports the specification of any Stationary kernel
+        as the base through the `basekern` parameter. If this is provided then
+        `input_dim`, `variance`, `lengthscales` and `active_dims` parameters are
+        taken from there.
+
+        If `basekern` is not provided, then a SquaredExponential is constructed
+        implicitly using the relevant arguments:
+
+            basekern = SquaredExponential(
```
st--: the following isn't quite right (it doesn't pass the period, it does pass the variance), but you could also just say "is constructed implicitly using the input_dim, variance, lengthscales, active_dims arguments".

gpflow/kernels.py
```diff
@@ -683,42 +683,75 @@ def Kdiag(self, X, presliced=False):
...
+                'variance={v}, lengthscales={l}, active_dims={a}), period={p}) '
+                'instead.'.format(i=input_dim, v=variance, l=lengthscales,
+                                  a=active_dims, p=period),
+                DeprecationWarning)
+            basekern = SquaredExponential(
+                input_dim,
+                variance=variance,
+                lengthscales=lengthscales,
+                active_dims=active_dims)
+
```
st--: maybe add a
```python
if (base is not None) and (input_dim is not None):
    raise ValueError("input_dim is defined by the base kernel")
```
or similar?

gpflow/kernels.py
jahall: Oops, will fix!

gpflow/kernels.py
jahall: Good idea

st--:
@jahall thanks for your contribution! I'd like to make sure that this feature will also be present in gpflow2 - I think there might be an issue that needs resolving first (#1077), but would you be up to converting this code to tensorflow2 (and gpflow2) and make a similar PR into the `develop-2.0` branch?

jahall:
@st-- Sure thing - let me take a look

PR #1060, opened 2019-09-25
remove default argument to set_trainable (was False)

st--:
Previously, calling `set_trainable(model)` actually set it to *non*trainable, which is rather confusing.


condnsdmatters:
:heavy_check_mark:  Verified that all calls of set_trainable are explicit (using `git grep set_trainable`)
LGTM subject to tests passing.

condnsdmatters:
Is that code coverage test report correct though?

PR #1062, opened 2019-09-25
Notebook for Variational Fourier Features

st--:
- Adds a notebook showcasing how to add new types of inter-domain inducing variables at the example of representing a sparse GP in the spectral domain using variational Fourier features.
- Cleans up some of the Kuf/Kuu multi-dispatch definitions
- Fixes a bug when computing the objective of completely non-trainable models that was uncovered in porting this notebook (includes regression test)


gpflow/covariances/kufs.py
```diff
@@ -1,16 +1,26 @@
...
 
 
-@Kuf.register(InducingPoints, Kernel, object)
-def _Kuf(inducing_variable: InducingPoints, kernel: Kernel, Xnew: tf.Tensor):
+@Kuf.register(InducingVariables, Kernel, TensorLike)
+def Kuf_fallback(inducing_variable, kernel, X):
+    from warnings import warn
+    warn('Kuf(inducing_variable, kernel, X) is deprecated, please use '
+         'Kuf(kernel, inducing_variable, X) instead.',
+         DeprecationWarning)
```
markvdw: Why did you add this warning? This isn't the case, is it? I.e. `Kuf(inducing_variable, kernel, X)` is *not* deprecated, right?

gpflow/covariances/kufs.py
```diff
@@ -1,16 +1,22 @@
...
+from ..inducing_variables import InducingVariables, InducingPoints, Multiscale, InducingPatches
 from ..kernels import Kernel, SquaredExponential, Convolutional
 from .dispatch import Kuf
 
 
-@Kuf.register(InducingPoints, Kernel, object)
-def _Kuf(inducing_variable: InducingPoints, kernel: Kernel, Xnew: tf.Tensor):
+@Kuf.register(InducingVariables, Kernel, TensorLike)
+def Kuf_fallback(inducing_variable, kernel, X):
+    return Kuf(kernel, inducing_variable, X)
```
markvdw: I think we should not have this fallback, which allows parameter types to be switched. Allowing functions to be called with reversed parameters presents an inconsistent interface.

PR #1065, opened 2019-10-01
Change points

jahall:
Introduction of a `ChangePoints` kernel to allow the identification of regime changes in 1d functions or timeseries. Examples provided in the associated notebook. I would welcome feedback on the design, and whether this would be something useful to have in the core library.


jahall:
Does anyone have a view on the value of this change?

frgsimpson:
> Does anyone have a view on the value on this enhancement? Or have any proposed alterations? (@st-- ?)

I think it would be a nice addition - I'll try to take a closer look soon!

jahall:
> > Does anyone have a view on the value on this enhancement? Or have any proposed alterations? (@st-- ?)
> 
> I think it would be a nice addition - I'll try to take a closer look soon!

Perfect, thanks!

gpflow/kernels.py
```diff
@@ -1005,6 +1006,94 @@ def Kdiag(self, X, presliced=False):
...
+
+    @params_as_tensors
+    def Kdiag(self, X):
+        return tf.matrix_diag_part(self.K(X))
+
+    def _sigmoids(self, X):
+        locations = tf.sort(self.locations)  # Ensure locations are ordered
+        locations = tf.reshape(locations, (1, 1, -1))
+        steepness = 2 * np.log(95.0) / tf.reshape(self.widths, (1, 1, -1))
+        return 0.5 + 0.5 * tf.tanh(0.5 * steepness * (X - locations))
```
frgsimpson: Is this equivalent to: 
 return tf.sigmoid(steepness * (X - locations))?

jahall: Yes! Good spot - this started out life in MATLAB which doesn't have that.

jahall: Fixed

gpflow/kernels.py
```diff
@@ -1005,6 +1006,94 @@ def Kdiag(self, X, presliced=False):
         return reduce(tf.multiply, [k.Kdiag(X) for k in self.kernels])
 
 
+class ChangePoints(Combination):
+    """
```
frgsimpson:
```suggestion
    r"""
```
Turn it into a raw string when using LaTeX commands (`\cite` below) to ensure the `\c` does not get interpreted as a special character (e.g. `\t` = tab, `\n` = newline).

gpflow/kernels.py
```diff
@@ -1005,6 +1006,94 @@ def Kdiag(self, X, presliced=False):
...
+            raise ValueError("ChangePoints kernel only valid for 1d inputs")
+
+        if not isinstance(locations, Iterable):
+            locations = [locations]
+        if not isinstance(widths, Iterable):
+            widths = [widths]
+
+        self.locations = Parameter(locations, transform=None, dtype=settings.float_type)
+        self.widths = Parameter(widths, transform=transforms.positive, dtype=settings.float_type)
+        self.activation_order = list(activation_order or range(len(kernels)))
```
frgsimpson: Generally it's more robust to make explicit use of the ternary if-else operator (and necessary when dealing with numpy arrays):
```suggestion
        self.activation_order = list(activation_order if activation_order is not None else range(len(kernels)))
```
What would happen if `activation_order` is passed in as a numpy array? Would that be supported?

frgsimpson: What was your thinking behind including `activation_order` as an explicit argument (instead of simply relying on the order of `kernels` passed in)?

gpflow/kernels.py
```diff
@@ -1005,6 +1006,94 @@ def Kdiag(self, X, presliced=False):
...
+        """
+        Total number of change-points defined by kernel.
+        """
+        return len(self.activation_order) - 1
+
+    @params_as_tensors
+    def K(self, X, X2=None):
+        sig_X = self._sigmoids(X)
+        sig_X2 = self._sigmoids(X2) if X2 is not None else sig_X
+        starters = sig_X * tf.transpose(sig_X2, perm=(1, 0, 2))
```
frgsimpson: would be good to have a comment here explaining what `starters` and `stoppers` are

gpflow/kernels.py
```diff
@@ -1005,6 +1006,94 @@ def Kdiag(self, X, presliced=False):
...
+                K *= starters[:, :, i-1]
+            if i < self.num_changepoints:
+                K *= stoppers[:, :, i]
+            regime_Ks.append(K)
+
+        return reduce(tf.add, regime_Ks)
+
+    @params_as_tensors
+    def Kdiag(self, X):
+        return tf.matrix_diag_part(self.K(X))
```
frgsimpson: Could this be evaluated without constructing the full `K` matrix, by calling into the individual kernels' `Kdiag` instead?

tests/test_kerns.py
```diff
@@ -577,6 +577,61 @@ def test(self):
...
+                gpflow.kernels.ChangePoints([gpflow.kernels.Constant(1), gpflow.kernels.Constant(1)],
+                                            activation_order=[0, 3])
+            with self.assertRaisesRegexp(ValueError, r"Number of active kernels \(3\) must be one more .*"):
+                gpflow.kernels.ChangePoints([gpflow.kernels.Constant(1), gpflow.kernels.Constant(1)],
+                                            activation_order=[0, 1, 0])
+            with self.assertRaisesRegexp(ValueError, r"Number of widths \(2\) does not match number of locations \(1\)"):
+                gpflow.kernels.ChangePoints([gpflow.kernels.Constant(1), gpflow.kernels.Constant(1)],
+                                            locations=[1.0], widths=[0.5, 1.0])
+
+    def test_single(self):
```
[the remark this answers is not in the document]
frgsimpson: To clarify: would be good to have test of a) `ChangePoints` kernel with more than two individual kernels, and b) a "hierarchical" kernel with a `ChangePoints` kernel where one of the individual kernels is itself a `ChangePoints` kernel (this should get flattened out by the `Combination` base class, but good to test anyways).

tests/test_kerns.py
```diff
@@ -577,6 +577,61 @@ def test(self):
         super().test()
 
 
+class TestChangePoints(GPflowTestCase):
```
frgsimpson: We're trying to move towards pytest (function) style tests instead of subclassing TestCase - this often makes tests much simpler, and will also make it easier to port to gpflow2, would be great if you could change these accordingly (please ask if you've got questions how or pointers to good examples)!

gpflow/kernels.py
```diff
@@ -1005,6 +1006,94 @@ def Kdiag(self, X, presliced=False):
...
+
+    The kernel is by multiplication and addition of the base kernels with
+    sigmoidal functions (σ). A single change-point kernel is defined as:
+
+        K₁(x, x') * (1 - σ(x)) * (1 - σ(x')) + K₂(x, x') * σ(x) * σ(x')
+
+    where K₁ is deactivated around the change-point and K₂ is activated. The
+    single change-point version can be found in \citet{ldgtg2014}.
+
+    @incollection{ldgtg2014,
```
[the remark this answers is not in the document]
jahall: Good plan - will do this

gpflow/kernels.py
jahall: Good catch

gpflow/kernels.py
jahall: The thinking behind activation order is to support the more general case when you want to define a "change window" i.e. k1 changes to k2 then back to k1. `activation_order` allows this by specifying: `ChangePoints([k1, k2], locations=[1., 2.], activation_order=[0, 1, 0])`

My suspicion was that if I did `ChangePoints([k1, k2, k1], locations=[1., 2.])` then trained the kernel, the weights of the two k1 kernels would become decoupled?

gpflow/kernels.py
jahall: will do

tests/test_kerns.py
jahall: Great stuff - yeah I've had a look at gpflow2 code (currently trying to port my generalised periodic kernel there) - looks like a great improvement over the current structure and design choices

awav:
I recommend  a couple of things to do:

* Contribute directly to the GPflow 2.0 (develop-2.0). The GPflow 1.0 will be in the maintenance mode very soon, that means that all new features should go to the 2.0 not in 1.0. It is also less work for you. But, of course, it is up to you.
* Focus *only* on `pytest` even for GPflow 1.0. The `UnitTest` deprecated in GPflow 2.0.

jahall:
> I recommend a couple of things to do:
> 
> * Contribute directly to the GPflow 2.0 (develop-2.0). The GPflow 1.0 will be in the maintenance mode very soon, that means that all new features should go to the 2.0 not in 1.0. It is also less work for you. But, of course, it is up to you.
> * Focus _only_ on `pytest` even for GPflow 1.0. The `UnitTest` deprecated in GPflow 2.0.

Thanks @awav, makes sense - I will address @frgsimpson's comments, port to develop-2.0, close this PR and raise a new one.

tests/test_kerns.py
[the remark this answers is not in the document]
jahall: `ChangePoints` currently won't support the hierarchical flattening - but I have plans to introduce that

gpflow/kernels.py
jahall: Yes absolutely - I was too lazy in this first iteration to figure out the right way to do that

PR #1070, opened 2019-10-02
Update install instructions in README.md

awav:
^^^


st--:
I've updated the setup to make it simpler in #1074 ! have a look at that instead please :)

PR #1072, opened 2019-10-03
GPflow2.0 notebook update: coregionalisation.ipynb

fleibfried:


fleibfried:
@st-- , I updated the notebook using the latest version from develop

st--:
Changing the scipy optimizer method to `CG` seemed to make it work - though it's still a bit worrisome that we can't get it to behave as well as in GPflow 1 (`develop`). Should we merge this into develop-2.0 anyways and leave addressing #1090 until later?

PR #1073, opened 2019-10-04
Sync notebooks with develop (includes moving files from notebooks2 back to doc/source/notebooks)

st--:
Having notebooks both in doc/source/notebooks and notebooks2/ is confusing and means it's not quite clear what the authoritative version actually is. It also makes it hard to see the changes from GPflow1 to GPflow2, and means we wouldn't see if textual improvements on the gpflow1 (`develop`) side get accidentally reverted in the move (because they were based off an older version already in `develop-2.0`). To ensure we don't break the build with non-converted notebooks, we can simply put them into the BLACKLISTED_NOTEBOOKS list, and then we will know we are done when the blacklist is empty again.


awav:
1. I know that the `notebook2` folder upsets you, but isn't better to wait until all notebooks are ported now and move them at once.
2. Please, use `git mv` instead of copying files, it will preserve the history who worked on notebooks.

st--:
1. I don't see the value of keeping the `notebook2` directory around any longer than need be (yes I don't understand why it was ever created in the first place). I think it's more valuable to ensure preserving all the copy-edit changes by @akhayes and @JamesALeedham, which this PR painstakingly does - there were already a couple other PRs (#1072, #1076) that didn't realize `develop`'s version of the notebook was more up-to-date than the version in the gpflow-2.0 branch, and they'll have to redo their work.
2. I did use `git mv` wherever possible, it wasn't possible in all cases. This is independent of whether they get moved now or later, though.

PR #1074, opened 2019-10-04
Clean up dependencies and setup instructions

st--:
This PR separates packages that are needed for running tests (notebook tests in particular) from the dependencies actually needed to `import gpflow`, updates the tensorflow and tfp dependencies to the latest stable releases (tf 2.0, tfp 0.8), and updates the README accordingly.


st--:
@awav  do the CircleCI tests still need to be based on top of the personal Docker image you built, or can we simply reuse a standard ubuntu / python base image that someone else is providing, now that tensorflow 2 is released and easily installable via pip ?

codecov.yml
```diff
@@ -12,5 +12,5 @@ coverage:
 ignore:
   - "*.py"
   - "tests/*.py"
-  - "gpflow/training/external_optimizer.py"
+  - "gpflow/ci_utils.py"
```
awav: The `ci_utils.py` is supposed to be in notebooks folder. It is a bunch of plain utility functions.

setup.py
```diff
@@ -11,22 +11,16 @@
...
+    'tensorflow-probability>=0.8',
     'tabulate',
-    'tfp-nightly',
-    'dataclasses'
 ]
 
 min_tf_version = '2.0.0'
-tf_cpu = 'tf-nightly-2.0-preview'
-tf_gpu = 'tf-nightly-gpu-2.0-preview'
+tf_cpu = 'tensorflow'
```
awav: That's a brave move :)

tests_require.txt
```diff
@@ -0,0 +1,9 @@
+pytest>=3.5.0
```
awav: Can we rename a file to something else, e.g. test_requirements.txt. Even better to move to the tox and pinned dependencies.

README.md
```diff
@@ -19,30 +19,29 @@ GPflow implements modern Gaussian process inference for composable kernels and l
...
 pip install -e .
 ```
 
+in a check-out of the `develop-2.0` branch of the GPflow github repository.
+
 
 ## Getting Started
 
-You can find the converted to GPflow 2.0 notebooks in the `notebooks` folder.
+There is an "intro to gpflow 2" Jupyter notebook in `doc/source/notebooks/intro_to_gpflow2.ipynb". Not all tutorial notebooks may be running on gpflow2 already, but we are working on it!
```
awav:
```suggestion
There is an "Intro to GPflow 2" Jupyter notebook in `doc/source/notebooks/intro_to_gpflow2.ipynb". Not all tutorial notebooks may be running on `gpflow2` already, but we are working on it!
```

README.md
```diff
@@ -19,30 +19,29 @@ GPflow implements modern Gaussian process inference for composable kernels and l
...
+## Install GPflow 2.0 alpha version
 
-1. Install TensorFlow (TF) and TensorFlow Probability (TFP) nightly packages. You _might_ experience issues with TF and TFP latest versions.
-
-```bash
-pip install -U gast==0.2.2 dataclasses tf-nightly-2.0-preview tfp-nightly
-```
-
-2. Install develop GPflow 2.0 version from `awav/gpflow-2.0` branch:
+With the release of tensorflow 2.0 and tensorflow-probability 0.8, you should
```
awav:
```suggestion
With the release of `tensorflow` 2.0 and `tensorflow-probability` 0.8, you should
```

tests/test_gplvm.py
```diff
@@ -22,7 +21,6 @@
 from gpflow.utilities.ops import pca_reduce
 
 
-@dataclass
```
awav: Why did you remove the `dataclass`?

codecov.yml
st--: I moved it to `gpflow/` because that makes it much easier to import, no having to mess about with `sys.path` which is kinda confusing in a tutorial notebook. And it makes it easier to move the notebook to a different category if needed.

setup.py
st--: Well, it's released now, no? and it seems to work.

tests/test_gplvm.py
st--: Because the code works exactly the same without, and then I don't need the extra dependency. Why would it have to be a `@dataclass`?

tests_require.txt
st--: I had named it `tests_require` so that tab completion is easier (with the `tests` directory next to it, so it doesn't clash with `test_`). I thought the file name is sufficiently obvious as it is? If you think it's important to call it `requirements` we can rename it. I don't understand what you mean by "move to the tox and pinned dependencies", can you explain please?

awav: * The `*requirements.txt` is standard naming for requirements files, let's stick to that name.
* I meant that ideally this is also intermediate solution and we should move to the tox https://tox.readthedocs.io/en/latest/ configuration. Pinned or locked dependency file is another thing which we need - usage of the `latest` versions for deps is quite a bad practice.

st--: Ok, I renamed it.

Why would it be bad practice to simply depend on e.g. `tensorflow` ? I thought we'd want to be compatible with latest version of tensorflow, scipy etc - so if a test suddenly fails because it pulled in a new dependency that's the quickest way of finding out & adjusting as needed, I would think. Also, otherwise we force people to use specific versions of the dependencies, doesn't seem very user-friendly?
Could you give some references to articles explaining why that would be bad practice, what would be considered wrong about that?

st--:
@awav do the CircleCI tests still need to be based on top of the personal Docker image you built, or can we simply reuse a standard ubuntu / python base image that someone else is providing, now that tensorflow 2 is released and easily installable via pip ?

PR #1078, opened 2019-10-04
GPflow2.0 notebook update: sanity_check.ipynb

awav:
Port sanity check notebook to GPflow 2.0


awav:
@jameshensman, I applied formatting to the svgp.py (accidentally). Looks like it didn't follow length line convention.

doc/source/notebooks/theory/Sanity_check.ipynb
```diff
@@ -67,61 +72,27 @@
...
-    "m4 = gpflow.models.SVGP(X, Y, gpflow.kernels.RBF(1),\n",
-    "                      likelihood=gpflow.likelihoods.Gaussian(),\n",
-    "                      Z=X.copy(), q_diag=False, whiten=True)\n",
-    "m4.feature.set_trainable(False)\n",
-    "m5 = gpflow.models.SGPR(X, Y, gpflow.kernels.RBF(1), Z=X.copy())\n",
-    "m5.feature.set_trainable(False)\n",
-    "m6 = gpflow.models.GPRFITC(X, Y, gpflow.kernels.RBF(1), Z=X.copy())\n",
-    "m6.feature.set_trainable(False)\n",
+    "data = (tf.convert_to_tensor(X, dtype=default_float()), tf.convert_to_tensor(Y, dtype=default_float()))\n",
+    "inducing_variable = tf.convert_to_tensor(X, dtype=default_float())\n",
```
st--: does this ensure it's decoupled from `X` (/ `data`)?

doc/source/notebooks/theory/Sanity_check.ipynb
```diff
@@ -67,61 +72,27 @@
...
+    "\n",
+    "m1 = gpflow.models.GPR(data, kernel=gpflow.kernels.RBF())\n",
```
st--: would be good to take this opportunity to rename RBF -> SquaredExponential ..:)

doc/source/notebooks/theory/Sanity_check.ipynb
st--: irrelevant in this case, as they're all set non-trainable anyways, but I'd like to hear the answer to find out :)

doc/source/notebooks/theory/Sanity_check.ipynb
```diff
@@ -67,61 +72,27 @@
...
+    "m2 = gpflow.models.VGP(data, kernel=gpflow.kernels.RBF(), likelihood=gpflow.likelihoods.Gaussian())\n",
+    "m3 = gpflow.models.SVGP(gpflow.kernels.RBF(), gpflow.likelihoods.Gaussian(), inducing_variable, q_diag=False)\n",
+    "set_trainable(m3.inducing_variable, False)\n",
+    "\n",
+    "m4 = gpflow.models.SVGP(gpflow.kernels.RBF(), gpflow.likelihoods.Gaussian(), inducing_variable, q_diag=False, whiten=True)\n",
```
st--: Nominally all models are reusing the same `inducing_variable` - in this case that shouldn't make a difference, as they're also all set to non-trainable - but would it be better to make explicit that they *should* be different if you want to train different models?

doc/source/notebooks/theory/Sanity_check.ipynb
```diff
@@ -134,164 +105,19 @@
...
+   "outputs": [],
    "source": [
-    "_ = [gpflow.train.ScipyOptimizer(method='BFGS').minimize(m, maxiter=100) for m in models]"
+    "for m in models:\n",
+    "    opt = gpflow.optimizers.Scipy()\n",
+    "    if isinstance(m, gpflow.models.SVGP):\n",
+    "        loss_fn = lambda: m.neg_log_marginal_likelihood(data[0], data[1])\n",
+    "    else:\n",
+    "        loss_fn = m.neg_log_marginal_likelihood\n",
+    "    opt.minimize(loss_fn, variables=m.trainable_variables, options=dict(maxiter=1000))"
```
st--: Should this use `BFGS` (instead of the `L-BFGS-B` default) as well?

Is the optimization fast enough not to affect notebook test time? Or would it be worthwhile replacing the maxiter=1000 with `maxiter=ci_niter(1000)`?

doc/source/notebooks/theory/Sanity_check.ipynb
```diff
@@ -378,9 +204,9 @@
       "  kernel variance     = 0.82561\n",
       "  likelihood variance = 0.16002\n",
       "SGPR\n",
-      "  kernel lengthscale  = 1.0774\n",
-      "  kernel variance     = 0.82561\n",
-      "  likelihood variance = 0.16002\n",
+      "  kernel lengthscale  = 4.5905e-09\n",
```
st--: This looks significantly different from the previous answer. Did this notebook get run after #1106 was merged into develop-2.0 (which fixes a bug in the `SGPR` model)?

doc/source/notebooks/theory/Sanity_check.ipynb
```diff
@@ -67,61 +72,27 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 3,
+   "execution_count": 8,
```
st--: Would be good to re-run the notebook with "Restart & Run All" before merging, so that the cell numbers are nicely consecutive (otherwise it can appear confusing to people reading through the notebook!).

doc/source/notebooks/theory/Sanity_check.ipynb
awav: > does this ensure it's decoupled from X (/ data)?

yes

doc/source/notebooks/theory/Sanity_check.ipynb
awav: ok

doc/source/notebooks/theory/Sanity_check.ipynb
awav: Yeah, the bug is still in the branch :) I was about to fix it, but decided to wait #1106.

doc/source/notebooks/theory/Sanity_check.ipynb
st--: Instead of changing it to all be explicit copies, I think it would be sufficient to just add a sentence or two explicitly explaining that, in this case - because all the inducing point locations are set non-trainable - it's fine to re-use the object, but also explaining how you *would* get a fresh copy if you wanted to train inducing point locations separately (i.e. ensuring they're not tied together between models).

doc/source/notebooks/theory/Sanity_check.ipynb
st--: :+1: ok:) I guess you'll just have to merge develop-2.0 back in then, now that the bugfix is in there:)

doc/source/notebooks/theory/Sanity_check.ipynb
awav: I will update maxiter with `ci_niter` and `L-BFGS-B` works fine.

doc/source/notebooks/theory/Sanity_check.ipynb
awav: ...okay :)

awav: done

doc/source/notebooks/theory/Sanity_check.ipynb
awav: done

doc/source/notebooks/theory/Sanity_check.ipynb
awav: My feeling is that this notebook is a wrong place to explain how `inducing_variable` works, so I'm keen on adding a reference to new tips & tricks notebook which will address this question properly in the future.

st--: Yes, adding a reference here to a better place to explain that works too! can you make an issue for including this point in tips&tricks notebook so that we cannot lose track of it ?

Issue #1082, opened 2019-10-04
GPflow2 likelihood tests as well as likelihoods are buggy


st--:
I tried to increase code coverage and ended up fixing some tests in #1079 - is_analytic in develop-2.0 actually returns the opposite of what it said (`actual_method is fallback_method` when it should be `is not`!), and it did not take into account that likelihoods not necessarily overwrite all of variational_expectations, predict_density, and predict_mean_and_var - but now the tests don't pass anymore. Would be great if someone else could help investigate what's going on there!

st--:
MWE is in #1083.

st--:
It's just the tests that are broken, and #1083 that fixes them is ready to merge.

PR #1083, opened 2019-10-04
GPflow2: Fix behaviour of likelihood quadrature tests

st--:
For now, this is the MWE for #1082


st--:
The equivalent changes to `develop` (GPflow 1) in #1081 pass (the is_analytic in there was correct, but also only tested for predict_density having been overwritten)

tests/test_likelihoods.py
```diff
@@ -28,18 +28,22 @@
...
+        assert not isinstance(likelihood, gpflow.likelihoods.MonteCarloLikelihood)
+        quadrature_fallback = getattr(Likelihood, method_name)
+        actual_method = getattr(likelihood.__class__, method_name)
+        return actual_method is not quadrature_fallback
+    return [l for l in likelihood_setups if is_analytic(l.likelihood)]
 
 
 class Datum:
     tolerance = 1e-06
-    Yshape = (10, 2)
```
st--: Why was Y intended to have *2* dimensions?

tests/test_likelihoods.py
```diff
@@ -28,18 +28,22 @@
 tf.random.set_seed(99012)
 
 
-def is_analytic(likelihood):
-    method = likelihood.__class__.predict_density
-    return method is Likelihood.predict_density
```
st--: This meant the only thing ever tested was quadrature against quadrature!

st--:
So turns out the bug was that the `Y` used for testing the likelihoods was random-normal values, but some likelihoods (e.g. Gamma, Exponential) can only handle positive values and others have further restrictions (e.g. Beta). I've now copied over the corresponding randomness choices from `develop`, and now all tests pass.

Caveat: there were some issues with the MultiClass likelihood. I've for now marked those tests as skipped - I think it's better to have that as an open issue & deal with later but merge this PR.

awav:
@st-- , you need to fix your editor to 100 chars per line (GPflow always used 100 lines) ;)

st--:
@awav all my changes are <100 character lines? I even fixed one line that had more beforehand. If you have any issue with any line in particular please let me know!

tests/test_likelihoods.py
```diff
@@ -61,16 +64,43 @@ def __repr__(self):
...
-    LikelihoodSetup(Beta()),
-    LikelihoodSetup(MultiClass(2), Y=tf.argmax(Datum.Y, 1).numpy().reshape(-1, 1), rtol=1e-3, atol=1e-3),
-    LikelihoodSetup(Ordinal(np.array([-1, 1])), Y=np.random.randint(0, 3, Datum.Yshape)),
-    LikelihoodSetup(Poisson(invlink=Datum.square)),
-    LikelihoodSetup(Exponential(invlink=Datum.square)),
-    LikelihoodSetup(Gamma(invlink=Datum.square)),
-    LikelihoodSetup(Bernoulli(invlink=tf.sigmoid)),
+    LikelihoodSetup(Beta(),
+                    Y=tf.random.uniform(Datum.Yshape, dtype=default_float())),
+    pytest.param(LikelihoodSetup(MultiClass(2),
```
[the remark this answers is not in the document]
awav: I think `pytest.param` with skip option is unnecessary here, and just filtering is enough. Once it is resolved, the filter function can be removed.

awav:
@st-- , why this changes with formatting happened then? Can you revert them back, it is actually distracting :(

st--:
@awav I have no idea what formatting changes you're talking about, and I can't revert what I don't know. It'd be more helpful if you added a comment to the code you are referring to, not here!

tests/test_likelihoods.py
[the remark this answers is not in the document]
st--: I thought that's exactly what marking things as skipped is meant to be used for. I'd prefer it over just commenting out the test, and over adding a special `[l for l in likelihood_setups if l.likelihood.__class__ is not MultiClass]` filter... I'll put it at the end of the list though.

tests/test_likelihoods.py
```diff
@@ -61,16 +64,44 @@ def __repr__(self):
...
+    LikelihoodSetup(Ordinal(np.array([-1, 1])),
+                    Y=tf.random.uniform(Datum.Yshape, 0, 3, dtype=default_int())),
+    LikelihoodSetup(Poisson(invlink=tf.square),
+                    Y=tf.random.poisson(Datum.Yshape, 1.0, dtype=default_float())),
+    LikelihoodSetup(Exponential(invlink=tf.square),
+                    Y=tf.random.uniform(Datum.Yshape, dtype=default_float())),
+    LikelihoodSetup(Gamma(invlink=tf.square),
+                    Y=tf.random.uniform(Datum.Yshape, dtype=default_float())),
+    LikelihoodSetup(Bernoulli(invlink=tf.sigmoid),
+                    Y=tf.random.uniform(Datum.Yshape, dtype=default_float())),
```
awav: I meant this block of code, but now I can see that you changed it. You could create two random variables and reuse them in likelihood setup. It would be also useful for debugging in the future and the code would be more readable.

st--: I could, but then we'd need four different `Y`s in the `Datum` class, which I don't think is an improvement over the status quo. If you have a better way of doing it in mind, please feel free to change it, maybe you could do that in a different PR?

PR #1085, opened 2019-10-07
Test on print_summary now checks string output straight

spascual:
Test on print_summary now checks string output straight

### DIFF
--- a/gpflow/utilities/utilities.py
+++ b/gpflow/utilities/utilities.py
@@ -77,6 +77,15 @@ def print_summary(module: tf.Module, fmt: str = None):
     Prints a summary of the parameters and variables contained in a tf.Module.
     """
 
+    if fmt == "notebook":
+        from IPython.core.display import display, HTML
+        tab = _print_summary_output_string(module, "html")
+        display(HTML(tab))
+    else:
+        print(_print_summary_output_string(module, fmt))
+
+
+def _print_summary_output_string(module: tf.Module, fmt: str = None) -> str:
     fmt = fmt if fmt is not None else summary_fmt()
     column_names = ['name', 'class', 'transform', 'trainable', 'shape', 'dtype', 'value']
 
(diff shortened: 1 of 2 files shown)


gpflow/utilities/utilities.py
```diff
@@ -77,6 +77,15 @@ def print_summary(module: tf.Module, fmt: str = None):
...
 
+    if fmt == "notebook":
+        from IPython.core.display import display, HTML
+        tab = _print_summary_output_string(module, "html")
+        display(HTML(tab))
+    else:
+        print(_print_summary_output_string(module, fmt))
+
+
+def _print_summary_output_string(module: tf.Module, fmt: str = None) -> str:
```
```suggestion
def tabulate_module_summary(module: tf.Module, tablefmt: str = None) -> str:
```
might be a more appropriate name, it doesn't actually print anything, the (table)fmt is passed straight through to `tabulate`, and it takes a module as an input? There might actually be use-cases in which you want this, instead of having it printed, so I think it ought to be fine as a public function.

gpflow/utilities/utilities.py
```diff
@@ -115,7 +117,10 @@ def leaf_components(input: tf.Module):
 
 def _merge_leaf_components(
         input: Dict[str, Union[tf.Tensor, Parameter]]) -> Dict[str, Union[tf.Tensor, Parameter]]:
-    if len(set(input.values())) == len(input):
+    input_values = set(
+        [value.experimental_ref() if isinstance(value, tf.Variable) else value for value in input.values()]
```
st--: what does this change do? (not mentioned in PR summary)

spascual: tf.Variables are not hashable, so you cannot take set of a list of them. Instead, tensorflow recommends to use  .experimental_ref(). You can replicate it yourself if you want :)

gpflow/utilities/utilities.py
```diff
@@ -77,6 +77,15 @@ def print_summary(module: tf.Module, fmt: str = None):
     Prints a summary of the parameters and variables contained in a tf.Module.
     """
 
+    if fmt == "notebook":
+        from IPython.core.display import display, HTML
+        tab = _print_summary_output_string(module, "html")
+        display(HTML(tab))
+    else:
+        print(_print_summary_output_string(module, fmt))
```
```suggestion
        print(tabulate_module_summary(module, fmt))
```

gpflow/utilities/utilities.py
```diff
@@ -77,7 +77,16 @@ def print_summary(module: tf.Module, fmt: str = None):
...
+    if fmt == "notebook":
+        from IPython.core.display import display, HTML
+        tab = tabulate_module_summary(module, "html")
+        display(HTML(tab))
+    else:
+        print(tabulate_module_summary(module, fmt))
+
+
+def tabulate_module_summary(module: tf.Module, tablefmt: str = None) -> str:
+    tablefmt = tablefmt if tablefmt is not None else summary_fmt()
```
st--: I would pass through `None` (meaning "use tabulate's default"); someone who wants to use `tabulate_module_summary` directly probably has a specific use-case.

Issue #1077, opened 2019-10-04
GPflow2: stationary kernels have `ard` flag but does not work


st--:
This flag should be removed as it is misleading, and without `input_dim` I don't see how it is possible to initialize the lengthscale correctly.
Moreover, there should be a check that if the lengthscale is NOT a scalar, that it matches up with the effective input dimensions.
Also, the K_r / K_r2 framework that is useful e.g. for the generalised Periodic kernel (#1052) seems to only be partially supported (e.g. SquaredExponential falls out of it), this needs fixing to be able to port #1052 to develop-2.0

st--:
@jahall do you reckon this issue is now solved and can be closed?

jahall:
> @jahall do you reckon this issue is now solved and can be closed?

Yes - all resolved

st--:
Thanks!

PR #1088, opened 2019-10-07
GPFlow 2.0: ARD and Stationaries Refactor

jahall:
Migrating the generalized periodic enhancement from 1.0. This is the initial groundwork to make that happen i.e. addresses the issues outlined in https://github.com/GPflow/GPflow/issues/1077. In addition some minor refactoring has been done e.g. to move Periodic out of misc.py and into periodic.py.

Will re-introduce the actual enhancement in a follow up commit tomorrow.


jahall:
@st-- Would welcome your feedback on the initial groundwork.

jahall:
@st-- May need your help tracking down the bugs thrown up in CI. I think they are related to my removal of the `ard` kwarg and associated ard validation check.

gpflow/kernels/base.py
```diff
@@ -117,6 +120,19 @@ def slice_cov(self, cov: tf.Tensor) -> tf.Tensor:
...
+    def validate_ard_parameter(self, X, parameter):
+        """
+        Check that the input dimension matches the dimension of the parameter.
+        """
+        if tf.rank(parameter) == 0:
+            return
+        X_shape = tf.shape(X)
+        l_shape = tf.shape(parameter)
+        if l_shape[-1] != X_shape[-1]:
+            # TODO: Maybe this should be a tf.errors.InvalidArgumentError
```
st--: I don't think we should use another package's exceptions, especially if they're that unspecific...

st--:
@jahall thanks for your work on this. I think (at least some of) the tests fail because the `ard` flag is actually used in the kernel expectations (`expectations/`). We could keep an `ard` property that checks whether the lengthscale parameter is a scalar (ard=False) or a vector (ard=True)...

I'm wondering if it'd be cleaner (& easier to review) to move the refactoring of ARD and `K_r2`/`K_r` into its own PR?

gpflow/kernels/base.py
jahall: Agreed - will remove comment

gpflow/kernels/linears.py
```diff
@@ -13,24 +13,30 @@ class Linear(Kernel):
     where σ²  is the variance parameter.
     """
 
-    def __init__(self, variance=1.0, active_dims=None, ard=None):
+    def __init__(self, variance=1.0, active_dims=None):
         """
         - input_dim is the dimension of the input to the kernel
         - variance is the (initial) value for the variance parameter(s)
-          if ard=True, there is one variance per input
+          defaults to 1.0 but supports ARD i.e. one variance per input.
```
st--: Could we have a more complete explanation? e.g. if you want ARD but initialised to 1 you'd have to pass in [1, 1, ..., 1] (where the length of the array is equal to the input dimension)

gpflow/kernels/base.py
```diff
@@ -117,6 +120,24 @@ def slice_cov(self, cov: tf.Tensor) -> tf.Tensor:
 
         return cov
 
+    def validate_ard_parameter(self, X, parameter):
```
st--: Could also check whether `active_dims` has the right length (if not None)?

gpflow/kernels/linears.py
st--: (Applies to all kernels with ard feature)

gpflow/kernels/stationaries.py
```diff
@@ -17,48 +17,54 @@ class Stationary(Kernel):
...
     """
 
-    def __init__(self, variance=1.0, lengthscale=1.0, active_dims=None, ard=None):
+    def __init__(self, variance=1.0, lengthscale=1.0, active_dims=None):
         """
         - input_dim is the dimension of the input to the kernel
         - variance is the (initial) value for the variance parameter
         - lengthscale is the initial value for the lengthscale parameter
-          defaults to 1.0 (ard=False) or np.ones(input_dim) (ard=True).
+          defaults to 1.0.
```
st--: explain how to get ard behaviour?

gpflow/kernels/stationaries.py
```diff
@@ -17,48 +17,54 @@ class Stationary(Kernel):
...
 
     def K_diag(self, X, presliced=False):
         return tf.fill((X.shape[:-1]), tf.squeeze(self.variance))
 
-    def K_r(self, r):
+    def K_r2(self, r2):
         """
-        Returns the kernel evaluated on `r`, which is the scaled Euclidean distance
-        Should operate element-wise on r
+        Returns the kernel evaluated on `r²`, which is the squared scaled Euclidean distance
```
```suggestion
        Returns the kernel evaluated on r² (`r2`), which is the squared scaled Euclidean distance
```

tests/test_features.py
```diff
@@ -30,9 +30,8 @@ def test_inducing_points_inducing_variable_len(N, D):
 
 _kernel_setups = [
     gpflow.kernels.SquaredExponential(variance=0.46,
-                                      lengthscale=np.random.uniform(0.5, 3., 5),
-                                      ard=True),
-    gpflow.kernels.Periodic(period=0.4, variance=1.8)
+                                      lengthscale=np.random.uniform(0.5, 3., 5)),
```
st--: should the lengthscale be an array to ensure it's actually testing ard=True ?

tests/test_kernels.py
```diff
@@ -137,15 +132,14 @@ def test_arccosine_non_implemented_order(order):
         gpflow.kernels.ArcCosine(order=order)
 
 
-@pytest.mark.parametrize('ard', [True, False])
```
st--: I would keep this in - maybe as
```python
@pytest.mark.parametrize('weight_variances', [
    1.,  # ard = False
    [1., 1., 1.],  # ard = True
])
```
(and then take `weight_variances` out of the following parametrization)
(assumes input_dim=3)

tests/test_kernels.py
```diff
@@ -165,8 +159,6 @@ def _assert_periodic_kern_err(lengthscale, variance, period, X):
     gram_matrix = kernel(X)
     reference_gram_matrix = _ref_periodic(X, lengthscale, variance, period)
 
-    assert_allclose(gram_matrix, reference_gram_matrix)
```
st--: why did this get removed?

tests/test_kernels.py
st--: or did this already get tested elsewhere ? :thinking:

tests/test_kernels.py
```diff
@@ -243,7 +235,7 @@ def test_coregion_slice(N, input_dim, output_dim, rank):
 kernel_setups_extended = kernel_setups + [
     SquaredExponential() + Linear(),
     SquaredExponential() * Linear(),
-    SquaredExponential() + Linear(ard=True, variance=rng.rand(_dim, 1).reshape(-1))
+    SquaredExponential() + Linear(variance=rng.rand(_dim, 1).reshape(-1))
```
st--: Why not
```suggestion
    SquaredExponential() + Linear(variance=rng.rand(_dim))
```
?

tests/test_kernels.py
```diff
@@ -362,7 +354,7 @@ def test_active_product(N, D):
         np.hstack([ls[:rand_idx], ls[rand_idx + 1:]]), ls[rand_idx], ls
     ]
     kernels = [
-        gpflow.kernels.SquaredExponential(lengthscale=lengthscale, active_dims=dims, ard=True)
```
st--: does ard=True still get tested?

tests/test_kernels.py
```diff
@@ -389,23 +381,9 @@ def test_ard_init_scalar(D):
...
         k1(rng.randn(N, 4))
-    with pytest.raises(tf.errors.InvalidArgumentError):
-        k2 = gpflow.kernels.SquaredExponential(lengthscale=np.ones(3), ard=ard)
+    with pytest.raises(ValueError):
+        k2 = gpflow.kernels.SquaredExponential(lengthscale=np.ones(3))
         k2(rng.randn(N, 2))
-
-
-@pytest.mark.parametrize('D', [4, 7])
-def test_ard_init_MLP(D):
```
st--: why did this disappear?

gpflow/kernels/linears.py
jahall: Makes sense - have done this

gpflow/kernels/base.py
jahall: I will add in a check on __init__ to ensure size of `active_dims` has the right length (if not None and if not a slice)

tests/test_kernels.py
jahall: Ha - yeah I'm not sure why that was defined that way. I'll change it.

tests/test_kernels.py
jahall: This disappeared because the test is no longer valid - it was checking that if you passed in a scalar `weight_variances` with `ard=True` that it got expanded to an array. We now rely on the user explicitly specifying an array.

tests/test_kernels.py
jahall: Actually - this is now sufficiently tested in `test_arccosine_kern_err` - this test can be removed

jahall:
@st-- Have addressed all comments. However, I cannot for the life of me figure out how to get this `validate_ard_parameter` method to work properly as `tf.cond` doesn't always lazily evaluate :( - see TODO in code. Other than that I think this is in reasonable shape. Just fixing the notebooks.

gpflow/kernels/base.py
```diff
@@ -117,6 +121,42 @@ def slice_cov(self, cov: tf.Tensor) -> tf.Tensor:
...
+        is_ard = isinstance(raw_parameter, Iterable)
+        if (is_ard
+                and self.active_dims is not None
+                and not isinstance(self.active_dims, slice)
+                and len(raw_parameter) != len(self.active_dims)):
+            raise ValueError(f"Dimension of `active_dims` {self.active_dims} does not match "
+                             f"dimension of the ard parameter ({len(raw_parameter)})")
+        return is_ard
+
+    def validate_ard_parameter(self, X, parameter):
```
awav: This is wierd function, it just returns True? Why did you add this TODO and commented out code?

tests/test_kernels.py
```diff
@@ -380,32 +364,38 @@ def test_ard_init_scalar(D):
...
-    variances_1 = kernel_1.weight_variances.read_value()
-    variances_2 = kernel_2.weight_variances.read_value()
-    assert np.allclose(variances_1, variances_2, atol=1e-10)
+def test_ard_invalid_active_dims():
+    msg = r"Dimension of `active_dims` \[1\] does not match dimension of the ard parameter \(2\)"
+    with pytest.raises(ValueError, match=msg):
+        gpflow.kernels.SquaredExponential(lengthscale=np.ones(2), active_dims=[1])
+
+
+#@pytest.mark.parametrize('N', [4, 7])
```
awav: Why this part is commented out?

jahall: This is the associated test for the aforementioned commented code.

gpflow/kernels/linears.py
```diff
@@ -13,24 +13,23 @@ class Linear(Kernel):
...
+            to induce ARD behaviour this must be initialised as an array the same
+            length as the the number of active dimensions e.g. [1., 1., 1.]
+        :param active_dims: a slice or list specifying which columns of X are used
         """
         super().__init__(active_dims)
-
-        # variance, self.ard = self._validate_ard_shape("variance", variance, ard)
-        self.ard = ard
         self.variance = Parameter(variance, transform=positive())
+        self.ard = self._is_ard(variance)
```
st--: I think it's better as a property (as you had before), e.g. if someone decides to replace the `variance` attribute with a different Parameter object. Same applies to the other kernels as well of course!

jahall: Good point - I have reintroduced `ard` as a parameter and improved the implementation by creating an `ARDKernel` base class to enable better code re-use.

gpflow/kernels/base.py
jahall: I have removed this altogether for now as I cannot figure out how to perform this validation correctly during graph execution.

awav:
Hello @jahall , I apologize for closing your PR. It was closed automatically when I moved `develop-2.0` to the `develop`. Could you re-open the PR? Thanks!

awav:
I think it is a bit of overkill to create a separate class because of the ard property of parameters (not even all of them) in a kernel :)

jahall:
> I think it is a bit of overkill to create a separate class because of the ard property of parameters (not even all of them) in a kernel :)

That's reasonable. I'll refactor to make the change less intrusive.

gpflow/kernels/linears.py
```diff
@@ -1,31 +1,36 @@
...
-        self.ard = ard
         self.variance = Parameter(variance, transform=positive())
+        self._validate_ard_active_dims(self.variance)
+
+    @property
+    def ard(self) -> bool:
+        """
+        Whether ARD behaviour is active.
+        """
+        return len(self.variance.shape) > 0
```
awav: self.variance.shape.ndims
```suggestion
        return self.variance.shape.ndims > 0
```

gpflow/kernels/misc.py
```diff
@@ -55,10 +51,16 @@ def __init__(self,
...
         self.weight_variances = Parameter(weight_variances,
                                           transform=positive())
+        self._validate_ard_active_dims(self.weight_variances)
+
+    @property
+    def ard(self) -> bool:
+        """
+        Whether ARD behaviour is active.
+        """
+        return len(self.weight_variances.shape) > 0
```
awav:
```suggestion
        return self.weight_variances.shape.ndims > 0
```

jahall:
@awav @st-- Any further changes required for this PR? I just ask as its been open over a month and contains necessary changes (i.e. the `K_r`/`K_r2` refactor in `stationaries`) to merge the generalized periodic kernel from v1 (see #1052)

PR #1099, opened 2019-10-11
fix bug in log_prior when a Parameter has a transform but no prior

st--:
When doing MLE instead of MAP (i.e., no prior on the hyperparameters), the objective should be independent of the parametrisation of the parameters. This means that the `log_prior` of a parameter with a `transform` (e.g. `positive()`) but without a `prior` should be zero. This was not the case, leading to difficult / wrong optimisation e.g. in #1087. Fixed in this PR.

### DIFF
--- a/gpflow/base.py
+++ b/gpflow/base.py
@@ -46,7 +46,7 @@ class Parameter(tf.Module):
                  name: Optional[str] = None):
         """
         Unconstrained parameter representation.
-        According to standart terminology `y` is always transformed representation or,
+        According to standard terminology `y` is always transformed representation or,
         in other words, it is constrained version of the parameter. Normally, it is hard
         to operate with unconstrained parameters. For e.g. `variance` cannot be negative,
         therefore we need positive constraint and it is natural to use constrained values.
--- a/gpflow/base.py
+++ b/gpflow/base.py
@@ -73,9 +73,9 @@ class Parameter(tf.Module):
         bijector = self.transform
         if self.prior is not None:
             out += tf.reduce_sum(self.prior.log_prob(x))
-        if self.transform is not None:
-            log_det_jacobian = bijector.forward_log_det_jacobian(y, y.shape.ndims)
-            out += tf.reduce_sum(log_det_jacobian)
+            if self.transform is not None:
+                log_det_jacobian = bijector.forward_log_det_jacobian(y, y.shape.ndims)
+                out += tf.reduce_sum(log_det_jacobian)
         return out
 
     @property
--- a/gpflow/base.py
+++ b/gpflow/base.py
@@ -66,17 +66,15 @@ class Parameter(tf.Module):
     def log_prior(self):
         x = self.read_value()
         y = self._unconstrained
-        dtype = x.dtype
 
-        out = tf.convert_to_tensor(0., dtype=dtype)
-
-        bijector = self.transform
         if self.prior is not None:
-            out += tf.reduce_sum(self.prior.log_prob(x))
+            out = tf.reduce_sum(self.prior.log_prob(x))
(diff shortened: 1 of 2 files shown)


gpflow/base.py
```diff
@@ -46,7 +46,7 @@ def __init__(self,
                  name: Optional[str] = None):
         """
         Unconstrained parameter representation.
-        According to standart terminology `y` is always transformed representation or,
+        According to standard terminology `y` is always transformed representation or,
         in other words, it is constrained version of the parameter. Normally, it is hard
         to operate with unconstrained parameters. For e.g. `variance` cannot be negative,
         therefore we need positive constraint and it is natural to use constrained values.
```
st--: This docstring seems inconsistent with the code below. @awav could you please have a look?

gpflow/base.py
```diff
@@ -73,9 +73,9 @@ def log_prior(self):
         bijector = self.transform
         if self.prior is not None:
             out += tf.reduce_sum(self.prior.log_prob(x))
-        if self.transform is not None:
-            log_det_jacobian = bijector.forward_log_det_jacobian(y, y.shape.ndims)
-            out += tf.reduce_sum(log_det_jacobian)
+            if self.transform is not None:
+                log_det_jacobian = bijector.forward_log_det_jacobian(y, y.shape.ndims)
+                out += tf.reduce_sum(log_det_jacobian)
```
st--: Only add log_det_jacobian if there IS a prior

tests/test_prior.py
```diff
@@ -0,0 +1,30 @@
...
+    X = 10 * np.random.randn(5,1)
+    Y = 10 * np.random.randn(5,1)
+
+def test_gpr_objective_equivalence():
+    data = (Datum.X, Datum.Y)
+    l_value = 1.3
+    l_variable = tf.Variable(l_value, dtype=gpflow.default_float(), trainable=True)
+    m1 = gpflow.models.GPR(data, kernel=gpflow.kernels.SquaredExponential(lengthscale=l_value))
+    m2 = gpflow.models.GPR(data, kernel=gpflow.kernels.SquaredExponential(lengthscale=l_variable))
+    m2.kernel.lengthscale._transform = None
```
st--: Seems to be the only way of undoing the transform if I want to pass in a tf.Variable that's already in unconstrained space... should we ignore the transform in the Parameter class constructor if passing in a tf.Variable directly? @awav

gpflow/base.py
st--: What would be a good way of testing that the log-det-Jacobian of the transform is correctly added to the objective when doing MAP? :thinking:

tests/test_prior.py
awav: The transform is embedded into `parameter`. You have to replace parameter with another one, instead of trying to update the current one.

st--: so just do `m2.kernel.lengthscale = Parameter(l_variable, transform=None)`?

gpflow/base.py
awav: Good catch! Yes, we should add the log det jacobian only if prior exists. I would even suggest that None must be returned when the prior is None.

tests/test_prior.py
awav: yes!

gpflow/base.py
st--: If we return `None` then the return type will be unstable, and all places that use this method would have to explicitly check that it didn't return None before adding it to the objective ... is that actually what we want?

awav: I would say that's logical. No prior - you can't return a value. `None` is a good indicator that there is no valid value for the prior of the parameter.

PR #1100, opened 2019-10-14
Migrate MCMC notebook to gpflow2.0 

condnsdmatters:
Migrate MCMC notebook to gpflow 2 using tensorflow probabilities.
Changes required a few changes across the rest of the codebase, outside of the notebook.

*  small change to leaf printing function (fix bug, which was preventing printing of composite kernels)
* `parameters` and `trainable_parameters` now return tuples, not generators, like tf implementation of `variables`, `trainable_variables`
* tfp.distributions now work with different dtypes (by wrapping of parameters), so now play nicely with gpflow.


gpflow/utilities/utilities.py
```diff
@@ -15,6 +15,7 @@
     "multiple_assign",
     "training_loop",
     "print_summary",
+    "leaf_components"
```
```suggestion
    "leaf_components",
```
to make future diffs nicer ;)

doc/source/notebooks/advanced/multiclass_classification.py
```diff
@@ -19,71 +19,71 @@ def plot_posterior_predictions(m):
...
-    for i, s in samples[burn::thin].iterrows():  # also possible as arguments to hmc.sample()
-        feed_dict = m.sample_feed_dict(s)
-        Ypred.append(m.predict_y(xx, feed_dict=feed_dict)[0])
-        Fpred.append(m.predict_f_samples(xx, 1, feed_dict=feed_dict).squeeze())
+    xx = np.linspace(X.min(), X.max(), 200).reshape(-1,1)
+
+    Fpred, Ypred = [], []
+    num_samples = len(samples[0])
+    for i in range(burn, num_samples, thin):
+        hparams = [hp[i] for hp in samples]
```
st--: what does `hp` stand for?

doc/source/notebooks/advanced/multiclass_classification.py
```diff
@@ -19,71 +19,71 @@ def plot_posterior_predictions(m):
...
+        [var.assign(hp) for var, hp in zip(m.trainable_variables, hparams)]
```
st--: How does this compare in speed to the tf1 with feed-dict?

gpflow/likelihoods/robustmax.py
```diff
@@ -22,7 +23,8 @@ class RobustMax(tf.Module):
     def __init__(self, num_classes, epsilon=1e-3, **kwargs):
         super().__init__(**kwargs)
         transform = tfp.bijectors.Sigmoid()
-        prior = tfp.distributions.Beta(0.2, 5.)
+        fdtype = lambda x: np.array(x, dtype=default_float())
```
st--: what does the `f` stand for?
if this is something needed throughout the code base, would it be worthwhile putting it into `gpflow.utilities` or so ?

gpflow/optimizers/mcmc.py
```diff
@@ -0,0 +1,118 @@
...
+from typing import Callable, List, Optional, TypeVar
+
+import tensorflow as tf
+
+from gpflow.base import Parameter, positive
+
+__all__ = ["positive_parameter", "SamplingHelper"]
+
+
+def positive_parameter(value: tf.Tensor):
```
st--: It seems like this shouldn't be in `mcmc.py` but better belongs somewhere else...?

gpflow/optimizers/mcmc.py
```diff
@@ -0,0 +1,118 @@
...
+
+
+@dataclass(frozen=True)
+class SamplingHelper:
+    """
+    Helper reads from variables being set with a prior and writes values back to the same variables.
+
+    Example:
+        model = <Create GPflow model>
+        hmc_helper = SamplingHelper(lambda: -model.neg_log_marginal_likelihood(), model.trainable_parameters)
```
st--: `- neg_log_marginal_likelihood` seems a bit bizarre, I guess this is so that you properly include the log-prior (including log-det-jacobian of any transform)? would it perhaps be cleaner to refactor that in the Model base class ?

gpflow/optimizers/mcmc.py
```diff
@@ -0,0 +1,118 @@
...
+        target_fn = hmc_helper.make_posterior_log_prob_fn()
+
+        @tf.function
+        def run_chain_fn():
+            hmc = mcmc.HamiltonianMonteCarlo(target_log_prob_fn=target_fn, num_leapfrog_steps=10, step_size=0.01)
```
st--: Where does this come from? Not this module!?

gpflow/optimizers/mcmc.py
```diff
@@ -0,0 +1,118 @@
...
+            adaptive_hmc = mcmc.SimpleStepSizeAdaptation(hmc, num_adaptation_steps=10, target_accept_prob=0.75,
+                                                         adaptation_rate=0.1)
+            return mcmc.sample_chain(num_results=hmc_parameters.num_samples, num_burnin_steps=100,
+                                     current_state=hmc_helper.variables, kernel=adaptive_hmc)
+
+
+    Args:
+        target_log_prob_fn: Python callable which represents log-density under the target distribution.
+        parameters: List of `Variable`'s or gpflow `Parameter`s used as a state of the Markov chain.
```
st--:
```suggestion
        parameters: List of `tensorflow.Variable`s or `gpflow.Parameter`s used as a state of the Markov chain.
```

gpflow/optimizers/mcmc.py
```diff
@@ -0,0 +1,118 @@
...
+        def run_chain_fn():
+            hmc = mcmc.HamiltonianMonteCarlo(target_log_prob_fn=target_fn, num_leapfrog_steps=10, step_size=0.01)
+            adaptive_hmc = mcmc.SimpleStepSizeAdaptation(hmc, num_adaptation_steps=10, target_accept_prob=0.75,
+                                                         adaptation_rate=0.1)
+            return mcmc.sample_chain(num_results=hmc_parameters.num_samples, num_burnin_steps=100,
+                                     current_state=hmc_helper.variables, kernel=adaptive_hmc)
+
+
+    Args:
+        target_log_prob_fn: Python callable which represents log-density under the target distribution.
```
st--: according to the line 58 below, this callable actually takes model parameters, maybe worth mentioning that here (when just reading the docstring I assumed it would be a zero-argument callable...)

gpflow/optimizers/mcmc.py
```diff
@@ -0,0 +1,118 @@
...
+    def variables(self):
+        """
+        Returns the same list of parameters as `parameters` property, but replaces gpflow `Parameter`s
+        with their unconstrained variables - `parameter.unconstrained_variable`.
+        """
+        return [p.unconstrained_variable if isinstance(p, Parameter) else p for p in self.parameters]
+
+    def assign_values(self, *values, unconstrained: Optional[bool] = True):
+        """
+        Assings (constrained or unconstrained) values to the parameter's variable.
```
st--:
```suggestion
        Assigns (constrained or unconstrained) values to the parameter's variable.
```

gpflow/optimizers/mcmc.py
```diff
@@ -0,0 +1,118 @@
...
+        Unconstrained values are assigned to the list of `variables` property.
```
st--:
```suggestion
        When assigning unconstrained values, the target is `self.variables`, otherwise `self.parameters`.
```

gpflow/optimizers/mcmc.py
```diff
@@ -0,0 +1,118 @@
...
+        parameters: List of `Variable`'s or gpflow `Parameter`s used as a state of the Markov chain.
+    """
+
+    target_log_prob_fn: Callable[[ModelParameters], tf.Tensor]
+    parameters: ModelParameters
+
+    @property
+    def variables(self):
+        """
+        Returns the same list of parameters as `parameters` property, but replaces gpflow `Parameter`s
```
st--:
```suggestion
        Returns the same list of parameters as `self.parameters`, but replaces gpflow `Parameter`s
```

gpflow/optimizers/mcmc.py
```diff
@@ -0,0 +1,118 @@
...
+        """
+        return [p.unconstrained_variable if isinstance(p, Parameter) else p for p in self.parameters]
+
+    def assign_values(self, *values, unconstrained: Optional[bool] = True):
+        """
+        Assings (constrained or unconstrained) values to the parameter's variable.
+        Unconstrained values are assigned to the list of `variables` property.
+        """
+        trainables = self.variables if unconstrained else self.parameters
+        assert len(values) == len(trainables)
```
st--: Replace with a proper check & exception?

gpflow/optimizers/mcmc.py
```diff
@@ -0,0 +1,118 @@
...
+        n = len(trainables)
+        for i in range(n):
+            trainables[i].assign(values[i])
+
+    def convert_to_constrained_values(self, *unconstrained_values):
+        """
+        Converts list of `unconstrained_values` to constrained versions. Each value in the list correspond to the
```
st--:
```suggestion
        Converts list of `unconstrained_values` to constrained versions. Each value in the list correspond to an entry in
```

gpflow/optimizers/mcmc.py
```diff
@@ -0,0 +1,118 @@
...
+        parameter and in case when an object in the same position has `Parameter` type, the `forward` method
```
st--:
```suggestion
        `self.parameters`; in case that object is a `gpflow.Parameter`, the `forward` method of its transform
```

gpflow/optimizers/mcmc.py
```diff
@@ -0,0 +1,118 @@
...
+        of transform will be applied.
```
st--:
```suggestion
        will be applied first.
```

gpflow/optimizers/mcmc.py
```diff
@@ -0,0 +1,118 @@
...
+        """
+        samples = []
+        for i, values in enumerate(unconstrained_values):
```
st--: Should there be a similar check as in `assign_values` to check the lengths match?

st--:
```suggestion
        for i, value in enumerate(unconstrained_values):
```

gpflow/optimizers/mcmc.py
```diff
@@ -0,0 +1,118 @@
...
+            param = self.parameters[i]
+            if isinstance(param, Parameter) and param.transform is not None:
+                sample = param.transform.forward(values)
```
st--:
```suggestion
                sample = param.transform.forward(value)
```

gpflow/optimizers/mcmc.py
```diff
@@ -0,0 +1,118 @@
...
+            else:
+                sample = values
```
st--:
```suggestion
                sample = value
```

gpflow/optimizers/mcmc.py
```diff
@@ -0,0 +1,118 @@
...
+            samples.append(sample.numpy())
```
st--: Is the casting to numpy needed? Would this slow down the execution?

gpflow/optimizers/mcmc.py
```diff
@@ -0,0 +1,118 @@
...
+        return samples
+
+    def make_posterior_log_prob_fn(self):
+        """
+        Make a differentiable posterior log-probability function using helper's `target_log_prob_fn` with respect to
+        passed `parameters`.
```
st--:
```suggestion
        `self.parameters`.
```
This function does not take any arguments so referring to "passed" something seems confusing

gpflow/optimizers/mcmc.py
```diff
@@ -0,0 +1,118 @@
...
+            return mcmc.sample_chain(num_results=hmc_parameters.num_samples, num_burnin_steps=100,
+                                     current_state=hmc_helper.variables, kernel=adaptive_hmc)
+
+
+    Args:
+        target_log_prob_fn: Python callable which represents log-density under the target distribution.
+        parameters: List of `Variable`'s or gpflow `Parameter`s used as a state of the Markov chain.
+    """
+
+    target_log_prob_fn: Callable[[ModelParameters], tf.Tensor]
```
st--: Actually, this is used as if it doesn't take any arguments, so maybe the type hint is wrong?

gpflow/optimizers/mcmc.py
```diff
@@ -0,0 +1,118 @@
...
+
+            variables_to_watch = self.variables
+            with tf.GradientTape(watch_accessed_variables=False) as tape:
+                tape.watch(variables_to_watch)
+                log_prob = self.target_log_prob_fn()
+
+            @tf.function
+            def grad_fn(in_grad: tf.Tensor, variables: Optional[tf.Variable] = None):
+                grad = tape.gradient(log_prob, variables_to_watch)
+                return grad, [None] * len(variables)
```
st--: What's the second value used for - why does it return a list of `None`s?

gpflow/optimizers/mcmc.py
```diff
@@ -0,0 +1,118 @@
...
+        Make a differentiable posterior log-probability function using helper's `target_log_prob_fn` with respect to
+        passed `parameters`.
+        """
+
+        @tf.custom_gradient
+        def log_prob_fn(*values):
+            self.assign_values(*values)
+
+            variables_to_watch = self.variables
+            with tf.GradientTape(watch_accessed_variables=False) as tape:
```
st--: why `watch_accessed_variables=False` - would it be worth adding a little comment to explain?

awav:
@condnsdmatters, Thanks for updating the notebook.

Could we not mix thing up in single PR?

> small change to leaf printing function (fix bug, which was preventing printing of composite kernels)

That must be separate PR with testing and cetera. I have a suspicion that it has been resolved already.

> parameters and trainable_parameters now return tuples, not generators, like tf implementation of variables, trainable_variables.

I don't think that `variables` return a tuple, it is a list. Why this change is important here?

> tfp.distributions now work with different dtypes (by wrapping of parameters), so now play nicely with gpflow.

I don't understand the problem with it. Could you explain what you are trying to do with it?

In general, when I wrote `SamplingHelper`, I adjusted it to the specific project for solving the needs. It requires much more thought about how the interface should look for a user. Let's discuss it offline.

Thanks!

condnsdmatters:
In reply to @awav, I think I misunderstood the helper you wrote - my understanding was I should use as is, rather than thinking about interfaces, etc. I'm very happy to discuss and change, I agree it seems to be a bit of an odd class atm. 

Vis a vis other changes:
* The printing does indeed seem to be fixed
* The `variables` and `trainable_variables` parameters onb `tf.Module` return tuples technically not lists: [see here](https://github.com/tensorflow/tensorflow/blob/1cf0898dd4331baf93fe77205550f2c2e6c90ee5/tensorflow/python/module/module.py#L143) However, the key problem here is that at the moment the module wrapper wwe have neither returns lists nor tuples, but rather a generator for `self._flatten`. If you dont wrap the generator in an iterable, when you pass the generator to the `frozenset` `SampleHelper`, the generator exhausts after its first use. eg:
    ```python
    params = m.trainable_parameters
    y = [p for p in params ]  # List[Params]
    z = [p for p in params]  # [] Empty! 
    ```
* The problem with `tf.distributions` is that there is no way to chose the dtype explicitly for the `Normal` and `Gamma` distributions, and there is no way to change the default dtype in tensorflow (as per the issue [you opened](https://github.com/tensorflow/tensorflow/issues/26033)).  `tfp` implemenatation of these distributions instead infers the dtype from the dtype of the parameters to the distribution eg [here](https://github.com/tensorflow/probability/blob/b959b26b7b3eee31711096d140b01cf58768e5e1/tensorflow_probability/python/distributions/normal.py#L134), and then passes these explicitly to the base `Distribution` class. The end result is that if you want to chose the dtype for the distributions (as we need to when operating in 64bit mode in gpflow), the only way you can do this is by wrapping the parameters before sending them in to the constructor. Most unsatisfactory I know.

 I am more than happyto put these into sepate commits and PRs, as I do agree that's better from a git commit histoy point of view. I can restructure these now.

@st--  I'll also address all the comments  in this PR before doing that.

doc/source/notebooks/advanced/multiclass_classification.py
condnsdmatters: no clue. Theres clearly iteration overhead to consider as well as the unknown speed of `assign`. I dont think we have a choice either way, though worth investigating

gpflow/optimizers/mcmc.py
condnsdmatters: i would agree with you, but i cant comment too much but i assumed there was a reason, we can ask @awav

awav:
Hello @condnsdmatters , I apologize for closing your PR. It was closed automatically when I moved `develop-2.0` to the `develop`. Could you re-open the PR? Thanks!

gpflow/optimizers/mcmc.py
awav: We don't want to watch all variables except these which user specified. It is clear from the `tf.GradientTape` documentation.

doc/source/notebooks/advanced/multiclass_classification.py
```diff
@@ -18,10 +18,10 @@ def plot_posterior_predictions(m, X, Y):
     xx = np.linspace(X.min(), X.max(), 200).reshape(-1,1)
```
awav: Multiclass classification notebook doesn't have MCMC, right?

gpflow/optimizers/mcmc.py
```diff
@@ -0,0 +1,111 @@
...
+
+class SamplingHelper:
+    """
+    Helper reads from variables being set with a prior and writes values back to the same variables.
+
+    Example:
+        model = <Create GPflow model>
+        hmc_helper = SamplingHelper(m.trainable_parameters, lambda: -model.neg_log_marginal_likelihood())
+
+
```
awav:
```suggestion
```

gpflow/optimizers/mcmc.py
```diff
@@ -0,0 +1,111 @@
...
+        def _target_log_prob_fn_closure(*variables):
+            for v_old, v_new in zip(variables_list, variables):
+                v_old.assign(v_new)
+
+            with tf.GradientTape(watch_accessed_variables=False) as tape:
+                tape.watch(variables_list)
+                log_prob = self._target_log_prob_fn()
+
+            @tf.function
+            def grad_fn(dy, variables = None):
```
awav:
```suggestion
            def grad_fn(dy, variables: Optional[tf.Tensor] = None):
```

doc/source/notebooks/advanced/multiclass_classification.py
awav: Comparing them would be wrong as they are different things

awav: @condnsdmatters , there is a utility function for multiple assign

gpflow/optimizers/mcmc.py
condnsdmatters: I tried this at one point and it didnt work, but can give another go.

doc/source/notebooks/advanced/multiclass_classification.py
condnsdmatters: not entirely sure what you mean but doesnt have mcmc, but effectively the mcmc notebook uses some plotting functionality from multiclass_classification

condnsdmatters:
@awav  I fundamentally cant change the grad function to remove None's due to the requirements that we differentiate w.r.t to the held params on the model. As it says in the [tf.custom_gradient](https://www.tensorflow.org/api_docs/python/tf/custom_gradient) documentation:

  

> If f uses Variables (that are not part of the inputs), i.e. through get_variable, then grad_fn should have signature g(*grad_ys, variables=None), where variables is a list of the Variables, and return a 2-tuple (grad_xs, grad_vars), where grad_xs is the same as above, and grad_vars is a list<Tensor> with the derivatives of Tensors in y with respect to the variables.

doc/source/notebooks/advanced/multiclass_classification.py
awav: got it, I thought it is a notebook

gpflow/optimizers/mcmc.py
```diff
@@ -0,0 +1,110 @@
...
+        adaptive_hmc = tfp.mcmc.SimpleStepSizeAdaptation(hmc, ...)
+
+        @tf.function
+        def run_chain_fn():
+            return mcmc.sample_chain(num_samples, num_burnin_steps, current_state, kernel=adaptive_hmc)
+
+        hmc_samples = run_chain_fn()
+        parameter_samples = hmc_helper.convert_samples_to_parameter_values(hmc_samples)
+
+
```
awav:
```suggestion
```

Issue #1080, opened 2019-10-04
GPflow2 set_default_float and set_default_int accept tf dtypes but that breaks code


st--:
Some functions such as the Gauss-Hermite quadrature code call `numpy_array.astype(default_float())`, which fails if `default_float()` returns a *tensorflow* dtype such as `tf.float64` instead of `np.float64`. However, `gpflow.config.set_default_float()` explicitly allows both numpy and tensorflow dtypes as arguments.

st--:
Also, it'd be really nice to have a config context-manager to be able to have localised changes to the config...

PR #1103, opened 2019-10-16
Ensure all tf_probability distributions take the correct dtype parameters

condnsdmatters:
This PR supercedes PR # 1100

TF Probabilities does not allow you to either choose a dtype for your distributions (see PR #1100 ) at the moment. Instead it is currently inferred by the dtype of the params. Tensorflow also has no way of setting default dtypes in the backend, and this is also an open issue.

There are two solutions to this: 
* One is simply to covert native values to the default dtype before sending them into the tfp distribution. This relies on the user knowing this is how tfp distributions work.
* The other is to create an intermediate wrapper around all the distributions in tfp, and allow users of gpflow not to have to worry about this.

Although the first is uglier (users will get bitten by this, possibly), if we create a wrapper around tfp.distributions.Gamma etc, we will have to maintain it, probably for a very long time (in particular, we may have to add distributions etc). 

Ultimately this is a tf problem and I think it best if we don't turn it into our problem as well - is there a way we could communicate this problem clearly to users? If so where would we put it?

### DIFF
--- a/gpflow/config/__config__.py
+++ b/gpflow/config/__config__.py
@@ -1,4 +1,5 @@
 import os
+import contextlib
 
 import numpy as np
 import tabulate
@@ -13,6 +14,9 @@ __all__ = [
     "set_default_float",
     "set_default_jitter",
     "set_default_int",
+    "to_default_float",
+    "to_default_int",
+    "config_session"
 ]
 
 _ENV_JITTER = "GPFLOW_JITTER"
@@ -51,6 +55,14 @@ def default_jitter():
     return __config._jitter
 
 
+def to_default_int(x):
+    return tf.cast(x, dtype=default_int())
+
+
+def to_default_float(x):
+    return tf.cast(x, dtype=default_float())
+
+
 def set_default_int(value_type):
     try:
         tf_dtype = tf.as_dtype(value_type)  # Test that it's a tensorflow-valid dtype
@@ -87,3 +99,16 @@ def set_summary_fmt(fmt: str):
         raise ValueError(f"Summary does not support '{fmt}' format")
 
     __config._summary_fmt = fmt
+
+
+@contextlib.contextmanager
+def config_session():
+    '''Ensure that global configs defaults, with a context manager. Useful
+    for testing.'''
+    float_dtype = default_float()
+    int_dtype = default_int()
+    jitter = default_jitter()
+    yield
+    set_default_float(float_dtype)
+    set_default_int(int_dtype)
+    set_default_jitter(jitter)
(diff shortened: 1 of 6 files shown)


gpflow/config/__config__.py
```diff
@@ -51,6 +55,14 @@ def default_jitter():
...
+    return tf.cast(x, dtype=default_int())
+
+
+def to_default_float(x):
+    return tf.cast(x, dtype=default_float())
+
+
 def set_default_int(value_type):
     try:
         tf_dtype = tf.as_dtype(value_type)  # Test that it's a tensorflow-valid dtype
```
st--: I know it's not part of the changes in this PR but wanted to flag up #1080 at this point in case it fits in with what you're thinking about already

condnsdmatters: we can pretty easily do a context manager to solve that issue, easier perhaps once we merge this?

tests/test_prior.py
```diff
@@ -75,3 +71,24 @@ def test_map_contains_log_det_jacobian():
...
+    likelihood = gpflow.likelihoods.Gaussian()
+    data = [np.arange(5), np.arange(5)]
+    return data, kernel, likelihood
+
+
+@pytest.mark.parametrize('model_class, args', [
+    (gpflow.models.GPMC, get_GPMC_model_params()),
+    #(gpflow.models.SGPMC, get_SGPMC_model_params()) # Fails due to inducing_variable=None bug
+    ])
+def test_V_prior_dtypes(model_class, args):
```
awav:
```suggestion
def test_v_prior_dtypes(model_class, args):
```

tests/test_prior.py
```diff
@@ -75,3 +71,24 @@ def test_map_contains_log_det_jacobian():
     assert np.allclose(- m1.neg_log_marginal_likelihood().numpy(),
                        - m2.neg_log_marginal_likelihood().numpy() + m1.log_scale), \
             "MAP objective should differ by log|Jacobian| of the transform"
+
+
+def get_GPMC_model_params():
```
awav:
```suggestion
def get_gpmc_model_params():
```

tests/test_prior.py
st--: the variable is called `V`, so I think it's actually clearer to call it `test_V_prior_dtypes`.

tests/test_prior.py
```diff
@@ -75,3 +71,24 @@ def test_map_contains_log_det_jacobian():
...
+
+def get_gpmc_model_params():
+    kernel = gpflow.kernels.Matern32()
+    likelihood = gpflow.likelihoods.Gaussian()
+    data = [np.arange(5), np.arange(5)]
+    return data, kernel, likelihood
+
+
+@pytest.mark.parametrize('model_class, args', [
+    (gpflow.models.GPMC, get_GPMC_model_params()),
```
condnsdmatters:
```suggestion
    (gpflow.models.GPMC, get_gpmc_model_params()),
```

PR #1106, opened 2019-10-17
Fix hidden bug in SGPR

condnsdmatters:
SGPR likelihoods were previously using full rank matrices instead of diagonal ones in both upper bound and likelihood calculation. Ie `Kdiag` was not "diag".

This error was being masked by the intentional deactivation of tests comparing to the SGPR to the GPR, and what appears to be a hack to make tests working on the upper bound case. It was discovered in attempting to migrate the VFEvsFITC notebook.

Both are fixed here, and tests reactivated.


gpflow/models/sgpr.py
```diff
@@ -211,6 +211,31 @@ def predict_f(self, X: tf.Tensor, full_cov=False,
             var = tf.tile(var[:, None], [1, self.num_latent])
         return mean + self.mean_function(X), var
 
+    def compute_qu(self):
+        """
+        Computes the mean and variance of q(u), the variational distribution on
+        inducing outputs. SVGP with this q(u) should predict identically to
+        SGPR.
+        :return: mu, A
```
st--: What is A? the covariance of q(u)? or its cholesky? would be good if this could be consistent in code & docstring, e.g. "q(u) = N(mean, cov)" and `:return: mean, cov` or `:return mean, cholesky(cov)` ...

st--: or `q_mu` and `q_sqrt` if it is the cholesky

tests/test_methods.py
```diff
@@ -103,16 +103,14 @@ def test_sgpr_qu():
 
     @tf.function
     def closure():
-        return - model.log_likelihood()
+        return model.neg_log_marginal_likelihood()
 
     gpflow.optimizers.Scipy().minimize(closure, variables=model.trainable_variables)
+    [qu_m, qu_A] = model.compute_qu()
```
st--:
```suggestion
    qu_m, qu_A = model.compute_qu()
```

tests/test_methods.py
```diff
@@ -103,16 +103,14 @@ def test_sgpr_qu():
...
+    [f_atZ_m, f_atZ_A] = model.predict_f(model.inducing_variable.Z, full_cov=True)
```
st--:
```suggestion
    f_at_Z_mean, f_at_Z_cov = model.predict_f(model.inducing_variable.Z, full_cov=True)
```
(or `f_Z_mean` or so if you want to be more concise)

tests/test_methods.py
```diff
@@ -103,16 +103,14 @@ def test_sgpr_qu():
...
 
-    q_mu = model.mean_function(model.inducing_variable.Z.read_value())
-    q_var = model.kernel(model.inducing_variable.Z.read_value(), full=True)
-    model_qu = [q_mu, tf.reshape(q_var, (1, 20, 20))]
-
-    for v1, v2 in zip(model_qu, model.predict_f(model.inducing_variable.Z.read_value(), full_cov=True)):
-        np.testing.assert_allclose(v1, v2, rtol=0, atol=1e-7)
+    np.testing.assert_allclose(qu_m, f_atZ_m, rtol=0, atol=1e-5)
+    np.testing.assert_allclose(tf.reshape(qu_A, (1, 20, 20)), f_atZ_A, rtol=0, atol=1e-5)
```
st--: can the tolerance be any tighter? maybe trading off a bit of rtol ~ 1e-7 against atol ~ 1e-7 or so if needed?

gpflow/models/sgpr.py
```diff
@@ -55,7 +55,7 @@ def upper_bound(self):
         x_data, y_data = self.data
         num_data = tf.cast(tf.shape(y_data)[0], default_float())
 
-        Kdiag = self.kernel(x_data)
+        Kdiag = self.kernel(x_data, full=False)
```
st--: Out of scope for this PR I guess but I'd much rather call the argument something else, `diag=True` perhaps (given that we want the default to be "full covariance" whereas in other parts of the code `full_cov=False` is the default).

condnsdmatters: @awav , just flagging up.

gpflow/models/sgpr.py
condnsdmatters:
```suggestion
        :return: mu, cov
```

gpflow/models/sgpr.py
```diff
@@ -211,6 +211,31 @@ def predict_f(self, X: tf.Tensor, full_cov=False,
             var = tf.tile(var[:, None], [1, self.num_latent])
         return mean + self.mean_function(X), var
 
+    def compute_qu(self):
+        """
+        Computes the mean and variance of q(u), the variational distribution on
```
condnsdmatters:
```suggestion
        Computes the mean and variance of q(u) = N(mu, cov), the variational distribution on
```

gpflow/models/sgpr.py
```diff
@@ -211,6 +211,31 @@ def predict_f(self, X: tf.Tensor, full_cov=False,
...
+
+        kuf = Kuf(self.inducing_variable, self.kernel, x_data)
+        kuu = Kuu(self.inducing_variable, self.kernel, jitter=default_jitter())
+
+        sig = kuu + (self.likelihood.variance ** -1) * tf.matmul(kuf, kuf, transpose_b=True)
+        sig_sqrt = tf.linalg.cholesky(sig)
+
+        sig_sqrt_kuu = tf.linalg.triangular_solve(sig_sqrt, kuu)
+
+        A = tf.linalg.matmul(sig_sqrt_kuu, sig_sqrt_kuu, transpose_a=True)
```
condnsdmatters:
```suggestion
        cov = tf.linalg.matmul(sig_sqrt_kuu, sig_sqrt_kuu, transpose_a=True)
```

gpflow/models/sgpr.py
```diff
@@ -211,6 +211,31 @@ def predict_f(self, X: tf.Tensor, full_cov=False,
...
+        err = y_data - self.mean_function(x_data)
+        mu = tf.linalg.matmul(
+            sig_sqrt_kuu, tf.linalg.triangular_solve(sig_sqrt, tf.linalg.matmul(kuf, err)),
+            transpose_a=True) * self.likelihood.variance ** -1.0
+
+        return mu, A
```
condnsdmatters:
```suggestion
        return mu, cov
```

tests/test_methods.py
condnsdmatters:
```suggestion
    np.testing.assert_allclose(tf.reshape(qu_A, (1, 20, 20)), f_atZ_A, rtol=0, atol=1e-6)
```

tests/test_methods.py
```diff
@@ -103,16 +103,14 @@ def test_sgpr_qu():
...
+    qu_mean, qu_cov = model.compute_qu()
+    f_at_Z_mean, f_at_Z_cov = model.predict_f(model.inducing_variable.Z, full_cov=True)
 
-    q_mu = model.mean_function(model.inducing_variable.Z.read_value())
-    q_var = model.kernel(model.inducing_variable.Z.read_value(), full=True)
-    model_qu = [q_mu, tf.reshape(q_var, (1, 20, 20))]
-
-    for v1, v2 in zip(model_qu, model.predict_f(model.inducing_variable.Z.read_value(), full_cov=True)):
-        np.testing.assert_allclose(v1, v2, rtol=0, atol=1e-7)
+    np.testing.assert_allclose(qu_m, f_atZ_m, rtol=0, atol=1e-5)
```
condnsdmatters:
```suggestion
    np.testing.assert_allclose(qu_mean, f_atZ_mean, rtol=0, atol=1e-5)
```

tests/test_methods.py
```diff
@@ -103,16 +103,14 @@ def test_sgpr_qu():
...
+    np.testing.assert_allclose(tf.reshape(qu_A, (1, 20, 20)), f_atZ_A, rtol=0, atol=1e-6)
```
condnsdmatters:
```suggestion
    np.testing.assert_allclose(tf.reshape(qu_cov, (1, 20, 20)), f_atZ_cov, rtol=0, atol=1e-6)
```

condnsdmatters:
> LGTM - only question, are the tolerances in the fixed test in `test_methods.py` as tight as they could be (i.e. any smaller and the test fails due to numerics)?

No i can't make them tighter but i think they are genuinely only meant to be approximately similar

PR #1108, opened 2019-10-20
Traverse function for DAG (tf.Module) structures

awav:
The `traverse_module` function generalizes a way of walking through `tf.Module` based structures applying a function to the target instances in the graph. The function helps to simplify printing and deepcopy procedures.


tests/test_deepcopy.py
```diff
@@ -0,0 +1,35 @@
...
+    def __call__(self, x):
+        return self.a(x)
+
+
+@pytest.mark.parametrize('module', [A(), B()])
+def test_deepcopy_component_works(module):
+    input = 1.
+    _ = module(input)
+    with pytest.raises(TypeError):
+        deepcopy(module)
```
st--: why should this raise a `TypeError`? could you add a comment explaining why that would be expected to not work?

tests/test_deepcopy.py
```diff
@@ -0,0 +1,35 @@
...
+    module_copy = deepcopy_components(module)
+    assert module.var == module_copy.var
```
st--: Could we also have some checks that assert that a copy actually happened? (currently this test would also be satisfied by `deepcopy_components = lambda x: x`...)

tests/test_deepcopy.py
spascual: I added an explanation on the docstring of the test :)

tests/test_deepcopy.py
spascual: adding an extra-line asserting `module.var is not module_copy.var`

tests/test_deepcopy.py
```diff
@@ -0,0 +1,41 @@
...
+    This is due to the fact that HashableWeakRef objects are not pickle-able, which raises a TypeError. Alternatively,
+    one can make use of `deepcopy_component` to deepcopy a module containing used bijectors.
+    """
+    input = 1.
+    _ = module(input)
+    with pytest.raises(TypeError):
+        deepcopy(module)
+    module_copy = deepcopy_components(module)
+    assert module.var == module_copy.var
+    assert module.var is not module_copy.var
```
st--: Would it ever be possible for this assert to hold, but `module.var.assign(...)` changing the value in `module_copy.var`?

spascual: Not really. They are two different objects after the deep copy

```python 
a = tf.Variable(0.)
b  = deepcopy(a)
a.assign(2.)
Out[13]: <tf.Variable 'UnreadVariable' shape=() dtype=float32, numpy=2.0>
b
Out[14]: <tf.Variable 'Variable_copy:0' shape=() dtype=float32, numpy=0.0>
```

tests/test_deepcopy.py
```diff
@@ -0,0 +1,41 @@
...
+
+import tensorflow as tf
+import tensorflow_probability as tfp
+from gpflow.utilities import deepcopy_components
+
+
+
+class A(tf.Module):
+    def __init__(self):
+        self.var = tf.Variable(tf.random.normal((1,)))
```
st--:
```suggestion
        self.var = tf.Variable([1.0])
```

tests/test_deepcopy.py
```diff
@@ -0,0 +1,41 @@
...
+        self.bijector = tfp.bijectors.Softplus()
+
+    def __call__(self, x):
+        return self.bijector(x)
+
+
+class B(tf.Module):
+    def __init__(self):
+        self.var = tf.Variable(tf.random.normal((1,)))
```
st--:
```suggestion
        self.var = tf.Variable([2.0])
```

tests/test_deepcopy.py
st--:
```suggestion
    assert module.var is not module_copy.var
    module_copy.var.assign([5.0])
    assert module.var != module_copy.var
```

tests/test_deepcopy.py
awav: We don't raise error anymore. How does this test pass? :)

st--: it's calling python stdlib `deepcopy` not the custom `deepcopy_components`

spascual: TypeError corresponds to the error raised by standard `deepcopy` when copying a bijector with cache. I explain it in the docstring.

gpflow/utilities/utilities.py
```diff
@@ -131,45 +141,85 @@ def _merge_leaf_components(
...
+    :return:
+    """
+    path, state = acc
+
+    new_state = state
+
+    if isinstance(m, target_types):
+        return update_cb(m, path, state)
+
+    if isinstance(m, ListWrapper):
```
condnsdmatters: TL; DR - let's change `ListWrapper` and `_DictWrapper` to something thats not from the deepest guts of tensorflow. Open to discussion to what.

I'm not a huge fan of the usage of these non-exposed classes `ListWrapper` and `_DictWrapper` for a couple of reasons:
1. Since these classes are private (undocumented and buried), this is explicitly depending on implementation, not interface, leaving us with all the natural problems that arise from this. For one, comments on the implementation source, suggest this code is  [marked there for work](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/training/tracking/data_structures.py#L384)
2. I think we are effectively using these classes because the names "sound like" `list` and `dict` - not because we actually want to check instances of these specific classes. Can we be truer to our aim by either checking against `list` or `dict`, or (slightly worse) by checking for certain attributes  `has_attribute('items')`... The latter is slightly ugly (all dicts `has_attribute('items')`, not all `has_attribute('items')` are dicts) but it seems to me this is a true reflection of what actually want to check (do we _really_ care if the things is a dict?)

I ran the tests locally with `dict` and `list` and didnt see any different failures.

gpflow/utilities/utilities.py
```diff
@@ -131,45 +141,85 @@ def _merge_leaf_components(
     return {key: item for item, key in tmp_dict.items()}
 
 
-def _get_leaf_components(input: tf.Module, prefix: Optional[str] = None):
+def _get_leaf_components(input: tf.Module):
```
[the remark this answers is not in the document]
spascual: Good point!

gpflow/utilities/utilities.py
```diff
@@ -131,45 +141,85 @@ def _merge_leaf_components(
...
+
+    def update_state(parameter_or_variable, path, state):
+        state[path] = parameter_or_variable
+        return state
+
+    state = traverse_module(input, accumulator, update_state, target_types)
+    return state
+
+
+def deepcopy_components(input: tf.Module):
```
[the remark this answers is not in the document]
spascual: Yup! It requires minimal effort from my part and it could potentially be helpful for some other stuff. I will implement it and push it

gpflow/utilities/utilities.py
spascual: I agree with this. I'm not a huge fan either. Perhaps, it can be addressed later on in a separate PR (?)

condnsdmatters: I think given these are literally 2 word changes, that we should go for it with `list` and `dict`, unless @awav has a good reason why not. Here's why:
1. [ListWrapper](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/training/tracking/data_structures.py#L389) indicates that that it inherits from list, simply to allow `isinstance` checks!
2. [_DictWrapper](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/training/tracking/data_structures.py#L696) inherits from a `wrapt.ObjectProxy`, before going on to instantiate the proxy with a `dict`. The end result should still `isinstance(dict) == True` according to the [wrapt documentation](https://wrapt.readthedocs.io/en/latest/wrappers.html)

gpflow/utilities/utilities.py
```diff
@@ -131,45 +141,95 @@ def _merge_leaf_components(
...
+    new_state = state
+
+    if isinstance(m, target_types):
+        return update_cb(m, path, state)
+
+    if isinstance(m, ListWrapper):
+        for term_idx, subterm in enumerate(m):
+            new_acc = (f"{path}[{term_idx}]", new_state)
+            new_state = traverse_module(subterm, new_acc, update_cb, target_types)
+    elif isinstance(m, _DictWrapper):
```
condnsdmatters:
```suggestion
    elif isinstance(m, dict):
```

gpflow/utilities/utilities.py
condnsdmatters:
```suggestion
    if isinstance(m, list):
```

gpflow/utilities/utilities.py
```diff
@@ -1,10 +1,13 @@
...
 from functools import lru_cache
-from typing import Callable, Dict, List, Optional, Union
+from typing import Callable, Dict, List, Optional, Union, TypeVar, Any, Tuple
 
 import numpy as np
 import tensorflow as tf
+import tensorflow_probability as tfp
 from tabulate import tabulate
+
 from tensorflow.python.training.tracking.data_structures import ListWrapper, _DictWrapper
```
```suggestion
```

Issue #1095, opened 2019-10-10
Consistency of optimizers


frgsimpson:
There seem to be some differences in the behaviour of the scipy optimizer in GPflow 1 vs GPflow 2.
This might be due to the different treatment of positive parameters (the positive transform sets a lower bound of 1e-6 in Gpflow 1, while in GPflow 2 it uses tfp.bijectors.Softplus without this bound).

st--:
GPflow1 also explicitly specifies `method='L-BFGS-B'` as the default, whereas gpflow2 just lets scipy.minimize pick its default, which in our use-case turns out to be `BFGS` (which is a full-Hessian optimizer, therefore may be more aggressive and struggle with the optimisation surface, and might also be slower for large models).

st--:
Related PRs: #1110, #1112

st--:
@frgsimpson did you have a MWE that showcases the difference in behaviour? have the recent changes removed the difference ?

frgsimpson:
This issue related to the lack of a default lower bound on the likelihood variance in gpflow 2, which has now been re-introduced.

PR #1110, opened 2019-10-22
Lower-bound default positive transform

st--:
Change `gpflow.base.positive()` to behave like GPflow-1's `gpflow.transforms.positive`, i.e. instead of a "pure" softplus (log(1+exp(x)), lower-bound it by adding a small value (default = 1e-6), which prevents positively-constrained hyperparameters such as lengthscales from going too small.

This PR gives gpflow2 the same default behaviour as gpflow1. Relates to issues such as #1095. Probably warrants some discussion before merging.


awav:
Hello @st--, I apologize for closing your PR. It was closed automatically when I moved `develop-2.0` to the `develop`. Could you re-open the PR? Thanks!

frgsimpson:
I'd be keen to make any default lower bound much smaller then 1e-6, as the hyperparameters of spectral mixture kernels routinely fall to around 1e-10.

One possibility would be to enforce parameters to have well defined priors, and a default lower bound can naturally be accommodated there. (I'm a bit hesitant to put it inside a transform where the user may not be expecting it.)

jahall:
I noticed when making changes for #1148 that there is a bug in current `positive` function. Specifically that the `lower` override arg is not getting applied. This _might_ be the root cause of errors...though not investigated further yet.

Current code:
```
def positive(lower: Optional[float] = None):
    lower_value = config.default_positive_minimum()
    if lower_value is None:
        return tfp.bijectors.Softplus()
    ...
```
Corrected:
```
def positive(lower: Optional[float] = None):
    if lower is None:
        lower = config.default_positive_minimum()
    if lower is None:
        return tfp.bijectors.Softplus()
    ...
```

jahall:
Actually, ignore me...the root of your problem is that when you compose the `tfp.bijectors.Chain` you need to perform the shift _after_ the positive bijection not before.

jahall:
Fixed in #1148

PR #1111, opened 2019-10-22
Change step_callback behaviour of Scipy optimizer in gpflow2 to behave as in gpflow1

st--:
Changes behaviour of `step_callback` so that the callback only gets called once per optimisation step, not once per objective function evaluation. Makes gpflow2's Scipy() behave more like gpflow1's ScipyOptimizer()/monitoring.

### DIFF
--- a/gpflow/optimizers/scipy.py
+++ b/gpflow/optimizers/scipy.py
@@ -21,7 +21,7 @@ class Scipy:
                  **scipy_kwargs) -> OptimizeResult:
         """
         Minimize is a wrapper around the `scipy.optimize.minimize` function
-        handling the packing and unpacking of lists of shaped variables on the
+        handling the packing and unpacking of a list of shaped variables on the
         TensorFlow side vs. the flat numpy array required on the Scipy side.
 
         Args:
--- a/gpflow/optimizers/scipy.py
+++ b/gpflow/optimizers/scipy.py
@@ -8,8 +8,7 @@ from scipy.optimize import OptimizeResult
 __all__ = ['Scipy']
 
 Loss = tf.Tensor
-Variables = List[tf.Variable]
-Gradients = List[tf.Tensor]
+Variables = Tuple[tf.Variable]
 StepCallback = Callable[[int, Variables, List[tf.Tensor]], None]
 LossClosure = Callable[..., Tuple[tf.Tensor, Variables]]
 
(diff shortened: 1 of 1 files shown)


gpflow/optimizers/scipy.py
```diff
@@ -34,27 +36,44 @@ def minimize(self,
...
         step = 0  # type: int
 
-        def _eval(x):
+        def _callback(x):
             nonlocal step
             cls.unpack_tensors(variables, x)
             loss, grads = _compute_loss_and_gradients(closure, variables)
-            if step_callback is not None:
-                step_callback(step=step, loss=loss, variables=variables, gradients=grads)
+            step_callback(step=step, loss=loss, variables=variables, gradients=grads)
```
condnsdmatters: so by splitting out the callback from the eval function, we need to compute the loss and gradients twice. is there anyway to avoid this? i suspect not, but this may well seriously slow down, depending on how computationally heavy it is to calc the objective (is there memoization?)

condnsdmatters: if not does it make any sense to attempt some memoization?

condnsdmatters: **Proposal**:
1. We keep the same function as before: that is the `eval_func` has the same signature `def eval_func(cls, closure, variables, step_callback)`
2. In the `eval_func` we have non_local variables which include the `loss` and `gradient`, and we proceed to make not one, but _two_ closures - just as you have done in your change. One for eval, one for callback.
3. The `eval` closure now also writes to the `non_local` `loss` and `gradient`. The `callback` closure now only reads from these values. 
4. Since the variables and `non_local`, they can't leak outside the closure (without serious work), and avoid global state, while saving on the function call.

awav:
I don't understand the title and I don't understand the changes. GPflow 1.0 uses default method option for the `scipy.minimize`, GPflow 2.0 follows the same logic. The intention always was to make gpflow `scipy.minimize` version as close as possible to actual `scipy`.

gpflow/optimizers/scipy.py
st--: Neat idea but I don't think would work as the last call to eval_func during a line search may not be the one that gives the optimal value.

Changed the step_callback interface so that instead of recomputing loss&gradient every step(!), to only pass variables and (unpacked) values (as tensors) so that user can assign as required (or not).

gpflow/optimizers/scipy.py
```diff
@@ -34,44 +34,65 @@ def minimize(self,
...
     @classmethod
-    def eval_func(cls, closure: LossClosure, variables: Variables, step_callback: Optional[StepCallback] = None):
-        step = 0  # type: int
-
+    def eval_func(cls, closure: LossClosure, variables: Variables):
         def _eval(x):
-            nonlocal step
-            cls.unpack_tensors(variables, x)
+            values = cls.unpack_tensors(variables, x)
+            cls.assign_tensors(variables, values)
```
awav: This will lead to a huge slowdown. The idea behind step_callback: the user decides when to do assigning of values to the models.

st--:
@awav I've split this PR into two (see #1112 for the other one).

Here, I want to get the callback behaviour to work as in GPflow 1, i.e., only call `step_callback` once per optimisation step, as noted in @condnsdmatters's first comment.

gpflow/optimizers/scipy.py
```diff
@@ -34,44 +34,65 @@ def minimize(self,
         if not callable(closure):
             raise TypeError('Callable object expected.')  # pragma: no cover
         initial_params = self.initial_parameters(variables)
-        func = self.eval_func(closure, variables, step_callback)
+        func = self.eval_func(closure, variables)
+        if step_callback is not None:
+            if 'callback' in scipy_kwargs:
+                raise ValueError("Callback passed both via `step_callback` and `callback`")
+
+            callback = self.callback_func(closure, variables, step_callback)
```
awav: If the default value is None, it should stay None. Otherwise, it is just calling of the empty callback wasting cpu clock time (python function call isn't like c function call :)

gpflow/optimizers/scipy.py
```diff
@@ -34,27 +36,44 @@ def minimize(self,
         if not callable(closure):
             raise TypeError('Callable object expected.')  # pragma: no cover
         initial_params = self.initial_parameters(variables)
-        func = self.eval_func(closure, variables, step_callback)
-        return scipy.optimize.minimize(func, initial_params, jac=True, **scipy_kwargs)
+        func = self.eval_func(closure, variables)
+
+        if step_callback is not None:
+            if 'callback' in scipy_kwargs:
+                raise ValueError("Callback passed both via `step_callback` and `callback`")
```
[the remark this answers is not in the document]
awav: @st-- , @condnsdmatters these are different callbacks! The `callback` is for scipy.minimize method, `step_callback` for our part. Don't mix them up! :)

gpflow/optimizers/scipy.py
st--: This is inside the eval_func, not the callback_func.

gpflow/optimizers/scipy.py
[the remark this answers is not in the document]
st--: It's the same thing, we just wrap the step_callback. Alternatively, we can ensure that `scipy_kwargs` does NOT contain a `callback` ever.

gpflow/optimizers/scipy.py
st--: I don't understand what you mean - that's why I have the `if step_callback is not None:` check?

st--: Or do you want to replace the `if 'callback' in scipy_kwargs:` with `if scipy_kwargs.get('callback', None) is not None:`?

gpflow/optimizers/scipy.py
st--: it's doing the same assign as before - just more explicitly!

Issue #1090, opened 2019-10-08
Model training in the GPflow-2-version of the coregionalisation notebook is very slow


fleibfried:
System information:

* Python version: 3.7.4
* TensorFlow installed from binary
* TensorFlow version: 2.0.0-dev20191002
* GPflow installed from source
* GPflow version: 1.9.999

... _Describe the current behavior_
Training a model with a coregionalisation kernel takes much longer compared to the GPflow-1-version.

... _Describe the expected behavior_
Faster model training.

... _Code to reproduce the issue Provide a reproducible test case that is the bare minimum necessary to generate the problem._
see branch 'felix/coregionalisation' in the directory 'notebooks2/advanced'

st--:
(See #1072 for the gpflow2 version of the coregionalisation.ipynb notebook)

st--:
It looks like it may be some weird interaction with the (L-)BFGS(-B) optimizer default -- it seems to work much better (faster, in terms of wallclock speed if not necessarily in loss vs iterations) with conjugate gradients (`method='CG'` argument to Scipy().minimize())...

st--:
Turns out the issue was that the default method of Scipy optimizer had changed from explicitly selecting `L-BFGS-B` in gpflow1 to letting scipy.minimize pick its default which ended up being `BFGS` (see #1111). BFGS constructs a full Hessian, which explains why this notebook was so slow (and why it used up so much memory when running). I've now fixed the method to be L-BFGS-B, so regardless of merging #1111 this notebook is now fast & gives the answers that we expect.

PR #1112, opened 2019-10-22
Change gpflow2 Scipy optimizer default method to L-BFGS-B (as in gpflow1)

st--:
Explicitly sets the default optimization method for Scipy() to L-BFGS-B (as in gpflow1), instead of letting scipy.minimize() figure it out. This resulted in BFGS, which may be too aggressive when encountering complex optimization surfaces, so L-BFGS-B may be a more sensible default.


awav:
Are we sure that we want this change? I would say that default value shoudl bfgs, we never discussed why gpflow scipy version had l-bfgs by default.

awav:
In current scipy.minimize method we could just reference to the documentation in https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.minimize.html, after that change we will have to make a note that our default value is different.

st--:
`BFGS` estimates the full Hessian at each step, which makes it super slow for models with a reasonably large number of parameters, as it constructs a #param x #param matrix (also gives a big memory hit). This is what resulted in the significant slow-down in the Coregionalisation notebook (#1072, #1090). Optimising models with 10^2 inducing points -> 10^4 parameters is fairly common. Operating with 10^4 x 10^4 matrices by default is painful. I think `L-BFGS-B` is a more sensible default (and reproduces gpflow1 behaviour). I do agree that we should carefully think about this - e.g. it makes lots of tests fail (though I also wonder how they passed before in gpflow1!), which is why I split it off from #1111...

markvdw:
I'm with Ti.

gpflow/optimizers/scipy.py
```diff
@@ -14,6 +14,8 @@
 
 
 class Scipy:
+    _DEFAULT_METHOD = 'L-BFGS-B'
```
awav:
```suggestion
```

gpflow/optimizers/scipy.py
```diff
@@ -14,6 +14,8 @@
...
+
```
awav:
```suggestion
```

gpflow/optimizers/scipy.py
```diff
@@ -14,6 +14,8 @@
...
     def minimize(self,
                  closure: LossClosure,
                  variables: Variables,
```
awav:
```suggestion
                  variables: Variables,
                  method: Optional[str] = "L-BFGS-B",
```

gpflow/optimizers/scipy.py
```diff
@@ -46,14 +48,18 @@ def minimize(self,
...
         func = self.eval_func(closure, variables)
         if step_callback is not None:
             if 'callback' in scipy_kwargs:
                 raise ValueError("Callback passed both via `step_callback` and `callback`")
 
             callback = self.callback_func(closure, variables, step_callback)
             scipy_kwargs.update(dict(callback=callback))
-        return scipy.optimize.minimize(func, initial_params, jac=True, **scipy_kwargs)
+
+        method = scipy_kwargs.pop('method', self._DEFAULT_METHOD)
```
awav:
```suggestion
```

st--:
Sure! I only did it this way because that's how it was in the old (gpflow1) `ScipyOptimizer`, so I thought there was a good reason to do it that way...

PR #1114, opened 2019-10-23
Set summary format is now fixed

spascual:
The logic for loading the default summary format in `print_summary` used to be loaded at `tabulate_module_summary`. 

Thus whenever we used:
```python
set_summary_fmt("notebook")
print_summary(my_module)
```
the format tabulate would read would not be `html`. Instead, it would be 'notebook' which is not listed in `tabulate.tabulate_formats`.

Moved loading the default summary format to `print_summary` directly to avoid this :)


gpflow/utilities/utilities.py
```diff
@@ -95,7 +95,6 @@ def print_summary(module: tf.Module, fmt: str = None):
 
 
 def tabulate_module_summary(module: tf.Module, tablefmt: str = None) -> str:
```
condnsdmatters:
```suggestion
def tabulate_module_summary(module: tf.Module, tablefmt: Optional[str] = None) -> str:
```

condnsdmatters:
```suggestion
def tabulate_module_summary(module: tf.Module, tablefmt: Optional[str] = None) -> str:
```

condnsdmatters: frustratingly cant change the one above but yeah

PR #1121, opened 2019-10-28
Unblacklist some notebooks so tests can run and delete the monitoring

condnsdmatters:
notebook, since it is not needed in gpflow2


condnsdmatters:
This notebook was deleted per discussion with @awav

awav:
@condnsdmatters you need to remove settings notebook as well.

condnsdmatters:
Looks like the natural gradients test is failing - will inspect

awav:
@condnsdmatters , it is fixed.

PR #1132, opened 2019-11-07
Use experimental_ref when using tf.Variable as key in Dicts

vdutor:
We can't put variables in a set/dictionary as `variable.__hash__()` is no longer available starting Tensorflow 2.0. This caused problems in the `utilities._merge_leaf_components` function. This PR fixes the issue and adds a test.


vdutor:
@awav tests fail because of wrong importing of `positive` transform in `changepoint.py`, which is a file I did not touch in this PR. I fixed the bug in #1133 , please merge that one first.

gpflow/utilities/utilities.py
```diff
@@ -155,25 +155,31 @@ def get_transform(v):
         get_transform(variable), variable.trainable, variable.shape, variable.dtype.name,
         _str_tensor_value(variable.numpy())
     ] for path, variable in merged_leaf_components.items()]
+    # bla
```
```suggestion
```

gpflow/kernels/stationaries.py
```diff
@@ -18,19 +18,23 @@ class Stationary(Kernel):
...
-        - active_dims is a list of length input_dim which controls which
-          columns of X are used.
         - if ard is not None, it specifies whether the kernel has one
           lengthscale per dimension (ard=True) or a single lengthscale
           (ard=False). Otherwise, inferred from shape of lengthscale.
+        - kwargs, accepts `name` and `active_dims`, which is a list of
+          length input_dim which controls which columns of X are used.
         """
-        super().__init__(active_dims)
+        for kwarg in kwargs:
```
awav:
```suggestion
        if not set(kwargs.keys()).issubset(['name', 'active_dims']):
```

awav: This also require removing of the next line.

awav: You haven't changed it, right?

PR #1135, opened 2019-11-07
Refactor config with a dataclass

awav:
* Refactor config with the dataclass and enums
* Add an option for the positive transform bias
* Make the global config readable

### DIFF
--- a/gpflow/config/__config__.py
+++ b/gpflow/config/__config__.py
@@ -9,20 +9,23 @@ import tabulate
 import tensorflow as tf
 
 __all__ = [
-    "summary_fmt", "default_float", "default_jitter", "default_int", "set_summary_fmt", "set_default_float",
-    "set_default_jitter", "set_default_int", "as_context", "config", "set_config", "positive_minimum",
-    "set_positive_minimum"
+    "default_summary_fmt", "default_float", "default_jitter", "default_int", "set_default_summary_fmt",
+    "set_default_float", "set_default_jitter", "set_default_int", "as_context", "config", "set_config",
+    "default_positive_minimum", "set_default_positive_minimum"
 ]
 
 __config = None
 
 
 class _Values(enum.Enum):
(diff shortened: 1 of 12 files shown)


gpflow/config/__config__.py
```diff
@@ -1,106 +1,142 @@
...
-    return __config._float
+    return config().float
 
 
 def default_jitter():
-    return __config._jitter
+    return config().jitter
+
+
+def positive_minimum():
```
vdutor: why not add `default` here to be consistent?

gpflow/config/__config__.py
```diff
@@ -1,106 +1,142 @@
...
+    return config().positive_minimum
+
+
+def set_config(new_config: Config):
```
vdutor: Would be nice to have a one line docstring here

gpflow/config/__config__.py
```diff
@@ -1,106 +1,142 @@
...
+import enum
+import os
+from dataclasses import dataclass, field, replace
+from typing import Optional
 
 import numpy as np
 import tabulate
 import tensorflow as tf
 
 __all__ = [
```
vdutor: A little example on how to use this config would be great! 
- how to get the default value
- how to change it

gpflow/config/__config__.py
```diff
@@ -1,106 +1,142 @@
...
+    jitter: float = field(default_factory=lambda: default(_Values.JITTER))
+    positive_minimum: float = field(default_factory=lambda: default(_Values.POSITIVE_MINIMUM))
+    summary_fmt: str = _Values.SUMMARY_FMT.value
+
+
+def config() -> Config:
+    return __config
 
 
 def summary_fmt():
```
vdutor: maybe add `default` here as well

gpflow/config/__config__.py
awav: Do you mean the documentation for the `config.py`?

PR #1138, opened 2019-11-08
GPflow2 small fixes

nfergu:
Some small typing and documentation fixes for GPflow 2

Regarding the `TensorLike` type definition not working with multipledispatch, I think that multipledispatch expects a class, rather than a type definition, which is why it was failing. In the cases where we were using multipledispatch with `TensorLike` I think we should have just been using `object`.


st--:
I'd be strongly in favour of not just multiple-dispatching on `object`, this is more confusing when looking at the source code itself. The `multipledispatch` package does handle unions of types, but unfortunately using tuples instead of `typing.Union[]`. Maybe we could make a PR to the multipledispatch package to support that? Should be a fairly simple fix in https://github.com/mrocklin/multipledispatch/blob/b62be264c89cdddb39eb4a4d0d0e525d99b5133b/multipledispatch/dispatcher.py#L211 ...

nfergu:
> I'd be strongly in favour of not just multiple-dispatching on `object`, this is more confusing when looking at the source code itself...

@st-- A couple of comments on this:

 * I think that the code could be made clearer by using type hints (for example, the `Kuf_matern32_fourierfeatures1d` function could be declared as `Kuf_matern32_fourierfeatures1d(inducing_variable: FourierFeatures1D,  kernel: Matern32, X: TensorLike)`) but this is separate from the types that are used by multipledispatch.
  
 * I think that multipledispatch expects a class, but `Union` is (as I understand it) a type definition. So I'm not sure it would be easy to make the two compatible

 * I think it would be possible to define a tuple of `(tf.Tensor, tf.Variable, np.ndarray)` for the purposes of woirking with multipledispatch, if you think this would make things clearer. Personally I think that `object` is a good way of marking paramaters that you don't care about (that you don't want to dispatch on), but I haven't used multipledispatch very much, so don't feel strongly about this.

st--:
> I think that the code could be made clearer by using type hints (for example, the Kuf_matern32_fourierfeatures1d function could be declared as Kuf_matern32_fourierfeatures1d(inducing_variable: FourierFeatures1D, kernel: Matern32, X: TensorLike)) but this is separate from the types that are used by multipledispatch.

`multipledispatch` can actually make use of type annotations to dispatch (see https://github.com/mrocklin/multipledispatch/blob/b62be264c89cdddb39eb4a4d0d0e525d99b5133b/multipledispatch/dispatcher.py#L204 and following lines)! so maybe we could unify this more.

> I think that multipledispatch expects a class, but Union is (as I understand it) a type definition. So I'm not sure it would be easy to make the two compatible
I do think it could be made compatible, but would involve a code change upstream in `multipledispatch` (basically, when a type is a `Union[]`, extract its subtypes, and loop over those, instead of only hard-coding this behaviour for tuples).

st--:
@nfergu could you split out the typo/docstring fixes in scipy.py (which I'd be happy to approve straight away) from the TensorLike changes (which haven't been resolved yet)?

PR #1147, opened 2019-11-13
Small fixup for GPLVM and VGP.

condnsdmatters:
Missing `full_output_cov` parameter prevents us from calling `predict_f_samples` from the base class. Crash occurs.

### DIFF
--- a/gpflow/models/vgp.py
+++ b/gpflow/models/vgp.py
@@ -197,7 +197,8 @@ class VGPOpperArchambeau(GPModel):
         v_exp = self.likelihood.variational_expectations(f_mean, f_var, y_data)
         return tf.reduce_sum(v_exp) - KL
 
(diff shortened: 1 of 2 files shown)


gpflow/models/vgp.py
```diff
@@ -197,7 +197,8 @@ def log_likelihood(self):
         v_exp = self.likelihood.variational_expectations(f_mean, f_var, y_data)
         return tf.reduce_sum(v_exp) - KL
 
-    def predict_f(self, predict_at: DataPoint, full_cov: bool = False):
+    def predict_f(self, predict_at: DataPoint, full_cov: bool = False,
```
awav: We have 120 char limit for the line ;)

condnsdmatters: gross!

Issue #1144, opened 2019-11-12
[Discussion] Contributing


jahall:
# Contributing
I would be keen to contribute more regularly to this project as I think it is awesome. Just wondering what the most useful way to go about this would be? Do you have a backlog of bugs/enhancements/documentation for which you would welcome additional help? Or maybe you have sufficient support, which is also fine...

awav:
Hello @jahall.

Thanks for your interest in the GPflow! We don't have a board where share our plans or intentions (yet). Github projects is a good place for planning and organizing work in the future. As for now, we have a couple of things in the todo list: 1. gpflow 1.0 positive transform had shift term equal 1e-6, this https://github.com/GPflow/GPflow/pull/1110 is returning it back, but some tests are failing and we need to figure out what's wrong with it. 2. Current positive transform uses softplus, but in some cases it is "preferable" to use exp instead; it would be nice to have an option in the gpflow.config, such that the user could choose global positive transformation.

jahall:
Great stuff. Thanks for this! Feel free to close the "issue". Wasn't sure the best way to reach out :)

jahall:
Both points should now be addressed in #1148

PR #1148, opened 2019-11-13
Add config option for positive bijector + re-introduce positive lower bound

jahall:
Based on @awav's second TODO list item in #1144: _"Current positive transform uses softplus, but in some cases it is "preferable" to use exp instead; it would be nice to have an option in the gpflow.config, such that the user could choose global positive transformation"_

Still a work in progress but putting it up here for early feedback (as I calibrate coding style for this project).

### DIFF
--- a/gpflow/config/__config__.py
+++ b/gpflow/config/__config__.py
@@ -31,7 +31,7 @@ class _Values(enum.Enum):
     INT = np.int32
     FLOAT = np.float64
     POSITIVE_BIJECTOR = "softplus"
-    POSITIVE_MINIMUM = None
+    POSITIVE_MINIMUM = 1e-10
     SUMMARY_FMT = None
     JITTER = 1e-6
 
--- a/gpflow/config/__config__.py
+++ b/gpflow/config/__config__.py
@@ -31,7 +31,7 @@ class _Values(enum.Enum):
     INT = np.int32
     FLOAT = np.float64
     POSITIVE_BIJECTOR = "softplus"
-    POSITIVE_MINIMUM = 1e-10
+    POSITIVE_MINIMUM = None
     SUMMARY_FMT = None
     JITTER = 1e-6
 
--- a/gpflow/config/__config__.py
+++ b/gpflow/config/__config__.py
@@ -144,11 +144,11 @@ _VALID_POSITIVE_BIJECTORS = {"exp", "softplus"}
 
 def set_default_positive_bijector(value: str):
     if not isinstance(value, str):
-        TypeError(f"Expected a string, but got a {type(value)}")
+        raise TypeError(f"Expected a string, but got a {type(value)}")
 
     value = value.lower()
     if value not in _VALID_POSITIVE_BIJECTORS:
-        ValueError(f"Value '{value}' not in {sorted(_VALID_POSITIVE_BIJECTORS)}")
+        raise ValueError(f"'{value}' not found in {sorted(_VALID_POSITIVE_BIJECTORS)}")
 
     set_config(replace(config(), positive_bijector=value))
 
--- a/gpflow/config/__config__.py
+++ b/gpflow/config/__config__.py
@@ -2,11 +2,12 @@ import contextlib
 import enum
 import os
 from dataclasses import dataclass, field, replace
-from typing import Optional
+from typing import Optional, Union
 
 import numpy as np
 import tabulate
 import tensorflow as tf
+import tensorflow_probability as tfp
 
 
 __all__ = [
@@ -45,6 +46,18 @@ def default(value: _Values):
     return os.getenv(value.name, default=value.value)
 
 
+def _to_bijector(value: str) -> tfp.bijectors.Bijector:
+    bijector_map = {
+        "exp": tfp.bijectors.Exp,
+        "softplus": tfp.bijectors.Softplus,
+    }
+    if isinstance(value, str):
+        value = value.lower()
+    if value not in bijector_map:
+        raise ValueError(f"`{value}` not in set of valid bijectors: {sorted(bijector_map)}")
+    return bijector_map[value]()
+
+
 @dataclass(frozen=True)
 class Config:
     """
@@ -64,7 +77,8 @@ class Config:
     int: type = field(default_factory=lambda: default(_Values.INT))
     float: type = field(default_factory=lambda: default(_Values.FLOAT))
     jitter: float = field(default_factory=lambda: default(_Values.JITTER))
-    positive_bijector: str = field(default_factory=lambda: default(_Values.POSITIVE_BIJECTOR))
+    positive_bijector: tfp.bijectors.Bijector = field(
+        default_factory=lambda: _to_bijector(default(_Values.POSITIVE_BIJECTOR)))
     positive_minimum: float = field(default_factory=lambda: default(_Values.POSITIVE_MINIMUM))
     summary_fmt: str = field(default_factory=lambda: default(_Values.SUMMARY_FMT))
 
@@ -139,15 +153,11 @@ def set_default_jitter(value: float):
     set_config(replace(config(), jitter=value))
 
 
-def set_default_positive_bijector(value: str):
-    if not isinstance(value, str):
-        raise TypeError(f"Expected a string, but got a {type(value)}")
-
-    valid_bijectors = {"exp", "softplus"}
-
-    value = value.lower()
-    if value not in valid_bijectors:
-        raise ValueError(f"'{value}' not found in {sorted(valid_bijectors)}")
+def set_default_positive_bijector(value: Union[str, tfp.bijectors.Bijector]):
+    if isinstance(value, str):
+        value = _to_bijector(value)
+    elif not isinstance(value, tfp.bijectors.Bijector):
+        raise TypeError(f"Expected a Bijector, but got a {type(value)}")
 
     set_config(replace(config(), positive_bijector=value))
 
--- a/gpflow/config/__config__.py
+++ b/gpflow/config/__config__.py
@@ -8,11 +8,18 @@ import numpy as np
 import tabulate
 import tensorflow as tf
 
+
 __all__ = [
-    "Config", "default_summary_fmt", "default_float", "default_jitter", "default_int", "set_default_summary_fmt",
-    "set_default_float", "set_default_jitter", "set_default_int", "as_context", "config", "set_config",
-    "default_positive_minimum", "set_default_positive_minimum"
+    "Config", "as_context",
+    "config", "set_config",
+    "default_float", "set_default_float",
+    "default_int", "set_default_int",
+    "default_jitter", "set_default_jitter",
+    "default_positive_bijector", "set_default_positive_bijector",
+    "default_positive_minimum", "set_default_positive_minimum",
+    "default_summary_fmt", "set_default_summary_fmt",
 ]
+
 
 __config = None
 
@@ -23,6 +30,7 @@ class _Values(enum.Enum):
     will be `GPFLOW_SUMMARY_FMT`."""
     INT = np.int32
     FLOAT = np.float64
+    POSITIVE_BIJECTOR = "softplus"
     POSITIVE_MINIMUM = None
     SUMMARY_FMT = None
     JITTER = 1e-6
@@ -33,7 +41,7 @@ class _Values(enum.Enum):
 
 
 def default(value: _Values):
-    """Checks if """
+    """Checks if value is set in the environment."""
     return os.getenv(value.name, default=value.value)
 
 
@@ -47,6 +55,8 @@ class Config:
         float: Float data type, float32 or float64
         jitter: Jitter value. Mainly used for for making badly conditioned matrices more stable.
             Default value is `1e-6`.
+        positive_bijector: Method for positive bijector, either "softplus" or "exp".
+            Default is "softplus".
         positive_minimum: Lower level for the positive transformation.
         summary_fmt: Summary format for module printing.
     """
@@ -54,6 +64,7 @@ class Config:
     int: type = field(default_factory=lambda: default(_Values.INT))
     float: type = field(default_factory=lambda: default(_Values.FLOAT))
     jitter: float = field(default_factory=lambda: default(_Values.JITTER))
+    positive_bijector: str = field(default_factory=lambda: default(_Values.POSITIVE_BIJECTOR))
     positive_minimum: float = field(default_factory=lambda: default(_Values.POSITIVE_MINIMUM))
     summary_fmt: str = field(default_factory=lambda: default(_Values.SUMMARY_FMT))
 
@@ -61,10 +72,6 @@ class Config:
 def config() -> Config:
     """Returns current active config."""
     return __config
-
-
-def default_summary_fmt():
-    return config().summary_fmt
 
 
 def default_int():
@@ -79,8 +86,16 @@ def default_jitter():
     return config().jitter
 
 
+def default_positive_bijector():
+    return config().positive_bijector
+
+
 def default_positive_minimum():
     return config().positive_minimum
+
+
+def default_summary_fmt():
+    return config().summary_fmt
 
 
 def set_config(new_config: Config):
@@ -124,12 +139,18 @@ def set_default_jitter(value: float):
     set_config(replace(config(), jitter=value))
 
 
-def set_default_summary_fmt(value: str):
-    formats = tabulate.tabulate_formats + ['notebook', None]
-    if value not in formats:
-        raise ValueError(f"Summary does not support '{value}' format")
+_VALID_POSITIVE_BIJECTORS = {"exp", "softplus"}
 
-    set_config(replace(config(), summary_fmt=value))
+
+def set_default_positive_bijector(value: str):
+    if not isinstance(value, str):
+        TypeError(f"Expected a string, but got a {type(value)}")
+
+    value = value.lower()
+    if value not in _VALID_POSITIVE_BIJECTORS:
+        ValueError(f"Value '{value}' not in {sorted(_VALID_POSITIVE_BIJECTORS)}")
+
+    set_config(replace(config(), positive_bijector=value))
 
 
 def set_default_positive_minimum(value: float):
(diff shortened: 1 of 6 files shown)


jahall:
Also re-introducing positive lower bound (set to `1e-10` as per @frgsimpson's comment in #1110 for now) and fixing associated bug.

gpflow/config/__config__.py
```diff
@@ -23,7 +30,8 @@ class _Values(enum.Enum):
     will be `GPFLOW_SUMMARY_FMT`."""
     INT = np.int32
     FLOAT = np.float64
-    POSITIVE_MINIMUM = None
+    POSITIVE_BIJECTOR = "softplus"
+    POSITIVE_MINIMUM = 1e-10
```
awav: I would recommend to keep same as before `1e-6`. @frgsimpson can change it to 1e-10, when he needs it :)

gpflow/config/__config__.py
```diff
@@ -124,12 +139,18 @@ def set_default_jitter(value: float):
     set_config(replace(config(), jitter=value))
 
 
-def set_default_summary_fmt(value: str):
-    formats = tabulate.tabulate_formats + ['notebook', None]
-    if value not in formats:
-        raise ValueError(f"Summary does not support '{value}' format")
+_VALID_POSITIVE_BIJECTORS = {"exp", "softplus"}
```
awav: Could you put it either into the function or in the beginning of the file. (I prefer function itself)

gpflow/utilities/bijectors.py
```diff
@@ -5,17 +5,50 @@
...
-    lower_value = config.default_positive_minimum()
-    if lower_value is None:
-        return tfp.bijectors.Softplus()
+_POSITIVE_BIJECTOR_MAP = {
+    "exp": tfp.bijectors.Exp,
+    "softplus": tfp.bijectors.Softplus,
+}
+
+
+class Shift(tfp.bijectors.AffineScalar):
```
awav: * Check the comment in the test about printing.
* I believe we don't need this custom class

gpflow/utilities/bijectors.py
```diff
@@ -5,17 +5,50 @@
...
+             bijector: Optional[str] = None) -> tfp.bijectors.Bijector:
+    """
+    Returns a positive bijector (a reversible transformation from real to positive numbers).
+
+    :param lower: lower bound override (defaults to config.default_positive_minimum())
+    :param bijector: bijector method override (defaults to config.default_positive_bijector())
+    :returns: a bijector instance
+    """
+    bijector = _get_base_positive_bijector(bijector)
+    if lower is None:
```
awav:
```suggestion
    default = config.default_positive_minimum()
    lower = default if lower is None else lower
```

awav: You need update two lines instead of one, github doesn't allow to do suggestions for multiple lines.

gpflow/utilities/bijectors.py
```diff
@@ -5,17 +5,50 @@
...
+        lower = config.default_positive_minimum()
+    if lower is not None:
+        # Apply lower bound shift after applying base positive bijector
+        shift = Shift(shift=to_default_float(lower))
+        bijector = tfp.bijectors.Chain([bijector, shift])
```
awav:
```suggestion
        bijector = tfp.bijectors.Chain([shift, bijector])
```

Shift should be first

gpflow/utilities/bijectors.py
```diff
@@ -5,17 +5,50 @@
...
-    bijectors = [tfp.bijectors.AffineScalar(shift=shift), tfp.bijectors.Softplus()]
-    return tfp.bijectors.Chain(bijectors)
+def _get_base_positive_bijector(bijector_name: Optional[str] = None):
+    if bijector_name is None:
+        bijector_name = config.default_positive_bijector()
+    return _POSITIVE_BIJECTOR_MAP[bijector_name]()
 
 
-triangular = tfp.bijectors.FillTriangular
+def triangular():
```
awav: Could you add type annotation for return value?

gpflow/utilities/bijectors.py
```diff
@@ -5,17 +5,50 @@
...
+        # Apply lower bound shift after applying base positive bijector
+        shift = Shift(shift=to_default_float(lower))
+        bijector = tfp.bijectors.Chain([bijector, shift])
+    return bijector
+
 
-    shift = to_default_float(lower_value)
-    bijectors = [tfp.bijectors.AffineScalar(shift=shift), tfp.bijectors.Softplus()]
-    return tfp.bijectors.Chain(bijectors)
+def _get_base_positive_bijector(bijector_name: Optional[str] = None):
```
awav: Could you add type annotation to return value?

tests/test_printing.py
```diff
@@ -186,32 +186,32 @@ def create_model():
...
------------------------------------------  ---------  -----------  -----------  -------  -------  -------\n\
-Product.kernels[0].kernels[0].variance     Parameter  Softplus     True         ()       float64        1\n\
-Product.kernels[0].kernels[0].lengthscale  Parameter  Softplus     False        ()       float64        2\n\
-Product.kernels[0].kernels[1].variance     Parameter  Softplus     True         ()       float64        1\n\
-Product.kernels[0].kernels[1].lengthscale  Parameter  Softplus     False        ()       float64        2\n\
-Product.kernels[1].kernels[0].variance     Parameter  Softplus     True         ()       float64        1\n\
-Product.kernels[1].kernels[0].lengthscale  Parameter  Softplus     False        ()       float64        2\n\
-Product.kernels[1].kernels[1].variance     Parameter  Softplus     True         ()       float64        1\n\
-Product.kernels[1].kernels[1].lengthscale  Parameter  Softplus     False        ()       float64        2"""
+name                                       class      transform         trainable    shape    dtype      value\n\
```
awav: I would recommend to set config lower value to None, this test printing is going to be pain in the neck in general. Let's just constrain it to limited settings.

gpflow/config/__config__.py
frgsimpson: @awav  I'd really advise against it. For example, think of the consequences of imposing the positive transform on likelihood.variance (which I guess is the default). A 1e-6 minimum means that no GPflow model can reach an absolute precision of better than 1e-3!

gpflow/utilities/bijectors.py
jahall: I think the current code is cleaner. It also only fetches the default value when necessary rather than on every call. Any strong preference for this suggestion?

gpflow/utilities/bijectors.py
jahall: Actually I think bijector needs to be first - I believe that was the cause of the original bug in #1110. Here is a counterexample (in some kind of pseudo-code...):
```
unconstrained = -20.0
constrained = softplus(unconstrained + 1e-6)
constrained > 1e-6  # False!
```
whereas
```
unconstrained = -20.0
constrained = softplus(unconstrained) + 1e-6
constrained > 1e-6  # True
```

tests/test_printing.py
jahall: I'm happy to set these tests back to lower value = None. But I would like to add an explicit test in for the case where it is not None and keep the more informative print string:
```
transform
----------------
Softplus + Shift
```
instead of the uninformative
```
transform
---------
Chain
```

gpflow/utilities/bijectors.py
jahall: I agree this custom class _might_ be overkill...but the benefit of an informative print string (especially as this will become the new default) is compelling.

tests/test_printing.py
jahall: ...though maybe this is still overkill

gpflow/utilities/bijectors.py
jahall: Resolving this conversation - can move discussion to `test_printing`

gpflow/config/__config__.py
jahall: I feel like erring on the side of having a lower default makes sense...but up to you guys.

gpflow/utilities/bijectors.py
awav: :) my preference would be to switch to python 3.8 and use walrus operator. As it is not really possible, therefore don't have a strong opinion about it.

gpflow/utilities/bijectors.py
awav: Indeed, I checked documentation, https://www.tensorflow.org/probability/api_docs/python/tfp/bijectors/Chain. It is clearer what's happening. Good catch!

gpflow/utilities/bijectors.py
jahall: Yes! Bring on the walrus...one day

gpflow/utilities/bijectors.py
awav: This class should be removed, it will confuse people. The print should reflect existing TFP bijectors, not GPflow's aliases.

jahall: Yes, this makes sense

gpflow/config/__config__.py
awav: Okay, I think we need to revert it back to the None value and think about the documentation for this in next PRs.

jahall: At least we know it doesn't break anything now.

gpflow/config/__config__.py
```diff
@@ -124,12 +139,17 @@ def set_default_jitter(value: float):
...
-def set_default_summary_fmt(value: str):
-    formats = tabulate.tabulate_formats + ['notebook', None]
-    if value not in formats:
-        raise ValueError(f"Summary does not support '{value}' format")
+def set_default_positive_bijector(value: str):
+    if not isinstance(value, str):
+        raise TypeError(f"Expected a string, but got a {type(value)}")
 
-    set_config(replace(config(), summary_fmt=value))
+    valid_bijectors = {"exp", "softplus"}
```
st--: Could this simply be taken as `gpflow.utilities.bijectors._POSITIVE_BIJECTOR_MAP.keys()` instead? More DRY.

gpflow/config/__config__.py
```diff
@@ -124,12 +139,17 @@ def set_default_jitter(value: float):
     set_config(replace(config(), jitter=value))
 
 
-def set_default_summary_fmt(value: str):
-    formats = tabulate.tabulate_formats + ['notebook', None]
-    if value not in formats:
-        raise ValueError(f"Summary does not support '{value}' format")
+def set_default_positive_bijector(value: str):
+    if not isinstance(value, str):
+        raise TypeError(f"Expected a string, but got a {type(value)}")
```
st--: This error message could be a bit confusing - it doesn't actually expect "just a string", but a specific set of values. As a user, it would be more helpful to directly get the message "It has to be 'A' or 'B', not 5.0". Could deal with the lowering by "if isinstance(value, str): value = value.lower()"; if it's not a string it still won't be in the set of valid options.

st--: (Similar to set_default_summary_fmt)

gpflow/config/__config__.py
awav: @st-- , @jahall 

Yes! But, it should be in the config itself as an immutable field. The other way around is wrong.

```python
@dataclass(frozen=True)
class Config:
    """
    Immutable object for storing global GPflow settings
    Args:
        int: Integer data type, int32 or int64.
        float: Float data type, float32 or float64
        jitter: Jitter value. Mainly used for for making badly conditioned matrices more stable.
            Default value is `1e-6`.
        positive_bijector: Method for positive bijector, either "softplus" or "exp".
            Default is "softplus".
        positive_minimum: Lower level for the positive transformation.
        summary_fmt: Summary format for module printing.
    """

    int: type = field(default_factory=lambda: default(_Values.INT))
    float: type = field(default_factory=lambda: default(_Values.FLOAT))
    jitter: float = field(default_factory=lambda: default(_Values.JITTER))
    positive_bijector: str = field(default_factory=lambda: default(_Values.POSITIVE_BIJECTOR))
    positive_minimum: float = field(default_factory=lambda: default(_Values.POSITIVE_MINIMUM))
    summary_fmt: str = field(default_factory=lambda: default(_Values.SUMMARY_FMT))
    positive_type: Dict[str, type] = dict(exp=tfp.bijectors.Exp, softplus=tfp.bijectors.Softplus)
```

The only thing I don't like is that dictionary itself here is immutable.

jahall: Yep, agree that config shouldn't be importing from other modules. Happy to implement @awav's suggestion.

jahall: I refactored so that the actual tfp.bijector.Bijector instance lives in `Config.positive_bijector` rather than a plain string so should resolve these issues...though maybe we don't want "complex" datastructures like these in the config?

gpflow/config/__config__.py
```diff
@@ -143,10 +173,17 @@ def set_default_positive_minimum(value: float):
...
+        raise ValueError(f"Summary does not support '{value}' format")
+
+    set_config(replace(config(), summary_fmt=value))
+
+
 @contextlib.contextmanager
 def as_context(temporary_config: Optional[Config] = None):
-    '''Ensure that global configs defaults, with a context manager. Useful
-    for testing.'''
+    """Ensure that global configs defaults, with a context manager. Useful for testing."""
```
st--: This 
```suggestion
    """Ensure that global configs defaults, with a context manager. Useful for testing."""!
```
still doesn't make sense to me :( not your code but wanted to flag it up now that I noticed if you don't mind figuring out the intended meaning...

gpflow/config/__config__.py
```diff
@@ -47,13 +68,17 @@ class Config:
...
             Default value is `1e-6`.
+        positive_bijector: Method for positive bijector, either "softplus" or "exp".
+            Default is "softplus".
         positive_minimum: Lower level for the positive transformation.
         summary_fmt: Summary format for module printing.
     """
 
     int: type = field(default_factory=lambda: default(_Values.INT))
     float: type = field(default_factory=lambda: default(_Values.FLOAT))
     jitter: float = field(default_factory=lambda: default(_Values.JITTER))
```
st--: I don't think this will ensure that the environment variables are converted to type objects / float values (and same for `positive_minimum` below)! again -not fault of this PR- but thought important to flag this up now that I spotted it

gpflow/utilities/bijectors.py
```diff
@@ -5,17 +5,31 @@
...
+        shift = tfp.bijectors.AffineScalar(shift=to_default_float(lower))
+        bijector = tfp.bijectors.Chain([bijector, shift])
+    return bijector
+
+
+def triangular() -> tfp.bijectors.Bijector:
+    """
+    Returns instance of a triangular bijector.
+    """
+    return tfp.bijectors.FillTriangular()
```
st--: Is it wrong to re-use bijectors ?
In that case, the code in `__config__.py` needs changing, too, as there the config's `positive` bijector is already the instance, not the type (`default_positve_bijector()` returns the config value, doesn't call it).

gpflow/utilities/bijectors.py
st--: It would be great if you could add this as a test!

gpflow/config/__config__.py
awav: Looks like I was drunk when were writing this :) @jahall skip it, I will fix it in a separate single PR with other doc fixes.

gpflow/utilities/bijectors.py
awav: Test for what, that people read the documentation :) ?

gpflow/utilities/bijectors.py
awav: Yes, it is wrong to re-use bijectors. They have cache inside.

gpflow/config/__config__.py
awav: Indeed, @jahall I will fix it in separate PR.

gpflow/config/__config__.py
```diff
@@ -33,10 +42,22 @@ def name(self):
...
+def _to_bijector(value: str) -> tfp.bijectors.Bijector:
+    bijector_map = {
+        "exp": tfp.bijectors.Exp,
+        "softplus": tfp.bijectors.Softplus,
+    }
+    if isinstance(value, str):
+        value = value.lower()
+    if value not in bijector_map:
+        raise ValueError(f"`{value}` not in set of valid bijectors: {sorted(bijector_map)}")
+    return bijector_map[value]()
```
st--: This already instantiates the bijector, so the config will always return the same object, which means sharing the bijector which according to @awav is actually wrong...

gpflow/utilities/bijectors.py
st--: Then have to change the __config__ code, too.

gpflow/config/__config__.py
jahall: Good point - I think having the actual bijectors in the config is maybe too complex - let me refactor.

st--: It should be fine to have the definitions in `bijectors.py` and still get the valid keys from there inside `__config__.py` without having to manually duplicate! :)

st--:
```suggestion
    return bijector_map[value]
```
and then instantiate it inside bijectors.py's `positive()` :)

gpflow/utilities/bijectors.py
jahall: Soooo, turns out I was totally wrong. `Chain` applies the transformations in reverse order so the original implementation was correct - good job you suggested adding a test!

tests/test_utilities.py
```diff
@@ -0,0 +1,43 @@
...
+        bijector = positive(base=override_bijector)
+        assert isinstance(bijector, expected_class)
+
+
+def test_positive_calculation_order():
+    value, lower = -10.0, 10.0
+    expected = np.exp(value) + lower
+    with as_context(Config(positive_bijector="exp", positive_minimum=lower)):
+        result = positive()(value).numpy()
+    assert np.isclose(result, expected)
```
st--:
```suggestion
    assert np.isclose(result, expected)
    assert result > lower
```

PR #1149, opened 2019-11-13
Gaussian process with neural network notebook

awav:
Update to the `gp-nn.ipynb` notebook. Convert it to gpflow 2.0.


doc/source/notebooks/tailor/gp_nn.ipynb
```diff
@@ -210,198 +180,132 @@
...
-    }
-   },
+   "metadata": {},
    "outputs": [
     {
      "name": "stdout",
      "output_type": "stream",
      "text": [
-      "acc is 98.5900\n"
+      "Accuracy is 20.6055%\n"
```
vdutor: We probably want to check why performance is so off before we merge this. Was the model trained long enough?

awav:
@vdutor , run it longer, got 97.2233

doc/source/notebooks/tailor/gp_nn.ipynb
awav: @vdutor , run it longer, got 97.2233

PR #1150, opened 2019-11-13
Update CircleCI config

awav:


awav:
@condnsdmatters the deploy initially was in gpflow 2.0 branch, it was turned off to avoid accident deploy

PR #1155, opened 2019-11-14
quick fix of documentation mathematics

condnsdmatters:
Boring but important doc error

### DIFF
--- a/gpflow/conditionals/util.py
+++ b/gpflow/conditionals/util.py
@@ -18,7 +18,7 @@ def base_conditional(Kmn: tf.Tensor,
 
       p(g1) = N(g1; 0, Knn)
 
-      p(g1 | g2) = N(g1; Knm(Kmm^-1)g2, Knn - Knm(Kmm^-1)Kmn)
+      p(g1 | g2) = N(g1; Knm (Kmm⁻¹) g2, Knn - Knm (Kmm⁻¹) Kmn)
 
     And
       q(g2) = N(g2; f, q_sqrt * q_sqrt^T)
--- a/gpflow/conditionals/util.py
+++ b/gpflow/conditionals/util.py
@@ -15,8 +15,10 @@ def base_conditional(Kmn: tf.Tensor,
     """
     Given a g1 and g2, and distribution p and q such that
       p(g2) = N(g2; 0, Kmm)
+
       p(g1) = N(g1; 0, Knn)
-      p(g1 | g2) = N(g1;0,Knm)
+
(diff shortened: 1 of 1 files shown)


gpflow/conditionals/util.py
```diff
@@ -15,8 +15,10 @@ def base_conditional(Kmn: tf.Tensor,
     """
     Given a g1 and g2, and distribution p and q such that
       p(g2) = N(g2; 0, Kmm)
+
       p(g1) = N(g1; 0, Knn)
-      p(g1 | g2) = N(g1;0,Knm)
+
+      p(g1 | g2) = N(g1; Knm(Kmm^-1)g2, Knn - Knm(Kmm^-1)Kmn)
```
st--:
```suggestion
      p(g1 | g2) = N(g1; Kₙₘ Kₘₘ⁻¹ g2, Kₙₙ - Kₙₘ Kₘₘ⁻¹ Kₘₙ)
```
(would be nice updating the other maths with unicode as well but not necessary for this PR!)

gpflow/conditionals/util.py
```diff
@@ -15,8 +15,10 @@ def base_conditional(Kmn: tf.Tensor,
     """
     Given a g1 and g2, and distribution p and q such that
       p(g2) = N(g2; 0, Kmm)
+
       p(g1) = N(g1; 0, Knn)
```
st--:
```suggestion
      p(g1) = N(g1; 0, Kₙₙ)
```

gpflow/conditionals/util.py
```diff
@@ -15,8 +15,10 @@ def base_conditional(Kmn: tf.Tensor,
     """
     Given a g1 and g2, and distribution p and q such that
       p(g2) = N(g2; 0, Kmm)
```
st--:
```suggestion
      p(g2) = N(g2; 0, Kₘₘ)
```

gpflow/conditionals/util.py
st--: Or just changing the inverse:
```suggestion
      p(g1 | g2) = N(g1; Knm (Kmm⁻¹) g2, Knn - Knm (Kmm⁻¹) Kmn)
```
given that `Kmm` is actually the variable name in the code...

gpflow/conditionals/util.py
condnsdmatters: discussed and ignoring

gpflow/conditionals/util.py
condnsdmatters: discussed and ignoring

PR #1158, opened 2019-11-17
Porting the generalized periodic kernel

jahall:
This is a port of the generalized periodic kernel (introduced to v1.0 in #1052). I have not attempted to make this one backward compatible as it would be part of v2.0 and therefore assuming a breaking change is ok (but I could be wrong).

### DIFF
--- a/doc/source/notebooks/advanced/kernels.ipynb
+++ b/doc/source/notebooks/advanced/kernels.ipynb
@@ -105,7 +105,7 @@
     "plotkernelsample(gpflow.kernels.Constant(), axes[1,0])\n",
     "plotkernelsample(gpflow.kernels.Linear(), axes[1,1])\n",
     "plotkernelsample(gpflow.kernels.Cosine(), axes[1,2])\n",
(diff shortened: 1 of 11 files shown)


doc/source/notebooks/advanced/kernels.ipynb
```diff
@@ -105,7 +105,7 @@
     "plotkernelsample(gpflow.kernels.Constant(), axes[1,0])\n",
     "plotkernelsample(gpflow.kernels.Linear(), axes[1,1])\n",
     "plotkernelsample(gpflow.kernels.Cosine(), axes[1,2])\n",
-    "plotkernelsample(gpflow.kernels.Periodic(), axes[1,3])\n",
+    "plotkernelsample(gpflow.kernels.Periodic(gpflow.kernels.RBF()), axes[1,3])\n",
```
st--:
```suggestion
    "plotkernelsample(gpflow.kernels.Periodic(gpflow.kernels.SquaredExponential()), axes[1,3])\n",
```
let's try and get away from `RBF` please :)

PR #1159, opened 2019-11-18
Stabilise upper bound of `SGPR`.

markvdw:
Improved numerical stability of upper bound.


gpflow/models/sgpr.py
```diff
@@ -59,10 +59,12 @@ def upper_bound(self):
         kuu = Kuu(self.inducing_variable, self.kernel, jitter=default_jitter())
         kuf = Kuf(self.inducing_variable, self.kernel, x_data)
 
-        KufKfu = tf.linalg.matmul(kuf, kuf, transpose_b=True)
+        I = tf.eye(tf.shape(kuu)[0], dtype=default_float())
 
         L = tf.linalg.cholesky(kuu)
-        LB = tf.linalg.cholesky(kuu + KufKfu / self.likelihood.variance)
+        A = tf.linalg.triangular_solve(L, kuf, lower=True)
+        B = I + tf.linalg.matmul(A, A, transpose_b=True) / self.likelihood.variance
```
awav: You can compute it only once:  tf.linalg.matmul(A, A, transpose_b=True)

markvdw: Will do.

doc/source/notebooks/theory/upper_bound.ipynb
```diff
@@ -322,8 +331,8 @@
      "name": "stdout",
      "output_type": "stream",
      "text": [
-      "Lower bound: -65.059355\n",
-      "Upper bound: -65.058996\n"
+      "Lower bound: -62.487148\n",
+      "Upper bound: -62.479897\n"
```
st--: do we know why it changed ? are we not setting the random seed? would be good to have notebooks be reproducible ...

gpflow/models/sgpr.py
```diff
@@ -49,6 +49,11 @@ class SGPRUpperMixin(GPModel):
         year={2014},
         month={Dec}
       }
+
+    The key quantity, the trace term, can be computed with
+            # To compute each individual element of the trace term, use...
```
st--: do we need both of these lines? seems redundant

PR #1162, opened 2019-11-22
Assign lower bound directly to likelihood variance.

frgsimpson:
For an easier transition from v1, and improved stability during optimisation, this sets a lower bound on the likelihood variance. 

Compared to imposing the lower bound within the positive transform, this is (a) more transparent to the user and (b) does not impose the restriction on other Parameters.

### DIFF
--- a/gpflow/likelihoods/likelihoods.py
+++ b/gpflow/likelihoods/likelihoods.py
@@ -164,9 +164,9 @@ class Gaussian(Likelihood):
     optimization process. A lower bound of 1e-6 is therefore imposed on the likelihood variance
     by default.
     """
-    def __init__(self, variance=1.0, **kwargs):
+    def __init__(self, variance=1.0, variance_lower_bound=1e-6, **kwargs):
         super().__init__(**kwargs)
-        self.variance = Parameter(variance, transform=positive(lower=1e-6))
+        self.variance = Parameter(variance, transform=positive(lower=variance_lower_bound))
 
     def log_prob(self, F, Y):
         return logdensities.gaussian(Y, F, self.variance)
--- a/gpflow/likelihoods/likelihoods.py
+++ b/gpflow/likelihoods/likelihoods.py
@@ -156,6 +156,14 @@ class Likelihood(tf.Module):
 
 
 class Gaussian(Likelihood):
+    r"""
+    The Gaussian likelihood is appropriate where uncertainties associated with the data are
+    believed to follow a normal distribution, with constant variance.
+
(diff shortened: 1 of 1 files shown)


gpflow/likelihoods/likelihoods.py
```diff
@@ -158,7 +158,7 @@ def variational_expectations(self, Fmu, Fvar, Y):
 class Gaussian(Likelihood):
     def __init__(self, variance=1.0, **kwargs):
         super().__init__(**kwargs)
-        self.variance = Parameter(variance, transform=positive())
+        self.variance = Parameter(variance, transform=positive(lower=1e-6))
```
awav: This line deserves an explanation in documentation.

st--:
@frgsimpson Right now, this change would be a no-op, as the default lower for the positive transform is also 1e-6. Did you mean to then change that to None or 1e-10 or something else smaller ?

jahall:
> @frgsimpson Right now, this change would be a no-op, as the default lower for the positive transform is also 1e-6. Did you mean to then change that to None or 1e-10 or something else smaller ?

Actually the default is still `None` as per the thread in #1148

awav:
@st-- ,

> Right now, this change would be a no-op

.The current default value for the positive shift is None and therefore no shift is applied. Default value was discussed with @frgsimpson, @markvdw. The obscurity of the fact that all positive values by default can't go below `1e-6` has big impact on results misinterpretation for a model. The point in favor of having `1e-6` threshold is that it **might** have positive effect on optimization. What if it is not and moreover we don't see much of the change in current tests for example. Now, we have different settings in the config for changing the lower positive value minimum and the type of positive transformation - setting 'right' (e.g. `1e-10`, `1e-6` and etc.) default values doesn't look right for this degree of freedom.

gpflow/likelihoods/likelihoods.py
awav: @frgsimpson ^^^

frgsimpson: @awav I've just added a docstring for the class, let me know if that's what you had in mind

st--: Actually, how about just turning this into a keyword argument to the class ?
```python
    def __init__(self, variance=1.0, variance_lower_bound=1e-6, **kwargs):
        super().__init__(**kwargs)
        self.variance = Parameter(variance, transform=positive(lower=variance_lower_bound))
```

frgsimpson: Done!

PR #1163, opened 2019-11-22
ELBO and log marginal likelhood of the SVGP has one data argument

awav:
* "Unify" API by removing separate X and Y from the ELBO and log marginal likelihood
* Update notebooks and tests accordingly


tests/test_uncertain_conditional.py
```diff
@@ -261,16 +240,15 @@ def mean_sq_fn(X):
...
-        white=white)
+    mean_analytic, var_analytic = uncertain_conditional(DataQuad.Xmu,
+                                                        DataQuad.Xvar,
+                                                        inducing_variable,
+                                                        kernel,
+                                                        DataQuad.q_mu,
+                                                        DataQuad.q_sqrt,
+                                                        mean_function=mean_function,
+                                                        full_output_cov=False,
+                                                        white=white)
```
st--: Do different people have different settings for their editors(' auto-rejustify)? It seems like unrelated code keeps changing indentation and whether arguments are stuffed on one or two lines vs one argument per line - I find this quite distracting from reviewing PRs, would it be possible to avoid having this happen? Like a dedicated PR that *only* changes the whitespace, and gets everything into one consistent format (that we would have to agree on beforehand!)?

tests/test_uncertain_conditional.py
```diff
@@ -64,9 +62,7 @@ def gen_L(n, *shape):
 
 
 def gen_q_sqrt(D_out, *shape):
-    return tf.convert_to_tensor(np.array(
-        [np.tril(rng.randn(*shape)) for _ in range(D_out)]),
-                                dtype=default_float())
+    return tf.convert_to_tensor(np.array([np.tril(rng.randn(*shape)) for _ in range(D_out)]), dtype=default_float())
```
st--: FWIW, I found the previous state more readable...

tests/test_uncertain_conditional.py
awav: `.style.yapf`


```
[style]
based_on_style: pep8
indent_width: 4
split_before_logical_operator = true
column_limit = 100
```

tests/test_methods.py
```diff
@@ -103,7 +102,7 @@ def test_sgpr_qu():
 
     @tf.function
     def closure():
-        return - model.log_marginal_likelihood()
+        return -model.log_marginal_likelihood()
```
st--: FWIW, I find it more helpful to have the space there to highlight that it's the negative. Without the space it's easy to miss. (I actually couldn't find a reference to unary minus operator in PEP-8...)

tests/test_uncertain_conditional.py
st--:
```suggestion
    return tf.convert_to_tensor(np.array(
        [np.tril(rng.randn(*shape)) for _ in range(D_out)]
        ), dtype=default_float())
```

tests/test_uncertain_conditional.py
```diff
@@ -237,21 +222,15 @@ def test_quadrature(white, mean):
...
 
     def mean_fn(X):
         return conditional_fn(X)[0] + effective_mean(X)
 
     def var_fn(X):
         return conditional_fn(X)[1]
 
-    quad_args = DataQuad.Xmu, DataQuad.Xvar, DataQuad.H, DataQuad.D_in, (
-        DataQuad.D_out, )
+    quad_args = DataQuad.Xmu, DataQuad.Xvar, DataQuad.H, DataQuad.D_in, (DataQuad.D_out, )
```
st--:
```suggestion
    quad_args = DataQuad.Xmu, DataQuad.Xvar, DataQuad.H, DataQuad.D_in, (DataQuad.D_out,)
```
PEP-8

st--: (actually there's a bunch more cases around the codebase that have the same pattern)

tests/test_method_equivalence.py
```diff
@@ -29,6 +29,7 @@ class Datum:
     Y = np.sin(X) + 0.9 * np.cos(X * 1.6) + rng.randn(*X.shape) * 0.8
     Y = np.tile(Y, 2)  # two identical columns
     Xtest = rng.rand(10, 1) * 10
+    data = X, Y
```
st--: I don't have any strong preference between `data = X, Y` as here or `data = (X, Y)` as below, but would be good to be consistent!

tests/test_likelihoods.py
```diff
@@ -405,16 +395,16 @@ def test__switched_likelihood_variational_expectations(Y_list, F_list, Fvar_list
...
+def test_switched_likelihood_regression_valid_num_latent(num_latent):
     """
     A Regression test when using Switched likelihood: the number of latent
     functions in a GP model must be equal to the number of columns in Y minus
     one. The final column of Y is used to index the switch. If the number of
     latent functions does not match, an exception will be raised.
     """
+    x = np.random.randn(100, 1)
+    y = np.hstack((np.random.randn(100, 1), np.random.randint(0, 3, (100, 1))))
+    data = x, y
```
st--: why not use `X` and `Y` as elsewhere, for consistency?

tests/test_likelihoods.py
```diff
@@ -64,31 +64,24 @@ def __repr__(self):
...
-    pytest.param(LikelihoodSetup(MultiClass(2),
-                                 Y=tf.argmax(Datum.Y, 1).numpy().reshape(-1, 1),
-                                 rtol=1e-3, atol=1e-3),
+    LikelihoodSetup(Beta(), Y=tf.random.uniform(Datum.Yshape, dtype=default_float())),
+    LikelihoodSetup(Ordinal(np.array([-1, 1])), Y=tf.random.uniform(Datum.Yshape, 0, 3, dtype=default_int())),
+    LikelihoodSetup(Poisson(invlink=tf.square), Y=tf.random.poisson(Datum.Yshape, 1.0, dtype=default_float())),
+    LikelihoodSetup(Exponential(invlink=tf.square), Y=tf.random.uniform(Datum.Yshape, dtype=default_float())),
+    LikelihoodSetup(Gamma(invlink=tf.square), Y=tf.random.uniform(Datum.Yshape, dtype=default_float())),
+    LikelihoodSetup(Bernoulli(invlink=tf.sigmoid), Y=tf.random.uniform(Datum.Yshape, dtype=default_float())),
+    pytest.param(LikelihoodSetup(MultiClass(2), Y=tf.argmax(Datum.Y, 1).numpy().reshape(-1, 1), rtol=1e-3, atol=1e-3),
```
st--: Would the previous formatting not make it clearer that the last one has different tolerances, and how the choice of Y relates to the likelihood distribution?

gpflow/models/svgp.py
```diff
@@ -132,26 +131,27 @@ def prior_kl(self):
                                           self.q_sqrt,
                                           whiten=self.whiten)
 
-    def log_likelihood(self, X: tf.Tensor, Y: tf.Tensor) -> tf.Tensor:
+    def log_likelihood(self, data: Tuple[tf.Tensor, tf.Tensor]) -> tf.Tensor:
         """
         This gives a variational bound on the model likelihood.
         """
+        x, y = data
```
st--:
```suggestion
        X, Y = data
```
I'm in favour of using the upper-case variables to indicate that these are matrices

doc/source/notebooks/advanced/convolutional.ipynb
```diff
@@ -367,19 +325,22 @@
      "name": "stdout",
      "output_type": "stream",
      "text": [
-      "Train acc: 100.0%\n",
-      "Test acc : 98.33333333333333%\n",
-      "conv elbo after training: -4.4511e+01\n"
+      "Train acc: 97.0%\n",
+      "Test acc : 98.0%\n",
+      "conv elbo after training: -4.2114e+01\n"
```
st--: why is this different now ?

tests/test_methods.py
awav: https://softwareengineering.stackexchange.com/a/189055, it is a general style rule for unary operators in many languages

doc/source/notebooks/advanced/convolutional.ipynb
awav: Different seed? It is a common problem, that we often use np.random without caring about seed and etc., but fixing it is out of the scope of this PR

tests/test_methods.py
st--: > No space should separate a unary operator and its operand except when the operator is a word such as typeof.

which is the case here. Also, clarity >> our own style conforming to other people's style (we're already deviating with upper-case one-character variable names, and for a good reason)

tests/test_uncertain_conditional.py
st--: We still shouldn't have drive-by formatting changes, this makes PRs significantly more unclear.

tests/test_methods.py
st--: > Avoid extraneous whitespace in the following situations: [...] Between a trailing comma and a following close parenthesis.
(ref: https://www.python.org/dev/peps/pep-0008/#pet-peeves)

tests/test_uncertain_conditional.py
awav: Yes, agree. Sorry for that, I try to keep changes minimal, but sometimes fail :)

tests/test_uncertain_conditional.py
st--: > Avoid extraneous whitespace in the following situations: [...] Between a trailing comma and a following close parenthesis.

(ref: https://www.python.org/dev/peps/pep-0008/#pet-peeves)

tests/test_method_equivalence.py
st--:
```suggestion
    data = (X, Y)
```

tests/test_uncertain_conditional.py
awav: To be honest the whole line is horrible and should be splitted into multiple variables :)

tests/test_uncertain_conditional.py
```diff
@@ -210,19 +201,15 @@ def test_monte_carlo_2_din(white, mean):
 
     @tf.function
     def closure():
-        return - model.log_marginal_likelihood(DataMC2.X, DataMC2.Y)
+        return -model.log_marginal_likelihood(DataMC2.data)
```
st--:
```suggestion
        return - model.log_marginal_likelihood(DataMC2.data)
```

tests/test_uncertain_conditional.py
```diff
@@ -178,19 +173,15 @@ def test_monte_carlo_1_din(white, mean):
 
     @tf.function
     def closure():
-        return - model.log_marginal_likelihood(DataMC1.X, DataMC1.Y)
+        return -model.log_marginal_likelihood(DataMC1.data)
```
st--:
```suggestion
        return - model.log_marginal_likelihood(DataMC1.data)
```

tests/test_uncertain_conditional.py
```diff
@@ -147,12 +145,9 @@ def test_no_uncertainty(white, mean):
 
     @tf.function
     def closure():
-        return - model.log_marginal_likelihood(Data.X, Data.Y)
+        return -model.log_marginal_likelihood(Data.data)
```
st--:
```suggestion
        return - model.log_marginal_likelihood(Data.data)
```

tests/test_methods.py
```diff
@@ -60,50 +68,53 @@ def _check_models_close(m1, m2, tolerance=1e-2):
...
 def test_sgpr_qu():
+    rng = Datum().rng
     X, Z = tf.cast(rng.randn(100, 2), default_float()), tf.cast(rng.randn(20, 2), default_float())
     Y = tf.cast(np.sin(X @ np.array([[-1.4], [0.5]])) + 0.5 * np.random.randn(len(X), 1), default_float())
     model = gpflow.models.SGPR((X, Y), kernel=gpflow.kernels.SquaredExponential(), inducing_variable=Z)
 
     @tf.function
     def closure():
-        return - model.log_marginal_likelihood()
+        return -model.log_marginal_likelihood()
```
st--:
```suggestion
        return - model.log_marginal_likelihood()
```

tests/test_likelihoods.py
awav: This is auto formatting, I don't think it is matter for the test to change. Pls, do not force to change it back :)

PR #1166, opened 2019-11-26
Fix printing test

awav:
subj.


awav:
Develop is failing, because @frgsimpson changed Gaussian likelihood default value for variance #1162. This is fixing for printing test case by setting to default value back. @st-- , @vdutor

awav:
For now, I'm merging it.

Issue #831, opened 2018-08-22
SGPR build likelihood is incorrect for Multioutput


sanket-kamthe:
The multiplication by output_dim in the likelihood is not needed if using multi-output kernels
link (https://github.com/GPflow/GPflow/blob/develop/gpflow/models/sgpr.py#L175)

See line 175 below: 

```
        bound += -0.5 * output_dim * tf.reduce_sum(Kdiag) / self.likelihood.variance
        bound += 0.5 * output_dim * tf.reduce_sum(tf.linalg.diag_part(AAT))
```
This is correct if `Y` is multidim for single kernel. 

There is no clean way I could think of handling this issue without a discussion. 

Also, should we have had a test to find this bug ?

hughsalimbeni:
How about something like `output_dim = tf.where(tf.less(tf.rank(Kdiag), 2), output_dim, 1)`?

sanket-kamthe:
Well, I would like to avoid putting conditions on the gradient path of the model, especially when we can determine this before compiling the graph. 

Are we anticipating a scenario where `rank(kdiag) ` changes at run time ?

vdutor:
I haven't tested this my self, but I think you can have a condition based on `len(Kdiag.shape) == 2` which should work at graph compiling time.

hughsalimbeni:
I'd suggest that `len(Kdiag.shape) == 2` can be dangerous if the shape is None, as at run time the shape could then be anything

st--:
@sanket-kamthe it's an old issue but still seems relevant - would be great if you'd be up for writing a test (that currently would fail, but that we can work towards fixing). Thank you!

st--:
Closing this in favour of the larger issue captured in #1209

PR #1167, opened 2019-12-02
Shared interface for number of output dimensions in multi-output kernels

condnsdmatters:
Multioutput base.


gpflow/kernels/mo_kernels.py
```diff
@@ -80,6 +86,10 @@ def __init__(self, kernel: Kernel, output_dimensionality: int):
         self.kernel = kernel
         self.P = output_dimensionality
 
+    @property
+    def num_output_dims(self):
+        return self.P
```
st--: Could we unify this to all be one thing? `self.output_dim = output_dim` in the `__init__`, then no further property is needed?:)

gpflow/kernels/mo_kernels.py
```diff
@@ -35,6 +35,12 @@ class MultioutputKernel(Kernel):
     `full_output_cov` is set to True the covariance matrix will be filled with zeros
     until the appropriate size is reached.
     """
+
+    @property
+    @abc.abstractmethod
```
st--: `@abc.abstractproperty`

condnsdmatters: Deprecated since version 3.3:

gpflow/kernels/mo_kernels.py
condnsdmatters: sure, it means creating an __init__ in some cases where there isn't one so far but yes. I like the abstract property because it means new subclasses can't forget to define the property.

gpflow/kernels/mo_kernels.py
st--: oh! had missed that!

gpflow/kernels/mo_kernels.py
st--: Mainly I want to avoid property, attribute, and `__init__` argument all being different names...

condnsdmatters: what do you mean? Also what do you think of `num_output_dims`? I think its a bit long, but ive met others who dont like `output_dims` (which generally i prefer)

st--: My first choice would be `output_dim` (no plural, it's just a scalar). `num_output_dim(s)` is ok too.

gpflow/kernels/mo_kernels.py
```diff
@@ -76,10 +85,11 @@ class SharedIndependent(MultioutputKernel):
     Use `gpflow.kernels` instead for more efficient code.
     """
     def __init__(self, kernel: Kernel, output_dimensionality: int):
```
st--:
```suggestion
    def __init__(self, kernel: Kernel, output_dim: int):
```

gpflow/kernels/mo_kernels.py
```diff
@@ -76,10 +85,11 @@ class SharedIndependent(MultioutputKernel):
...
-        super().__init__()
+        super().__init__(output_dim=output_dimensionality)
```
st--:
```suggestion
        super().__init__(output_dim=output_dim)
```

gpflow/kernels/base.py
```diff
@@ -167,8 +167,8 @@ class Combination(Kernel):
 
     _reduction = None
 
-    def __init__(self, kernels: List[Kernel], name: Optional[str] = None):
-        super().__init__(name=name)
+    def __init__(self, kernels: List[Kernel], name: Optional[str] = None, **kwargs):
+        super().__init__(name=name, **kwargs)
```
st--: If you're already just passing through, might as well drop the `name` argument (I don't think anyone should be setting that positionally anyways).

This raises an issue with documentation - how will end users know what the accepted arguments are in the end? Sphinx would have to parse the `__init__` of every single parent class. Does it actually do that?

gpflow/kernels/mo_kernels.py
```diff
@@ -35,6 +35,15 @@ class MultioutputKernel(Kernel):
...
     until the appropriate size is reached.
     """
+
+    def __init__(self, output_dim: int, **kwargs):
+        self._output_dim = output_dim
+        super().__init__(**kwargs)
+
+    @property
+    def output_dim(self):
+        raise NotImplementedError
```
st--:
```suggestion
        return self._output_dim
```
no?

gpflow/kernels/mo_kernels.py
```diff
@@ -76,10 +85,11 @@ class SharedIndependent(MultioutputKernel):
     Use `gpflow.kernels` instead for more efficient code.
     """
     def __init__(self, kernel: Kernel, output_dimensionality: int):
-        super().__init__()
+        super().__init__(output_dim=output_dimensionality)
         self.kernel = kernel
         self.P = output_dimensionality
```
st--:
```suggestion
```

st--: Replace usages with `.output_dim` instead

gpflow/kernels/mo_kernels.py
```diff
@@ -100,7 +110,8 @@ class SeparateIndependent(MultioutputKernel, Combination):
     - Independent: Latents are uncorrelated a priori.
     """
     def __init__(self, kernels, name=None):
-        Combination.__init__(self, kernels, name)
+        super().__init__(output_dim=len(kernels), kernels=kernels, name=name)
```
st--: pass through `**kwargs`

gpflow/kernels/mo_kernels.py
```diff
@@ -137,7 +148,7 @@ class LinearCoregionalization(IndependentLatent, Combination):
     Linear mixing of the latent GPs to form the output.
     """
     def __init__(self, kernels, W, name=None):
-        Combination.__init__(self, kernels, name)
+        super().__init__(output_dim=len(kernels), kernels=kernels, name=name)
```
st--: pass through `**kwargs`.

(also something that we should do for *all* kernels? :thinking: )

tests/test_multioutput.py
```diff
@@ -307,6 +307,25 @@ def test_fully_correlated_conditional_repeat_shapes(func, R):
...
+
+    k1 = mk.LinearCoregionalization(kern_list, W=data.W)
+    assert k1.num_output_dims == data.L
+
+    k2 = mk.SeparateIndependent(kern_list)
+    assert k2.num_output_dims == data.L
+
+    dims = 5
+    k3 = mk.SharedIndependent(SquaredExponential(), dims)
+    assert k3.num_output_dims == dims
```
st--: should this be `k1/2/3.output_dim` instead?

gpflow/kernels/mo_kernels.py
awav: We could follow a standard convention, instead of making our own:

https://www.tensorflow.org/api_docs/python/tf/keras/backend/ndim
https://docs.scipy.org/doc/numpy/reference/generated/numpy.ndarray.ndim.html

gpflow/kernels/mo_kernels.py
awav: https://github.com/GPflow/GPflow/pull/1167#discussion_r353256138

awav: Where `kwargs` come from?

gpflow/kernels/mo_kernels.py
awav: https://github.com/GPflow/GPflow/pull/1167/files/cd98ac251922e80a1c374e1213434445347aba53#r353256360

gpflow/kernels/mo_kernels.py
awav: we don't really use `name`... should we remove it?

gpflow/kernels/mo_kernels.py
st--: from `__init__`

gpflow/kernels/mo_kernels.py
st--: I don't think that would be sufficiently unambiguous, given that there's input_dim as well (even if that one can now mostly be implicit)

gpflow/kernels/mo_kernels.py
st--: Given that `tf.Module` defines `name`, I think we should keep it to remain compatible (but we can replace it by pass-through of `**kwargs` here)

gpflow/kernels/mo_kernels.py
awav: Why init has `kwargs`? Why it has been changed in a combination kernel? It looks like a big change in a small PR hidden away.

condnsdmatters:
# Q: Why are we updating lots of __init__s to accept kwargs?
In our kernels, we use multiple inheritance, inheriting from both `MultioutputKernel` and `Combination`. However, we do not initialise the `MultioutputKernel`, instead just effectively treating it as an interface. This become problematic when we want to add common stateful attributes to the object. Both classes need to be inited, but we only want to initialise the base `Kernel` class once - a classic diamond inheritance structure. A good was to do this is to simply use super, which will follow the mro of the instantiated object and initialise the correct object. HOWEVER, how to avoid every inherited class having to know exactly what its parents need - we use kwargs, where the kwargs strip off the necessary argument for that interface (e.g. `output_dimension`) and passes all the other arguments to the next class that is found by `super`. If we dont do kwargs, but still want to use super, then the multioutput kernel would need to know to pass whatever arguments the `Combination` kernel needs - even though it doesnt inherit from it. This way - that is not necessary.

Ideally, the concrete, final child class would know exactly what every parent needs and not allow kwargs, while every intermediate layer strips off the argument it needs and passes the rest off to the next class. I believe this is a widely used pattern of multiple inheritance in Python, though open to discussion as per.

gpflow/kernels/mo_kernels.py
condnsdmatters: confusingly both of those actually use `ndim`  in a differnet way to us using `output_dim`.  in both cases `ndim` is actually the *rank* of the tensor, where as we want the *dimensionality* of a rank-1 tensor (vector), which normally corresponds to one element of `.shape`. I think the current version is unambiguous.

gpflow/kernels/base.py
condnsdmatters: Documentationwise, it doesnt document the all the args. The concrete 'leaf/child' cases should know whats necessary and pass to all parents... Would that work?

gpflow/kernels/mo_kernels.py
condnsdmatters: I think if we know that the this class (not abstract) will not be inherited from we might not need the kwargs. See my big comment below

gpflow/kernels/mo_kernels.py
condnsdmatters: same as my comment though i agree its not _truly_ cooperative w `super()`

gpflow/kernels/mo_kernels.py
condnsdmatters: yes i think we should keep, though sadly our own module does nothing with `name`, i think - though it should if it doesnt!

gpflow/kernels/mo_kernels.py
condnsdmatters: all abstract classes, yes. maybe all kernels, subject to discussion.

st--:
@condnsdmatters could this PR help address (fix) #831 ?

gpflow/kernels/base.py
st--:
```suggestion
        super().__init__(**kwargs)
```

gpflow/kernels/base.py
```diff
@@ -167,8 +167,8 @@ class Combination(Kernel):
 
     _reduction = None
 
-    def __init__(self, kernels: List[Kernel], name: Optional[str] = None):
-        super().__init__(name=name)
+    def __init__(self, kernels: List[Kernel], name: Optional[str] = None, **kwargs):
```
st--:
```suggestion
    def __init__(self, kernels: List[Kernel], **kwargs):
```

gpflow/kernels/mo_kernels.py
st--: well, the superclass tf.Module uses it

st--:
@condnsdmatters are you still interested in getting this merged into develop ?

gpflow/kernels/base.py
```diff
@@ -167,8 +167,8 @@ class Combination(Kernel):
 
     _reduction = None
 
-    def __init__(self, kernels: List[Kernel], name: Optional[str] = None):
-        super().__init__(name=name)
+    def __init__(self, kernels: List[Kernel], **kwargs):
```
awav: Can you explain the purpose of the `**kwargs`?

condnsdmatters: https://github.com/GPflow/GPflow/pull/1167#issuecomment-561240349

awav: On a wider scale, I do not understand the purpose of this PR

awav: @condnsdmatters, btw, I cannot load the link for some reason

condnsdmatters: The purpose of this whole PR is to give abstract parent `MultioutputKernel` a property `output_dims`, that derived classes set at instantiation.  

The reason for kwargs is that we have Multiple Inheritance, with a "diamond structure". That is, some Kernels (eg `SeparateIndependent`) need to inherit from both `Combination` and  `MultioutputKernel`. 

Naively, you would need to explicitly instantiate both parents, with the right arguments. Before - there were no arguments that were uniquely for `MultioutputKernel`; now there are. INstead of explicitly instantiating the parents, the pythonic way to do this is to use `super()` to deal with the MRO and the allow the`__init__` of each class to strip off the arguments it needs. Naturally, all intermediated parent `__init__` methods in this diamond need a kwargs. The children/leafs do not need kwargs. Hope this helps.

st--: On that topic, an _excellent_ talk by Raymond Hettinger: [Super considered super!](https://www.youtube.com/watch?v=EiOglTERPEo) (based on this [older blog post](https://rhettinger.wordpress.com/2011/05/26/super-considered-super/))

gpflow/kernels/mo_kernels.py
```diff
@@ -75,22 +80,26 @@ class SharedIndependent(MultioutputKernel):
     Note: this class is created only for testing and comparison purposes.
     Use `gpflow.kernels` instead for more efficient code.
     """
-    def __init__(self, kernel: Kernel, output_dimensionality: int):
+    def __init__(self, kernel: Kernel, output_dim: int):
         super().__init__()
         self.kernel = kernel
-        self.P = output_dimensionality
+        self._output_dim
```
st--:
```suggestion
        self._output_dim = output_dim
```

gpflow/kernels/mo_kernels.py
```diff
@@ -100,7 +109,12 @@ class SeparateIndependent(MultioutputKernel, Combination):
     - Independent: Latents are uncorrelated a priori.
     """
     def __init__(self, kernels, name=None):
-        Combination.__init__(self, kernels, name)
+        super().__init__(kernels=kernels, name=name)
+        self._output_dim = len(kernels)
+
+    @property
+    def output_dim(self):
+        return self._output_dim
```
st--:
```suggestion
        return len(self.kernels)
```

condnsdmatters: done

gpflow/kernels/mo_kernels.py
condnsdmatters: done

gpflow/kernels/mo_kernels.py
```diff
@@ -75,22 +82,26 @@ class SharedIndependent(MultioutputKernel):
     Note: this class is created only for testing and comparison purposes.
     Use `gpflow.kernels` instead for more efficient code.
     """
-    def __init__(self, kernel: Kernel, output_dimensionality: int):
+    def __init__(self, kernel: Kernel, num_latents: int):
```
vdutor: in this class `num_latents` equals `output_dimensionality`, so why are we changing this? It deprecates the shape comments as well. So I'm not a big fan of this.

vdutor: I think the best things to do here is to to have both `num_latent` and `output_dim` and show that they are the same.

gpflow/kernels/mo_kernels.py
```diff
@@ -75,22 +82,28 @@ class SharedIndependent(MultioutputKernel):
     Note: this class is created only for testing and comparison purposes.
     Use `gpflow.kernels` instead for more efficient code.
     """
-    def __init__(self, kernel: Kernel, output_dimensionality: int):
+    def __init__(self, kernel: Kernel, output_dim: int):
         super().__init__()
         self.kernel = kernel
-        self.P = output_dimensionality
+        self.output_dim = output_dim
+
```
st--:
```suggestion
```

gpflow/kernels/mo_kernels.py
```diff
@@ -35,6 +35,13 @@ class MultioutputKernel(Kernel):
     `full_output_cov` is set to True the covariance matrix will be filled with zeros
     until the appropriate size is reached.
     """
+
+    @property
+    @abc.abstractmethod
+    def num_latents(self):
+        """The number of latent gps in the multioutput kernel"""
```
st--:
```suggestion
        """The number of latent GPs in the multioutput kernel"""
```

gpflow/kernels/mo_kernels.py
```diff
@@ -75,22 +82,28 @@ class SharedIndependent(MultioutputKernel):
...
+    def __init__(self, kernel: Kernel, output_dim: int):
         super().__init__()
         self.kernel = kernel
-        self.P = output_dimensionality
+        self.output_dim = output_dim
+
+
+    @property
+    def num_latents(self):
+        # In this case num of latent gps (L) == output_dim (P)
```
st--:
```suggestion
        # In this case the number of latent GPs (L) == output_dim (P)
```

gpflow/kernels/mo_kernels.py
```diff
@@ -75,22 +82,28 @@ class SharedIndependent(MultioutputKernel):
...
+        return self.output_dim
 
     def K(self, X, X2=None, full_output_cov=True):
         K = self.kernel.K(X, X2)  # [N, N2]
         if full_output_cov:
-            Ks = tf.tile(K[..., None], [1, 1, self.P])  # [N, N2, P]
+            Ks = tf.tile(K[..., None], [1, 1, self.self.output_dim])  # [N, N2, P]
```
st--:
```suggestion
            Ks = tf.tile(K[..., None], [1, 1, self.output_dim])  # [N, N2, P]
```

gpflow/kernels/mo_kernels.py
```diff
@@ -100,7 +113,11 @@ class SeparateIndependent(MultioutputKernel, Combination):
     - Independent: Latents are uncorrelated a priori.
     """
     def __init__(self, kernels, name=None):
```
st--: Should this support `active_dims`? :thinking:

gpflow/kernels/mo_kernels.py
```diff
@@ -137,9 +154,13 @@ class LinearCoregionalization(IndependentLatent, Combination):
     Linear mixing of the latent GPs to form the output.
     """
     def __init__(self, kernels, W, name=None):
```
st--: What is this kernel's `output_dim`?...

condnsdmatters: we care about num latents not output dims

gpflow/kernels/mo_kernels.py
condnsdmatters: if you think so, you should open a PR!

gpflow/kernels/mo_kernels.py
condnsdmatters: this PR commit will reflect this

Issue #1086, opened 2019-10-07
GPflow2: kernel interface


st--:
Working on converting the `kernel_design` notebook to gpflow2, I've run into a couple of questions:
- ~inconsistency between `X, Y` and `X, X2`~ fixed in #1170
- ~`presliced` argument: is this still needed? some kernels ignore it, others don't - unlike gpflow1, in gpflow2 you MUST define it though, otherwise e.g. GPR does not work~ fixed in #1235 
- how to define a kernel that *only* works with certain input dimensions - what is the best way of handling that? -> separate issue #1240 
- ~base `Kernel` class uses `@abc.abstractmethod`, but it does not have ABC as a metaclass so you can instantiate kernels even if e.g. `K_diag` isn't defined~ fixed in #1170
- ~even though kernel function and samples look just as in gpflow1, in gpflow2 optimising the likelihood variance in the Brownian kernel example seems to go wrong~ fixed by #1099
 
~(#1077 is related, too.)~ fixed by #1088

st--:
@awav @condnsdmatters @vdutor any thoughts on the remaining two issues:

- `presliced` argument: is this still needed? some kernels ignore it, others don't - unlike gpflow1, in gpflow2 you MUST define it though, otherwise e.g. GPR does not work

- how to define a kernel that *only* works with certain input dimensions - what is the best way of handling that?

st--:
`presliced` is addressed in #1235.

Thoughts on constraining kernels to certain input dimensions (e.g. only 1D, only 2D)?

st--:
Before closing this issue we'll need to update the kernel_design notebook (e.g. I hadn't spotted in #1235 that it defines a kernel that's still using `presliced`...)

st--:
kernel_design notebook updated in #1242

PR #1170, opened 2019-12-03
Clean up kernel interface

st--:
Addresses several points in #1086

### DIFF
--- a/gpflow/kernels/base.py
+++ b/gpflow/kernels/base.py
@@ -26,7 +26,7 @@ import numpy as np
 import tensorflow as tf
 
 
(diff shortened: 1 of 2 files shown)


gpflow/kernels/base.py
```diff
@@ -26,7 +26,7 @@
 import tensorflow as tf
 
 
-class Kernel(tf.Module):
+class Kernel(tf.Module, metaclass=abc.ABCMeta):
```
st--: would it be better to do
```suggestion
class Kernel(tf.Module, abc.ABC):
```
instead ?

Issue #1176, opened 2019-12-04
Transforms influencing priors


frgsimpson:
It seems that adding a Uniform prior to a Parameter with a transform leads to a non-uniform prior.  This arises because the prior probability density is evaluated on the unconstrained parameter, via the log_det_jacobian, where the prior is no longer uniform.

Example code provided below:

```python
from gpflow import Parameter
from gpflow.kernels import Matern12

import numpy as np
from gpflow.utilities import positive
from tensorflow_probability.python.distributions import Uniform

LENGTHSCALES = [0.1, 1.0]
wide_uniform_prior = Uniform(low=np.float64(1e-10), high=np.float64(1e10))

kernel = Matern12()
kernel.lengthscale = Parameter(1.0, transform=positive(), prior=wide_uniform_prior)

for length in LENGTHSCALES:
    kernel.lengthscale.assign(length)
    print(kernel.lengthscale.log_prior().numpy(), 'at l=', length)
```

-----------------------------------------------------------------
Output:
 (Tested on 'develop':  ef5a7ca1aad46878735a6e30df3c43480cb7dd6b)
-----------------------------------------------------------------
-25.37801937681603 at l= 0.1
-23.48452607532754 at l= 1.0

frgsimpson:
Related to https://github.com/GPflow/GPflow/pull/1177 which makes explicit how the prior is defined and evaluated.

st--:
@frgsimpson has this issue been resolved by your PR #1177 ?

PR #1177, opened 2019-12-06
Defining priors with transforms

frgsimpson:
To address issue https://github.com/GPflow/GPflow/issues/1176 

Previously, adding a Uniform prior to a Parameter with a transform led to a non-uniform prior. This was due to the prior being imposed on the constrained version of the parameter, while the prior density was evaluated on the unconstrained version of the parameter.  In order to ensure Uniform priors do not become distorted in this fashion, the default behaviour has now been changed to evaluate the prior density on the constrained version of the parameter.

To give the user greater flexibility on how priors are defined, Parameters now have an 'prior_on' property, which specifies whether a Parameter's prior is imposed upon the constrained or unconstrained variable. In either case, the prior density is still evaluated in the constrained space. 

Other changes include:

- An update to the mcmc helper, since it requires the density in the unconstrained space.

- test_prior.py: It used to be stated that the MAP objective should differ by log|Jacobian| of the transform. However since the default behaviour is to evaluate MAP in the constrained space, the objective should be independent of the choice of transform.


gpflow/base.py
```diff
@@ -39,15 +39,17 @@ def __init__(self,
...
                  dtype: Optional[DType] = None,
                  name: Optional[str] = None):
         """
-        Unconstrained parameter representation.
-        According to standard terminology `y` is always transformed representation or,
-        in other words, it is constrained version of the parameter. Normally, it is hard
-        to operate with unconstrained parameters. For e.g. `variance` cannot be negative,
-        therefore we need positive constraint and it is natural to use constrained values.
+        Representation of a constrained parameter.
+        According to standard terminology `x` denotes the transformed representation or,
```
st--: I don't think it's sufficiently standard to not have to link to anything - e.g. we might say that we are using tfp.bijectors and therefore follow their terminology - according to https://www.tensorflow.org/probability/api_docs/python/tfp/bijectors/Bijector `x` is the input, and `y` is the transformed representation (`y = bijector.forward(x)`).

gpflow/base.py
```diff
@@ -39,15 +39,17 @@ def __init__(self,
...
+        in other words, it is the constrained version of the parameter. It is often challenging
+        to operate with unconstrained parameters. For example a variance cannot be negative,
+        therefore we need a positive constraint and it is natural to use constrained values.
+        A prior can be imposed either on the constrained or unconstrained
```
st--: sounds like it's finishing mid-sentence - is it missing something here?

gpflow/base.py
```diff
@@ -39,15 +39,17 @@ def __init__(self,
                  *,
                  transform: Optional[Transform] = None,
                  prior: Optional[Prior] = None,
+                 prior_on_x: bool = True,
```
st--: Can we make this more obvious? `unconstrained_space_prior=False` maybe? I would expect the default to be that you put priors on the space on which you can easily reason about, i.e. constrained space with "physical" units.

gpflow/base.py
```diff
@@ -59,17 +61,29 @@ def __init__(self,
             self._unconstrained = tf.Variable(value, dtype=dtype, name=name, trainable=trainable)
 
         self.prior = prior
+        self.prior_on_x = prior_on_x
         self._transform = transform
 
-    def log_prior(self):
+    def log_prior(self, evaluate_on_x: bool = False):
```
st--:
```suggestion
    def log_prior(self, evaluate_in_constrained_space: bool = False):
```
(or something similar)

gpflow/base.py
```diff
@@ -59,17 +61,29 @@ def __init__(self,
...
+        """ Prior probability density.
+        This can be evaluated either on the constrained or unconstrained variable. """
```
st--: Can you add a bit more explanation to the docstring what the difference is ?

When would you use evaluate_on_x=True ?

frgsimpson: Yep have updated docstring with a worked example. 

I've now set the default to True to resolve https://github.com/GPflow/GPflow/issues/1176.  It's also useful to try both in order to expose the sensitivity of MAP to the choice of parameterisation.

gpflow/base.py
frgsimpson: Agreed, I've reduced the usage of x, and reversed the definition (x, y) where it does still appear, in def log_prior()

gpflow/base.py
```diff
@@ -59,17 +60,35 @@ def __init__(self,
...
 
-    def log_prior(self):
-        x = self.read_value()
-        y = self._unconstrained
+    def log_prior(self, evaluate_on_constrained: bool = True):
+        """ Prior probability density.
+        This can be evaluated either on the constrained or unconstrained variable.
+        For example if transform = Exp(), prior = Uniform(), then log_prior will either be
+        uniform (when evaluate_on_constrained is True), or scale as 1/value when
+        evaluate_on_constrained is set to False.
```
st--: This is specific to the `prior_on_constrained=True` setting (which is now the default), is that correct? Would it be worth mentioning that to be clear?

gpflow/base.py
```diff
@@ -59,17 +60,35 @@ def __init__(self,
             self._unconstrained = tf.Variable(value, dtype=dtype, name=name, trainable=trainable)
 
         self.prior = prior
+        self.prior_on_constrained = prior_on_constrained
         self._transform = transform
 
-    def log_prior(self):
-        x = self.read_value()
-        y = self._unconstrained
+    def log_prior(self, evaluate_on_constrained: bool = True):
```
st--: This is the line that changes the behaviour of GPflow, correct? I.e. "backwards-compatible" would be `prior_on_constrained=True` but `evaluate_on_constrained=False`. If we want to change this, now (transition to gpflow-2.0) is probably the right time, just want to make sure we advertise this sufficiently loudly.

gpflow/models/model.py
```diff
@@ -44,8 +44,12 @@ def neg_log_marginal_likelihood(self, *args, **kwargs) -> tf.Tensor:
     def log_marginal_likelihood(self, *args, **kwargs) -> tf.Tensor:
         return self.log_likelihood(*args, **kwargs) + self.log_prior()
 
-    def log_prior(self) -> tf.Tensor:
-        log_priors = [p.log_prior() for p in self.trainable_parameters]
+    def log_marginal_likelihood_on_unconstrained(self, *args, **kwargs) -> tf.Tensor:
+        return self.log_likelihood(*args, **kwargs) + self.log_prior(evaluate_on_constrained=False)
```
st--: Why not have this as an argument to `log_marginal_likelihood`?

gpflow/base.py
frgsimpson: Yes exactly, if we set this to False then we recover the previous behaviour

gpflow/base.py
frgsimpson: Correct, the docstring is now updated to reflect this.

gpflow/models/model.py
```diff
@@ -44,8 +44,12 @@ def neg_log_marginal_likelihood(self, *args, **kwargs) -> tf.Tensor:
     def log_marginal_likelihood(self, *args, **kwargs) -> tf.Tensor:
```
```suggestion
    def log_marginal_likelihood(self, *args, evaluate_prior_on_constrained=True, **kwargs) -> tf.Tensor:
```

gpflow/models/model.py
frgsimpson: Done, thanks!

markvdw:
I approve the added functionality. However, some care must be taken on how this interacts with MCMC code. As I remember, the MCMC works on the unconstrained space, so that any transition kernel always leads to proposals which have defined marginal likelihoods (e.g. we want to avoid proposing negative lengthscales, which can lead to NaNs).

My view, is that this can be merged if the MCMC code is adapted to have the same behaviour as before.

frgsimpson:
> My view, is that this can be merged if the MCMC code is adapted to have the same behaviour as before.

@markvdw  Ah that's a very good point - I've updated SamplingHelper in mcmc.py to explicitly call target_log_prob_fn in the unconstrained space.

The only caveat is that this assumes 'target_log_prob_fn' will always have an argument called evaluate_on_constrained. Is there a use case where we don't use log_marginal_likelihood as target_log_prob_fn?

awav:
@frgsimpson, **I think the prior on unconstrained part should not exist at all**. I consider this PR as a fix for a bug that a prior was set on top of the unconstrained. The `Parameter` class provides an abstraction to work with **constrained** variables **only**. All interfaces are intended to work with constrained representations and introducing `prior_on_unconstrained` implies that these interfaces should change to reflect that. Guess, that's a reason why you are changing complexity to functions like `log_prior` and adding other functions as `log_marginal_likelihood_on_unconstrained`.
Long story short. Let's fix the bug, first, and then discuss whether the unconstrained logic around the prior should be added.
Thoughts?

frgsimpson:
@awav -  I agree that it would be quite elegant to have the unconstrained space hidden where possible. But as @markvdw pointed out, the mcmc code operates directly on the unconstrained variables (by making use of the unconstrained_variable property).  In order to correctly navigate the unconstrained space, we need log_prior and log_marginal_likelihood to be evaluated in that space too. So I don't think we can escape adding some complexity if we wish to retain the current functionality. 

(There are also use cases beyond mcmc, for example the standard prior for a dimensionful quantity p is a uniform prior over log p. In the current set up, I'm not sure how to achieve that other than by combining a uniform prior on the unconstrained variable with an Exp transform.)

awav:
> But as @markvdw pointed out, the mcmc code operates directly on the unconstrained variables (by making use of the unconstrained_variable property)

Yes, that's right!

>  In order to correctly navigate the unconstrained space, we need log_prior and log_marginal_likelihood to be evaluated in that space too. 

That's a problem. The parameter class implementation should not entail changes in the model class, adding implicit dependencies. My proposal was: "let's fix the bug and discuss *how* you can implement it properly". E.g. I haven't seen implementations where people cared about priors on top of the unconstrained space, a simple example is https://pyro.ai/. If you do care about it, then you want a different object, not the parameter - may be a modified version of it.

tests/test_prior.py
```diff
@@ -67,12 +105,12 @@ def log_likelihood(self):
...
 
-def test_map_contains_log_det_jacobian():
+def test_map_invariance_to_transform():
     m1 = DummyModel(with_transform=True)
     m2 = DummyModel(with_transform=False)
     assert np.allclose(m1.log_marginal_likelihood().numpy(),
-                       m2.log_marginal_likelihood().numpy() + m1.log_scale), \
-                       "MAP objective should differ by log|Jacobian| of the transform"
+                       m2.log_marginal_likelihood().numpy()), \
+                       "MAP objective should not be affected by a transform"
```
st--: Could you add some explanation (not in this part of the code, but perhaps in the PR description - so that it ends up in the git log when merging) why we completely change this around?

tests/test_prior.py
```diff
@@ -46,6 +49,41 @@ def test_log_prior_with_no_prior():
     assert param.log_prior().numpy() == 0.0
 
 
+def test_log_prior_for_uniform_prior():
+    """
+    A parameter with a uniform prior should have uniform log-prior,
```
st--: Can you clarify what you mean here? It doesn't make sense to me how a uniform prior should be a uniform log-prior...

tests/test_prior.py
```diff
@@ -46,6 +49,41 @@ def test_log_prior_with_no_prior():
...
+    param.assign(10.)
+    high_value = param.log_prior().numpy()
+
+    assert np.isclose(low_value, high_value)
+
+
+def test_log_prior_on_unconstrained():
+    """
+    A parameter with an Exp transform, and a uniform prior on its unconstrained, should have a
+    prior which scales as  1/value.
```
st--:
```suggestion
    prior in the constrained space that scales as 1/value.
```
is this what you mean?

tests/test_mcmc_helper.py
```diff
@@ -52,18 +53,31 @@ def test_mcmc_helper_target_function():
...
 
-    assert model.log_marginal_likelihood() == target_log_prob_fn()
+    def compute_jacobian_correction_term():
+        correction_term = tf.constant(0., dtype=default_float())
+        for param in model.trainable_parameters:
+            if param.transform is not None:
+                value = param.read_value()
+                log_det_jacobian = param.transform.forward_log_det_jacobian(value, value.shape.ndims)
+                correction_term += tf.reduce_sum(log_det_jacobian)
+        return correction_term
```
st--: This is just the same code again as in mcmc.py - what could a test look like that actually checks *behaviour*?

gpflow/models/model.py
```diff
@@ -45,7 +45,8 @@ def log_marginal_likelihood(self, *args, **kwargs) -> tf.Tensor:
         return self.log_likelihood(*args, **kwargs) + self.log_prior()
 
     def log_prior(self) -> tf.Tensor:
-        log_priors = [p.log_prior() for p in self.trainable_parameters]
+        log_priors = [p.log_prior()
+                      for p in self.trainable_parameters]
```
```suggestion
```

gpflow/base.py
```diff
@@ -33,25 +34,33 @@ def trainable_parameters(self):
...
                  dtype: Optional[DType] = None,
                  name: Optional[str] = None):
         """
-        Unconstrained parameter representation.
-        According to standard terminology `y` is always transformed representation or,
-        in other words, it is constrained version of the parameter. Normally, it is hard
-        to operate with unconstrained parameters. For e.g. `variance` cannot be negative,
-        therefore we need positive constraint and it is natural to use constrained values.
+        A parameter retains both constrained and unconstrained
+        representations, If no transforms is provided, these two values will be the same.
```
st--:
```suggestion
        representations. If no transform is provided, these two values will be the same.
```

gpflow/base.py
```diff
@@ -33,25 +34,33 @@ def trainable_parameters(self):
...
+        It is often challenging to operate with unconstrained parameters. For example a variance cannot be negative,
```
st--:
```suggestion
        It is often challenging to operate with unconstrained parameters. For example, a variance cannot be negative,
```

gpflow/base.py
```diff
@@ -33,25 +34,33 @@ def trainable_parameters(self):
...
+        therefore we need a positive constraint and it is natural to use constrained values.
+        A prior can be imposed either on the constrained or unconstrained version of the parameter.
```
st--:
```suggestion
        A prior can be imposed either on the constrained version (default) or on the unconstrained version of the parameter.
```

gpflow/base.py
```diff
@@ -60,18 +69,26 @@ def __init__(self,
...
-        self.prior = prior
-
     def log_prior(self):
-        x = self.read_value()
-        y = self._unconstrained
+        """ Prior probability density.
+        This can be evaluated either on the constrained or unconstrained variable.
+        For example if we set transform = Exp(), prior = Uniform(), prior_on_constrained=True
+        Then log_prior will either be uniform (when evaluate_on_constrained is True),
+        or log_prior scales as 1/value when evaluate_on_constrained is set to False.
```
st--: docstring seems to be out of date

gpflow/base.py
```diff
@@ -60,18 +69,26 @@ def __init__(self,
...
-            return out
+            y = self.read_value()
+            if self.prior_on == PriorOn.CONSTRAINED:
+                log_p = tf.reduce_sum(self.prior.log_prob(y))
+            else:
+                x = self._unconstrained
+                log_p = tf.reduce_sum(self.prior.log_prob(x))
+                if self.transform is not None:
+                    log_det_jacobian = self.transform.inverse_log_det_jacobian(y, y.shape.ndims)
+                    log_p += tf.reduce_sum(log_det_jacobian)
```
st--: I thought this branch (prior on unconstrained) was the default behaviour before - but this code has x and y swapped compared to the old version?

gpflow/optimizers/mcmc.py
```diff
@@ -84,6 +84,13 @@ def _target_log_prob_fn_closure(*variables):
             with tf.GradientTape(watch_accessed_variables=False) as tape:
                 tape.watch(variables_list)
                 log_prob = self._target_log_prob_fn()
+                # Now need to correct for the fact that the prob fn is evaluated on the
+                # constrained space while we wish to evaluate it in the unconstrained space
+                for param in self._model_parameters:
+                    if isinstance(param, Parameter) and param.transform is not None:
+                        value = param.read_value()
```
st--: should this not be the *un*constrained version here ?
Also, should this have a check on whether param.prior_on is CONSTRAINED or UNCONSTRAINED?

gpflow/base.py
```diff
@@ -33,13 +34,18 @@ def trainable_parameters(self):
         return tuple(self._flatten(predicate=_IS_TRAINABLE_PARAMETER))
 
 
+class PriorOn(Enum):
+     CONSTRAINED = 1
```
st--:
```suggestion
     CONSTRAINED = 'constrained'
```

gpflow/base.py
```diff
@@ -33,13 +34,18 @@ def trainable_parameters(self):
...
+     UNCONSTRAINED = 2
```
st--:
```suggestion
     UNCONSTRAINED = 'unconstrained'
```

gpflow/base.py
```diff
@@ -53,6 +59,8 @@ def __init__(self,
         super().__init__()
 
         self._transform = transform
+        self.prior = prior
+        self.prior_on = prior_on
```
st--:
```suggestion
        self.prior_on = PriorOn(prior_on)
```
together with the changes to the PriorOn class suggested above, this will allow a user to call
```python
Parameter(..., prior=..., prior_on="unconstrained")
```
instead of having to import the PriorOn enum class and the more cumbersome
```python
Parameter(..., prior=..., prior_on=PriorOn.UNCONSTRAINED)
```

tests/test_prior.py
frgsimpson: Yep, I've updated the description to explain this change and that of the mcmc helper.

tests/test_prior.py
frgsimpson: Sorry bad wording - has been updated now (log-prior here referred to the function call log(p) rather than a log distribution.)

awav:
@st--, We had a chat with @frgsimpson a while ago about this PR. The outcome was:

1. Introduce a flag for prior choosing where it is set, constrained or unconstrained space.
2. Update MCMC method for log_marginal_likelihood generation.
3. Make docs clear about the changes. In particular, how to use MCMC and why it modifies the log_marginal_likelihood.

st--:
@awav yes, I completely agree with all those changes! that's what I was referring to.
You make a very good point that this PR should also update the MCMC notebook accordingly though :)

gpflow/optimizers/mcmc.py
frgsimpson: Yep thanks, fixed! I don't think it needs to check param.prior_on because evaluation of the prior is always on the constrained variable, so a correction always needs to be applied.

tests/test_mcmc_helper.py
frgsimpson: Yep agreed, the tests are now updated!

gpflow/base.py
frgsimpson: Right, the x and y have swapped definitions to maintain consistency with how they are defined in tensorflow bijectors. (Is this what you meant?) It is also a bit more intuitive if we think of eg y=exp(x) for the Exp() bijector

gpflow/optimizers/mcmc.py
st--: Right- the point is that when prior is on unconstrained then the correction here will simply cancel the one that is applied in `log_prior`?

gpflow/optimizers/mcmc.py
```diff
@@ -84,6 +84,14 @@ def _target_log_prob_fn_closure(*variables):
             with tf.GradientTape(watch_accessed_variables=False) as tape:
                 tape.watch(variables_list)
                 log_prob = self._target_log_prob_fn()
+                # Now need to correct for the fact that the prob fn is evaluated on the
+                # constrained space while we wish to evaluate it in the unconstrained space
+                for param in self._model_parameters:
+                    if isinstance(param, Parameter) and param.transform is not None:
+                        x = param.unconstrained_variable
+                        log_det_jacobian = param.transform.forward_log_det_jacobian(x,
```
st--: `forward_log_det_jacobian` may not be defined if it's a non-injective bijector. Why not instead subtract the `inverse_log_det_jacobian` to cancel the (wrong, in this case) correction in `param.log_prior()`?

st--:
Will merging this PR close #1176?

frgsimpson:
> Will merging this PR close #1176?

Yes, I've checked that the script quoted in #1176 now produces the desired output (i.e. the Uniform prior is uniform):

-23.025850929940457 at l= 0.1
-23.025850929940457 at l= 1.0

gpflow/optimizers/mcmc.py
frgsimpson: Could do! I haven't thought about the case of non-injective bijectors - it might be safest to assert that the transform must be injective if a prior is specified.

st--: So what would be the (dis)advantages of having this code as follows instead:
```python
                for param in self._model_parameters:
                    if isinstance(param, Parameter) and param.transform is not None:
                        y = param.read_value()
                        log_det_jacobian = param.transform.inverse_log_det_jacobian(y,
                                                                                    y.shape.ndims)
                        log_prob -= tf.reduce_sum(log_det_jacobian)
```
exactly mirroring the behaviour in `param.log_prior()` ?

frgsimpson: Ah because in param.log_prior() the inverse_log_det_jacobian is only applied when the prior is defined on the unconstrained variable. If the prior is defined on the constrained variable, then we aren't trying to cancel a term, we're applying a correction for the first time.

frgsimpson: And that correction is (I think) potentially pathological if the transform isn't bijective.

tests/test_mcmc_helper.py
```diff
@@ -43,27 +46,110 @@ def test_mcmc_helper_parameters():
...
+
+        expected_log_prior += np.log(prior_density_on_unconstrained)
+
+    log_likelihood = model.log_likelihood().numpy()
+    expected_log_prob = log_likelihood + expected_log_prior
+
+    hmc_helper = gpflow.optimizers.SamplingHelper(
+        model.trainable_parameters, model.log_marginal_likelihood
+    )
+    print(hmc_helper.target_log_prob_fn())
```
frgsimpson:
```suggestion
```

tests/test_mcmc_helper.py
```diff
@@ -43,27 +46,109 @@ def test_mcmc_helper_parameters():
...
+        expected_log_prior += np.log(prior_density_on_unconstrained)
+
+    log_likelihood = model.log_likelihood().numpy()
+    expected_log_prob = log_likelihood + expected_log_prior
+
+    hmc_helper = gpflow.optimizers.SamplingHelper(
+        model.trainable_parameters, model.log_marginal_likelihood
+    )
+
+    assert np.isclose(expected_log_prob, target_log_prob_fn())
```
frgsimpson:
```suggestion
    np.testing.assert_allclose(expected_log_prob, target_log_prob_fn())
```

tests/test_mcmc_helper.py
```diff
@@ -43,27 +46,109 @@ def test_mcmc_helper_parameters():
 
 
 def test_mcmc_helper_target_function():
+    """ Set up priors on the model parameters such that we can
+    readily compute their expected values. """
     data = build_data()
     model = build_model(data)
 
+    prior_width = 200
```
st--:
```suggestion
    prior_width = 200.0
```
(It's used as a float)

tests/test_mcmc_helper.py
```diff
@@ -43,27 +46,109 @@ def test_mcmc_helper_parameters():
...
+        prior_density_on_unconstrained = prior_density_on_constrained * param.value()
+
+        expected_log_prior += np.log(prior_density_on_unconstrained)
+
+    log_likelihood = model.log_likelihood().numpy()
+    expected_log_prob = log_likelihood + expected_log_prior
+
+    hmc_helper = gpflow.optimizers.SamplingHelper(
+        model.trainable_parameters, model.log_marginal_likelihood
+    )
```
st--: Is this code doing anything? can it just be removed?

tests/test_mcmc_helper.py
```diff
@@ -43,27 +46,109 @@ def test_mcmc_helper_parameters():
...
     hmc_helper = gpflow.optimizers.SamplingHelper(
         model.trainable_parameters, model.log_marginal_likelihood
     )
+    target_log_prob_fn = hmc_helper.target_log_prob_fn
+
+    # Priors which are set on the constrained space
+    expected_log_prior = 0.
+    for param in model.trainable_parameters:
+        if param.value() < 1e-3:
+            param.assign(1.0)
```
st--: could you add a comment why this is necessary / what the intent is ?

tests/test_mcmc_helper.py
```diff
@@ -43,27 +46,109 @@ def test_mcmc_helper_parameters():
...
+
+        param.transform = Exp()
+
+        low_value = -100
+        high_value = low_value + prior_width
```
st--: How does the prior being on y ∈ [-100, 100] interact with the transform ensuring that y > 0 always?

tests/test_mcmc_helper.py
```diff
@@ -43,27 +46,109 @@ def test_mcmc_helper_parameters():
...
+
+def test_mcmc_helper_target_function_unconstrained():
+    """ Verifies the objective for a set of priors which are defined on the unconstrained space.
+    """
+    data = build_data()
+    model = build_model(data)
+
+    # Set up priors on the model parameters such that we can readily compute their expected values.
+    expected_log_prior = 0.
+    prior_width = 200
```
st--:
```suggestion
    prior_width = 200.
```

tests/test_mcmc_helper.py
```diff
@@ -43,27 +46,109 @@ def test_mcmc_helper_parameters():
...
+
+    hmc_helper = gpflow.optimizers.SamplingHelper(
+        model.trainable_parameters, model.log_marginal_likelihood
+    )
+
+    for param in model.trainable_parameters:
+        low_value = -100
+        high_value = low_value + prior_width
+        param.prior = Uniform(low=np.float64(low_value), high=np.float64(high_value))
+        param.prior_on = 'unconstrained'
```
st--:
```suggestion
        param.prior_on = gpflow.base.PriorOn.UNCONSTRAINED
```

tests/test_mcmc_helper.py
```diff
@@ -43,27 +46,109 @@ def test_mcmc_helper_parameters():
...
+
+        prior_density = 1 / prior_width
+        expected_log_prior += np.log(prior_density)
 
     target_log_prob_fn = hmc_helper.target_log_prob_fn
+    expected_log_prob = model.log_likelihood().numpy() + expected_log_prior
+
+    assert np.isclose(expected_log_prob, target_log_prob_fn())
```
st--:
```suggestion
    np.testing.assert_allclose(target_log_prob_fn(), expected_log_prob)
```
(signature is `assert_allclose(actual, desired)`)

tests/test_mcmc_helper.py
```diff
@@ -43,27 +46,109 @@ def test_mcmc_helper_parameters():
...
+        expected_log_prior += np.log(prior_density_on_unconstrained)
+
+    log_likelihood = model.log_likelihood().numpy()
+    expected_log_prob = log_likelihood + expected_log_prior
+
+    hmc_helper = gpflow.optimizers.SamplingHelper(
+        model.trainable_parameters, model.log_marginal_likelihood
+    )
+
+    np.testing.assert_allclose(expected_log_prob, target_log_prob_fn())
```
st--:
```suggestion
    np.testing.assert_allclose(target_log_prob_fn(), expected_log_prob)
```

tests/test_mcmc_helper.py
```diff
@@ -43,27 +46,109 @@ def test_mcmc_helper_parameters():
 
 
 def test_mcmc_helper_target_function():
```
st--:
```suggestion
def test_mcmc_helper_target_function_constrained():
```

tests/test_mcmc_helper.py
```diff
@@ -43,27 +46,109 @@ def test_mcmc_helper_parameters():
...
-    assert model.log_marginal_likelihood() == target_log_prob_fn()
+        prior_density = 1 / prior_width
+        expected_log_prior += np.log(prior_density)
 
-    # test the wrapped closure
+    log_likelihood = model.log_likelihood().numpy()
+    expected_log_prob = log_likelihood + expected_log_prior
+    target_log_prob_fn = hmc_helper.target_log_prob_fn
+
+    assert np.isclose(expected_log_prob, target_log_prob_fn())
```
st--:
```suggestion
    np.testing.assert_allclose(target_log_prob_fn(), expected_log_prob)
```

tests/test_mcmc_helper.py
```diff
@@ -43,27 +46,109 @@ def test_mcmc_helper_parameters():
...
+    for param in model.trainable_parameters:
+        if param.value() < 1e-3:
+            param.assign(1.0)
+
+        param.transform = Exp()
+
+        low_value = -100
+        high_value = low_value + prior_width
+
+        param.prior = Uniform(low=np.float64(low_value), high=np.float64(high_value))
```
st--:
```suggestion
        param.prior = Uniform(low=np.float64(low_value), high=np.float64(high_value))
        param.prior_on = gpflow.base.PriorOn.CONSTRAINED
```
just in case the default ever changes, we want both cases to be tested explicitly...

tests/test_mcmc_helper.py
```diff
@@ -43,27 +46,109 @@ def test_mcmc_helper_parameters():
...
+        prior_density = 1 / prior_width
+        expected_log_prior += np.log(prior_density)
 
     target_log_prob_fn = hmc_helper.target_log_prob_fn
+    expected_log_prob = model.log_likelihood().numpy() + expected_log_prior
+
+    assert np.isclose(expected_log_prob, target_log_prob_fn())
+
+
+def test_mcmc_helper_target_function_no_transforms():
```
st--:
```suggestion
@pytest.mark.parametrize('prior_on', [gpflow.base.PriorOn.CONSTRAINED, gpflow.base.PriorOn.UNCONSTRAINED])
def test_mcmc_helper_target_function_no_transforms(prior_on):
```
should be the same regardless of where the prior is, right?

tests/test_mcmc_helper.py
```diff
@@ -43,27 +46,109 @@ def test_mcmc_helper_parameters():
...
+    hmc_helper = gpflow.optimizers.SamplingHelper(
+        model.trainable_parameters, model.log_marginal_likelihood
+    )
 
-    model.likelihood.variance.assign(1)
+    for param in model.trainable_parameters:
+        param.transform = None
+        low_value = -100
+        high_value = low_value + prior_width
+        param.prior = Uniform(low=np.float64(low_value), high=np.float64(high_value))
```
st--:
```suggestion
        param.prior = Uniform(low=np.float64(low_value), high=np.float64(high_value))
        param.prior_on = prior_on
```

tests/test_mcmc_helper.py
```diff
@@ -43,27 +46,109 @@ def test_mcmc_helper_parameters():
...
+    assert np.isclose(expected_log_prob, target_log_prob_fn())
+
+
+def test_mcmc_helper_target_function_no_transforms():
+    """ Verifies the objective for a set of priors where no transforms are set.
+    """
+    data = build_data()
+    model = build_model(data)
+    expected_log_prior = 0.
+    prior_width = 200
```
st--:
```suggestion
    prior_width = 200.
```

st--:
Hi @frgsimpson, thanks for your patience with all these review comments - I hope it makes sense that given the sensitive nature of these changes to the core of gpflow we're being a bit more cautious!

One final request: could you please add a few lines on PriorOn and the change in behaviour to the MCMC notebook? (Note that I just merged #1239 so you will need to install jupytext in your dev environment if not installed already, and to make sure you sync the .pct.py version.)

tests/test_mcmc_helper.py
awav: This is correct usage and testing that is right.

tests/test_mcmc_helper.py
awav:
```suggestion
@pytest.mark.parametrize("prior_on", ["constrained", "unconstrained"])
def test_mcmc_helper_target_function_no_transforms():
```

gpflow/base.py
```diff
@@ -42,25 +43,33 @@ def _repr_pretty_(self, p, cycle):
         p.text(tabulate_module_summary(self, tablefmt=''))
 
 
+class PriorOn(Enum):
+     CONSTRAINED = 'constrained'
```
awav:
```suggestion
     CONSTRAINED = "constrained"
```

gpflow/base.py
```diff
@@ -42,25 +43,33 @@ def _repr_pretty_(self, p, cycle):
...
+     UNCONSTRAINED = 'unconstrained'
```
awav:
```suggestion
     UNCONSTRAINED = "unconstrained"
```

gpflow/base.py
```diff
@@ -42,25 +43,33 @@ def _repr_pretty_(self, p, cycle):
...
+
+
 class Parameter(tf.Module):
     def __init__(self,
                  value,
                  *,
                  transform: Optional[Transform] = None,
                  prior: Optional[Prior] = None,
+                 prior_on: PriorOn = PriorOn.CONSTRAINED,
```
awav:
```suggestion
                 prior_on: Union[str, PriorOn] = PriorOn.CONSTRAINED,
```

tests/test_mcmc_helper.py
```diff
@@ -43,27 +46,112 @@ def test_mcmc_helper_parameters():
...
+        prior_density_on_constrained = 1 / prior_width
+        prior_density_on_unconstrained = prior_density_on_constrained * param.value()
+
+        expected_log_prior += np.log(prior_density_on_unconstrained)
+
+    log_likelihood = model.log_likelihood().numpy()
+    expected_log_prob = log_likelihood + expected_log_prior
+
+    hmc_helper = gpflow.optimizers.SamplingHelper(
+        model.trainable_parameters, model.log_marginal_likelihood
```
awav:
```suggestion
```

tests/test_mcmc_helper.py
```diff
@@ -43,27 +46,112 @@ def test_mcmc_helper_parameters():
...
+    )
```
awav:
```suggestion
```

Issue #1179, opened 2019-12-06
static shapes vs dynamic shapes


marcoadurno:
Hi,

The codebase heavily relieson static shapes (eg t.shape) as opposed to dynamic shapes (eg tf.shape(t)). 

## Bug case:
Using static shapes makes it impossible to use unknown batch size:
```
import tensorflow as tf
import gpflow

@tf.function
def K_diag_dynamic(X):
    variance = 1.0
    return tf.fill(tf.shape(X)[:-1],tf.squeeze(variance))

@tf.function
def K_diag_static(X):
    variance = 1.0
    return tf.fill(X.shape[:-1],tf.squeeze(variance))

if __name__=='__main__':
    K_diag_dynamic.get_concrete_function(X=tf.TensorSpec([None,2], tf.float64))
    K_diag_static.get_concrete_function(X=tf.TensorSpec([None,2], tf.float64))
```
Getting the concrete function for ```K_diag``` succeeds when using dynamic shapes and throws when using static shapes. At the moment we often rely on static shapes, like [here](https://github.com/GPflow/GPflow/blob/develop/gpflow/kernels/stationaries.py#L61). Wouldn't be better to use dynamic shapes?

## System information:

Python version: 3.7
TensorFlow installed from (source or binary): binary
TensorFlow version (use command below): 'v2.0.0-rc2-26-g64c3d382ca', '2.0.0'
GPflow installed from (source or binary): source
GPflow version: develop, 24b5733

Best Regards,

Marco

st--:
I completely agree, and I had even changed some of them already because I needed it to get something else to work... let me dig up where that was and start a PR -- would be great if you wanted to help out with that!

awav:
@marcoadurno, I do agree about dynamic shapes. But, it is not obvious for me what is the relation between the dynamic shape and a saver here. Could you remove mentioning about saver? It echoes the #1127 issue.

marcoadurno:
@awav: done. I only really use dynamic shapes when saving models but I do agree that there might be other scenarios.

marcoadurno:
@st-- Happy to help but I'm traveling to neurips tomorrow, I'll try and do what I can.

Thanks,

Marco

marcoadurno:
@st-- I might have something to share. Ho do you want me to proceed?

st--:
[...]

Issue #1091, opened 2019-10-09
MultiClass likelihood tests broken


st--:
#1083 marks the tests of the `MultiClass` likelihood as skipped, as there were some issues. We should investigate, fix the tests (or likelihood, as required), and un-skip them!

st--:
@awav why did you close this issue?

st--:
Remaining failures are due to #966

PR #1181, opened 2019-12-06
use dynamic shape evaluation instead of static shapes

st--:
Addresses #1179


st--:
This PR still needs tests - there is already some use of tf.function() around the code, but specifically we need to test *with* passing `input_signature` to the tf.function() using `tf.TensorSpec(shape=None)`. Any help appreciated!

st--:
The patch coverage is low due to the change in gpflow.likelihoods.MultiClass.log_prob(), which isn't tested, as the quadrature vs analytic tests in tests/test_likelihoods.py are currently skipped (see #1091).

PR #1186, opened 2019-12-09
Update SGPR_notes.ipynb

tadejkrivec:
I was going trough the derivation and I think that there is no inverse in the predictive distribution.


markvdw:
You are correct.

Issue #1185, opened 2019-12-09
GPflow-2.0: Notebooks are missing tf.function use


st--:
We should demonstrate in all notebooks how to (correctly) use tf.function. This would unearth bugs e.g. due to use of static instead of dynamic shapes (#1179) that aren't currently covered by any tests.

PR #1188, opened 2019-12-11
From static shapes to dynamic shapes

marcoadurno:
## PR content:

* [Title](#title)
* [Description](#description)

### Description

We prepare the code base to support usage of dynamic shapes and tf.function with unknown sizes.

* Bad example: `This is a bug-fix found at #1179.`
* Good example: `We allow usage of tf.function with unknown sizes.`

TODO: dedicated tests

### DIFF
--- a/gpflow/conditionals/uncertain_conditionals.py
+++ b/gpflow/conditionals/uncertain_conditionals.py
@@ -49,9 +49,7 @@ def uncertain_conditional(Xnew_mu: tf.Tensor,
     pXnew = Gaussian(Xnew_mu, Xnew_var)
 
     num_data = tf.shape(Xnew_mu)[0]  # number of new inputs (N)
-    num_ind = q_mu.shape[0]  # number of inducing points (M)
-    num_func = q_mu.shape[1]  # output dimension (D)
-
+    num_ind, num_func = tf.unstack(tf.shape(q_mu), num=2, axis=0) . # number of inducing points (M), output dimension (D)
     q_sqrt_r = tf.linalg.band_part(q_sqrt, -1, 0)  # [D, M, M]
 
     eKuf = tf.transpose(expectation(pXnew, (kernel, inducing_variable)))  # [M, N] (psi1)
--- a/gpflow/conditionals/uncertain_conditionals.py
+++ b/gpflow/conditionals/uncertain_conditionals.py
@@ -49,7 +49,7 @@ def uncertain_conditional(Xnew_mu: tf.Tensor,
     pXnew = Gaussian(Xnew_mu, Xnew_var)
 
     num_data = tf.shape(Xnew_mu)[0]  # number of new inputs (N)
-    num_ind, num_func = tf.unstack(tf.shape(q_mu), num=2, axis=0) . # number of inducing points (M), output dimension (D)
+    num_ind, num_func = tf.unstack(tf.shape(q_mu), num=2, axis=0)  # number of inducing points (M), output dimension (D)
     q_sqrt_r = tf.linalg.band_part(q_sqrt, -1, 0)  # [D, M, M]
 
     eKuf = tf.transpose(expectation(pXnew, (kernel, inducing_variable)))  # [M, N] (psi1)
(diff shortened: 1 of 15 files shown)


gpflow/conditionals/uncertain_conditionals.py
```diff
@@ -48,7 +48,7 @@ def uncertain_conditional(Xnew_mu: tf.Tensor,
 
     pXnew = Gaussian(Xnew_mu, Xnew_var)
 
-    num_data = Xnew_mu.shape[0]  # number of new inputs (N)
+    num_data = tf.shape(Xnew_mu)[0]  # number of new inputs (N)
     num_ind = q_mu.shape[0]  # number of inducing points (M)
     num_func = q_mu.shape[1]  # output dimension (D)
```
st--: should these by dynamic shapes as well ?

marcoadurno: Within the context of a defined model no, I agree that M&D are known, but if you were interested in reusing the function with generic tensors then maybe yes. Happy to revert.

st--: let me rephrase: is there anything we would *loose* by having these be `num_ind, num_func = tf.unstack(tf.shape(q_mu), num=2, axis=0)` as well?

gpflow/conditionals/util.py
```diff
@@ -51,11 +51,11 @@ def base_conditional(Kmn: tf.Tensor,
         0)  # [N]
     Kmn = tf.transpose(Kmn, perm)  # [..., M, N]
 
-    leading_dims = Kmn.shape[:-2]
+    leading_dims = tf.shape(Kmn)[:-2]
```
st--: Does the `[:-2]` still work in graph mode (wrapped in `@tf.function`)? :thinking:

gpflow/conditionals/util.py
```diff
@@ -86,7 +86,7 @@ def base_conditional(Kmn: tf.Tensor,
             L_shape = tf.shape(L)
             L = tf.broadcast_to(L, tf.concat([leading_dims, L_shape], 0))
 
-            shape = [*leading_dims, num_func, M, N]
+            shape = tf.concat([leading_dims, [num_func, M, N]], 0)
```
st--:
```suggestion
            shape = tf.concat([leading_dims, [num_func, M, N]], axis=0)
```

gpflow/conditionals/util.py
st--: No I'd be happy with it I just vaguely remember some issues with slicing tf.shape() results in tf1 not working properly... so wanted to make sure it does work. (-> we need tests that wrap code in `tf.function()` and check it works the same as without...)

I can't access your mwe link - can you put it as a github gist or post in here?

marcoadurno: [here](https://gist.github.com/marcoadurno/97fc32c4387d2c77b97cbba9cab04b55)?

marcoadurno: Regarding the tests, similar to Example Usage [here](https://www.tensorflow.org/api_docs/python/tf/function#used_in_the_tutorials)?

Shall the tests be tied up with issue #1185 ?

st--: I think we should add tf.function with input_signature set to `tf.TensorSpec(shape=None)` in the tests.

This is distinct from #1185 where the intent is just to add `tf.function()` in the way one would usually use it (without passing `input_signature`, so not testing whether it would work with minibatching most of the time).

st--:
I'm going to merge this into my other branch and then we can add tests in there (feel free to open another PR into the same branch if you want to help - would be much appreciated :))

st--:
@marcoadurno I've now added an initial test to the main PR #1181 - if you've got any ideas for how further to test that dynamic shapes are working now, or if you can think of some cases that this test doesn't yet cover (even if you can't think of how you WOULD cover it), please leave a message there!

I'd prefer to add more comprehensive tests first and only then merge #1181, but equally I don't want to put it off too much further...

PR #1190, opened 2019-12-16
Add jit flag to gpflow.optimizers.Scipy

st--:
As the Scipy optimizer combines numpy with tensorflow, we can't simply wrap its `minimize()` method in `tf.function()`. Instead, this PR adds a `jit` flag (True by default) which wraps the evaluation of objective and its gradients in `tf.function()`.

### DIFF
--- a/gpflow/optimizers/scipy.py
+++ b/gpflow/optimizers/scipy.py
@@ -38,6 +38,9 @@ class Scipy:
                 above, and `values` is the corresponding list of tensors of
                 matching shape that contains their value at this optimisation
                 step.
+            jit: If True, wraps the evaluation function (the passed `closure` as
+                well as its gradient computation) inside a `tf.function()`,
+                which will improve optimization speed in most cases.
 
             scipy_kwargs: Arguments passed through to `scipy.optimize.minimize`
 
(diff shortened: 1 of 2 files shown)


awav:
@st-- , the thing is that we had `jit` in scipy and removed it.

condnsdmatters:
> @st-- , the thing is that we had `jit` in scipy and removed it.

what was the motivation for getting rid of it? it looks faster for SVGP though not GPR

awav:
@st-- , the reason for not having a jit flag is that `tf.function` doesn't work with many gpflow models. At the same time, there are cases when you need a special case gradient, or use might want to specialize either `tf.function` or `tf.function` without autograph.

awav:
> > @st-- , the thing is that we had `jit` in scipy and removed it.
> 
> what was the motivation for getting rid of it? it looks faster for SVGP though not GPR

@condnsdmatters, check previous message

condnsdmatters:
@awav which models do we know for a fact this _doesn't_ work with? sounds like we should fix if poss.

awav:
@condnsdmatters every single one with multiple dispatch, it is not an easy fix. We can discuss it in person.

st--:
All the tests pass, including the notebook tests, and I've not actually encountered any issues with the conditional(), so I don't think this is an immediate issue.

If you know of cases that would not work for this, please add it as a test case, so that we can work on fixing it!

Issue #1191, opened 2019-12-28
`assign()` for new parameters


markvdw:
I just saw @st--'s stackoverflow post, where it was written "In the MWE you assign a new value to a Parameter that is already existing (and does not have the logistic transform)". I was doing exactly this the other day, since in [models.ipynb](https://github.com/GPflow/GPflow/blob/b0e09ee6d669b2761cb14778507a903c38a9e0ff/doc/source/notebooks/understanding/models.ipynb) cell 10 (below), it recommends to assign a `Parameter` object using `.assign()`.

Is this an error in the documentation? It did not work for me, so I changed it to ` = `, but I didn't follow up on it any further.

```
old_parameter = m.kernel.kernels[0].lengthscale
new_parameter = gpflow.Parameter(old_parameter.value(),
                                 trainable=old_parameter.trainable,
                                 prior=old_parameter.prior,
                                 name=old_parameter.name.split(':')[0],  # tensorflow is weird and adds ':0' to the name
                                 transform=tfp.bijectors.Exp())
m.kernel.kernels[0].lengthscale.assign(new_parameter)
```

awav:
Hello, @markvdw, this is not a bug. The `assign` were never supposed to work with Parameter, it is a part of the Tensor interface. You are doing everything right with replacing a parameter with an equal sign. Just to clarify: `=` sign is a setter for the module's attributes and `assign` is a part of `tf.Variable` interface to be able to change value of the variable/parameter.

st--:
It's a bug in the documentation, which is misleading.

awav:
The issue addressed at #1193

PR #1193, opened 2020-01-02
Add Parameter.assign docstring

awav:
Addressing #1191


st--:
Thanks for the docstring improvement! To actually close #1191 I think we still need to change cell [10] in [doc/source/notebooks/understanding/models.ipynb](https://github.com/GPflow/GPflow/blob/develop/doc/source/notebooks/understanding/models.ipynb) from `m.kernel.kernels[0].lengthscale.assign(new_parameter)` to `m.kernel.kernels[0].lengthscale = new_parameter`. Do you want to do that in this PR ?

awav:
@st-- , yep, will do that!

gpflow/base.py
```diff
@@ -112,8 +112,27 @@ def trainable(self, flag: Union[bool, int]):
...
+            ```
+
+        :param value: Constrained tensor-like value.
+        :param use_locking: If `True`, use locking during the assignment.
+        :param name: The name of the operation to be created.
+        :param read_value: if True, will return something which evaluates to the new
+            value of the variable; if False will return the assign op.
+        """
         value = _verified_value(value, self.dtype)
         unconstrained_value = _to_unconstrained(value, self.transform)
```
st--: Is this missing a `return` on the call to `self._unconstrained.assign()` below?

doc/source/notebooks/understanding/models.ipynb
```diff
@@ -24,7 +24,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": 27,
```
st--: Would be great if you could "Restart&Run all" the notebook to make sure it all runs correctly & so the cells are numbered in a nice sequence!

awav: done

gpflow/base.py
awav: Yes, it is. It was intentional, but for protocol compatibility, I added it back.

awav:
@st--, Thanks! I addressed your comments! Ready to merge it, waiting for tests to pass.

Issue #1197, opened 2020-01-07
GPflow1 to GPflow2: more Cholesky decomposition problems


tdewolff:
GPflow: 2.0.0rc1
Tensorflow: 2.0.0

We're in the process of moving our code from GPflow1 to GPflow2, but while doing so it seems that Cholesky decomposition problems occur much more often than previously. One reason was that the likelihood variance in the GPR model became VERY small all of a sudden. This variance is added to the diagonala, but if it is smaller than 1e-12 for example Cholesky fails. In general, parameter estimation through the SciPy/Adam optimizers gives poorer results. Sometimes parameters become NaN while optimizing.

Has anything fundamentally changed between GPflow1 and GPflow2 that we need to be aware of? For example, the inner working of `gpflow.transforms.positive` is similar to the new `gpflow.utilities.positive()`? The latter seems to more easily result in NaN (when input is negative). Anything we need to be aware of?

PS: overall GPflow2 is a big improvement! Thanks for the work!

awav:
@tdewolff, thanks a lot! 

In GPflow 1.0, positive values had a lower bound 1e-6. It was removed in favour of a more customisable solution. Now you can set a lower positive bound to a value which you think is most appropriate.

```python
assert gpflow.config.default_positive_minimum() is None  # by default it is `None`
gpflow.config.set_default_positive_minimum(1e-6)
```

https://github.com/GPflow/GPflow/blob/develop/gpflow/config/__config__.py#L153

You also can change the positive bijector type (by default it is `softplus`). You have two choices "exp" and "softplus".

```python
assert gpflow.config.default_positive_bijector() == "softplus"
gpflow.config.set_default_positive_bijector("exp")
```

https://github.com/GPflow/GPflow/blob/develop/gpflow/config/__config__.py#L143

st--:
I actually just ran into a related issue where in my custom model I had a zero-initialized positive-constrained parameter, `p = gpflow.Parameter(0.0, transform=gpflow.utilities.positive())`. This is represented as `-inf` in the unconstrained space, which is fine in the forward pass (back to 0.0), but results in `nan`s when computing gradients. What can we do to improve the user experience around positive transforms ?

tdewolff:
[...]

PR #1198, opened 2020-01-08
Add validation to Parameter.assign() to increase robustness of transforms

st--:
In GPflow2, `Parameter` behaves differently from GPflow1, as noted in #1197.
Specifically, when using a positive transform, initializing or assigning the value 0.0 is perfectly fine in the forward pass: `p = gpflow.Parameter(0.0, transform=gpflow.utilities.positive())`, and it looks like nothing is wrong; the constrained representation is 0.0 as expected. (In GPflow1, it would implicitly force the value to be the lower bound of 1e-6.) However, the *unconstrained* representation is then `-inf`, which leads to `nan`s when computing gradients in the backward pass. This is error-prone and hard to debug. To make this issue more obvious, this PR adds an error check.

### DIFF
--- a/gpflow/base.py
+++ b/gpflow/base.py
@@ -56,7 +56,7 @@ class Parameter(tf.Module):
         if isinstance(value, tf.Variable):
             self._unconstrained = value
         else:
-            unconstrained_value = self.verified_unconstrained_value(value, dtype)
+            unconstrained_value = self.validate_unconstrained_value(value, dtype)
             self._unconstrained = tf.Variable(unconstrained_value,
                                               dtype=dtype, name=name, trainable=trainable)
 
@@ -113,11 +113,11 @@ class Parameter(tf.Module):
     def initial_value(self):
         return self._unconstrained.initial_value
 
-    def verified_unconstrained_value(self, value: tf.Tensor, dtype: DType) -> tf.Tensor:
+    def validate_unconstrained_value(self, value: tf.Tensor, dtype: DType) -> tf.Tensor:
         value = _cast_to_dtype(value, dtype)
         unconstrained_value = _to_unconstrained(value, self.transform)
         return tf.debugging.assert_all_finite(unconstrained_value,
-                message=f"gpflow.Parameter.verified_unconstrained_value")
+                message=f"gpflow.Parameter.validate_unconstrained_value")
 
 
     def assign(self, value: tf.Tensor, use_locking=False, name=None, read_value=True) -> tf.Variable:
@@ -141,7 +141,7 @@ class Parameter(tf.Module):
         :param read_value: if True, will return something which evaluates to the new
             value of the variable; if False will return the assign op.
         """
-        unconstrained_value = self.verified_unconstrained_value(value, self.dtype)
+        unconstrained_value = self.validate_unconstrained_value(value, self.dtype)
         return self._unconstrained.assign(unconstrained_value,
                 use_locking=use_locking, name=name, read_value=read_value)
 
--- a/gpflow/base.py
+++ b/gpflow/base.py
@@ -120,7 +120,6 @@ class Parameter(tf.Module):
                   "has NaN or Inf and cannot be assigned."
         return tf.debugging.assert_all_finite(unconstrained_value, message=message)
 
-
     def assign(self, value: tf.Tensor, use_locking=False, name=None, read_value=True) -> tf.Variable:
         """
         Assigns constrained `value` to the unconstrained parameter's variable.
--- a/gpflow/base.py
+++ b/gpflow/base.py
@@ -51,12 +51,12 @@ class Parameter(tf.Module):
         """
         super().__init__()
 
-        value = _verified_value(value, dtype)
         if isinstance(value, tf.Variable):
             self._unconstrained = value
         else:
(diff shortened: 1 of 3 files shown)


gpflow/base.py
```diff
@@ -112,6 +113,13 @@ def trainable(self, flag: Union[bool, int]):
     def initial_value(self):
         return self._unconstrained.initial_value
 
+    def validate_unconstrained_value(self, value: tf.Tensor, dtype: DType) -> tf.Tensor:
+        value = _cast_to_dtype(value, dtype)
+        unconstrained_value = _to_unconstrained(value, self.transform)
+        return tf.debugging.assert_all_finite(unconstrained_value,
+                message=f"gpflow.Parameter.validate_unconstrained_value")
```
awav: Could you change the message to something more descriptive?

gpflow/base.py
```diff
@@ -225,9 +232,7 @@ def _run_op(a, *args, **kwargs):
 tf.register_tensor_conversion_function(Parameter, lambda x, *args, **kwds: x.read_value())
 
 
-def _verified_value(value: VariableData, dtype: Optional[DType] = None) -> np.ndarray:
-    if isinstance(value, tf.Variable):
-        return value
+def _cast_to_dtype(value: VariableData, dtype: Optional[DType] = None) -> tf.Tensor:
```
awav:
```suggestion
def _cast_to_default_dtype(value: VariableData, dtype: Optional[DType] = None) -> tf.Tensor:
```

gpflow/base.py
```diff
@@ -51,15 +51,16 @@ def __init__(self,
         """
         super().__init__()
 
-        value = _verified_value(value, dtype)
+        self._transform = transform
```
awav: Could you move it back to the end? The reason is that attribute assign should happen after all potential failure cases.

gpflow/base.py
st--: What would you suggest?

gpflow/base.py
st--: It's not the default_dtype if the `dtype` argument is actually passed in. We could add a docstring, if you want, though it's an internal (module-private) function and very short so I thought that'd be overkill.

gpflow/base.py
st--: No I can't, I moved it to the front so that it's ready for validate_unconstrained_value() before the Variable construction.

st--:
This change means it is no longer possible to actually assign `inf` or `nan` to a gpflow.Parameter. Are there any use-cases where we would actually want to do this ?

jameshensman:
This all looks very sensible, and the test look like it covers the case you're interested in. 

I think the only thing to discuss is the behaviour. When doing `Parameter(0.0, positive)`:

1) the old behaviour is to get `1e-6` implicitly
2) the current (pre-this PR) is to get a NaN (bad! must fix.)
3) this PR would give you an exception
4) I wonder if it would be nicer to get a Warning and a change to `1e-6`?

gpflow/base.py
awav: "Unconstrained value of passed value has `NaN` or `Inf` and cannot be assigned"

markvdw:
I agree with @st-- that an error would be the cleanest thing to do.

It's rare to want to initialise your parameters at the total edge of the boundary condition anyway, and if you are, it's probably best to be aware that you're not really able to do so!

markvdw:
Although, just a thought... A related issue sometimes happens when the optimisation pushes the value to the utmost extreme. It would be annoying to optimise your model for ages, only for it to fail at the stage where the parameters are assigned.

I think this is the only use-case where I think that an error may not be the most appropriate thing? It would be nice if the result of the optimisation wouldn't be lost somehow...

st--:
@markvdw I don't think this would be an issue as the optimizers only work in unconstrained space and directly assign to the actual (unconstrained_)variable, not the (constrained) parameter. So the _validate_unconstrained_value won't be called, and no assert error can be triggered.

st--:
(we can also change it up again if it turns out to be an issue for anyone)

markvdw:
Ah right, if that's the case, then that's great! Things change so quickly... I think in the past `assign()` was called on finishing the optimisation...

st--:
Is there anything that needs doing on this PR then? If not please approve so we can merge it in!

gpflow/base.py
awav: Ok, then this is a separate issue, dependencies in the constructor calls and existing functions.

gpflow/base.py
```diff
@@ -112,6 +113,14 @@ def trainable(self, flag: Union[bool, int]):
     def initial_value(self):
         return self._unconstrained.initial_value
 
+    def validate_unconstrained_value(self, value: tf.Tensor, dtype: DType) -> tf.Tensor:
+        value = _cast_to_dtype(value, dtype)
+        unconstrained_value = _to_unconstrained(value, self.transform)
+        message = "gpflow.Parameter: unconstrained value of passed value " \
+                  "has NaN or Inf and cannot be assigned."
+        return tf.debugging.assert_all_finite(unconstrained_value, message=message)
+
```
awav:
```suggestion
```

PR #1200, opened 2020-01-09
reinstate project links in README.md that disappeared when changing develop to 2.0

st--:


st--:
Should we add a column stating gpflow 1/2 compatibility ?

PR #1211, opened 2020-01-16
Fix issue with tf.cast() and implicit dtype conversion

st--:
TensorFlow's tf.cast(value, dtype) implicitly does a tf.convert_to_tensor(value)
*before* the cast when the value is not a tensor already. When value is a python float,
this results in the following behaviour:

```python
>>> tf.cast(0.2, tf.float64)
<tf.Tensor: id=37, shape=(), dtype=float64, numpy=0.20000000298023224>
```

instead of the expected expansion of 0.2 to float64 precision that you get when
passing in an object that already carries dtype information, such as a numpy array
(which has float64 precision by default):

```python
>>> tf.cast(np.array(0.2), tf.float64)
<tf.Tensor: id=40, shape=(), dtype=float64, numpy=0.2>
```

This affected *all* gpflow.Parameter objects, resulting in numerical discrepancies
between GPflow 1 and 2, due to the pass through _cast_to_dtype, which is fixed by this PR.

### DIFF
--- a/gpflow/base.py
+++ b/gpflow/base.py
@@ -253,7 +253,10 @@ tf.register_tensor_conversion_function(Parameter, lambda x, *args, **kwds: x.rea
 def _cast_to_dtype(value: VariableData, dtype: Optional[DType] = None) -> tf.Tensor:
     if dtype is None:
         dtype = default_float()
-    return tf.cast(value, dtype)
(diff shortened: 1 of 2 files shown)


st--:
This might be an issue in other parts of GPflow; I'm investigating.

PR #1213, opened 2020-01-16
Fix more potential tf.cast() loss-of-precision issues

st--:
Follow-up to #1211 to extend the tf.cast() workaround for tensorflow/tensorflow#35938 to the rest of the GPflow codebase. Replaces more tf.cast() with to_default_float/to_default_int/gpflow.base.cast().


st--:
I'm a bit surprised by the failure in test_scipy_jit -- looks like the changes to cast ever so slightly *reduced* precision? :thinking:

st--:
The commit that makes the test fail is 7f00191.

awav:
@st-- , I skimmed through the PR. Please, move cast operation to utilities/ops.py

st--:
> @st-- , I skimmed through the PR. Please, move cast operation to utilities/ops.py
@awav that is exactly where I put them first. Turns out it does not work, because of circular imports. Importing anything from gpflow.utilities.ops first imports the parent module, gpflow.utilities, which in turn imports gpflow.utilities.utilities, which depends on gpflow.base for the Parameter.

I would agree it would be good to disentangle the module structure, but I think that's scope for another PR.

gpflow/base.py
```diff
@@ -15,6 +15,14 @@
 Prior = tfp.distributions.Distribution
 
 
+def cast(value, dtype):
```
awav: I will see on Monday what we can do with it. The cast should stay in util :)

awav: I will see on Monday what we can do with it. The cast should stay in util :)

gpflow/base.py
```diff
@@ -263,9 +263,10 @@ def _cast_to_dtype(value: VariableData, dtype: Optional[DType] = None) -> tf.Ten
     if dtype is None:
         dtype = default_float()
     if tf.is_tensor(value):
+        # TODO(awav) TF2.2 resolves issue with cast.
+        # `tf.cast` can be used alone instead this auxilary function.
         return tf.cast(value, dtype)
-    else:
-        return tf.convert_to_tensor(value, dtype)
+    return tf.convert_to_tensor(value, dtype=dtype)
```
st--:
```suggestion
    else:
        return tf.convert_to_tensor(value, dtype=dtype)
```

gpflow/base.py
```diff
@@ -263,9 +263,10 @@ def _cast_to_dtype(value: VariableData, dtype: Optional[DType] = None) -> tf.Ten
     if dtype is None:
         dtype = default_float()
     if tf.is_tensor(value):
+        # TODO(awav) TF2.2 resolves issue with cast.
+        # `tf.cast` can be used alone instead this auxilary function.
```
st--:
```suggestion
        # From TF2.2, `tf.cast` can be used alone instead of this auxiliary function.
        # workaround for https://github.com/tensorflow/tensorflow/issues/35938
```

gpflow/utilities/ops.py
```diff
@@ -2,9 +2,17 @@
 from typing import List, Optional, Union
 
 import tensorflow as tf
-import tensorflow_probability as tfp
 import numpy as np
-from ..base import cast
+
+
+def cast(
```
st--: @awav what makes this code duplication better than importing it from base ?

awav: The `base` should not export `cast`, this is simply a very bad place for a utility function.
The `tf.cast` is TF issue, we should not break gpflow integrity because of TF mistakes.

st--: I'm still trying to understand here... if it's used all over the place, why is `base` a bad module for it ?

awav: > if it's used all over the place, why is base a bad module for it?

The cast is a utility function and proper place is the utility module. Making it public though base module is wrong in terms of package hierarchy.

Because it is a temporary issue, I could make a cast function inside the base and use it in utilities, but export only utility function. This would help to avoid duplication. 

@st-- do you want me to make this change?

st--: No it's fine - hopefully it'll be fixed upstream soon and we can remove all of it anyways! - I was just trying to understand your reasoning :)

PR #1219, opened 2020-01-20
assert_shapes in code instead of just comments

st--:
Adds shape asserts to some of the modules listed in #1241:
- conditionals.mo_conditionals (not 100%)
- conditionals.util
- kullback_leiblers
- logdensities (only `multivariate_normal`)


gpflow/conditionals/util.py
```diff
@@ -7,7 +7,7 @@
 def base_conditional(Kmn: tf.Tensor,
                      Kmm: tf.Tensor,
                      Knn: tf.Tensor,
-                     function: tf.Tensor,
+                     f: tf.Tensor,
```
awav: I agree that `function` name might not be appropriate. However, `f` is clearly not self-explanatory and depends a lot on the context. Moreover, we don't follow pep8 in naming, but still, we need to stick to those rules as far as it is possible.

awav: I agree that `function` name might not be appropriate. However, `f` is clearly not self-explanatory and depends a lot on the context. Moreover, we don't follow pep8 in naming, but still, we need to stick to those rules as far as it is possible.

gpflow/conditionals/util.py
```diff
@@ -51,6 +51,16 @@ def base_conditional(Kmn: tf.Tensor,
...
 
+    shape_constraints = [
+        (Kmn, [..., 'M', 'N']),
+        (Kmm, ['M', 'M']),
+        (Knn, [..., 'N', 'N'] if full_cov else [..., 'N']),
+        (f, ['M', 'R']),
+    ]
+    if q_sqrt is not None:
+        shape_constraints.append((q_sqrt, ['R', 'M', 'M']))
+    tf.debugging.assert_shapes(shape_constraints)
```
awav: This is super cool and worth of making it as a separate internal module function.

st--: Thanks to GPflow user @bodin-e (tensorflow#25422)... :)
What do you mean by making it an internal function?

gpflow/conditionals/util.py
st--: Happy to discuss name changes (I also find the module name unhelpful) - this was meant to bring it in line with the other functions in this file at least.

gpflow/conditionals/mo_conditionals.py
```diff
@@ -50,8 +50,7 @@ def shared_independent_conditional(
     """
     Kmm = covariances.Kuu(inducing_variable, kernel, jitter=default_jitter())  # [M, M]
     Kmn = covariances.Kuf(inducing_variable, kernel, Xnew)  # [M, N]
-    Knn = kernel(Xnew, full=full_cov, full_output_cov=False)
-    Knn = Knn[0, ...] if full_cov else Knn[..., 0]  # [N, N] or [N]
+    Knn = kernel.kernel(Xnew, full=full_cov)
```
st--: There was an inefficiency in the code in which the SharedIndependent kernel used tile() to construct a [P, N, N] matrix, of which this conditional implementation then selected the first slice. Instead, we can simply call the base kernel and only construct the [N, N] matrix.

gpflow/conditionals/mo_conditionals.py
```diff
@@ -96,12 +95,12 @@ def single_gp_conditional(t):
     rmu, rvar = tf.map_fn(single_gp_conditional, (Kmms, Kmns, Knns, fs, q_sqrts),
                           (default_float(), default_float()))  # [P, N, 1], [P, 1, N, N] or [P, N, 1]
 
-    fmu = rollaxis_left(rmu[..., 0], 1)  # [N, P]
+    fmu = rollaxis_left(tf.squeeze(rmu, axis=-1), 1)  # [N, P]
```
st--: tf.squeeze() includes a check that this axis has length 1, it's good to have even a small additional safe guard.

gpflow/conditionals/util.py
```diff
@@ -51,6 +51,18 @@ def base_conditional(Kmn: tf.Tensor,
         0)  # [N]
     Kmn = tf.transpose(Kmn, perm)  # [..., M, N]
```
st--: We can't have the assert_shapes check at the start of the function body as the function signature expects Kmn with shape [M, ..., N], but assert_shapes() only handles *leading* Ellipsis dimensions - so we need to add the shape constraint *after* rolling it to [..., M, N]. Would it not be cleaner to simply expect the input to be [..., M, N] in the first place? or if we want the ... to be next to the N, change it to take `Knm` instead of `Kmn` and expect shape [..., N, M]?

gpflow/conditionals/util.py
```diff
@@ -296,66 +348,89 @@ def fully_correlated_conditional_repeat(Kmn,
         - variance: [R, N, P], [R, N, P, P], [R, P, N, N], [R, N, P, N, P]
     """
     R = tf.shape(f)[1]
-    M, N, K = tf.unstack(tf.shape(Kmn), num=Kmn.shape.ndims, axis=0)
+    M, N, P = tf.unstack(tf.shape(Kmn), num=Kmn.shape.ndims, axis=0)
```
st--: K -> P for consistency with other functions in this module

gpflow/conditionals/util.py
awav: I wonder if it were more readable with tf.debugging.assert_shapes per each shape instead of having a list.

tests/test_multioutput.py
```diff
@@ -291,8 +291,11 @@ def test_sample_conditional_mixedkernel():
     np.testing.assert_array_almost_equal(np.cov(value, rowvar=False), np.cov(value2, rowvar=False), decimal=1)
 
 
-@pytest.mark.parametrize('R', [1, 5])
-@pytest.mark.parametrize("func", [fully_correlated_conditional_repeat, fully_correlated_conditional])
```
st--: The case with R=5 and func=fully_correlated_conditional is not valid as fully_correlated_conditional requires R=1

gpflow/conditionals/util.py
```diff
@@ -196,6 +228,16 @@ def independent_interdomain_conditional(Kmn,
     """
     M, L, N, P = tf.unstack(tf.shape(Kmn), num=Kmn.shape.ndims, axis=0)
 
+    shape_constraints = [
```
st--: This function isn't actually tested - probably not scope of this PR to add a test for that, but should we turn it into a separate issue ?

gpflow/conditionals/util.py
st--: Then you can't cross-check that the dimensions match each other (e.g. that the "M" of Kmn is the same as the "M" of Kmm).

awav:
I think this PR should be split into many small tasks rather than one big change. Could you make an issue and put it on GPflow feature board (it is enhancing code feature)?

Thanks

st--:
How do you think this PR should be split up? (As far as GPflow PRs go, this doesn't seem to be a particularly large one?)

st--:
I've made a new issue #1241 to handle the rest of the code base (hopefully by an army of volunteer contributors ;)) - can we review this PR and get it merged in then?

gpflow/conditionals/util.py
condnsdmatters: agree we should move away from `f` . can I suggest `f_mu` and `q_sqrt` being `f_sqrt`. In general `f` looks like it s either, a `q_mu` or apparently just a set of points (dirac points, when `q_sqrt` is None). Another possibility is `q_mu`.

gpflow/conditionals/util.py
condnsdmatters: seeing as its not a critical change I wouldn't change this important interface at this point for a mildly aesthetic improvement.

condnsdmatters:
if you add some tests that will raise testing asserts with dodgy shapes that would also be good, if poss

gpflow/conditionals/util.py
```diff
@@ -375,38 +450,59 @@ def rollaxis_right(A, num_rolls):
...
+        cov_shape = "NPP"
 
     elif not full_cov and not full_output_cov:  # g_var is [..., N, L]
         W_squared = W**2  # [P, L]
         f_var = tf.tensordot(g_var, W_squared, [[-1], [-1]])  # [..., N, P]
+        cov_shape = "NP"
+
+    shape_constraints.extend([
+        (f_mean, "*NP"),
+        (f_var, "*"+cov_shape),
```
awav:
```suggestion
        (f_var, f"*{cov_shape}"),
```

gpflow/conditionals/util.py
st--: I'm happy to defer that to another PR but I think it's an important clarity-of-design issue - "leading" dimensions in the middle seems rather confusing!

gpflow/conditionals/util.py
st--: how about f_mean and f_cov_chol ?

gpflow/conditionals/util.py
```diff
@@ -100,7 +112,15 @@ def base_conditional(Kmn: tf.Tensor,
...
         fvar = tf.linalg.adjoint(fvar)  # [N, R]
 
-    return fmean, fvar  # [N, R], [R, N, N] or [N, R]
+    shape_constraints = [
+        (Kmn, [..., 'M', 'N']),  # for N
+        (f, [..., 'M', 'R']),  # for R
+        (fmean, [..., 'N', 'R']),
+        (fvar, [..., 'R', 'N', 'N'] if full_cov else [..., 'N', 'R']),
+    ]
+    tf.debugging.assert_shapes(shape_constraints)
```
joelberkeley-pio: I very much don't think we should be checking return values

gpflow/conditionals/util.py
vdutor: I'm in favour of keeping `f` and `q_sqrt` for *now*. I agree that they are not the most appropriate names, but we should not ignore the cost of changing them to something else sub-optimal. I would rather spend time bundling the `f` and `q_sqrt` params in an appropriate class.

gpflow/conditionals/mo_conditionals.py
joelberkeley-pio: is this related to the PR title?

gpflow/conditionals/util.py
joelberkeley-pio: is this related to the PR title?

gpflow/conditionals/util.py
```diff
@@ -30,14 +31,14 @@ def base_conditional(Kmn: tf.Tensor,
     :param Knn: [..., N, N]  or  N
     :param f: [M, R]
     :param full_cov: bool
-    :param q_sqrt: None or [R, M, M] (lower triangular)
+    :param q_sqrt: None or [R, M, M] (lower triangular) or [M, R] (diagonal)
```
joelberkeley-pio: I'd suggest adding a type hint to `q_sqrt`. I guess it's `Optional[tf.Tensor]`. Then you can say
```
     :param q_sqrt: If not `None`, must have shape [R, M, M] (lower triangular) or [M, R] (diagonal) 
```
which I think reads a little more clearly. For example, isn't `None` a shape in some contexts?

gpflow/conditionals/util.py
```diff
@@ -51,6 +52,18 @@ def base_conditional(Kmn: tf.Tensor,
...
+        (Kmn, [..., 'M', 'N']),
+        (Kmm, ['M', 'M']),
+        (Knn, [..., 'N', 'N'] if full_cov else [..., 'N']),
+        (f, ['M', 'R']),
+    ]
+    if q_sqrt is not None:
+        shape_constraints.append(
+            (q_sqrt, (['M', 'R'] if q_sqrt.shape.ndims == 2 else ['R', 'M', 'M']))
+        )
+    assert_shapes(shape_constraints)
```
joelberkeley-pio: todo joel review this shape check

gpflow/conditionals/util.py
stefanosele: My opinion is that if you choose to rename `q_sqrt` globally in the repo then feel free to change it as well here. Now given the fact that you won't do that since I don't see many good reasons for it, and we keep discussing about having a distribution object for the posterior instead of `q_mu` and `q_sqrt` (this is the only good reason which makes a lot of sense but it is not trivial) I would suggest to keep it as is. Same goes for `f`

gpflow/conditionals/util.py
joelberkeley-pio: `tf.debugging.assert_shapes` is still good for function arguments, while `utilities.assert_shapes` is for internal code consistency checks

gpflow/conditionals/util.py
```diff
@@ -100,7 +113,15 @@ def base_conditional(Kmn: tf.Tensor,
...
         fvar = tf.linalg.adjoint(fvar)  # [N, R]
 
-    return fmean, fvar  # [N, R], [R, N, N] or [N, R]
+    shape_constraints = [
+        (Kmn, [..., 'M', 'N']),  # for N
+        (f, [..., 'M', 'R']),  # for R
+        (fmean, [..., 'N', 'R']),
+        (fvar, [..., 'R', 'N', 'N'] if full_cov else [..., 'N', 'R']),
+    ]
+    assert_shapes(shape_constraints)
```
joelberkeley-pio: todo joel review this check

gpflow/conditionals/util.py
```diff
@@ -113,7 +134,14 @@ def sample_mvn(mean, cov, cov_structure=None, num_samples=None):
     - "full": cov holds the full covariance matrix (without jitter)
     :return: sample from the MVN of shape [..., (S), N, D], S = num_samples
     """
-    assert cov_structure == "diag" or cov_structure == "full"
+    if cov_structure not in ("diag", "full"):
+        raise ValueError("cov_structure must be 'diag' or 'full'")
```
joelberkeley-pio: not for this PR, but an enum might be good here

gpflow/conditionals/util.py
```diff
@@ -113,7 +134,14 @@ def sample_mvn(mean, cov, cov_structure=None, num_samples=None):
...
+
+    shape_constraints = [
+        (mean, [..., 'N', 'D']),
+        (cov, [..., 'N', 'D'] if cov_structure == 'diag' else [..., 'N', 'D', 'D']),
+    ]
+    assert_shapes(shape_constraints)
```
joelberkeley-pio: as above `tf.debugging.assert_shapes` on args

gpflow/conditionals/util.py
joelberkeley-pio: not fussed - this PR or another, but yeah somewhere

gpflow/conditionals/util.py
```diff
@@ -206,14 +249,18 @@ def independent_interdomain_conditional(Kmn,
...
         At = tf.reshape(tf.transpose(Ar), (P, N, M * L))  # [P, N, L]
         fvar = Knn - tf.linalg.matmul(At, At, transpose_b=True)  # [P, N, N]
+        cov_shape = "PNN"
     elif not full_cov and full_output_cov:
         At = tf.reshape(tf.transpose(Ar, [2, 3, 1, 0]), (N, P, M * L))  # [N, P, L]
         fvar = Knn - tf.linalg.matmul(At, At, transpose_b=True)  # [N, P, P]
+        cov_shape = "NPP"
     elif not full_cov and not full_output_cov:
         fvar = Knn - tf.reshape(tf.reduce_sum(tf.square(A), [0, 1]), (N, P))  # Knn: [N, P]
+        cov_shape = "NP"
```
joelberkeley-pio: rename `cov_shape` to `intended_cov_shape` or similar, to make it clear it's the intended not necessarily actual shape

gpflow/conditionals/util.py
```diff
@@ -196,6 +229,16 @@ def independent_interdomain_conditional(Kmn,
...
 
+    shape_constraints = [
+        (Kmn, "MLNP"),
+        (Kmm, "LMM"),
+        (f, "ML"),
+    ]
+    if q_sqrt is not None:
+        shape_constraints.append(
+            (q_sqrt, "ML" if q_sqrt.shape.ndims == 2 else "LMM")
+        )
```
joelberkeley-pio: assert args shape here ...

gpflow/conditionals/util.py
```diff
@@ -240,6 +287,14 @@ def independent_interdomain_conditional(Kmn,
             fvar = fvar + tf.linalg.matmul(LTAr, LTAr, transpose_a=True)  # [N, P, P]
         elif not full_cov and not full_output_cov:
             fvar = fvar + tf.reshape(tf.reduce_sum(tf.square(LTA), (0, 1)), (N, P))
+
+    shape_constraints.extend([
+        (Knn, cov_shape),
+        (fmean, "NP"),
+        (fvar, cov_shape),
+    ])
+    assert_shapes(shape_constraints)
```
joelberkeley-pio: ... and return value shapes here

gpflow/conditionals/util.py
```diff
@@ -251,23 +306,21 @@ def fully_correlated_conditional(Kmn, Kmm, Knn, f, *, full_cov=False, full_outpu
...
-                                               q_sqrt=q_sqrt,
-                                               white=white)
-    return m[0, ...], v[0, ...]
+    mean, var = fully_correlated_conditional_repeat(
+        Kmn, Kmm, Knn, f,
+        full_cov=full_cov,
+        full_output_cov=full_output_cov,
+        q_sqrt=q_sqrt,
+        white=white)
+    return tf.squeeze(mean, axis=0), tf.squeeze(var, axis=0)
```
joelberkeley-pio: is `tf.squeeze` as performant as `Tensor.__getitem__`?

gpflow/conditionals/util.py
```diff
@@ -296,66 +349,89 @@ def fully_correlated_conditional_repeat(Kmn,
...
+
+    shape_constraints = [
+        (Kmn, "MNP"),
+        (Kmm, "MM"),
+        (f, "MR"),
+    ]
+    if q_sqrt is not None:
+        shape_constraints.append(
+            (q_sqrt, "MR" if q_sqrt.shape.ndims == 2 else "RMM")
+        )
```
joelberkeley-pio: check args here ...

gpflow/conditionals/util.py
```diff
@@ -296,66 +349,89 @@ def fully_correlated_conditional_repeat(Kmn,
...
+            fvar = fvar[None, ...] + addvar  # [R, N, P]
     else:
         fvar = tf.broadcast_to(fvar[None], tf.shape(fmean))
+
+    shape_constraints.extend([
+        (Knn, cov_shape),
+        (fmean, "RNP"),
+        (fvar, "R"+cov_shape),
+    ])
+    assert_shapes(shape_constraints)
```
joelberkeley-pio: ... and return value here

gpflow/conditionals/util.py
```diff
@@ -375,38 +451,59 @@ def rollaxis_right(A, num_rolls):
     return tf.transpose(A, perm)
 
 
-def mix_latent_gp(W, g_mu, g_var, full_cov, full_output_cov):
+def mix_latent_gp(W, g_mean, g_var, full_cov, full_output_cov):
```
joelberkeley-pio: why this change?

gpflow/utilities/utilities.py
```diff
@@ -39,6 +41,20 @@ def to_default_float(x):
     return tf.cast(x, dtype=default_float())
 
 
+class ShapeError(AssertionError):
+    pass
+
+
+def assert_shapes(shapes, *args, **kwargs):
```
vdutor: do we really need this wrapper? I, personally, like to be as close to tensorflow as possible.

gpflow/conditionals/mo_conditionals.py
vdutor: not really, but it's a good catch.

tests/test_multioutput.py
vdutor: how come it was working before?

gpflow/utilities/utilities.py
vdutor: Is this custom `assert_shapes` really necessary? I, personally, like to stay as close to raw tensorflow as much as possible.

gpflow/utilities/utilities.py
```diff
@@ -39,6 +41,20 @@ def to_default_float(x):
...
+    """
+    Wrapper around tf.debugging.assert_shapes() to replace ValueError with ShapeError
+    """
+    try:
+        tf.debugging.assert_shapes(shapes, *args, **kwargs)
+    except ValueError as e:
+        raise ShapeError(e)
```
joelberkeley-pio:
```suggestion
        if __debug__:
            raise ShapeError(e)
```

joelberkeley-pio: so that users can turn off debugging

gpflow/utilities/utilities.py
joelberkeley-pio: ti wants to use assertions to check code consistency, and I get that, but I don't think inconsistent code should throw an error that one might reaonsably catch. `tf.debugging.assert_shapes` throws a `ValueError` so we're converting it to throw a different error

joelberkeley-pio: i agree we should be as close to tf as possible, but I think their functionality is limited here

gpflow/conditionals/util.py
st--: I think we should consistently use one of them, not mix&match tf.debugging.assert_shapes and our wrapper utilities.assert_shapes. If we are saying we want our own wrapper to not always raise ValueErrors, wouldn't it be better to simply give it an argument that says whether to replace ValueError with ShapeError or not?

gpflow/conditionals/util.py
```diff
@@ -100,7 +113,15 @@ def base_conditional(Kmn: tf.Tensor,
     if not full_cov:
         fvar = tf.linalg.adjoint(fvar)  # [N, R]
 
-    return fmean, fvar  # [N, R], [R, N, N] or [N, R]
+    shape_constraints = [
+        (Kmn, [..., 'M', 'N']),  # for N
```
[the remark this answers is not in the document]
st--: This tensor got checked in the assert_shapes at the start of function already. I include it in here so I can get hold of the `N` dimension - but it could have been any other tensor that has an `N` somewhere in its dimensions. Same for `f` below.

gpflow/conditionals/util.py
st--: `tf.squeeze` checks that the dimension you're squeezing actually only has length 1. I think this extra check is valuable, and I don't expect it to have a significant performance impact relative to any of the matmuls.

gpflow/conditionals/util.py
st--: to be more explicit

gpflow/utilities/utilities.py
st--: @joelberkeley-pio pointed out that ValueErrors are commonly try:except:ed and that this would hide the shape asserts. I think ideally TensorFlow's `assert_shapes` would raise a custom exception, not a ValueError; this wrapper makes clearer what went wrong. I think it'll make it a bit more user-friendly.

Can take it out again if that's the majority preference...

st--: @joelberkeley-pio pointed out that ValueErrors are commonly try:except:ed and that this would hide the shape asserts. I think ideally TensorFlow's `assert_shapes` would raise a custom exception, not a ValueError; this wrapper makes clearer what went wrong. I think it'll make it a bit more user-friendly.

Can take it out again if that's the majority preference...

tests/test_multioutput.py
st--: because it simply did `[0]` instead of `tf.squeeze()`.

gpflow/conditionals/util.py
joelberkeley-pio: yeah could do that

joelberkeley-pio:
> Do you have any idea if all these checks slow down our code?

I would have thought most of the execution time in GPflow code would be in data manipulation. And I'd expect (but I don't really know) `tf.Tensor` to store their shapes as values on the object, and so would be really quick to access. The rest of the check logic is probably insignificant.

If however they do slow things down, it's a good reason to wrap them all along with some kind of flag to turn them on and off. The obvious flag when checking code consistency (e.g. return value shapes) would be python's in built `__debug__`. We could have another config flag for argument checks.

gpflow/kullback_leiblers.py
```diff
@@ -43,7 +44,7 @@ def gauss_kl(q_mu, q_sqrt, K=None):
           p(x) = N(0, K)    if K is not None
           p(x) = N(0, I)    if K is None
 
-    We assume N multiple independent distributions, given by the columns of
+    We assume L multiple independent distributions, given by the columns of
     q_mu and the last dimension of q_sqrt. Returns the sum of the divergences.
```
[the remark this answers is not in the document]
st--: depends on whether q_sqrt has rank 2 (last dimension) or 3 (first dimension).... :/

gpflow/conditionals/util.py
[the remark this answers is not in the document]
st--: made the comment clearer

gpflow/conditionals/util.py
st--: what's wrong with always raising a ShapeError though ?

gpflow/conditionals/util.py
st--: Put into an issue: #1295

gpflow/conditionals/util.py
st--: Moved into new issue: #1296

gpflow/conditionals/util.py
joelberkeley-pio: It's a good question, and I'm not entirely clear on the reasons, but essentially I believe it boils down to the fact that the two errors have occurred for different reasons. See [this question](https://stackoverflow.com/q/117171/5986907) for popular answers.

The error at the beginning of the function is a result of erroneous user input, which could have resulted from any number of things out of our control. Perhaps a malformed CSV has been uploaded to some ML pipeline and it's missing a column. Crucially, it could be that they can recover from that situation.

The error at the end of the function is a check of the GPflow code. If there's a bug in GPflow, the user can't do anything about that.

gpflow/utilities/utilities.py
```diff
@@ -40,6 +42,20 @@ def to_default_float(x):
     return tf.cast(x, dtype=default_float())
 
 
+class ShapeError(AssertionError):
+    pass
+
+
+def assert_shapes(shapes, *args, **kwargs):
```
joelberkeley-pio: can we keep the argument names and types? It will at least help with autocompletion in the editor

gpflow/utilities/utilities.py
```diff
@@ -40,6 +42,20 @@ def to_default_float(x):
...
+    """
+    Wrapper around tf.debugging.assert_shapes() to replace ValueError with ShapeError
+    """
+    try:
+        tf.debugging.assert_shapes(shapes, *args, **kwargs)
+    except ValueError as e:
+        raise ShapeError(e)
```
joelberkeley-pio:
```suggestion
    if __debug__:
        try:
            tf.debugging.assert_shapes(shapes, *args, **kwargs)
        except ValueError as e:
            raise ShapeError(e)
```

joelberkeley-pio: so that users can turn off debugging

gpflow/utilities/utilities.py
st--: what if tensorflow just goes and changes them ? (not for our code, but if someone uses this part of gpflow with extra arguments that we're not testing for, and suddenly it won't work anymore?)

st--:
@condnsdmatters can we merge this in now ?

gpflow/conditionals/mo_conditionals.py
```diff
@@ -75,8 +75,7 @@ def shared_independent_conditional(
     """
     Kmm = covariances.Kuu(inducing_variable, kernel, jitter=default_jitter())  # [M, M]
     Kmn = covariances.Kuf(inducing_variable, kernel, Xnew)  # [M, N]
-    Knn = kernel(Xnew, full_cov=full_cov, full_output_cov=False)
-    Knn = Knn[0, ...] if full_cov else Knn[..., 0]  # [N, N] or [N]
+    Knn = kernel.kernel(Xnew, full_cov=full_cov)
```
joelberkeley-pio: why this change?

gpflow/conditionals/util.py
```diff
@@ -54,6 +57,18 @@ def base_conditional(
     )  # [N]
     Kmn = tf.transpose(Kmn, perm)  # [..., M, N]
 
+    shape_constraints = [
+        (Kmn, [..., "M", "N"]),
```
joelberkeley-pio: is this going to raise an error that contradicts the docstring, cos we're asserting on the altered shape?

gpflow/conditionals/util.py
```diff
@@ -103,7 +118,15 @@ def base_conditional(
     if not full_cov:
         fvar = tf.linalg.adjoint(fvar)  # [N, R]
 
-    return fmean, fvar  # [N, R], [R, N, N] or [N, R]
+    shape_constraints = [
+        (Kmn, [..., "M", "N"]),  # tensor included again for N dimension
+        (f, [..., "M", "R"]),  # tensor included again for R dimension
```
joelberkeley-pio: what do these comments mean?

gpflow/conditionals/mo_conditionals.py
joelberkeley-pio: i'd suggest leaving this kind of change to a separate PR, it makes reviewing a lot easier as I need to look out for fewer things

gpflow/conditionals/mo_conditionals.py
st--: because the code was wrong and this was spotted by the asserts

gpflow/conditionals/util.py
st--: yes. and there's nothing we can do about it

gpflow/conditionals/util.py
st--: that I'm including this tensor in the list so I get a handle on the `N` and `R` respectively.... I tried to improve it following previous review comments!

gpflow/conditionals/mo_conditionals.py
st--: This was covered by previous review comments

gpflow/conditionals/mo_conditionals.py
st--: well it kind of goes with the asserting shapes - it's asserting that that dimension has length 1!

gpflow/conditionals/util.py
joelberkeley-pio: sure you can. you can add an explanation to the shape assertion explaining that the expected type isn't what the function expects

gpflow/conditionals/util.py
```diff
@@ -373,38 +445,56 @@ def rollaxis_right(A, num_rolls):
...
     r"""Takes the mean and variance of an uncorrelated L-dimensional latent GP
     and returns the mean and the variance of the mixed GP, `f = W g`,
     where both f and g are GPs, with W having a shape [P, L]
 
     :param W: [P, L]
-    :param g_mu: [..., N, L]
+    :param g_mean: [..., N, L]
     :param g_var: [..., N, L] (full_cov = False) or [L, ..., N, N] (full_cov = True)
-    :return: f_mu and f_var, shape depends on `full_cov` and `full_output_cov`
+    :return: f_mean and f_var, shape depends on `full_cov` and `full_output_cov`
```
joelberkeley-pio: return value don't have names. i suggest "mean and variance of f"

gpflow/utilities/utilities.py
joelberkeley-pio: i guess they could add further kwargs and be backwards compatible. For any other changes, i'd be worried we're not pinning our version of tensorflow tightly enough. if tensorflow's changing their API under our feet that's a problem, as our own usage of tf.function might be broken

i won't block the PR on this

gpflow/conditionals/util.py
st--: I'd appreciate if you could suggest (in code) how you think that could look like!

gpflow/conditionals/util.py
```diff
@@ -54,6 +57,18 @@ def base_conditional(
...
+        (Kmn, [..., "M", "N"]),
+        (Kmm, ["M", "M"]),
+        (Knn, [..., "N", "N"] if full_cov else [..., "N"]),
+        (f, ["M", "R"]),
+    ]
+    if q_sqrt is not None:
+        shape_constraints.append(
+            (q_sqrt, (["M", "R"] if q_sqrt.shape.ndims == 2 else ["R", "M", "M"]))
+        )
+    assert_shapes(shape_constraints)
```
joelberkeley-pio:
```suggestion
    assert_shapes(shape_constraints, message="Note that this check verifies the shape of an alternative representation of Kmn. See the docs for the actual expected shape")
```

gpflow/conditionals/util.py
joelberkeley-pio: added

gpflow/conditionals/util.py
```diff
@@ -373,38 +445,56 @@ def rollaxis_right(A, num_rolls):
...
-    :return: f_mu and f_var, shape depends on `full_cov` and `full_output_cov`
+    :return: f_mean and f_var, shape depends on `full_cov` and `full_output_cov`
     """
-    f_mu = tf.tensordot(g_mu, W, [[-1], [-1]])  # [..., N, P]
+    shape_constraints = [
+        (W, "PL"),
+        (g_mean, "*NL"),
+    ]
+    if not full_cov:
+        shape_constraints.append((g_var, "*NL"))
```
joelberkeley-pio: why would g_var not have constraints if not full_cov?

gpflow/kullback_leiblers.py
```diff
@@ -71,32 +73,45 @@ def gauss_kl(q_mu, q_sqrt, K=None, *, K_cholesky=None):
     is_white = (K is None) and (K_cholesky is None)
     is_diag = len(q_sqrt.shape) == 2
 
-    M, B = tf.shape(q_mu)[0], tf.shape(q_mu)[1]
+    shape_constraints = [
+        (q_mu, ["M", "L"]),
+        (q_sqrt, (["M", "L"] if is_diag else ["L", "M", "M"])),
```
joelberkeley-pio: above you've used `"ML"` etc instead of `["M", "L"]` etc. do we want to stick with one?

gpflow/kullback_leiblers.py
```diff
@@ -71,32 +73,45 @@ def gauss_kl(q_mu, q_sqrt, K=None, *, K_cholesky=None):
...
+    ]
+    if not is_white:
+        if K is not None:
+            shape_constraints.append((K, (["L", "M", "M"] if len(K.shape) == 3 else ["M", "M"])))
+        else:
+            shape_constraints.append(
+                (K_cholesky, (["L", "M", "M"] if len(K_cholesky.shape) == 3 else ["M", "M"]))
+            )
```
joelberkeley-pio: just to give you some idea of how nuts idris is, you could write this all as part of the type and it would check it at compile time. yes, including the function calls.

gpflow/utilities/utilities.py
```diff
@@ -44,6 +46,21 @@ def to_default_float(x):
...
+
+
+def assert_shapes(shapes, *args, **kwargs):
+    """
+    Wrapper around tf.debugging.assert_shapes() to replace ValueError with ShapeError
+    """
+    if __debug__:
+        try:
+            tf.debugging.assert_shapes(shapes, *args, **kwargs)
+        except ValueError as e:
```
joelberkeley-pio: you mentioned it could also raise InvalidArgumentError
```suggestion
        except (tf.errors.InvalidArgumentError, ValueError) as e:
```

st--: Yes, but that won't actually work as the try:except: is parsed at trace-time, whereas the InvalidArgumentError would only be raised at graph-run-time, so it will still be an InvalidArgumentError (I just tried it out locally).

gpflow/kullback_leiblers.py
st--: I'd love that...

gpflow/conditionals/util.py
st--: because we can't check if there are inner "leading" dimensions (like above, see https://github.com/GPflow/GPflow/issues/1296). in one case, g_var is rolled, and the shape_constraint is added below

st--: (I've added a comment to clarify)

gpflow/utilities/utilities.py
joelberkeley-pio: i'm starting to question the sanity of something like `tf.function` in an interpreted language

joelberkeley-pio: so i guess the problem is that try except doesn't work as expected in `tf.function`. OK, if we can't make this consistent with and without `tf.function`, maybe we should abandon the custom error.

How about we just use `tf.debugging.assert_shapes` as is (delete the custom `assert_shapes` function), and apply `if __debug__` where checks are for code consistency, then at least users can just turn those ones off and catch the `InvalidArgumentError` or `ValueError` as normal from outside the `tf.function`.

Sorry we've had to go round the houses. I'm confident it was worth looking into, but alas I think we're limited here by python/tensorflow functionality

st--: I don't think the `if __debug__:` all around our code are worth it. I'm happy to kick out the wrapper function and replace with tf.debugging.assert_shapes again though.

joelberkeley-pio: ok. i think it could be a _bit_ of a problem for users, but let's just do it and get the PR in

gpflow/conditionals/util.py
```diff
@@ -294,66 +352,87 @@ def fully_correlated_conditional_repeat(
...
-            fvar = fvar[None, ...] + addvar  # [R, N, K]
+            addvar = tf.reshape(tf.reduce_sum(tf.square(LTA), axis=1), (R, N, P))  # [R, N, P]
+            fvar = fvar[None, ...] + addvar  # [R, N, P]
     else:
         fvar = tf.broadcast_to(fvar[None], tf.shape(fmean))
+
+    shape_constraints.extend(
+        [(Knn, intended_cov_shape), (fmean, ["R", "N", "P"]), (fvar, ["R"] + intended_cov_shape),]
+    )
+    tf.debugging.assert_shapes(shape_constraints, message="fully_correlated_conditional_repeat()")
```
joelberkeley-pio: no need to change it, cos I'd rather the PR's in, but I would have thought the stack trace would say what function it was raised in

st--: you'd hope so , but in tensorflow function land forget about stacktraces ...

PR #1220, opened 2020-01-20
Remove TFP deprecated bijector

awav:
The `tfp.bijectors.AffineScalar` bijector gives annoying deprecated messages :)


awav:
@st--, indeed. I guess that TFP runs tests against TF2.1, but it works fine with TF2.0 as they have it as a dependency. They are just lazy :)

It feels like TF people should have never released 2.0 and started straight from 2.1. There are so many functional changes and fixes there (e.g. is_tensor_like). Should we switch support to the `2.1`? I do not want to have many things like `if ... else ...` checking package versions.

condnsdmatters:
could we not promote rc2 to be something easily accessible for 2.0 users, and then change support for gpflow 2.1 to be 2.1 compatible. smth like that?

Issue #846, opened 2018-09-07
Doc(string) updates


st--:
Conditional() does not actually require q_sqrt to be lower-triangular. (gauss_kl still does)
Should update the docs accordingly...
There are a few other things to update as well, e.g. consistent notation (some places use L (for latent), some B (for batch), some P (for outPut), sometimes in the same function e.g. gauss_kl), nicer docstrings (unicode)...

st--:
@condnsdmatters has this been addressed by #1223?

PR #1223, opened 2020-01-21
Inconisistency on whether q_sqrt is triangular fixed.

condnsdmatters:
Problem:
* KL divergence calculations only take lower triangle parts of q_sqrt in
their calculation (implying a lower triangular parameterisation of
covariance)
* Conditional calculations perform calculations using the whole of
q_sqrt.
* The end result is if you forgot to add the triangular transform
to the q_sqrt, your forward pass may be the same (if you initialise with
a triangular matrix), but your gradients will immediately stop q_sqrt
being triangular.

Fix:
* Conditional calulcations now only take the lower triangle parts of q_sqrt
in their calculation

### DIFF
--- a/gpflow/conditionals/util.py
+++ b/gpflow/conditionals/util.py
@@ -82,7 +82,7 @@ def base_conditional(Kmn: tf.Tensor,
         if q_sqrt_dims == 2:
             LTA = A * tf.expand_dims(tf.transpose(q_sqrt), 2)  # [R, M, N]
         elif q_sqrt_dims == 3:
-            L = q_sqrt
+            L =  tf.linalg.band_part(q_sqrt, -1, 0)  # force lower triangle # [R, M, M]
             L_shape = tf.shape(L)
             L = tf.broadcast_to(L, tf.concat([leading_dims, L_shape], 0))
 
--- a/tests/test_conditionals.py
+++ b/tests/test_conditionals.py
@@ -18,6 +18,8 @@ import tensorflow as tf
 from numpy.testing import assert_allclose
 
 import gpflow
+from gpflow import Parameter
+from gpflow.utilities.bijectors import triangular
 from gpflow.conditionals import conditional
 from gpflow.config import default_float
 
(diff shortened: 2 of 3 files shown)


tests/test_kldiv.py
```diff
@@ -17,13 +17,40 @@
...
 
 rng = np.random.RandomState(0)
 
+# ------------------------------------------
+# Fixtures
+# ------------------------------------------
+
+Ln = 2
+Nn = 10
+Mn = 50
```
st--: Would the test be faster but just as useful if these were smaller ?

tests/test_kldiv.py
```diff
@@ -150,3 +177,37 @@ def test_unknown_size_inputs():
     unknown_shape = gauss_kl(mu, sqrt)
 
     np.testing.assert_allclose(known_shape, unknown_shape)
+
+
+@pytest.mark.parametrize('white', [True, False])
+def test_q_sqrt_constraints(inducing_points, kernel, mu, white):
```
st--: Would it be worth unifying the different ways of specifying input arrays (pytest fixtures vs Datum class)? I don't have any preference either way...

tests/test_conditionals.py
```diff
@@ -128,3 +130,40 @@ def test_gaussian_whiten(Xdata, Xnew, kernel, mu, sqrt):
...
+    q_sqrt_constrained = Parameter(tril, transform=triangular())
+    q_sqrt_unconstrained = Parameter(tril)
+
+    diff_before_gradient_step = (q_sqrt_constrained - q_sqrt_unconstrained).numpy()
+    assert_allclose(diff_before_gradient_step, 0)
+
+    Fstars = []
+    for q_sqrt in [q_sqrt_constrained, q_sqrt_unconstrained]:
+
+        with tf.GradientTape() as g:
```
awav:
```suggestion
        with tf.GradientTape() as tape:
```

tests/test_conditionals.py
```diff
@@ -128,3 +130,40 @@ def test_gaussian_whiten(Xdata, Xnew, kernel, mu, sqrt):
...
+            _, Fstar_var = conditional(Xnew,
+                                       Xdata,
+                                       kernel,
+                                       mu,
+                                       q_sqrt=q_sqrt,
+                                       white=white)
+
+        grad = g.gradient(Fstar_var, q_sqrt.unconstrained_variable)
```
awav:
```suggestion
        grad = tape.gradient(Fstar_var, q_sqrt.unconstrained_variable)
```

tests/test_conditionals.py
```diff
@@ -128,3 +130,40 @@ def test_gaussian_whiten(Xdata, Xnew, kernel, mu, sqrt):
 
     assert_allclose(mean_diff, 0, atol=4)
     assert_allclose(var_diff, 0, atol=4)
+
+@pytest.mark.parametrize('white', [True, False])
+def test_q_sqrt_constraints(Xdata, Xnew, kernel, mu, white):
```
awav: What's about the previous code, would this test pass?

tests/test_kldiv.py
```diff
@@ -150,3 +177,37 @@ def test_unknown_size_inputs():
...
+    q_sqrt_constrained = Parameter(tril, transform=triangular())
+    q_sqrt_unconstrained = Parameter(tril)
+
+    diff_before_gradient_step = (q_sqrt_constrained - q_sqrt_unconstrained).numpy()
+    assert_allclose(diff_before_gradient_step, 0)
+
+    kls = []
+    for q_sqrt in [q_sqrt_constrained, q_sqrt_unconstrained]:
+
+        with tf.GradientTape() as g:
```
awav:
```suggestion
        with tf.GradientTape() as tape:
```

tests/test_kldiv.py
```diff
@@ -150,3 +177,37 @@ def test_unknown_size_inputs():
...
+            kl = prior_kl(inducing_points, kernel, mu, q_sqrt, whiten=white)
+
+        grad = g.gradient(kl, q_sqrt.unconstrained_variable)
```
awav:
```suggestion
        grad = tape.gradient(kl, q_sqrt.unconstrained_variable)
```

awav:
@condnsdmatters, I left some comments for you ;)

gpflow/conditionals/util.py
```diff
@@ -82,7 +82,7 @@ def base_conditional(Kmn: tf.Tensor,
         if q_sqrt_dims == 2:
             LTA = A * tf.expand_dims(tf.transpose(q_sqrt), 2)  # [R, M, N]
         elif q_sqrt_dims == 3:
-            L = q_sqrt
+            L =  tf.linalg.band_part(q_sqrt, -1, 0)  # force lower triangle # [R, M, M]
```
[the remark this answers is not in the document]
st--: Yes, using proper distributions and passing a q(u) to conditionals is definitely something that we should move towards to :)

tests/test_kldiv.py
st--: (it may not be much difference for any individual test, but with lots of tests even milliseconds can start to matter)

st--:
@condnsdmatters does this address #846 ?

tests/test_conditionals.py
condnsdmatters: no

tests/test_kldiv.py
condnsdmatters: pytests fixtures have the advantage of only running once then making the most of pytests in built caching. In general we should move to fixtures, though I also want to look at different ways you can sort of parameterize fixtures. (ie run with multiple diff fixtures)

Issue #1202, opened 2020-01-13
Question about RobustMax


anh-tong:
Hi, 

Can you explain the intuition of making epsilon to be `Parameter`. This quantity is not trainable. Why don't we make it a constant tensor?

https://github.com/GPflow/GPflow/blob/b17df8b62ccf00713622464f5b2b28f9ba326740/gpflow/likelihoods/robustmax.py#L27

Thanks!

jameshensman:
Hi @anh-tong , 

There's no reason why epsilon shouldn't be a trainable parameter. It represents the fraction of 'errors' in the labels of the dataset. I think it would be a *hard* parameter to optimize, and for that reason I think most people simply set it to a small value. 

The Robustmax likelihood is attractive because it has an analytic form for `variational_expectations`, which means no mean for Monte Carlo and so no (additional) variance in the elbo. But for most problems, I'd actually recommend the softmax form: this doesn't have tractable `variational_expectations`, so we have to use Monte Carlo, which introduces variance into the elbo, but empirically I've found the resulting elbo to be much more straightforward to optimize, and the variance is small compared to that introduced by minibatching (if you're using SVGP). 

Hope that helps.

st--:
@anh-tong if you want to re-state your question on [stackoverflow.com with the gpflow tag](stackoverflow.com/tags/gpflow) we could cross-post the answer there which will make it much more helpful and accessible to other users in the future.

anh-tong:
Thanks for your detailed answer.

PR #1228, opened 2020-01-24
Improve explanation of RobustMax likelihood

st--:
Saves @jameshensman's answer from #1202 in code so we won't have to answer similar ones again...

### DIFF
--- a/gpflow/likelihoods/robustmax.py
+++ b/gpflow/likelihoods/robustmax.py
@@ -24,7 +24,7 @@ class RobustMax(tf.Module):
 
     def __init__(self, num_classes, epsilon=1e-3, **kwargs):
         """
-        epsilon represents the fraction of 'errors' in the labels of the
+        `epsilon` represents the fraction of 'errors' in the labels of the
         dataset. This may be a hard parameter to optimize, so by default
         it is set un-trainable, at a small value.
         """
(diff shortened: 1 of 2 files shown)


gpflow/likelihoods/robustmax.py
```diff
@@ -16,11 +16,18 @@ class RobustMax(tf.Module):
...
-          eps/(k-1)  otherwise.
+    y_i = (1-epsilon)  i == argmax(f)
+          epsilon/(k-1)  otherwise
+
+    where k is the number of classes.
     """
 
     def __init__(self, num_classes, epsilon=1e-3, **kwargs):
+        """
+        epsilon represents the fraction of 'errors' in the labels of the
```
awav:
```suggestion
        The ``epsilon`` represents the fraction of 'errors' in the labels of the
```

gpflow/likelihoods/robustmax.py
```diff
@@ -16,11 +16,18 @@ class RobustMax(tf.Module):
 
     with
 
-    y_i = (1-eps)  i == argmax(f)
-          eps/(k-1)  otherwise.
+    y_i = (1-epsilon)  i == argmax(f)
```
awav: I'm surprised that you didn't use unicode here. You can also add to the docs that Ɛ is ``epsilon`` property.

```suggestion
    y_i = (1 - Ɛ)  i == argmax(f)
```

st--: I didn't use unicode because the parameter is also called `epsilon` not `Ɛ` (I'd be open to a discussion about the pros and cons of unicode-variable names in code though :wink: )

PR #1231, opened 2020-01-25
Improve documentation and behaviour of Periodic kernel

### LINKED ISSUE #1230: Syntax for setting active_dims on Periodic kernel is unclear
It is unclear from the documentation what the correct syntax is for creating a periodic Matern kernel, say:

k1 = gpflow.kernels.Matern12(1, active_dims = [0])
k = gpflow.kernels.Periodic(base = k1)

or 

k1 = gpflow.kernels.Matern12(1, active_dims = [0])
k2 = gpflow.kernels.Periodic(1, active_dims = [0])
k = k1 * k2

st--:
Addresses unclear/insufficient documentation (closes #1230).

Previously, `Periodic` did not take `active_dims`, and only the first code example in #1230 would have worked - but the `Kernel` base class specifies `active_dims` as part of the interface, so this PR adds getter/setter that pass through to the `base` kernel, as well as accepting `active_dims` in the `Periodic()` constructor but raising an error when it is not consistent with `base.active_dims`.

### DIFF
--- a/gpflow/kernels/base.py
+++ b/gpflow/kernels/base.py
@@ -40,9 +40,15 @@ class Kernel(tf.Module, metaclass=abc.ABCMeta):
         :param name: optional kernel name.
         """
         super().__init__(name=name)
-        if isinstance(active_dims, list):
-            active_dims = np.array(active_dims)
-        self._active_dims = active_dims
+        self._active_dims = self._normalize_active_dims(active_dims)
+
+    @staticmethod
+    def _normalize_active_dims(value):
+        if value is None:
+            value = slice(None, None, None)
+        if not isinstance(value, slice):
+            value = np.array(value, dtype=int)
+        return value
 
     @property
     def active_dims(self):
@@ -50,11 +56,7 @@ class Kernel(tf.Module, metaclass=abc.ABCMeta):
 
     @active_dims.setter
     def active_dims(self, value):
-        if value is None:
-            value = slice(None, None, None)
-        if not isinstance(value, slice):
-            value = np.array(value, dtype=int)
-        self._active_dims = value
+        self._active_dims = self._normalize_active_dims(value)
 
     def on_separate_dims(self, other):
         """
--- a/gpflow/kernels/base.py
+++ b/gpflow/kernels/base.py
@@ -35,7 +35,8 @@ class Kernel(tf.Module, metaclass=abc.ABCMeta):
                  active_dims: Optional[Union[slice, list]] = None,
                  name: Optional[str] = None):
         """
-        :param active_dims: active dimensions, has the slice type.
+        :param active_dims: active dimensions, either a slice or list of
+            indices into the columns of X.
         :param name: optional kernel name.
         """
         super().__init__(name=name)
--- a/gpflow/kernels/periodic.py
+++ b/gpflow/kernels/periodic.py
@@ -9,6 +9,9 @@ from .base import Kernel
 from .stationaries import Stationary
 
 
+_ACTIVE_DIMS_ON_BASE_KERNEL = 'active_dims should be specified through the base kernel'
+
+
 class Periodic(Kernel):
     """
     The periodic family of kernels. Can be used to wrap any Stationary kernel
@@ -35,7 +38,8 @@ class Periodic(Kernel):
     is absorbed into the lengthscale hyperparameter).
     """
 
-    def __init__(self, base: Stationary, period: Union[float, List[float]] = 1.0):
+    def __init__(self, base: Stationary, period: Union[float, List[float]] = 1.0,
+                 active_dims=_ACTIVE_DIMS_ON_BASE_KERNEL):
         """
         :param base: the base kernel to make periodic; must inherit from Stationary
             Note that active_dims should be specified in the base kernel.
@@ -45,6 +49,10 @@ class Periodic(Kernel):
         """
         if not isinstance(base, Stationary):
             raise TypeError("Periodic requires a Stationary kernel as the `base`")
+
+        if active_dims != _ACTIVE_DIMS_ON_BASE_KERNEL:
+            if active_dims != base.active_dims:
+                raise ValueError("active_dims must be consistent with base.active_dims")
 
         super().__init__()
         self.base = base
--- a/gpflow/kernels/periodic.py
+++ b/gpflow/kernels/periodic.py
@@ -77,6 +77,7 @@ class Periodic(Kernel):
             X2 = X
 
         # Introduce dummy dimension so we can use broadcasting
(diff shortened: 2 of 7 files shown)


st--:
The reason the tests fail is because of numeric errors in constructing kernel matrices. By normalizing `active_dims` from `None` to `slice(None, None, None)`, there are now cases where `X2` is `X`, not `None`, and this uses a different branch of code in `gpflow.utilities.ops.square_dist()`.

gpflow/kernels/base.py
```diff
@@ -36,25 +36,28 @@ def __init__(self,
...
         :param name: optional kernel name.
         """
         super().__init__(name=name)
-        if isinstance(active_dims, list):
-            active_dims = np.array(active_dims)
-        self._active_dims = active_dims
+        self._active_dims = self._normalize_active_dims(active_dims)
+
+    @staticmethod
+    def _normalize_active_dims(value):
```
awav: From the name is not obvious what the function does.

gpflow/kernels/base.py
```diff
@@ -36,25 +36,28 @@ def __init__(self,
...
+        if value is None:
+            value = slice(None, None, None)
+        if not isinstance(value, slice):
+            value = np.array(value, dtype=int)
```
awav:
```suggestion
            value = tf.constant(value, dtype=tf.int32)
```

gpflow/kernels/periodic.py
```diff
@@ -9,6 +9,9 @@
 from .stationaries import Stationary
 
 
+_ACTIVE_DIMS_ON_BASE_KERNEL = 'active_dims should be specified through the base kernel'
```
awav: What is that? And why is that a default value for active_dims?
actives_dims is either None, a slice or an array, but not a string 100% :)

gpflow/kernels/periodic.py
```diff
@@ -32,35 +35,49 @@ class Periodic(Kernel):
     γ is the period parameter.
 
     (note that usually we have a factor of 4 instead of 0.5 in front but this
-    is absorbed into lengthscale hyperparameter).
+    is absorbed into the lengthscale hyperparameter).
     """
 
-    def __init__(self, base: Stationary, period: Union[float, List[float]] = 1.0):
+    def __init__(self, base: Stationary, period: Union[float, List[float]] = 1.0,
```
jameshensman: I'm not sure why we need to handle the active dims argument here. why can't we just not _have_ the active dims argument, but respect the one in the base kernel?

gpflow/kernels/base.py
st--: Suggestions for a better name welcome. Is "canonize" better? How do you call the thing that takes a bunch of inputs and makes sure regardless of what it started out with it's now a more limited form so that we don't have to special-case everywhere?

gpflow/kernels/base.py
st--: What's the advantage of having it be a tensor? Why had you made it a numpy array before?

gpflow/kernels/periodic.py
st--: That's how it was before. But it confused users (https://github.com/GPflow/GPflow/issues/1230), and the `Kernel` base-class defines active dims as part of the kernel interface - and all child classes should have a consistent interface. This makes it easier for users, and they get a clear error message when they try and set active_dims to something inconsistent with base.active_dims.

gpflow/kernels/periodic.py
st--: It's a sentinel value to detect whether someone passes an `active_dims` argument to the `Periodic` constructor. And makes it obvious when looking at e.g. `Periodic?` in IPython that you shouldn't actually pass anything in there. The issue is that currently `None` doesn't mean "active_dims hasn't been passed in" but "apply to all dimensions", so passing `active_dims=None` to the periodic kernel when the base kernel has active_dims set to something else should throw an error.

awav: The sentinel value cannot violate type restrictions, the active dim is an **optional** argument, and can be only a **slice** or an **array**. Here you can say that periodic class accepts only none value and doesn't tolerate anything else; state it in the docstring and raise an exception in case of when it is not none.

awav: I don't understand the argument with `IPython`.

gpflow/kernels/base.py
awav: The same reason as to why `tf.function` should receive only tensors - it is a safety precaution from graph regeneration.

st--: It's kinda weird though that the `Kernel` constructor specifies numpy array, and then it gets turned into a tensor...?

gpflow/kernels/periodic.py
st--:
```
In [2]: gpflow.kernels.Periodic?                                                                                
Init signature:
gpflow.kernels.Periodic(
    base: gpflow.kernels.stationaries.Stationary,
    period: Union[float, List[float]] = 1.0,
    active_dims='active_dims should be specified through the base kernel',
)
```

st--: Alternatively we can use **kwargs and check whether they contain `"active_dims"` (and then check consistency with `base.active_dims`)?

st--: > Here you can say that periodic class accepts only none value and doesn't tolerate anything else; state it in the docstring and raise an exception in case of when it is not none.

My point was that because `None` stands for "all input dimensions" we *cannot* use it to indicate that this is the only accepted value.

awav: Let's not allow `IPython` to rule our coding standards. The string is unacceptable. It should accept only `None` value. What we need to do: 1. change type annotation to emphasise it, 2. make docstring clear about it. 

Sorry, but IPython has nothing to do with it.

gpflow/kernels/base.py
awav: No, it is not weird, it is internal *only* valid representation. When you pass ndarray to the parameter, you don't mind for it being converted to tf.variable ;)

gpflow/kernels/base.py
awav: Let me think...

gpflow/kernels/base.py
st--: True. Was there any reason behind your choice to originally make it a np.array()? (Just trying to clarify/understand because you're asking me to change a line of code that is actually unrelated to the rest of my PR's content, and it's one that you authored in the first place.)

gpflow/kernels/periodic.py
st--: changed to **kwargs and corresponding check - happy?

PR #1233, opened 2020-01-28
Improve pretty-printing of GPflow objects in IPython shell and jupyter notebook

st--:
Makes all GPflow objects (kernels, likelihoods, ...) inherit from `gpflow.base.Module` (not `tf.Module`), and adds `_repr_html_` and `_repr_pretty_` methods to `gpflow.base.Module` that return appropriate rich representation for jupyter notebook and IPython shell, respectively.

### DIFF
--- a/gpflow/base.py
+++ b/gpflow/base.py
@@ -32,6 +32,15 @@ class Module(tf.Module):
     def trainable_parameters(self):
         return tuple(self._flatten(predicate=_IS_TRAINABLE_PARAMETER))
 
+    def _repr_html_(self):
+        from IPython.display import HTML
+        from .utilities import tabulate_module_summary
+        return HTML(tabulate_module_summary(self, tablefmt='html'))
+
+    def _repr_pretty_(self, p, cycle):
+        from .utilities import tabulate_module_summary
+        p.text(tabulate_module_summary(self, tablefmt=''))
+
 
 class Parameter(tf.Module):
     def __init__(self,
--- a/.coveragerc
+++ b/.coveragerc
@@ -5,7 +5,7 @@ exclude_lines =
     def __repr__
     def __str__
     def _repr_html_
-    def _html_table_rows
+    def _repr_pretty_
     if self.debug:
     if settings.DEBUG
     raise AssertionError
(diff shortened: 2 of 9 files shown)


gpflow/base.py
```diff
@@ -32,6 +32,15 @@ def parameters(self):
     def trainable_parameters(self):
         return tuple(self._flatten(predicate=_IS_TRAINABLE_PARAMETER))
 
+    def _repr_html_(self):
```
awav: What if we would make just `gpflow.print_summary()` alias function instead of `gpflow.utilities.print_summary()`. People do `import gpflow` anyway, the call will be shorter.

awav: @jameshensman ^^^

jameshensman: That doesn't solve the issue that this PR addresses, @awav, which is that GPflow objects hould play nice in jupyter environments

awav: @jameshensman, it does work for notebooks as well, you can set the format output, globally and in the function itself.

awav: @jameshensman, @st-- the PR duplicated what implemented here https://github.com/GPflow/GPflow/blob/develop/gpflow/utilities/utilities.py#L132. It also introduces a circular module dependency between utilities and base classes.

awav:
https://github.com/GPflow/GPflow/pull/1233#discussion_r371738132

gpflow/base.py
st--: print_summary() doesn't address that in IPython or a jupyter notebook, _evaluating an object_ (i.e. just `model` at the end of a cell) should play nicely.

Circular dependency: That's why the imports are only _inside_ the `_repr_html_` and `_repr_pretty_` methods.

Code duplication: We could remove `print_summary()` entirely, add IPython as a dependency, and tell people to use `IPython.display()` instead. `print_summary()` might still be valuable if there is a need for plotting Parameters inside something that subclasses `tf.Module` directly, not `gpflow.base.Module`.

(Should we expose `gpflow.base.Module` in the `gpflow` namespace directly - i.e. `gpflow.Module`?)

PR #1235, opened 2020-01-29
Clean up `presliced` argument to kernel calls

st--:
In GPflow 1, each kernel's `K` and `K_diag` method was supposed to take a `presliced` argument. Not all kernels actually had this signature, and not all kernels who had this signature actually did anything with the argument. This PR makes `presliced` an argument solely to the kernel's `__call__`, with the intent of making the entire set-up a lot cleaner.

### DIFF
--- a/doc/source/notebooks/tailor/gp_nn.ipynb
+++ b/doc/source/notebooks/tailor/gp_nn.ipynb
@@ -120,14 +120,14 @@
     "            \n",
     "            self.cnn.build()\n",
     "    \n",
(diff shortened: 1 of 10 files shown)


st--:
Should fix test issue in #1231

jameshensman:
I can't see any disadvantages of doing it this way, @st-- , and it does make the code _so_much cleaner. Really nice. 

Are we sure we've got tests got sums/products of kernels with different active_dims? That's the most likely thing to be affected.

PR #1236, opened 2020-01-29
kernel call: replace `not full` with `diag`

st--:
This PR defines a Kernel's `__call__` method to use `diag=False` as the default (compute kernel matrix), and `diag=True` can be passed in when requiring only the diagonal of the kernel matrix:
* `kernel(X)`- full kernel matrix cov(f(X), f(X))
* `kernel(X, X2)` - full kernel matrix cov(f(X), f(X2))
* `kernel(X, diag=True)` - diagonal part of the kernel matrix
* `kernel(X, X2, diag=True)`: not supported, raises an exception with error message

This makes the notation more consistent with the use of "diag" in linear algebra code, e.g. tf.linalg.diag_part to obtain the diagonal of a matrix in TensorFlow". `kernel(X, diag=True)` gives the same answer as `tf.linalg.diag_part(kernel(X))`, but without computation of the off-diagonal elements.


st--:
We decided to close this PR in favour of renaming the `full` argument to `Kernel.__call__()` to `full_cov` (consistent with `conditional()`); it will also become a keyword-only argument.

PR #1237, opened 2020-01-29
Reorder inducing_variables and covariances code

st--:
# Reorder inducing variables and covariances

As someone who works quite closely with the GPflow internals, including working on conditionals and implementing new types of inducing variables, I find it very jarring that the definitions of `Kuu` and `Kuf` for some InducingVariable subclass are far away from the class definition - as they are "kinda" like class methods (not quite, because we also need to dispatch on the kernel type), I think it is much more logical to have the covariance implementation next to the type definition. This PR splits up the inducing variable definitions into one file per class (one file for all multi-output inducing variables), and `covariances.py` is now just a single file which contains the Dispatchers.


awav:
@condnsdmatters, @st-- Let's discuss this PR and #1236 Thursday this week.

condnsdmatters:
> @condnsdmatters, @st-- Let's discuss this PR and #1236 Thursday this week.

Sure!

st--:
closing for now

PR #1238, opened 2020-02-02
Clear chain bijectors

awav:
Hello all, chain of bijectors was not properly handled in deepcopy method. The PR fixes it. Thanks!


gpflow/utilities/utilities.py
```diff
@@ -228,12 +228,21 @@ def reset_cache_bijectors(input_module: tf.Module) -> tf.Module:
     target_types = (tfp.bijectors.Bijector, )
     accumulator = ('', None)
 
+    def clear_cache(b: tfp.bijectors.Bijector):
+        if isinstance(b, tfp.bijectors.Bijector):
```
st--: This is a bit confusing - the signature's type annotation does not make sense together with the isinstance check: if the isinstance is needed, then the types this function takes must be more than just tfp.bijectors.Bijector ... could you clarify?

awav:
It does not change the output of bijector's actions (forward, inverse, logdet). I don't think it is a reason for changing the name.
It is in docstring https://github.com/GPflow/GPflow/pull/1238/files#diff-03f230066686ef8ee54c56b03a33b4e3R223.

gpflow/utilities/utilities.py
```diff
@@ -228,12 +228,21 @@ def reset_cache_bijectors(input_module: tf.Module) -> tf.Module:
     target_types = (tfp.bijectors.Bijector, )
     accumulator = ('', None)
 
+    def clear_cache(b: tfp.bijectors.Bijector):
```
```suggestion
    def clear_cache(b):
```

gpflow/utilities/utilities.py
awav: It can be any type

gpflow/utilities/utilities.py
```diff
@@ -228,12 +228,21 @@ def reset_cache_bijectors(input_module: tf.Module) -> tf.Module:
...
-        bijector._from_x.clear()
-        bijector._from_y.clear()
+        clear_cache(bijector)
+        if isinstance(bijector, tfp.bijectors.Chain):
+            for m in bijector.submodules:
+                clear_cache(m)
         return state
 
-    _ = traverse_module(input_module, accumulator, clear_bijector, target_types)
+    traverse_module(input_module, accumulator, clear_bijector, target_types)
```
st--: Is it not clearer to show that you are intentionally not making use of `traverse_module`'s return value?

PR #1242, opened 2020-02-03
quickfix kernel_design notebook

st--:
#1235 removed the `presliced` argument from a Kernel's `K` and `K_diag` methods, but hadn't amended the custom `Brownian` kernel in the kernel_design notebook: fixed by this PR.


st--:
I have run `make` to ensure the .ipynb is synced with the .pct.py.

PR #1243, opened 2020-02-03
fix link function bug in binary classification notebook

### LINKED ISSUE #1172: binary classification notebook assumes wrong link function
Bernoulli() by default uses probit link function, but basic classification notebook assumes it is logit, which means that the p samples are actually wrong.

st--:
Fixes #1172 - improves explanation of link function for Bernoulli in binary classification notebook and fixes random seeds to make figures reproducible.


st--:
Is my argument for the reason behind gpflow's Bernoulli likelihood using the inv-probit as its link function (`p = (1 + erf(f / √2))/2`) instead of inv-logistic (`exp(f)/(1+exp(f))`) correct?

PR #1247, opened 2020-02-05
fix Parameter.__repr__ for tf.function mode

st--:
`Parameter.__repr__` called `.numpy()` on the constrained and unconstrained representations - this throws an error when in graph mode (tf.function-wrapped), which can lead to errors-in-exception-handling, so this PR makes use of `tf.eagerly_executing()` to check whether we are in graph mode, and if so, only print shapes and dtype, not values.

### DIFF
--- a/gpflow/base.py
+++ b/gpflow/base.py
@@ -206,13 +206,19 @@ class Parameter(tf.Module):
     def __repr__(self):
         unconstrained = self.unconstrained_variable
         constrained = self.read_value()
-        info = f"dtype={self.dtype.name} " \
-               f"unconstrained-shape={unconstrained.shape} " \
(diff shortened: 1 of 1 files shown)


elvijs:
Thanks for doing this! It will make our stack traces shorter and more readable

st--:
I tried coming up with a test, but you can't define a Parameter inside a tf.function()-wrapped function, and once execution leaves that function it will be back to eager mode anyways - the only way repr() gets triggered is in certain circumstances when an exception is thrown and somewhere in the error handling the value is included. But I couldn't figure out how to actually reproduce that... @elvijs if you come up with one let's open a new PR to add a test.

Issue #1039, opened 2019-08-29
Sphinx documentation does not include `__init__` docstring


st--:
Particularly evident for the Coregion kernel: https://gpflow.readthedocs.io/en/develop/gpflow/kernels/index.html#gpflow-kernels-coregion does not contain anything, even though the code is well-documented.

joelberkeley-pio:
worth mentioning here that most of the `Coregion.__init__` docstring looks like it should be in the class docstring not the `__init__` docstring

PR #1253, opened 2020-02-11
include special methods in docstrings

joelberkeley-pio:
# Standard GPflow pull request

## PR content:

* Include special methods in docs
* Implements #1039. With this change, all special methods _and_ functions, not only `__init__` will be included in the docs
* For an MWE, build the docs and look at the Coregion kernel docs

### DIFF
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -18,6 +18,7 @@
 #
 import os
 import sys
+import types
 sys.path.insert(0, os.path.abspath('./ext'))
 
 # on_rtd is whether we are on readthedocs.org, this line of code grabbed from docs.readthedocs.org
@@ -368,3 +369,19 @@ texinfo_documents = [
 # If true, do not generate a @detailmenu in the "Top" node's menu.
 #
 # texinfo_no_detailmenu = False
+
+
+def setup(app):
+    """ Entry point to sphinx build customisation. """
+    app.connect("autodoc-skip-member", autodoc_skip_member_callback)
+
+
+def autodoc_skip_member_callback(app, what, name, obj, skip, options):
+    """
+    Only skip special methods and functions, including `__init__`, if they have no docstring.
+    """
+    if isinstance(obj, (types.FunctionType, types.MethodType)):
+        if getattr(obj, "__doc__", None) is not None:
+            return False  # never skip methods containing a docstring
+
+    return skip


doc/source/conf.py
```diff
@@ -368,3 +368,19 @@
 # If true, do not generate a @detailmenu in the "Top" node's menu.
 #
 # texinfo_no_detailmenu = False
+
+
+def setup(app):
+    """ Entry point to sphinx build customisation. """
+    app.connect('autodoc-skip-member', special_methods_callback)
```
st--:
```suggestion
    app.connect('autodoc-skip-member', autodoc_skip_member_callback)
```

doc/source/conf.py
```diff
@@ -368,3 +368,19 @@
...
+
+
+def special_methods_callback(app, what, name, obj, skip, options):
```
st--:
```suggestion
def autodoc_skip_member_callback(app, what, name, obj, skip, options):
```

doc/source/conf.py
```diff
@@ -368,3 +368,19 @@
...
+    """
+    Include all method and function docstrings in the docs if they have a custom docstring. This
+    overrides the default behaviour of omitting special methods or functions from the docs.
+    """
+    if hasattr(obj, '__doc__') and isinstance(obj, (types.FunctionType, types.MethodType)):
+        return False
```
st--:
```suggestion
        return False  # never skip when a function/method has a docstring
```

joelberkeley-pio: i'm averse to comments - do you not feel the docstring covers it?

doc/source/conf.py
joelberkeley-pio: done

doc/source/conf.py
st--: I added the comment because it took me a bit to understand what exactly is going on here - I could figure it out by myself, but wanted to save future-me and other people having to go through that extra effort

st--:
We should check that this change won't pull in all the default docstrings such as `__repr__` -> `"Return repr(self)."` ...

doc/source/conf.py
joelberkeley-pio: i've edited the docstring - is that clearer?

joelberkeley-pio:
> We should check that this change won't pull in all the default docstrings such as `__repr__` -> `"Return repr(self)."` ...

`Parameter.__repr__` doesn't appear. Is that sufficient?

doc/source/conf.py
```diff
@@ -368,3 +369,18 @@
...
+    """ Entry point to sphinx build customisation. """
+    app.connect("autodoc-skip-member", autodoc_skip_member_callback)
+
+
+def autodoc_skip_member_callback(app, what, name, obj, skip, options):
+    """
+    Only skip special methods and functions, including `__init__`, if they have no docstring.
+    """
+    if isinstance(obj, (types.FunctionType, types.MethodType)):
+        return getattr(obj, "__doc__", None) is None
```
st--:
```suggestion
        if getattr(obj, "__doc__", None) is not None:
            return False  # never skip methods with docstring
```
That

joelberkeley-pio: alright, alright, i'll add the comment - though never forget how diametrically opposed to the entire concept of comments I am and forever will be

joelberkeley-pio: also good catch on the logic error there

PR #1255, opened 2020-02-13
Improve documentation and behaviour of Coregion kernel

st--:
Improves documentation and behaviour of the `Coregion` kernel:
- clean up of the documentation
- changes initialization from W=zeros() to W=0.1*ones() which makes it easier to use out of the box - no longer requires manual assigning of symmetry-breaking values to the W matrix after kernel construction
- adds output_covariance() method that computes the covariance between outputs (and output_variance() for its diagonal)


st--:
It seems a bit silly to initialize `W` to all-zeros if that's the worst possible initialization: would it not be better to default-initialize it to something that already breaks the symmetry? @jameshensman

gpflow/kernels/misc.py
```diff
@@ -98,37 +98,42 @@ def K_diag(self, X):
...
 
-        NB. There is a symmetry between the elements of W, which creates a
-        local minimum at W=0. To avoid this, it's recommended to initialize the
-        optimization (or MCMC chain) using a random W.
+    NB. There is a symmetry between the elements of W, which creates a local
+    minimum at W=0. To avoid this, it is recommended to initialize the
+    optimization (or MCMC chain) using a random W.
+    """
+
+    def __init__(self, output_dim: int, rank: int, **kwargs):
```
joelberkeley-pio: any reason you've opted for kwargs here?

gpflow/kernels/misc.py
```diff
@@ -98,37 +98,42 @@ def K_diag(self, X):
...
-        (we cast them from floats as needed) which usually specify the
-        *outputs* of a Coregionalization model.
+    """
+    A Coregionalization kernel. The inputs to this kernel are _integers_ (we
+    cast them from floats as needed) which usually specify the *outputs* of a
+    Coregionalization model.
 
-        The parameters of this kernel, W, kappa, specify a positive-definite
-        matrix B.
+    The kernel function is an indexing of a positive-definite matrix, so
```
joelberkeley-pio:
```suggestion
    The kernel function is an indexing of a positive-definite matrix:
```

gpflow/kernels/misc.py
```diff
@@ -139,10 +144,19 @@ def K(self, X, X2=None):
             X2 = X
         else:
             X2 = tf.cast(X2[:, 0], tf.int32)
+
+        shape_constraints = [
+            (X, [..., 'N', 1]),
+            (X2, [..., 'M', 1]),
+        ]
+        tf.debugging.assert_shapes(shape_constraints)
```
joelberkeley-pio: I prefer having argument validation as the very first thing in a function, partly so that feedback can be given on the exact data the caller provided, and partly as a good practice of catching errors early

gpflow/kernels/misc.py
st--: Because it hadn't passed through `name` before

gpflow/kernels/misc.py
st--: Won't work because X2 may be None though

gpflow/kernels/misc.py
joelberkeley-pio: i think it's almost always the case that kwargs can and should be avoided. I realise it's used frequently in GPflow, but I strongly encourage moving away from it. I won't block the PR on it

tests/test_coregion.py
```diff
@@ -62,6 +62,7 @@ def _prepare_models():
     kc.lengthscale.trainable = False
     kc.variance.trainable = False  # variance is handles by the coregion kernel
     coreg = gpflow.kernels.Coregion(output_dim=2, rank=1, active_dims=[2])
+    coreg.W.assign(np.zeros((2, 1)))  # zero correlation between outputs
```
joelberkeley-pio: why do we want W to be zero here?

gpflow/kernels/misc.py
```diff
@@ -98,51 +98,66 @@ def K_diag(self, X):
...
+    number of outputs in a coregionalization model. We refer to the number of
+    columns on W as 'rank': it is the number of degrees of correlation between
+    the outputs.
 
-        NB. There is a symmetry between the elements of W, which creates a
-        local minimum at W=0. To avoid this, it's recommended to initialize the
-        optimization (or MCMC chain) using a random W.
+    NB. There is a symmetry between the elements of W, which creates a local
+    minimum at W=0. To avoid this, it is recommended to initialize the
+    optimization (or MCMC chain) using a random W.
```
joelberkeley-pio: it would be nice to have this clarified, but not a blocker

gpflow/kernels/misc.py
```diff
@@ -98,51 +98,66 @@ def K_diag(self, X):
...
         """
 
         # assert input_dim == 1, "Coregion kernel in 1D only"
-        super().__init__(active_dims)
+        super().__init__(**kwargs)
 
         self.output_dim = output_dim
         self.rank = rank
-        W = np.zeros((self.output_dim, self.rank))
+        W = np.ones((self.output_dim, self.rank))
```
joelberkeley-pio: as I said before, I don't know the implications of randn vs ones vs sth else - you'll need to ask sbd else about this

gpflow/kernels/misc.py
```diff
@@ -98,51 +98,73 @@ def K_diag(self, X):
...
 
         self.output_dim = output_dim
         self.rank = rank
-        W = np.zeros((self.output_dim, self.rank))
+        W = 0.1 * np.ones((self.output_dim, self.rank))
         kappa = np.ones(self.output_dim)
         self.W = Parameter(W)
         self.kappa = Parameter(kappa, transform=positive())
 
+    def compute_B(self):
```
joelberkeley-pio: can this be private?

gpflow/kernels/misc.py
```diff
@@ -98,51 +98,73 @@ def K_diag(self, X):
...
+        B = tf.linalg.matmul(self.W, self.W, transpose_b=True) + tf.linalg.diag(self.kappa)
+        return B
+
+    def compute_B_diag(self):
```
joelberkeley-pio: can this be private?

doc/source/notebooks/advanced/coregionalisation.pct.py
```diff
@@ -159,7 +160,7 @@ def plot(m):
 # The covariance matrix between outputs is as follows:
 
 # %%
-B = coreg.W.numpy() @ coreg.W.numpy().T + np.diag(coreg.kappa.numpy())
+B = coreg.compute_B().numpy()
```
joelberkeley-pio: i'm wary of exposing internals for demonstrations or debugging. it increases the change of breaking changes in the API

doc/source/notebooks/advanced/coregionalisation.ipynb
```diff
@@ -64,7 +64,7 @@
...
...jsaYa+M/j68bY/4p3+foBxn8n55hjHneGPP7+P/rxkKcZzEzxvzCGHPEGPPHIZ43xpi749f4D8aY+Tk9AWttyXzh9Nt+G5gNBIF/A+amHHMzcG/89leABwt93sX2leF1vASoit/+G13H0V/D+HGTgR1AC1Bf6PMutq8MfxbPBn4PTInfP7nQ511sXxlex3XA38RvzwUOFvq8i+0LWATMB/44xPONwFOAARqAl3P5/qU2wr4A2Get3W+tjQK/ApakHLMEeCB++xHgUmOMyeM5+sGI19Fa+7y1NhK/2wKcludzLHaZ/CwC/D3wD8DxfJ6cj2RyHW8Amqy1nQDW2iN5Pkc/yOQ6WuDE+O2TgH/P4/n5grV2B3B0mEOWAButowX4mDFmWq7ev9QC9nTg3YT778UfS3uMtbYP+BCoycvZ+Ucm1zHR13A+VcqAEa9hfLrsdGvtk/k8MZ/J5GdxDjDHGPOSMabFGHN53s7OPzK5jn8H/LUx5j1gG/Bf8nNqJWW0vztHZVKuXkgmJmPMXwP1wMWFPhc/McaUAT8Gri/wqZSCSTjT4p/DmenZYYyZZ639c0HPyn+WAfdba//RGLMA2GSM+aS1tr/QJyaOUhthvw+cnnD/tPhjaY8xxkzCmfrpyMvZ+Ucm1xFjzGXA3wJXW2t78nRufjHSNZwMfBL4jTHmIM561xNKPBskk5/F94AnrLW91toDwB6cAC4DMrmOXwMeArDWNgMn4DS1kMxl9LtzrEotYL8KnG2MmWWMCeIklT2RcswTwMr47S8Cz9l4toB4RryOxpjzgPtwgrXWDAcb9hpaaz+01tZaa2daa2fi5AFcba1tLczpFq1M/k8/hjO6xhhTizNFvj+fJ+kDmVzHPwGXAhhjPoETsNvyepb+9wSwIp4t3gB8aK09lKsXL6kpcWttnzHmG8B2nKzIX1hrXzfG3Am0WmufAH6OM9WzDyd54CuFO+PilOF1/F9ANfBwPGfvT9baqwt20kUmw2soI8jwOm4H/tIYsxuIAd+x1mrWLEGG1/H/A9YbY27FSUC7XoOZZMaYLTgfDmvja/13AOUA1tp7cdb+G4F9QAT4ak7fX/8eIiIixa/UpsRFRERKkgK2iIiIDyhgi4iI+IACtoiIiA8oYIuIiPiAAraIiIgPKGCLiIj4wP8F/P8gu3wsYUMAAAAASUVORK5CYII=\n",
```
joelberkeley-pio: how do you review a file like this? (I realise this is what you're trying desperately, and quite rightly, to move away from). For now, what should I be looking at?

tests/test_coregion.py
joelberkeley-pio: i suppose what i mean is: why is the local minimum at 0 not a problem here?

st--: because here we want to compare equivalence between separate independent GPs and a coregion GP that enforces zero correlations (W=0).

gpflow/kernels/misc.py
st--: double-checked with @jameshensman

gpflow/kernels/misc.py
st--: it's actually useful for users of this kernel! e.g. see the coregion notebook.

maybe we want to call it something more descriptive ... `output_covariances()` and `output_variances()` maybe?

gpflow/kernels/misc.py
st--: see above

doc/source/notebooks/advanced/coregionalisation.pct.py
st--: yeah- that's why I changed it! it's now requiring *less* exposure to internals (e.g. it would still work if we change the parametrisation of B to something else).

doc/source/notebooks/advanced/coregionalisation.ipynb
st--: You don't - you check out the PR branch, remove the notebook .ipynb file, run `make` in the doc/source/notebooks directory, and check that it doesn't actually introduce any changes to the .ipynb file: this ensures that the .ipynb is up to date with the percent-py jupytext representation. (It requires the notebook to be deterministic, which if it's not already the case the PR author should add to an edited notebook)

gpflow/kernels/misc.py
joelberkeley-pio: i thought the kernel was the covariance. what's the covariance then? and the variance?

awav: > maybe we want to call it something more descriptive ... output_covariances() and output_variances() maybe?

Yes, the name can be improved. copute_B looks cryptic :). I'd propose singular names `output_covariance` & `output_variance`.

joelberkeley-pio: or
```python3
@property
def covariance(self) -> tf.Tensor:
   ...
```
etc

st--: as this does computation, I think it's important to not hide that behind a property. renamed to output_covariance as I suggested.

joelberkeley-pio: Everything does computation. I've been thinking about this in general for a few days now, as a more general question, and I've come to the conclusion that naming shouldn't be related to performance, because performance is typically an implementation detail and relied heavily on architecture.

As an aside, I can't as a reader tell if `output_` is a noun or verb

doc/source/notebooks/advanced/coregionalisation.ipynb
joelberkeley-pio: wow, just, wow. I feel like I've been taken back to the days of ad hoc tests that are never committed to version control

PR #1261, opened 2020-02-16
Remove masking kwargs

awav:
The PR removes kwargs from Periodic kernel. The periodic kernel uses active_dims from the base kernel and should not accept active_dims as an argument. Currently, kwargs mask active_dims argument adding less clarity to a constructor signature.


gpflow/kernels/periodic.py
```diff
@@ -31,26 +31,23 @@ class Periodic(Kernel):
     σ² is the variance parameter,
     γ is the period parameter.
 
-    (note that usually we have a factor of 4 instead of 0.5 in front but this
-    is absorbed into the lengthscale hyperparameter).
+    NOTE: that usually we have a factor of 4 instead of 0.5 in front but this
```
joelberkeley-pio:
```suggestion
    NOTE: usually we have a factor of 4 instead of 0.5 in front but this
```

gpflow/kernels/periodic.py
```diff
@@ -31,26 +31,23 @@ class Periodic(Kernel):
...
+        the constructor doesn't have it as an argument.
     """
 
-    def __init__(self, base: Stationary, period: Union[float, List[float]] = 1.0,
-                 **kwargs):
+    def __init__(self, base: Stationary, period: Union[float, List[float]] = 1.0):
         """
         :param base: the base kernel to make periodic; must inherit from Stationary
-            Note that active_dims should be specified in the base kernel.
+            Note that `active_dims` should be specified in the base kernel.
```
joelberkeley-pio: This line isn't necessary

gpflow/kernels/periodic.py
```diff
@@ -31,26 +31,23 @@ class Periodic(Kernel):
     σ² is the variance parameter,
     γ is the period parameter.
 
-    (note that usually we have a factor of 4 instead of 0.5 in front but this
-    is absorbed into the lengthscale hyperparameter).
+    NOTE: that usually we have a factor of 4 instead of 0.5 in front but this
+        is absorbed into the lengthscale hyperparameter.
+    NOTE: periodic kernel uses `active_dims` of a base kernel, therefore
+        the constructor doesn't have it as an argument.
```
joelberkeley-pio: I don't think you need this last note

gpflow/kernels/periodic.py
st--: Yes it is, users were confused (see #1230)

gpflow/kernels/periodic.py
st--: We do, users were confused (see #1230)

tests/test_kernels.py
```diff
@@ -547,22 +547,20 @@ def test_kernel_call_diag_and_X2_errors(kernel):
 def test_periodic_active_dims_mismatch_check():
     active_dims_1 = [0]
     active_dims_2 = [1]
-    with pytest.raises(ValueError):
-        base = gpflow.kernels.SquaredExponential(active_dims=active_dims_1)
+    base = gpflow.kernels.SquaredExponential(active_dims=active_dims_1)
+    with pytest.raises(TypeError):
```
joelberkeley-pio: i think you can just delete this test now

gpflow/kernels/periodic.py
awav: I kind of agree that this line is not necessary. The signature of the constructor is self-explanatory. Unfortunately, https://github.com/GPflow/GPflow/issues/1230 doesn't have many details about what was wrong with the documentation. However, I prefer to keep this mention to make it crystal clear that a periodic kernel uses active_dims from a base kernel.

joelberkeley-pio: ok. happy to leave as is

gpflow/kernels/periodic.py
joelberkeley-pio: ok. happy to leave as is

PR #1262, opened 2020-02-16
Fix environment variables handling by config

### LINKED ISSUE #1180: Config does not handle environment variables correctly
In gpflow2, `gpflow/config/__config__.py` has code for overwriting the default values for float/int dtype, jitter, etc using environment variables (instead of a gpflowrc file as in previous versions). However, this is currently broken as it does not actually convert the strings returned by os.getenv() into the required types (e.g. `"float32"` -> `np.float32` or `"0.0001"` -> `1e-4`).

awav:
Fix environment variables in `gpflow.config`. I would appreciate if somebody recommended a way of testing env variables with reloading a module.

This PR fixes #1180.


gpflow/config/__config__.py
```diff
@@ -162,7 +220,7 @@ def set_default_positive_minimum(value: float):
 
 
 def set_default_summary_fmt(value: str):
-    formats = tabulate.tabulate_formats + ['notebook', None]
+    formats = tabulate.tabulate_formats + ["notebook", None]
```
st--: Why do you change the type of string quotes? I've never spelled it out explicitly anywhere but I've always thought of `'` for "string-constants" and `"` for "message-strings" - i.e. if it's something that should be exactly this, single-quotes, and if it's something that's just an example or wouldn't change behaviour if changed, double-quotes...

gpflow/config/__config__.py
```diff
@@ -41,11 +49,59 @@ def name(self):
...
+    if bijector_type not in positive_bijector_type_map().keys():
+        raise TypeError("Config cannot set the passed value as a default positive bijector.")
+    return bijector_type
+
+
+def _default_positive_minimum_factory():
+    try:
+        value = _default(_Values.POSITIVE_MINIMUM)
+        if value is not None:
+            value = float(value)
```
st--: What happens with `None`? If that's a useful, intended case: shouldn't the empty string (`export GPFLOW_POSITIVE_MINIMUM=`) also map to `None`?

gpflow/config/__config__.py
awav: I don't change it, the black does - https://black.readthedocs.io/en/stable/the_black_code_style.html#strings. `black` became a golden standard in the python world. Generally, I prefer double quotes and I don't know the language (except python and SQL) that represents strings with single quotes. For me as C/C++ programmer, it is even more confusing, single quotes used for char type.

gpflow/config/__config__.py
awav: I think this should never happen, "GPFLOW_POSITIVE_MINIMUM=" must be a number, otherwise don't set it. There is also some interference with other values and how they behave. The empty string might mean different things depending on the context. E.g. None or just default value. I think the rule of thumb should be: _set it or get the hardcoded default value_.

gpflow/config/__config__.py
awav: > I've never spelled it out explicitly anywhere but I've always thought of ' for "string-constants"

Btw, what you are describing for single quotes is an atom type. Sadly, python doesn't have it and python can't check it at compile or run time.

gpflow/config/__config__.py
st--: but the hardcoded default value is also `None`?...

awav: `None` is a special case, exists only for switching off `shift` bijector. Maybe we need to be even more restrictive and set default value to 0 instead of `None`. That is a slightly different question.

tests/test_config.py
```diff
@@ -25,6 +26,30 @@
...
+    config = gpflow.config.Config()
+    assert config.int == np.int64
+    assert config.float == np.float32
+    assert config.positive_bijector == "exp"
+    assert config.positive_minimum == 1e-3
+    assert config.summary_fmt == "simple"
+    assert config.jitter == 1e-2
+
+
+@pytest.mark.parametrize("key", ["INT", "FLOAT", "POSITIVE_BIJECTOR", "POSITIVE_MINIMUM", "JITTER"])
```
st--: Could we test this programmatically - e.g. taking Config.__dict__.keys() or something like that, and manually exclude summary_fmt - to make sure new cases will be tested if at some point someone adds a new field to the config ?

tests/test_config.py
```diff
@@ -25,6 +26,30 @@
 from gpflow.utilities import to_default_float, to_default_int
 
 
+def test_env_variables():
+    os.environ["GPFLOW_INT"] = "int64"
+    os.environ["GPFLOW_FLOAT"] = "float32"
+    os.environ["GPFLOW_POSITIVE_BIJECTOR"] = "exp"
+    os.environ["GPFLOW_POSITIVE_MINIMUM"] = "1e-3"
+    os.environ["GPFLOW_SUMMARY_FMT"] = "simple"
+    os.environ["GPFLOW_JITTER"] = "1e-2"
```
st--: This test relies on all of these actually being different from the default values, is that correct? Would it be sensible to test for two different values each to make sure the test will be independent of external state?

gpflow/config/__config__.py
```diff
@@ -2,50 +2,106 @@
...
+        value = float(_default(_Values.JITTER))
+    except ValueError:
+        raise TypeError("Config cannot set the jitter value with non float type.")
+    return value
+
+
+def _default_positive_bijector_factory():
+    bijector_type = _default(_Values.POSITIVE_BIJECTOR)
+    if bijector_type not in positive_bijector_type_map().keys():
+        raise TypeError("Config cannot set the passed value as a default positive bijector.")
```
st--: might be helpful to include in the exception message what the read-in value was and which keys are valid ?

tests/test_config.py
awav: done

tests/test_config.py
awav: done

gpflow/config/__config__.py
awav: I don't it is necessary to match a full message. Messages are different and comparing all of them adds extra complexity to the testing, _it can be messy_. Message can change and it does not necessarily mean that we need to update tests, it doesn't change the framework's behaviour after all. _The review of the message content is much more important._

st--: Not in the test, in *here*!

tests/test_config.py
```diff
@@ -12,19 +12,60 @@
...
+@pytest.mark.parametrize("attr_name, value, expected_value", _env_values)
+def test_env_variables(attr_name, value, expected_value):
+    env_name = f"GPFLOW_{attr_name.upper()}"
+    with mock.patch.dict("os.environ", {env_name: value}):
+        assert os.environ[env_name] == value
+        config = gpflow.config.Config()
+        assert getattr(config, attr_name) == expected_value
+
+
+@pytest.mark.parametrize("attr_name", set(list(zip(*_env_values))[0][1:]))
```
st--: instead of the `[1:]` I'd recommend an explicit check (can be in the test below!) e.g.
```python
if attr_name == "summary_fmt":
    pytest.skip("cannot easily validate summary_fmt")
```

tests/test_config.py
```diff
@@ -12,19 +12,60 @@
...
+    ("positive_bijector", "softplus", "softplus"),
+    ("positive_minimum", "1e-3", 1e-3),
+    ("jitter", "1e-2", 1e-2),
+]
+
+
+@pytest.mark.parametrize("attr_name, value, expected_value", _env_values)
+def test_env_variables(attr_name, value, expected_value):
+    env_name = f"GPFLOW_{attr_name.upper()}"
+    with mock.patch.dict("os.environ", {env_name: value}):
```
st--: nice!

gpflow/config/__config__.py
st--:
```suggestion
        valid_values = list(positive_bijector_type_map.keys())
        raise TypeError(f"Config cannot set the passed value {bijector_type} as a default positive bijector. Valid values: {valid_values}")
```

awav:
@st--, I was thinking about the same. The issue is that an invalid value can be so bad that it can screw up the error message itself. E.g. it can have chars that will make the printed message unreadable.

st--:
I just realised this feature doesn't actually seem to be documented very much - should this be added to a notebook? Or just add a good module-level docstring that will get pulled into the Sphinx documentation?

awav:
> I just realised this feature doesn't actually seem to be documented very much - should this be added to a notebook? Or just add a good module-level docstring that will get pulled into the Sphinx documentation?

@st-- , it will be done here: https://github.com/GPflow/GPflow/pull/1267

PR #1264, opened 2020-02-18
Remove the use of generator in dispatcher

vdutor:
GPflow makes use of a multipledispatch Dispatcher that internally uses a generator.  However, according to TensorFlow [Capabilities and Limitations](https://github.com/tensorflow/tensorflow/blob/560e2575ecad30bedff5b192f33f6d06b19ccaeb/tensorflow/python/autograph/LIMITATIONS.md) generators are not supported by AutoGraph and probably will never be. Thus, compiling code that passed though the dispatcher led to the following warnings:
```bash
WARNING:tensorflow:Entity <bound method Dispatcher.dispatch_iter of <dispatched sample_conditional>> appears to be a generator function. It will not be converted by AutoGraph.
WARNING: Entity <bound method Dispatcher.dispatch_iter of <dispatched sample_conditional>> appears to be a generator function. It will not be converted by AutoGraph.
WARNING:tensorflow:Entity <bound method Dispatcher.dispatch_iter of <dispatched conditional>> appears to be a generator function. It will not be converted by AutoGraph.
WARNING: Entity <bound method Dispatcher.dispatch_iter of <dispatched conditional>> appears to be a generator function. It will not be converted by AutoGraph.
```

This PR still uses the same dispatcher, but overwrites the problematic method that uses python generators and replaces it by a simple list.

### NOTE
As of this PR we do not need to write `autograph=False` in `tf.function` anymore, and all the code inside dispatching gets compiled the same way as everything else :) !

PR #1265, opened 2020-02-19
Make notebooks deterministic

st--:
Adds random seed setting as required to the various notebooks under doc/source/notebooks to make sure they're as reproducible as possible (i.e., re-running shouldn't change the git diff).

There are some issues with
- tailor/gp_nn (tensor shape warnings, and bad accuracy)
- tailor/mixture_density_networks (autograph warnings despite setting autograph=False?)


awav:
@vdutor works on moving documentation to the separate repo. This is a good candidate for being a part his chunk of change.

st--:
This change came out of a discussion around #1264 and I had agreed with @vdutor to split it up into a separate PR. It seems more sensible to have smaller self-contained chunks of change rather than one bigger PR!

awav:
@vdutor, @st-- this is not a big PR, this is huge PR and mostly, as well as #1264, it affects the exact same parts of the code - notebooks. I don't see a problem with having this change there, it reduces my time on checking that output of notebooks in both cases.

Issue #1127, opened 2019-11-01
GPflow 2.0 & tf.saved_model


marcoadurno:
System information:

* Python version: 3.7
* TensorFlow installed from (source or binary): binary
* TensorFlow version (use command below): 'v2.0.0-rc2-26-g64c3d382ca', '2.0.0'
* GPflow installed from (source or binary): source
* GPflow version: develop_2.0, f070a8b

I tried to export a GPflow model using the `tf.saved_model` API by simply decorating the `predict_f` method with the `@tf.function` decorator. The export fails.

I'd expect the export to succeed.

The first obstacle seems to be the Parameter class. Even the following simple export fails:

```
import gpflow
import tensorflow as tf

x = tf.Variable(1.)
p = gpflow.base.Parameter(x)
tf.saved_model.save(p, "test")
```
Simply exporting the `tf.Variable` does work, exporting the `gpflow.base.Parameter` throws (Traceback below).

I've read the [tf.Module rfc](https://github.com/tensorflow/community/blob/master/rfcs/20190117-tf-module.md) and I gather that this is also used by GPflow [gpflow.base.Module](https://github.com/GPflow/GPflow/blob/develop-2.0/gpflow/base.py#L29) and [gpflow.base.Parameter](https://github.com/GPflow/GPflow/blob/develop-2.0/gpflow/base.py#L39). 

Is there any clear reasons why the export fails?

<pre>
Traceback (most recent call last):
  File "test.py", line 10, in <module>
    tf.saved_model.save(p, "test")
  File "C:\Users\madurno\AppData\Local\Continuum\anaconda3\envs\gpflow\lib\site-packages\tensorflow_core\python\saved_model\save.py", line 893, in save
    meta_graph_def, saveable_view, signatures, options.namespace_whitelist)
  File "C:\Users\madurno\AppData\Local\Continuum\anaconda3\envs\gpflow\lib\site-packages\tensorflow_core\python\saved_model\save.py", line 559, in _fill_meta_graph_def
    object_map, resource_map, asset_info = saveable_view.map_resources()
  File "C:\Users\madurno\AppData\Local\Continuum\anaconda3\envs\gpflow\lib\site-packages\tensorflow_core\python\saved_model\save.py", line 269, in map_resources
    new_variable = resource_variable_ops.copy_to_graph_uninitialized(obj)
[...]

PR #1266, opened 2020-02-20
Parameters to variables

marcoadurno:
Hi, 

I've added a function named `parameters_to_variables` inside `gpflow.utilities` in order to convert all `gpflow.base.Parameter` instances to `tf.Variable`. This in order to allow exporting model with `tf.saved_model` API.

This would partially solve #1127.

Working example is available [here](https://gist.github.com/marcoadurno/c53298ffe8313be22ed9bf6a19e8174a). The working example also demos the usage of `tf.function` (#1185).

Best Regards,

Marco

### DIFF
--- a/doc/source/notebooks/intro_to_gpflow2.ipynb
+++ b/doc/source/notebooks/intro_to_gpflow2.ipynb
@@ -20,11 +20,12 @@
    "outputs": [],
    "source": [
     "from typing import Tuple, Optional\n",
-    "from pathlib import Path\n",
     "\n",
     "import datetime\n",
     "import io\n",
+    "import os\n",
     "import matplotlib.pyplot as plt\n",
+    "import tempfile\n",
     "\n",
     "import numpy as np\n",
     "import tensorflow as tf\n",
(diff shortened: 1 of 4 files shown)


awav:
Hello @marcoadurno! Thanks for the PR. Strange, I was about to make a similar PR with a different implementation.

Remarks:
The `parameters_to_variables` should not change the state of the current model/module/parameter, it must create a copy of an object with parameters replaced with constants. If a user would use your implementation for him/her it would not be clear that training of this model is no longer possible.

st--:
@marcoadurno how about adding the working example you posted into the intro_to_gpflow2 notebook (remember to `pip install jupytext` first if you want to edit it in jupyter, so that it syncs the .pct.py automatically) to document this new feature?

@awav I think it's fine to replace the Parameter instances with constants as long as that's clearly reflected in the name & documentation. How about calling it "replace_parameters_with_constants"? also, would it be worth explicitly replacing with tf.constant() not tf.Variable() to make clear it's no longer trainable ?

awav:
@marcoadurno can you join [GPflow slack](https://join.slack.com/t/gpflow/shared_invite/enQtOTE5MDA0Nzg5NjA2LTYwZWI3MzhjYjNlZWI1MWExYzZjMGNhOWIwZWMzMGY0YjVkYzAyYjQ4NjgzNDUyZTgyNzcwYjAyY2QzMWRmYjE) to discuss the PR?

marcoadurno:
Hi @awav , Hi @st-- ,

I've updated the code to enforce the `clone_and_freeze` behavior suggested by @awav .

I'm happy to update the intro notebook as suggested by @st-- , and join slack as suggested by @awav , but if ok I'll do it at the weekend,

Thanks for the awesome tool you're developing.

Marco

awav:
@marcoadurno, Thanks a lot for your PR! I'm closing it in favour #1293. Your approach cannot handle indices and, unfortunately, the solution is not that trivial as in your case. You can reintroduce changes in notebooks in a separate PR or create one for explaining how to store models.

Thanks!

PR #1270, opened 2020-02-20
use natgrad optimizer in varying noise notebook

### LINKED ISSUE #1173: Put natgrads back into varying_noise.ipynb
varying_noise notebook was ported when gpflow2 did not yet have natgrads - now that they are there, this should be fixed

joelberkeley-pio:
Put natgrads back into the varying noise notebook.
Resolves #1173

### DIFF
--- a/doc/source/notebooks/advanced/varying_noise.pct.py
+++ b/doc/source/notebooks/advanced/varying_noise.pct.py
@@ -32,6 +32,7 @@ import numpy as np
 import tensorflow as tf
 import gpflow
 from gpflow.ci_utils import ci_niter
+from gpflow.optimizers import NaturalGradient
 import matplotlib
 import matplotlib.pyplot as plt
 # %matplotlib inline
@@ -114,10 +115,12 @@ model = gpflow.models.VGP((X, Y_data), kernel=kernel, likelihood=likelihood, num
 def objective_closure():
     return - model.log_marginal_likelihood()
 
-opt = gpflow.optimizers.Scipy()
-opt.minimize(objective_closure,
-             model.trainable_variables,
-             options=dict(maxiter=ci_niter(1000)))
+natgrad_opt = NaturalGradient(gamma=1.0)
+scipy_opt = gpflow.optimizers.Scipy()
+
+for _ in range(ci_niter(1000)):
+    scipy_opt.minimize(objective_closure, model.kernel.trainable_variables, options={'maxiter': 1})
+    natgrad_opt.minimize(objective_closure, [(model.q_mu, model.q_sqrt)])
 
 # %%
 # let's do some plotting!
@@ -132,7 +135,7 @@ plt.plot(xx, mu - 2*np.sqrt(var), 'C0', lw=0.5)
 
 plt.errorbar(X.squeeze(), Y.squeeze(), yerr=2*(np.sqrt(NoiseVar)).squeeze(),
              marker='x', lw=0, elinewidth=1., color='C1')
-plt.xlim(-5, 5);
+plt.xlim(-5, 5)
 
 # %% [markdown]
 # ### Questions for the reader
@@ -204,10 +207,8 @@ from gpflow.utilities import set_trainable
 set_trainable(model.kernel, False)
 set_trainable(model.likelihood, False)
 
-opt = gpflow.optimizers.Scipy()
-opt.minimize(objective_closure,
-             model.trainable_variables,
-             options=dict(maxiter=ci_niter(1000)))
+for _ in range(ci_niter(1000)):
+    natgrad_opt.minimize(objective_closure, [(model.q_mu, model.q_sqrt)])
 
 # %% [markdown]
 # We've now fitted the VGP model to the data, but without optimizing over the hyperparameters. Plotting the data, we see that the fit is not terrible, but hasn't made use of our knowledge of the varying noise. 
@@ -224,7 +225,7 @@ ax.plot(xx, mu + 2*np.sqrt(var), 'C0', lw=0.5)
 ax.plot(xx, mu - 2*np.sqrt(var), 'C0', lw=0.5)
 
 ax.plot(X, Y, 'C1x', mew=2)
-ax.set_xlim(-5, 5);
+ax.set_xlim(-5, 5)
 
 # %% [markdown]
 # ### Optimizing the noise variances
@@ -240,9 +241,9 @@ model = gpflow.models.VGP((X, Y_data), kernel=kernel, likelihood=likelihood, num
 def objective_closure():
     return - model.log_marginal_likelihood()
 
-opt.minimize(objective_closure,
-             model.trainable_variables,
-             options=dict(maxiter=ci_niter(1000)))
+for _ in range(ci_niter(1000)):
+    scipy_opt.minimize(objective_closure, model.kernel.trainable_variables, options={'maxiter': 1})
+    natgrad_opt.minimize(objective_closure, [(model.q_mu, model.q_sqrt)])
 
 # %% [markdown]
 # ### Plotting the fitted model
@@ -265,7 +266,7 @@ ax.plot(xx, mu - 2*np.sqrt(var), 'C0', lw=0.5)
 
 ax.plot(X, Y, 'C1x', mew=2)
 ax.set_xlim(-5, 5)
-ax.plot(xx, 2.5 * np.sin(6 * xx) + np.cos(3 * xx), 'C2--');
+ax.plot(xx, 2.5 * np.sin(6 * xx) + np.cos(3 * xx), 'C2--')
 
 # %%
 from gpflow.utilities import print_summary
--- a/doc/source/notebooks/advanced/varying_noise.pct.py
+++ b/doc/source/notebooks/advanced/varying_noise.pct.py
@@ -203,10 +203,6 @@ model = gpflow.models.VGP((X, Y_data), kernel=kernel, likelihood=likelihood, num
 def objective_closure():
     return - model.log_marginal_likelihood()
 
-from gpflow.utilities import set_trainable
-set_trainable(model.kernel, False)
-set_trainable(model.likelihood, False)
-
 for _ in range(ci_niter(1000)):
     natgrad_opt.minimize(objective_closure, [(model.q_mu, model.q_sqrt)])
 
--- a/doc/source/notebooks/advanced/varying_noise.pct.py
+++ b/doc/source/notebooks/advanced/varying_noise.pct.py
@@ -118,8 +118,9 @@ def objective_closure():
 natgrad_opt = NaturalGradient(gamma=1.0)
 scipy_opt = gpflow.optimizers.Scipy()
 
-model.q_mu.set_trainable(False)
-model.q_sqrt.set_trainable(False)
+from gpflow.utilities import set_trainable
+set_trainable(model.q_mu, False)
+set_trainable(model.q_sqrt, False)
 
 for _ in range(ci_niter(1000)):
     scipy_opt.minimize(objective_closure, model.trainable_variables, options={'maxiter': 1})
@@ -240,8 +241,8 @@ model = gpflow.models.VGP((X, Y_data), kernel=kernel, likelihood=likelihood, num
 def objective_closure():
     return - model.log_marginal_likelihood()
 
-model.q_mu.set_trainable(False)
-model.q_sqrt.set_trainable(False)
+set_trainable(model.q_mu, False)
+set_trainable(model.q_sqrt, False)
 
 for _ in range(ci_niter(1000)):
     scipy_opt.minimize(objective_closure, model.trainable_variables, options={'maxiter': 1})
--- a/doc/source/notebooks/advanced/varying_noise.pct.py
+++ b/doc/source/notebooks/advanced/varying_noise.pct.py
@@ -33,6 +33,7 @@ import tensorflow as tf
 import gpflow
 from gpflow.ci_utils import ci_niter
 from gpflow.optimizers import NaturalGradient
+from gpflow.utilities import set_trainable
 import matplotlib
 import matplotlib.pyplot as plt
 # %matplotlib inline
@@ -102,6 +103,8 @@ class HeteroskedasticGaussian(gpflow.likelihoods.Likelihood):
 # Here we'll build a variational GP model with the previous likelihood on the dataset that we generated. We'll use the natural gradient optimizer (see [Natural gradients](natural_gradients.ipynb) for more information).
 #
 # The variational GP object is capable of variational inference with any GPflow-derived likelihood. Usually, the inference is an inexact (but pretty good) approximation, but in the special case considered here, where the noise is Gaussian, it will achieve exact inference. Optimizing over the variational parameters is easy using the natural gradients method, which provably converges in a single step. 
+#
+# Note: We mark the variational parameters as not trainable so that they are not included in the `model.trainable_variables` when we optimize using the Adam optimizer. We train the variational parameters separately using the natural gradient method.
 
 # %%
 # model construction (notice that num_latent is 1)
@@ -118,13 +121,12 @@ def objective_closure():
 natgrad = NaturalGradient(gamma=1.0)
 adam = tf.optimizers.Adam()
 
-from gpflow.utilities import set_trainable
 set_trainable(model.q_mu, False)
 set_trainable(model.q_sqrt, False)
 
 for _ in range(ci_niter(1000)):
-    adam.minimize(objective_closure, model.trainable_variables)
     natgrad.minimize(objective_closure, [(model.q_mu, model.q_sqrt)])
+    adam.minimize(objective_closure, model.trainable_variables)
 
 # %%
 # let's do some plotting!
@@ -139,7 +141,7 @@ plt.plot(xx, mu - 2*np.sqrt(var), 'C0', lw=0.5)
 
 plt.errorbar(X.squeeze(), Y.squeeze(), yerr=2*(np.sqrt(NoiseVar)).squeeze(),
              marker='x', lw=0, elinewidth=1., color='C1')
-plt.xlim(-5, 5)
+plt.xlim(-5, 5);
 
 # %% [markdown]
 # ### Questions for the reader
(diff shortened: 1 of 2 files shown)


doc/source/notebooks/advanced/varying_noise.pct.py
```diff
@@ -114,10 +115,12 @@ def variational_expectations(self, Fmu, Fvar, Y):
...
 
-opt = gpflow.optimizers.Scipy()
-opt.minimize(objective_closure,
-             model.trainable_variables,
-             options=dict(maxiter=ci_niter(1000)))
+natgrad_opt = NaturalGradient(gamma=1.0)
+scipy_opt = gpflow.optimizers.Scipy()
+
+for _ in range(ci_niter(1000)):
+    scipy_opt.minimize(objective_closure, model.kernel.trainable_variables, options={'maxiter': 1})
```
vdutor: I think it would be more useful in this example to pass `model.trainable_variables` to the optimiser and explicitly do `set_trainable(model.q_mu, False)` and `set_trainable(model.q_sqrt, False)` for the natgrad optimiser. This to make sure that the natgrad and scipy optimiser are not acting on the same variables and to allow scipy to optimise over all the non-variational parameters.

doc/source/notebooks/advanced/varying_noise.pct.py
```diff
@@ -204,10 +207,8 @@ def objective_closure():
 set_trainable(model.kernel, False)
 set_trainable(model.likelihood, False)
```
vdutor: Is it necessary to set these to `trainable=False`? They are not being optimised as far as I can tell?

vdutor: Is it necessary to set these to `trainable=False`? They are not being optimised as far as I can tell?

doc/source/notebooks/advanced/varying_noise.pct.py
joelberkeley-pio: do you mean do all the scipy optimization first, then do the natgrad optimization, or to
```
model.q_mu.set_trainable(False)
model.q_srt.set_trainable(False)
scipy_opt.optimize(..., model.trainable_variables)
model.q_mu.set_trainable(True)
model.q_srt.set_trainable(True)
natgrad_opt.optimize(..., [(model.q_mu, model.q_sqrt)])

File: doc/source/notebooks/advanced/varying_noise.pct.py (hunk 2 above)
Comment (joelberkeley-pio) [reply doc/source/notebooks/advanced/varying_noise.pct.py]: no. good catch. will remove

File: doc/source/notebooks/advanced/varying_noise.pct.py (hunk 1 above)
Comment (vdutor) [reply doc/source/notebooks/advanced/varying_noise.pct.py]: Typically you want to interleave training of "normal" parameters and hyper-parameters as follows:
```python
model.q_mu.set_trainable(False)
model.q_srt.set_trainable(False)
for i in range(iters):
    scipy_opt.optimize(..., model.trainable_variables)
    natgrad_opt.optimize(..., [(model.q_mu, model.q_sqrt)])
```
The reason we set `q_mu` and `q_sqrt` to non trainable is for the scipy optmiser to ignore them. Even in this state, will the natgrad optimiser still optimise them.

joelberkeley-pio: ah yes, I forgot that `.trainable_variables` denoted whether they're marked as trainable, not whether they are actually mutable

joelberkeley-pio: i notice you've omitted maxiter of 1. was that for brevity or cos it's not needed?

joelberkeley-pio: it's worth mentioning that `set_trainable` reads to me as "this value can't be mutated" rather than "this value won't be gathered by `.trainable_variables`

vdutor: Good catch - you do want `maxiter=1` as I believe the default behaviour of scipy would be to run until convergence.

gpflow/optimizers/natgrad.py
```diff
@@ -105,11 +109,9 @@ def minimize(self, loss_fn: Callable, var_list: List[Parameter]):
         self._natgrad_steps(loss_fn, parameters)
 
     def _natgrad_steps(self, loss_fn: Callable, parameters: List[Tuple[Parameter, Parameter, XiTransform]]):
-        def natural_gradient_step(q_mu, q_sqrt, xi_transform):
-            self._natgrad_step(loss_fn, q_mu, q_sqrt, xi_transform)
-
         with tf.name_scope(f"{self._name}/natural_gradient_steps"):
-            list(map(natural_gradient_step, *zip(*parameters)))
+            for q_mu, q_sqrt, xi_transform in parameters:
+                self._natgrad_step(loss_fn, q_mu, q_sqrt, xi_transform)
```
joelberkeley-pio: @vdutor in trying to understand the natgrad code, I realised this wrapper can be written much more simply.

joelberkeley-pio: and another small change

st--:
@joelberkeley-pio it may be useful to refer to the original version with natgrads in the develop-1.0 branch: https://github.com/GPflow/GPflow/blob/develop-1.0/doc/source/notebooks/advanced/varying_noise.ipynb

doc/source/notebooks/advanced/varying_noise.pct.py
```diff
@@ -200,14 +206,8 @@ def generate_data(N=100):
...
 
-from gpflow.utilities import set_trainable
-set_trainable(model.kernel, False)
-set_trainable(model.likelihood, False)
-
-opt = gpflow.optimizers.Scipy()
-opt.minimize(objective_closure,
-             model.trainable_variables,
-             options=dict(maxiter=ci_niter(1000)))
+for _ in range(ci_niter(1000)):
```
vdutor: why aren't we interleaving the optimisation of natgrad with scipy here?

doc/source/notebooks/advanced/varying_noise.pct.py
```diff
@@ -114,10 +115,15 @@ def variational_expectations(self, Fmu, Fvar, Y):
...
     return - model.log_marginal_likelihood()
 
-opt = gpflow.optimizers.Scipy()
-opt.minimize(objective_closure,
-             model.trainable_variables,
-             options=dict(maxiter=ci_niter(1000)))
+natgrad_opt = NaturalGradient(gamma=1.0)
+scipy_opt = gpflow.optimizers.Scipy()
+
+model.q_mu.set_trainable(False)
```
vdutor: this is not allowed, you should do the following
```
from gpflow.utilities import set_trainable
set_trainable(model.q_mu, False)
```

Probably best to run the notebook and visually check the ipynb before committing to catch these errors.

doc/source/notebooks/advanced/varying_noise.pct.py
vdutor: why aren't we adding scipy steps here?

gpflow/optimizers/natgrad.py
vdutor: This is indeed much nicer! Have you checked `_natgrad_steps` isn't used anywhere else?

joelberkeley-pio: yup

doc/source/notebooks/advanced/varying_noise.pct.py
joelberkeley-pio: oh yeah

do we not have tests that check the notebook doesn't error?

vdutor: yes we do, and they failed :)

vdutor:
Feel free to merge once the tests pass.

doc/source/notebooks/advanced/varying_noise.pct.py
joelberkeley-pio: cos only q_mu and q_sqrt were left as trainable, so I guess there's nothing else to train

see
```
set_trainable(model.kernel, False)
set_trainable(model.likelihood, False)
```
in original

joelberkeley-pio:
@vdutor @st-- i've looked through the new notebook and it looks similar to the old (at least the plots do). I've also fixed up the todos and moved over to using Adam, as the text says.

btw that's a good notebook. really good documentation

gpflow/optimizers/natgrad.py
```diff
@@ -102,14 +106,10 @@ def minimize(self, loss_fn: Callable, var_list: List[Parameter]):
...
-        self._natgrad_steps(loss_fn, parameters)
-
-    def _natgrad_steps(self, loss_fn: Callable, parameters: List[Tuple[Parameter, Parameter, XiTransform]]):
-        def natural_gradient_step(q_mu, q_sqrt, xi_transform):
-            self._natgrad_step(loss_fn, q_mu, q_sqrt, xi_transform)
 
         with tf.name_scope(f"{self._name}/natural_gradient_steps"):
-            list(map(natural_gradient_step, *zip(*parameters)))
+            for q_mu, q_sqrt, xi_transform in parameters:
+                self._natgrad_step(loss_fn, q_mu, q_sqrt, xi_transform)
```
awav: This was written for `tf.function`. The `autograph=True` might recognise `map` function and apply parallerisable map function instead of sequential `for` loop.

joelberkeley-pio: here's the optimised code for the original
```
def tf___natgrad_steps(loss_fn: Callable, parameters: List[Tuple[Parameter, Parameter, XiTransform]]):
  with ag__.FunctionScope('_natgrad_steps', 'fscope', ag__.ConversionOptions(recursive=True, user_requested=True, optional_features=(), internal_convert_user_code=True)) as fscope:

    @ag__.do_not_convert_internal
    def natural_gradient_step(q_mu, q_sqrt, xi_transform):
      with ag__.FunctionScope('natural_gradient_step', 'fscope_1', ag__.ConversionOptions(recursive=True, user_requested=True, optional_features=(), internal_convert_user_code=True)) as fscope_1:
        ag__.converted_call(self._natgrad_step, (loss_fn, q_mu, q_sqrt, xi_transform), None, fscope_1)
    with tf.name_scope(f'{self._name}/natural_gradient_steps'):
      ag__.converted_call(list, (ag__.converted_call(map, (natural_gradient_step,) + tuple(ag__.converted_call(zip, tuple(parameters), None, fscope)), None, fscope),), None, fscope)
```
and the new version
```
def tf__f():
  with ag__.FunctionScope('f', 'fscope', ag__.ConversionOptions(recursive=True, user_requested=True, optional_features=(), internal_convert_user_code=True)) as fscope:
    with tf.name_scope(f'{self._name}/natural_gradient_steps'):

      def get_state():
        return ()

      def set_state(_):
        pass

      def loop_body(iterates):
        q_mu, q_sqrt, xi_transform = iterates
        ag__.converted_call(self._natgrad_step, (loss_fn, q_mu, q_sqrt, xi_transform), None, fscope)
        return ()
      ag__.for_stmt(parameters, None, loop_body, get_state, set_state, (), (), ())
```
not sure how to read them. will have a look

joelberkeley-pio: do we _need_ it to recognise `map`? If so, I propose we use a tensorflow function so that we're sure

joelberkeley-pio: it looks like, from the [docs](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/autograph/g3doc/reference/control_flow.md)

> Key Point: Only statements that are conditioned on, or iterate over, a TensorFlow object such as tf.Tensor, are converted into TensorFlow ops.

that neither version is likely to be converted to a tf op.

joelberkeley-pio: but since I can't be sure, I'm just going to put it back how it was for now

doc/source/notebooks/advanced/varying_noise.pct.py
```diff
@@ -114,10 +115,16 @@ def variational_expectations(self, Fmu, Fvar, Y):
...
-opt = gpflow.optimizers.Scipy()
-opt.minimize(objective_closure,
-             model.trainable_variables,
-             options=dict(maxiter=ci_niter(1000)))
+natgrad = NaturalGradient(gamma=1.0)
+adam = tf.optimizers.Adam()
+
+from gpflow.utilities import set_trainable
+set_trainable(model.q_mu, False)
+set_trainable(model.q_sqrt, False)
```
st--: Could you add a comment explaining why we do this (so that the Adam optimizer doesn't keep moving the variational parameters as well)? (and would the import be better at the top of the file perhaps?)
I think that is important to people new to natural gradients and the way the optimizers work.

doc/source/notebooks/advanced/varying_noise.pct.py
```diff
@@ -132,7 +139,7 @@ def objective_closure():
 
 plt.errorbar(X.squeeze(), Y.squeeze(), yerr=2*(np.sqrt(NoiseVar)).squeeze(),
              marker='x', lw=0, elinewidth=1., color='C1')
-plt.xlim(-5, 5);
+plt.xlim(-5, 5)
```
st--: The ; at the end was intentional so that jupyter doesn't print the return value
```suggestion
plt.xlim(-5, 5);
```

doc/source/notebooks/advanced/varying_noise.pct.py
st--: Here we are explicitly only optimising variational parameters, not hyperparameters (see line 213/214)

doc/source/notebooks/advanced/varying_noise.pct.py
```diff
@@ -224,11 +225,11 @@ def objective_closure():
 ax.plot(xx, mu - 2*np.sqrt(var), 'C0', lw=0.5)
 
 ax.plot(X, Y, 'C1x', mew=2)
-ax.set_xlim(-5, 5);
+ax.set_xlim(-5, 5)
```
st--:
```suggestion
ax.set_xlim(-5, 5);
```

doc/source/notebooks/advanced/varying_noise.pct.py
```diff
@@ -240,9 +241,12 @@ def objective_closure():
 def objective_closure():
     return - model.log_marginal_likelihood()
 
-opt.minimize(objective_closure,
-             model.trainable_variables,
-             options=dict(maxiter=ci_niter(1000)))
+set_trainable(model.q_mu, False)
+set_trainable(model.q_sqrt, False)
```
st--: again, add comment why setting to non-trainable

doc/source/notebooks/advanced/varying_noise.pct.py
```diff
@@ -265,7 +269,7 @@ def objective_closure():
 
 ax.plot(X, Y, 'C1x', mew=2)
 ax.set_xlim(-5, 5)
-ax.plot(xx, 2.5 * np.sin(6 * xx) + np.cos(3 * xx), 'C2--');
+ax.plot(xx, 2.5 * np.sin(6 * xx) + np.cos(3 * xx), 'C2--')
```
st--:
```suggestion
ax.plot(xx, 2.5 * np.sin(6 * xx) + np.cos(3 * xx), 'C2--');
```

gpflow/optimizers/natgrad.py
```diff
@@ -79,7 +79,12 @@ def __init__(self, gamma: Scalar, name=None):
         super().__init__(name)
         self.gamma = gamma
 
-    def minimize(self, loss_fn: Callable, var_list: List[Parameter]):
+    def minimize(
+            self,
+            loss_fn: Callable,
+            var_list: List[
+                Union[Tuple[Parameter, Parameter], Tuple[Parameter, Parameter, XiTransform]]]
```
st--: :+1:

doc/source/notebooks/advanced/varying_noise.pct.py
```diff
@@ -240,9 +241,12 @@ def objective_closure():
...
 
-opt.minimize(objective_closure,
-             model.trainable_variables,
-             options=dict(maxiter=ci_niter(1000)))
+set_trainable(model.q_mu, False)
+set_trainable(model.q_sqrt, False)
+
+for _ in range(ci_niter(1000)):
+    adam.minimize(objective_closure, model.trainable_variables)
+    natgrad.minimize(objective_closure, [(model.q_mu, model.q_sqrt)])
```
st--: Shouldn't natgrad come first ?

doc/source/notebooks/advanced/varying_noise.pct.py
```diff
@@ -114,10 +115,16 @@ def variational_expectations(self, Fmu, Fvar, Y):
...
+natgrad = NaturalGradient(gamma=1.0)
+adam = tf.optimizers.Adam()
+
+from gpflow.utilities import set_trainable
+set_trainable(model.q_mu, False)
+set_trainable(model.q_sqrt, False)
+
+for _ in range(ci_niter(1000)):
+    adam.minimize(objective_closure, model.trainable_variables)
+    natgrad.minimize(objective_closure, [(model.q_mu, model.q_sqrt)])
```
st--: shouldn't natgrad come first?

gpflow/optimizers/natgrad.py
st--: It looks like you caught an inefficiency in our natgrad code though - could you open an issue for this ?

doc/source/notebooks/advanced/varying_noise.pct.py
joelberkeley-pio: i have absolutely no idea. should it?

doc/source/notebooks/advanced/varying_noise.pct.py
joelberkeley-pio: :sad face: ok. will revert

gpflow/optimizers/natgrad.py
joelberkeley-pio: i did notice in this that this type signature is incompatible with that of `tf.optimizers.Optimizer`. we can come back to that

doc/source/notebooks/advanced/varying_noise.pct.py
joelberkeley-pio: Would
> We mark the variational parameters as not trainable so that they are not included in the `model.trainable_variables` when we optimize using the Adam optimizer. We train the variational parameters separately using the natural gradient method.

doc/source/notebooks/advanced/varying_noise.pct.py
joelberkeley-pio: also curious - if it should, why?

doc/source/notebooks/advanced/varying_noise.pct.py
st--: Yes, that's great!

doc/source/notebooks/advanced/varying_noise.pct.py
st--: I think NatGrad should come first, for the following motivation: in the conjugate case (Gaussian likelihood), a single natgrad step with gamma=1.0 brings the variational distribution q(u) to its optimum. The ELBO is only the right objective for variational parameters, whereas there might be a bias in the hyperparameters. If we bring the q(u) closer to its optimum, there will be better gradients for the adam optimization of the hyperparameters. In practice this would only make a difference for the first (& last) step but conceptually I think it's useful to revert the order here and make natgrad.minimize() come first.

PR #1271, opened 2020-02-20
Move docs to dedicated repo: GPflow/docs

vdutor:
Hi all,

This PR moves all of GPflow's documentation to its dedicated repo: [GPflow/docs](https://github.com/GPflow/docs). This means that as of this PR we won't have to deal with ipynb files anymore - something we've all been dreaming of for a while :) ...

### Most important changes:
- The entire content of `GPflow/doc/` except for the the notebooks in `GPflow/doc/source/notebooks/` has been moved to the repo GPflow/docs.
- Only the source code of our notebooks (i.e. *.md, *.pct.py and *.py) still lives in this repo along the code. The files have been moved to GPflow/notebooks.
- Upon merge into develop in GPflow are the actual notebooks (ipynb) re-generated and hosted with the other documentation in the  [GPflow/docs](https://github.com/GPflow/docs) repo.
- Our documentation on readthedocs now points to GPflow/docs.

For people interested in how the notebooks are generated in the other repo:
https://github.com/GPflow/docs/blob/master/.circleci/config.yml

#### TODO:

- [ ] Copy over static data from notebooks repo to docs/notebooks.
- [ ] Set up alert system in slack if docs are not properly build.
- [ ] Allow for documentation at different tags of the source code.


awav:
I love it

gpflow/ci_utils.py
```diff
@@ -18,7 +18,17 @@
```
awav: Pls, move this file to notebooks folder

.circleci/config.yml
```diff
@@ -79,6 +94,13 @@ workflows:
           filters:
             tags:
               only: /^v[0-9]+(\.[0-9]+)*(-rc[0-9]+)?/
+      - trigger-docs:
```
awav:
```suggestion
      - trigger-docs-generation:
```

gpflow/ci_utils.py
vdutor: That would require changing all the notebooks. Let's do this once this branch is in.

joelberkeley-pio:
I assume this doesn't include docstrings

gpflow/ci_utils.py
```diff
@@ -18,7 +18,17 @@
 
 
 def is_continuous_integration():
-    return os.environ.get('CI', None) is not None
+    """
+    Returns `True` if notebooks are executed on CI system, otherwise `False`.
+    Also returns `False` is `DOCS` environment variable is set,
```
joelberkeley-pio:
```suggestion
    Also returns `False` if `DOCS` environment variable is set,
```

gpflow/ci_utils.py
```diff
@@ -18,7 +18,17 @@
...
+    even when notebooks are executed on CI.
+    """
+    ci = os.environ.get('CI', None)
```
joelberkeley-pio:
```suggestion
    ci = os.environ.get('CI')
```

gpflow/ci_utils.py
```diff
@@ -18,7 +18,17 @@
...
+    docs = os.environ.get('DOCS', None)
```
joelberkeley-pio:
```suggestion
    docs = os.environ.get('DOCS')
```

gpflow/ci_utils.py
```diff
@@ -18,7 +18,17 @@
...
+    if docs is not None:
+        return False
+
+    return (ci is not None)
```
joelberkeley-pio: why the parentheses?

joelberkeley-pio: could be
```
if 'DOCS' in os.environ:
    return False

return 'CI' in os.environ
```

gpflow/ci_utils.py
st--: You can't move it to the notebooks folder because then you can't import it in the notebooks without hacking the `sys.path`, which is super hacky.

gpflow/ci_utils.py
st--: :+1: to the first suggestion (I think the one-liner makes it less clear what's going on)

vdutor: This is neat - thanks!

joelberkeley-pio:
I'm very glad the .ipynb files are in a different repo. I mildly prefer the .pct.py files to stay in GPflow/GPflow, and for the machinery/config to build the .ipynb files from .pct.py files to live in the new repo

vdutor:
Thanks for your comments - I see where you are coming from, but I think having the all the docs in GPflow/docs is a good solution for a couple reasons:
- I think it is neater to have all the docs config and generator files in one place: if we ever want to upgrade to another system or write a lot more documentation it won't clutter GPflow source code. 
- the code in GPflow/docs only consists of very static files so we won't be changing them often.
- it is a practice often done in other packages: most notably https://github.com/Tensorflow/docs and I like consistency.
- It reduces gpflow clone time (a bit).

vdutor:
see #1275

PR #1275, opened 2020-02-24
Second attempt: auto-generate docs

vdutor:
Hi all,

This PR (superseding #1271) enables auto-generation of all of GPflow's documentation. Upon a new merge in `develop`, we trigger our new repo [GPflow/docs](https://github.com/GPflow/docs) to:
- copy the complete content of GPflow/doc
- compile the notebooks `.ipynb` from the `.pct.py` files
- create the `rst` files for the api documentation required for readthedocs.

#### TODO:
- [x] Set up alert system in slack if docs are not properly build.
- [ ] Allow for documentation at different tags of the source code.


gpflow/ci_utils.py
```diff
@@ -18,7 +18,14 @@
 
 
 def is_continuous_integration():
-    return os.environ.get('CI', None) is not None
+    """
+    If the environment variable `DOCS` is set, always returns False.
+    Otherwise, returns `True` if notebooks are executed on CICD, else `False`.
```
st--:
```suggestion
    Otherwise, returns `True` if notebooks are executed on CI, else `False`.
```
?

gpflow/ci_utils.py
```diff
@@ -31,3 +38,4 @@ def ci_range(n: int, test_n: int = 2):
 
 def ci_list(lst: list, test_n=2):
     return lst[:test_n] if is_continuous_integration() else lst
+
```
st--:
```suggestion
```

gpflow/ci_utils.py
vdutor: continuous integration and continuous deployment

setup.py
```diff
@@ -3,22 +3,29 @@
...
 
+import os
 import sys
 from pathlib import Path
 
 from pkg_resources import parse_version
 from setuptools import find_packages, setup
 
 is_py37 = sys.version_info.major == 3 and sys.version_info.minor == 7
+on_rtd = os.environ.get('READTHEDOCS', None) == 'True'  # copied from the docs
```
vdutor: I followed this workflow: https://docs.readthedocs.io/en/stable/faq.html#i-get-import-errors-on-libraries-that-depend-on-c-modules

gpflow/ci_utils.py
```diff
@@ -18,7 +18,14 @@
 
 
 def is_continuous_integration():
-    return os.environ.get('CI', None) is not None
+    """
+    If the environment variable `DOCS` is set, always returns False.
+    Otherwise, returns `True` if notebooks are executed on CI, else `False`.
+    """
+    if 'DOCS' in os.environ:
```
st--: would it make sense to change this to `"READTHEDOCS"` (to be consistent with what RTD sets already)? (I'm happy with it as it is, too)

setup.py
st--: would it be useful to put that link into the comment?

st--: (I doubt anyone's going to manage to find these PR comments in the future)

gpflow/ci_utils.py
vdutor: This is actually a environment variable used by circleci

PR #1276, opened 2020-02-24
refactor training objective methods

st--:
This gives all `BayesianModel` subclasses a consistent interface both for optimization (MLE/MAP) and MCMC. Models are required to implement `maximum_log_likelihood_objective`, which is to be maximized for model training.

Optimization: The `_training_loss` method is defined as `- (maximum_log_likelihood_objective + log_prior_density)`. This is exposed by the InternalDataTrainingLossMixin and ExternalDataTrainingLossMixin classes.

For models that keep hold of the data internally, `training_loss` can directly be passed as a closure to an optimizer's `minimize`, for example:
```python
model = gpflow.models.GPR(data, ...)
gpflow.optimizers.Scipy().minimize(model.training_loss, model.trainable_variables)
```

If the model objective requires data to be passed in, a closure can be constructed on the fly using `model.training_loss_closure(data)`, which returns a no-argument closure:
```python
model = gpflow.models.SVGP(...)
gpflow.optimizers.Scipy().minimize(
    model.training_loss_closure(data), model.trainable_variables, ...
)
```

The training_loss_closure() method provided by both InternalDataTrainingLossMixin and ExternalDataTrainingLossMixin takes a boolean `compile` argument (default: True) that wraps the returned closure in tf.function(). Note that the return value should be cached in a variable if the minimize() step is run several times to avoid re-compilation in each step!

MCMC: The `log_posterior_density` method can be directly passed to the `SamplingHelper`. By default, `log_posterior_density` is implemented as `maximum_log_likelihood_objective + log_prior_density`. Models can override this if needed. Example:
```python
model = gpflow.models.GPMC(...)
hmc_helper = gpflow.optimizers.SamplingHelper(
    model.log_posterior_density, model.trainable_parameters
)
hmc = tfp.mcmc.HamiltonianMonteCarlo(
    target_log_prob_fn=hmc_helper.target_log_prob_fn, ...
)
```
In this case, the function that runs the MCMC chain should be wrapped in tf.function() (see MCMC notebook).


st--:
I agree it's better to keep purpose of PRs separate so I've already separated out other renamings.

I think the type changes should go in here because they're closely linked to the training loss and what type of data the different models depend on.

Having the `has_own_data` property means model implentations will have to be explicit about it, using `hasattr(model, "data")` seems more hacky and more likely to break because someone called it `self.training_data` instead of `self.data`.

I made all `training_loss` methods have a data argument because @yiwei-prowler said it would make their code much simpler with less having to special-case by model.

yiwei-prowler:
1. I would prefer the name `log_parameter_likelihood` instead of training_loss. This is because `log_parameter_likelihood` reminds us what this formula is, and I think it is approximate when the objective function for parameter learning is elbo as well. `training_loss` is too general in my opinion.

2. The objective function for parameter learning (training_loss, or log_parameter_likelihood, or other names we decide) should have a single parameter data, so all GPflow models that inherit from GPModel conforms to the Liskov Substitution Principle.  

Liskov substitution principle says for an abstract class GPModel with descendant GPR and SVGP, a client AutoML of the abstract class GPModel should continue to work when we swap the descendants.
For example a client may do:
```
def foo(model: GPModel, data: Tuple[tf.Tensor, tf.Tensor]):
    optimize(model.log_likelihood(data))
```
This is what AutoML wants to do, but since GPR.log_likelihood and SVGP.log_likelihood have different APIs (different arguments), we cannot do that with a single piece of code. We have to write two pieces of code to handle them -- Every client of GPflow, for example, bayesian optimisation team, needs to write two different pieces of code to handle this situation.

Passing data into GPR.training_loss(data) is indeed a bit hacky, because this method will simply ignore the passed-in data. We need to explicitly mention this in its documentation. It is a compromise, but I believe it is a compromise in the right direction -- better object-oriented API (=happier clients) without sacrificing performance.

3. Can you give a sample code demonstrating how `has_own_data` is to be used?

st--:
@ 1. - for some models (GPR) we can evaluate the log marginal likelihood (log parameter likelihood) exactly, for others (e.g. SGPR, VGP, SVGP) we can only evaluate a lower bound to this (elbo). In some cases you may want to use either, but I think it's important to make clear that these are different things, so there should not be a common interface for it. However, in each case there is a `training_loss` that can be minimized by an optimizer to train the model, and for that case I think it's useful to have a common interface to make models more composable.

@ 2. - yes, that's why in the current draft `training_loss` always allows you to pass the `data` argument - or am I missing something?

@ 3. - I thought `has_own_data` might be useful to programmatically distinguish between cases where you *must* pass `data` (e.g. SVGP) and where you don't need to (e.g. SGPR). But I can remove it if people think it's not needed.

frgsimpson:
I like the introduction of the training loss. And I broadly agree with @yiwei-prowler that a universal interface across models is very useful. Just to clarify, I believe it would be desirable to have both a training loss and a likelihood (the difference being that only the former has a dependency on the priors of the hyperparameters). Based upon previous discussions, these definitions could be expressed as:

log [p(Y|X, θ, m)] is the log parameter likelihood [currently model.log_likelihood()]
- log [p(Y|X θ, m) x p(θ)]  is the training loss [currently -model.log_marginal_likelihood()]
where Y, X are training data, θ is the set of model parameters, m is model. Or some other naming convention - my preference would be to omit 'marginal likelihood' as it is often used to describe different quantities.

> >  for some models (GPR) we can evaluate the log marginal likelihood, for others (e.g. SGPR, VGP, SVGP) we can only evaluate a lower bound to this

A possible compromise here would be to adopt the name estimate_log_likelihood(), then it is clear that the answer will not necessarily be exact, yet we could still retain the very useful universal interface.

yiwei-prowler:
"In some cases you may want to use either, but I think it's important to make clear that these are different things, so there should not be a common interface for it." Sure.

To clarify myself -- when I was talking about `log_parameter_likelihood`, or `training_loss`. I'm think about the objective function for parameter learning. We need this have a single API, like `objective_for_parameter_learning(data: Tuple[tf.Tensor, tf.Tensor])`. Or using the originally suggested name, `training_loss(data)`.  Because parameter learning is common to all models (from the client's point of view).

Of course, descendant classes  such as SVGP can have their own queries. In fact, SVGP has a query `elbo` right now. That's ok. It is just not part of the common API that a client who needs to swap different models should know about.

st--:
How about this for `BayesianModel` interface:
- abstract method `mle_objective` (or `maximum_likelihood_objective`), which will be *maximized* - this would be log-marginal likelihood for GPR, log-likelihood for (S)GPMC, or ELBO - this would need to be overwritten by each concrete model
- `log_posterior_density` which returns `self.mle_objective() + self.log_prior_density()` which can be used with the SamplingHelper for MCMC
- `training_loss` which returns `- (self.mle_objective() + self.log_prior_density())` which can be used with optimizers' `minimize()` for MLE (no priors) or MAP (priors on parameters)

frgsimpson:
>How about this for BayesianModel interface 
That sounds great to me, @st-- !

awav:
> Liskov substitution principle says for an abstract class GPModel with descendant GPR and SVGP, a client AutoML of the abstract class GPModel should continue to work when we swap the descendants.

@jameshensman, @st--, @yiwei-prowler, @vdutor, @alexisboukouvalas 

Indeed, SOLID are good principles and we must try to follow it whenever it is possible. Although, on practice, it might be very hard. That is our case. Training procedures for GPR style models are different from SVGP style models. In the latter case, we use stochastic optimization algorithms with mini batching.

The current solution for training loss unification masks the problem and introduces more worrying issues to GPR style models.

* Constructors of GPR style models assume that **data and parameters are interrelated**. The training loss with data breaks this integrity.
* Prediction for these models uses **encapsulated at construction time data**. The training loss method with the `data` argument misleads the user about what parameters learnt. It entails to possible Cholesky errors; it can develop wrong newcomer or inexperienced user intuition on how GPR works.

A quote about Liskov principle: _"If it looks like a duck and quacks like a duck but it needs batteries, you probably have a wrong abstraction"._ As far as I'm concerned, the problem is not in the training loss protocol and that it doesn't have a common `data` argument for every Bayesian model in GPflow. The root of the problem lies in the fact that some models have state (data) and others don't._ We discussed that with @st-- and @jameshensman, and I thought we found an agreement that we should pursue a solution with stateless models (details upon request)._

In this PR we should try to address naming **only** and provide self-descriptive methods. I agree with https://github.com/GPflow/GPflow/pull/1276#issuecomment-591567399 and would use `maximum_likelihood` instead of `mle` shortcut.

joelberkeley-pio:
i'm going to throw this in:

we don't need every class to support this right off the bat. it doesn't need to be right at the bottom of the class heirarchy. we can do it for classes where the sensible interface doesn't conflict with the current implementation, and add it to others as and when it makes sense. perhaps it will still end up in the root of the heirarchy, but it won't have been rushed.

gpflow/models/sgpmc.py
```diff
@@ -79,14 +79,26 @@ def __init__(self,
...
+    def training_loss(self, data: Optional[RegressionData] = None) -> tf.Tensor:
+        return - self.log_posterior_density(data)
+
+    def log_posterior_density(self, data: Optional[RegressionData] = None) -> tf.Tensor:
+        return self.log_likelihood(data) + self.log_prior_density()
+
+    def maximum_likelihood_objective(self, data: Optional[RegressionData] = None) -> tf.Tensor:
+        return self.log_likelihood(data)
+
+    def log_likelihood(self, data: Optional[RegressionData] = None) -> tf.Tensor:
```
st--: is this correct?

joelberkeley-pio:
If this happens to be part of this work, can we please please please drop `(*args, **kwargs)` in `BayesianModel`

st--:
> If this happens to be part of this work, can we please please please drop `(*args, **kwargs)` in `BayesianModel`?

I'm not sure what you're talking about - the refactored version of BayesianModel does not have any `*args, **kwargs` as far as I can tell?

joelberkeley-pio:
> > If this happens to be part of this work, can we please please please drop `(*args, **kwargs)` in `BayesianModel`?
> 
> I'm not sure what you're talking about - the refactored version of BayesianModel does not have any `*args, **kwargs` as far as I can tell?

You're completely right - I didn't think to look at the code changes :/

gpflow/models/training_mixins.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+class InternalDataTrainingLossMixin:
+    def training_loss(self) -> tf.Tensor:
+        """
+        Training loss for models that encapsulate their data.
+        """
+        return -self.maximum_a_posteriori_objective()
+
+    def training_loss_closure(self, jit=True) -> Callable[[], tf.Tensor]:
```
joelberkeley-pio: i'm not so sure about this name. If I have some model `model`, I could easily think `model.training_loss_closure` is the closure itself, in part because it's a noun. I reckon `build_training_loss_closure` might be more appropriate

joelberkeley-pio: tensorflow appear to use the word 'jit' as compiling for/with XLA. Isn't that different from `tf.function`. Should we be using a different word, one that they use, like `eager`? Really I'd prefer we leave it to the user to `tf.function` things and not have this argument at all

gpflow/models/training_mixins.py
```diff
@@ -0,0 +1,58 @@
...
+from typing import Callable, Optional, Tuple, TypeVar, Union
+
+
+import tensorflow as tf
+import numpy as np
+
+InputData = tf.Tensor
+OutputData = tf.Tensor
+RegressionData = Tuple[InputData, OutputData]
+Data = TypeVar("Data", RegressionData, InputData, OutputData)
```
joelberkeley-pio:
```suggestion
Data = Union[RegressionData, InputData, OutputData]
```

gpflow/models/training_mixins.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+class InternalDataTrainingLossMixin:
+    def training_loss(self) -> tf.Tensor:
+        """
+        Training loss for models that encapsulate their data.
```
joelberkeley-pio: 'encapsulate' isn't quite the right word here. it doesn't have to _hide_ the data, which is implied by 'encapsulation'. It just has to have access to their data by other means

gpflow/models/training_mixins.py
```diff
@@ -0,0 +1,58 @@
...
+        self, data: Union[Data, collections.abc.Iterator], jit=True
+    ) -> Callable[[], tf.Tensor]:
+        training_loss = self.training_loss
+        if jit:
+            training_loss = tf.function(
+                training_loss
+            )  # TODO need to add correct input_signature here to allow for differently sized minibatches
+
+        def closure():
+            batch = next(data) if isinstance(data, collections.abc.Iterator) else data
```
joelberkeley-pio: see my [comment](https://github.com/GPflow/GPflow/pull/1362/files#r397557765) on eric's PR about this

gpflow/models/training_mixins.py
```diff
@@ -0,0 +1,58 @@
...
+    def training_loss(self, data: RegressionData) -> tf.Tensor:
+        """
+        Training loss for models that do not encapsulate the data.
+        
+        :param data: the data to be used for computing the model objective.
+        """
+        return -self.maximum_a_posteriori_objective(data)
+
+    def training_loss_closure(
+        self, data: Union[Data, collections.abc.Iterator], jit=True
```
joelberkeley-pio: will this work for all `Iterators`?

joelberkeley-pio: as an aside, this should perhaps be `typing.Iterator`

gpflow/models/training_mixins.py
```diff
@@ -0,0 +1,58 @@
...
+    ) -> Callable[[], tf.Tensor]:
+        training_loss = self.training_loss
+        if jit:
+            training_loss = tf.function(
+                training_loss
+            )  # TODO need to add correct input_signature here to allow for differently sized minibatches
```
joelberkeley-pio: what's happening with this TODO?

gpflow/models/training_mixins.py
```diff
@@ -0,0 +1,58 @@
...
+import collections
+from typing import Callable, Optional, Tuple, TypeVar, Union
+
+
+import tensorflow as tf
+import numpy as np
+
+InputData = tf.Tensor
+OutputData = tf.Tensor
+RegressionData = Tuple[InputData, OutputData]
```
joelberkeley-pio: is `RegressionData` still accurate if we're doing GPC?

gpflow/models/training_mixins.py
```diff
@@ -0,0 +1,58 @@
+import abc
+import collections
+from typing import Callable, Optional, Tuple, TypeVar, Union
+
+
+import tensorflow as tf
+import numpy as np
+
+InputData = tf.Tensor
+OutputData = tf.Tensor
```
joelberkeley-pio: what are we gaining with these types? As a user, I might be a bit bewildered by all these types which just turn out to all be aliases for `tf.Tensor`

gpflow/models/training_mixins.py
```diff
@@ -0,0 +1,58 @@
...
+RegressionData = Tuple[InputData, OutputData]
+Data = TypeVar("Data", RegressionData, InputData, OutputData)
+
+
+class InternalDataTrainingLossMixin:
+    def training_loss(self) -> tf.Tensor:
+        """
+        Training loss for models that encapsulate their data.
+        """
+        return -self.maximum_a_posteriori_objective()
```
joelberkeley-pio: I spoke with @condnsdmatters about making this an abstract class with an abstract `maximum_a_posteriori_objective` method. I think it's the right thing to do, eric seemed far from convinced. I don't want it to block this PR cos it doesn't change any functionality and can very easily be added later.

In the mean time though, I would like the docs to explain that an `maximum_a_posteriori_objective` with a particular signature must be implemented for classes that mix in this, err..  mixin

gpflow/models/training_mixins.py
joelberkeley-pio: `Data` isn't used in any places that require a `TypeVar`

st--: I believe TypeVar is the correct thing in this context.

gpflow/models/training_mixins.py
st--: Please suggest a better way of phrasing it.

gpflow/models/training_mixins.py
st--: `compile`?

st--: I think it's sufficiently unambigous - training_loss is the function that returns the training loss as a tensor and training_loss_closure is the corresponding function that returns you a closure which returns the training loss as a tensor. I like that they have the same prefix as this makes code completion easy.

gpflow/models/training_mixins.py
st--: For optimization use-cases, the dataset iterators only really make sense when being `.repeat()`ed, so they never run out. If someone doesn't do that and next() raises, it's up to the caller to handle it - I don't think we need to do any more here.

gpflow/models/training_mixins.py
st--: We could change the type signature, I just thought it might as well be consistent with the `isinstance` check in the function body.

gpflow/models/training_mixins.py
joelberkeley-pio: we need it to be unambiguous for most users, and I estimate I'd be one of those who would find it unclear

gpflow/models/training_mixins.py
st--: It was already called RegressionData before, it just got moved from model.py to trainin_mixins.py so the Mixin classes have access to it. I agree it's not quite perfect; the key distinction here is between models that take both X and Y and models that only take X (point processes) or Y (latent variable models). If you can come up with a better name, please make a suggestion.

gpflow/models/training_mixins.py
st--: A mixin by definition doesn't make sense without a class to mix it into. All GPflow models subclass `BayesianModel`, which provides the maximum_a_posteriori_objective (note that the abstract method to implement is maximum_likelihood_objective). Adding an abstract method in the mixin would mean people had to be very careful about the multiple-inheritance order, so I am strongly against adding that (now or in the future).

st--: I've added a module-level docstring to clarify. Please re-open if you think that's still not clear enough.

gpflow/models/training_mixins.py
st--: Make the models a bit more self-documenting in regards to whether you pass in an input or an output or something else? Note that they only got moved in this PR.

gpflow/models/training_mixins.py
st--: Added docstrings to clarify this. I would hope that the combination of return type signature, docstring, and example use in the notebooks makes this sufficiently clear for the level of users we are targetting with GPflow.

gpflow/models/training_mixins.py
joelberkeley-pio: note [`repeat`](https://www.tensorflow.org/api_docs/python/tf/data/Dataset#repeat) has an argument `count` to specify the number of repeats

gpflow/models/training_mixins.py
joelberkeley-pio: yeah let's do that

gpflow/models/training_mixins.py
joelberkeley-pio: I've created a ticket re 'jit' #1373 . What do you think about my second point about users doing the `tf.function`-ing themselves. since we have the above closure, this just becomes
`tf.function(model.training_loss)` which is only 3 more characters than `model.training_loss_closure()`, and feel is mildly clearer

gpflow/models/training_mixins.py
st--: it needed doing. it's now done.

gpflow/models/training_mixins.py
st--: done

gpflow/models/training_mixins.py
joelberkeley-pio:
```
class InternalDataTrainingLossMixin:
    """
    Mixin to provide convenience methods for training loss. Subclasses will typically own their own
    data (this contrasts with :class:`ExternalDataTrainingLossMixin`).
    """
    def training_loss(self) -> tf.Tensor:
        """
        The training loss for this model, as used in :class:`gpflow.optimizers.Scipy` and subclasses
        of `tf.optimizers.Optimizer`.
        """
        ...
```
and similar for `ExternalDataTrainingLossMixin`

gpflow/models/training_mixins.py
st--: I still think it's best to simply not handle the StopIteration and let the caller figure out what to do with it. Seems like the most pythonic way to me, anyways

gpflow/models/training_mixins.py
st--: Alas it's not so simple for the ExternalDataTrainingLossMixin (see the latest changes), so I think it needs to stay in there, and then for myself as a user I like having available the corresponding equivalent way of doing the same for models with internal data. (This also makes it easier to do e.g. `training_loss_closure(jit=number_of_calls>10)` or something like that, when you only want the compile overhead when you know you'll call multiple times...)

joelberkeley-pio: can it not be added as a separate PR, especially since it's an optional argument and this PR's already 51 files?

gpflow/models/training_mixins.py
joelberkeley-pio: actually that's not quite right, and has just highlighted to me that this PR's doing two things:

  - creating a *uniform API* for training loss across all models (to the extent that's possible). This is `training_loss`
  - adding a *convenience method* that constructs the closure for passing into optimizers. This is `training_loss_closure`

joelberkeley-pio:
```python3
class InternalDataTrainingLossMixin:
    """
    Mixin for models that own their own data. It provides

      - a uniform API for the training loss :meth:`training_loss`
      - a convenience method for constructing the closure expected by various optimizers, namely
        :class:`gpflow.optimizers.Scipy` and subclasses of `tf.optimizers.Optimizer`.

    See :class:`ExternalDataTrainingLossMixin` for an equivalent mixin for models that do **not**
    own their own data.
    """
    def training_loss(self) -> tf.Tensor:
        """ Return the training loss for this model. """
        ...

    def training_loss_closure(self, jit: bool = True) -> Callable[[], tf.Tensor]:
        """
        Convenience method. Returns a closure which can be passed to
        :class:`gpflow.optimizers.Scipy` and subclasses of `tf.optimizers.Optimizer`

        :param jit: If `True`, compile the training loss function in a TensorFlow graph.
        """
```

gpflow/models/training_mixins.py
st--: > can it not be added as a separate PR [...]?

What are you referring to? The `jit` flag? We definitely need it for the `training_loss_closure` for the ExternalData... version. It makes sense to me to have it for the InternalData... version, too, then. Most of the files are fixing use-cases, so separating it out into another PR doesn't seem very practical to me (lots of changes in this PR, plus lots of files changed in the other PR as well, for marginally removing the number of lines changed in this PR, at the cost of an overall much larger number of lines changed).

gpflow/models/training_mixins.py
st--: thanks. done

st--: and all that documentation made the patch-coverage test pass :sweat_smile:

gpflow/models/training_mixins.py
joelberkeley-pio: why do you say we definitely need it? what functionality is it providing that we need in this PR?

joelberkeley-pio: essentially I'm asking: what is the aim of this PR, and how does this functionality achieve that aim?

joelberkeley-pio: my concern is that there's too much here to get the fundamentals in quick

gpflow/optimizers/scipy.py
```diff
@@ -134,8 +133,9 @@ def assign_tensors(to_tensors: Iterable[tf.Variable], values: Iterable[tf.Tensor
...
 
-def _compute_loss_and_gradients(loss_cb: LossClosure, variables: V) -> Tuple[tf.Tensor, V]:
-    with tf.GradientTape() as tape:
-        loss = loss_cb()
-    grads = tape.gradient(loss, variables)
+def _compute_loss_and_gradients(loss_closure: LossClosure, variables: V) -> Tuple[tf.Tensor, V]:
+    with tf.GradientTape(watch_accessed_variables=False) as tape:
+        tape.watch(variables)
+        loss = loss_closure()
+        grads = tape.gradient(loss, variables)
```
joelberkeley-pio: why this change?

gpflow/models/model.py
```diff
@@ -57,8 +52,29 @@ def log_prior_density(self) -> tf.Tensor:
...
+        SGPMC). It assumes that maximum_likelihood_objective() is defined
+        sensibly.
+        """
+        return self.maximum_likelihood_objective(*args, **kwargs) + self.log_prior_density()
+
+    def maximum_a_posteriori_objective(self, *args, **kwargs) -> tf.Tensor:
+        """
+        Maximum a-posteriori objective. Assumes that maximum_likelihood_objective() is the log marginal likelihood and adds the log density of all priors.
+        """
+        return self.maximum_likelihood_objective(*args, **kwargs) + self.log_prior_density()
```
joelberkeley-pio: remind me why there are two of these with the same default implementation

st--: I wanted to call it _training_loss and have it contain the minus sign already, but I think in reviewing my other proposal (which got subsequently merged in here) that wasn't deemed acceptable, so we called it something else ?

gpflow/models/model.py
```diff
@@ -57,8 +52,29 @@ def log_prior_density(self) -> tf.Tensor:
...
+        This may be the posterior with respect to the hyperparameters (e.g. for
+        GPR) or the posterior with respect to the function (e.g. for GPMC and
+        SGPMC). It assumes that maximum_likelihood_objective() is defined
+        sensibly.
+        """
+        return self.maximum_likelihood_objective(*args, **kwargs) + self.log_prior_density()
+
+    def maximum_a_posteriori_objective(self, *args, **kwargs) -> tf.Tensor:
+        """
+        Maximum a-posteriori objective. Assumes that maximum_likelihood_objective() is the log marginal likelihood and adds the log density of all priors.
```
joelberkeley-pio: you say "assumes". what happens when it's not e.g. when it's the elbo?

gpflow/models/model.py
joelberkeley-pio: SGPMC and GPMC are the only cases where they're overridden, but there these methods differ by `self.log_prior_density()`. Are these defaults correct?

st--: They don't differ?

st--: If you think we can simply call it _training_loss here I'd honestly prefer that.

gpflow/models/model.py
st--: that's fine - I was thinking more "it has to be the *log* marginal likelihood not the non-logged actual marginal likelihood"

gpflow/optimizers/scipy.py
st--: a clean-up due to sorting out the notebooks, I can move this into a separate PR.

gpflow/models/training_mixins.py
st--: > do we have notebooks which can't be expressed easily without it?

Yes- they are all significantly shorter (and cleaner) now, don't you think?

gpflow/models/model.py
joelberkeley-pio: why wasn't it deemed acceptable? i'm generally in favour of making everything and anything private, so `_training_loss` sounds promising

gpflow/models/model.py
```diff
@@ -27,27 +37,12 @@
...
 from ..utilities import ops, to_default_float
 
-InputData = tf.Tensor
-OutputData = tf.Tensor
-RegressionData = Tuple[InputData, OutputData]
 MeanAndVariance = Tuple[tf.Tensor, tf.Tensor]
 
 
-class BayesianModel(Module):
+class BayesianModel(Module, metaclass=abc.ABCMeta):
```
joelberkeley-pio: you can shorten this with
```suggestion
class BayesianModel(Module, abc.ABC):
```

joelberkeley-pio: just a suggestion

gpflow/models/model.py
```diff
@@ -57,8 +52,29 @@ def log_prior_density(self) -> tf.Tensor:
...
+
+    def maximum_a_posteriori_objective(self, *args, **kwargs) -> tf.Tensor:
+        """
+        Maximum a-posteriori objective. Assumes that maximum_likelihood_objective() is the log marginal likelihood and adds the log density of all priors.
+        """
+        return self.maximum_likelihood_objective(*args, **kwargs) + self.log_prior_density()
+
     @abc.abstractmethod
-    def log_likelihood(self, *args, **kwargs) -> tf.Tensor:
+    def maximum_likelihood_objective(self, *args, **kwargs) -> tf.Tensor:
```
joelberkeley-pio: given your 'assumes' above, would `maximum_log_likelihood_objective` be better?

gpflow/models/training_mixins.py
```diff
@@ -0,0 +1,129 @@
...
+            if isinstance(data, DatasetOwnedIterator):
+                tensor_specs = [
+                    tf.TensorSpec(shape, dtype)
+                    for shape, dtype in zip(data.output_shapes, data.output_types)
+                ]
+                input_signature = [tuple(tensor_specs)]
+            training_loss = tf.function(training_loss, input_signature=input_signature)
+
+        def closure():
+            batch = next(data) if isinstance(data, Iterator) else data
```
joelberkeley-pio: I believe you can't do `isinstance` checks with members of the typing package, and you can't specify generics with members of collections.abc, so the only option is to import and use both

gpflow/optimizers/scipy.py
st--: moved into #1377

gpflow/models/model.py
```diff
@@ -57,8 +52,30 @@ def log_prior_density(self) -> tf.Tensor:
...
+        sensibly.
+        """
+        return self.maximum_log_likelihood_objective(*args, **kwargs) + self.log_prior_density()
+
+    def _training_loss(self, *args, **kwargs) -> tf.Tensor:
+        """
+        Training loss definition. To allow MAP (maximum a-posteriori) estimation,
+        adds the log density of all priors to maximum_log_likelihood_objective().
+        """
+        return -(self.maximum_log_likelihood_objective(*args, **kwargs) + self.log_prior_density())
```
joelberkeley-pio: yeah i'm pro this - I think fewer names means users will be less doubtful about which method they should be calling, and private names gives us a chance to have some time with the new API and make any adjustments down the line if we need

gpflow/models/training_mixins.py
```diff
@@ -0,0 +1,129 @@
...
+        :param input_signature: to be passed to tf.function() when jit=True;
+            will be inferred from the iterator if `data` is an iterator on a
+            tf.data.Dataset.
+        """
+        training_loss = self.training_loss
+        if jit:
+            if isinstance(data, DatasetOwnedIterator):
+                tensor_specs = [
+                    tf.TensorSpec(shape, dtype)
+                    for shape, dtype in zip(data.output_shapes, data.output_types)
```
joelberkeley-pio: these attributes appear to be deprecated ... see https://github.com/tensorflow/tensorflow/blob/v2.1.0/tensorflow/python/data/ops/iterator_ops.py

`tf.compat.v1.data.get_output_shapes`
`tf.compat.v1.data.get_output_types`

gpflow/models/training_mixins.py
```diff
@@ -0,0 +1,129 @@
...
+                ]
+                input_signature = [tuple(tensor_specs)]
```
joelberkeley-pio: why this line? shouldn't `input_signature` be a list of `TensorSpec`s?

gpflow/models/training_mixins.py
```diff
@@ -0,0 +1,129 @@
...
+            training_loss = tf.function(training_loss, input_signature=input_signature)
```
joelberkeley-pio: can we have tests for this functionality? It's hard to know if it's going to work just by looking at it

joelberkeley-pio: alternatively, can we leave the input_signature stuff for another PR?

gpflow/models/training_mixins.py
st--: yes. _training_loss has a single argument, so input_signature has length one. this argument is a tuple of tensors, with specs given by tensor_specs.

gpflow/models/training_mixins.py
st--: what do you suggest we replace it with?

st--: ah, there's now `element_specs`! much neater

st--: thank you, great catch

gpflow/models/training_mixins.py
st--: added a test

gpflow/models/gpmc.py
```diff
@@ -62,7 +63,13 @@ def __init__(
             loc=to_default_float(0.0), scale=to_default_float(1.0)
         )
 
-    def log_likelihood(self, *args, **kwargs) -> tf.Tensor:
+    def log_posterior_density(self) -> tf.Tensor:
+        return self.log_likelihood() + self.log_prior_density()
+
+    def maximum_log_likelihood_objective(self) -> tf.Tensor:
+        return self.log_likelihood()
```
joelberkeley-pio: this will need updating with `_training_loss`

gpflow/models/gpr.py
```diff
@@ -50,7 +51,10 @@ def __init__(
         super().__init__(kernel, likelihood, mean_function, num_latent_gps=Y_data.shape[-1])
         self.data = data
 
-    def log_likelihood(self) -> tf.Tensor:
+    def maximum_log_likelihood_objective(self) -> tf.Tensor:
+        return self.log_marginal_likelihood()
```
joelberkeley-pio: and here

gpflow/models/training_mixins.py
```diff
@@ -0,0 +1,125 @@
...
+
+        return closure
+
+
+class MCMCTrainingLossMixin(InternalDataTrainingLossMixin):
+    def training_loss(self) -> tf.Tensor:
+        """
+        Training loss for gradient-based relaxation of MCMC models.
+        """
+        return -self.log_posterior_density()
```
joelberkeley-pio: am i missing sth here? `_training_loss` on BayesianModel is the negative `log_posterior_density`, so this appears to be the same as `InternalDataTrainingLossMixin`

doc/source/notebooks/understanding/models.pct.py
```diff
@@ -199,14 +195,29 @@ def closure():
 #
 # In this very simple demo, we'll implement linear multiclass classification.
 #
-# There are two parameters: a weight matrix and a bias (offset). The key thing to implement the `log_likelihood` method, which returns a TensorFlow scalar that represents the (log) likelihood. You can use parameter objects inside `log_likelihood`.
+# There are two parameters: a weight matrix and a bias (offset). You can use
+# Parameter objects directly, like any TensorFlow tensor.
+#
+# The training objective depends on the type of model; it may be possible to
```
vdutor: Is this still up-to-date with the new Mixin classes?

vdutor: If so, wouldn't it be nice to mention them? Also, maybe say something about the `training_loss_closure`?

doc/source/notebooks/basics/regression.pct.py
```diff
@@ -117,14 +117,14 @@
 
 
 # %% [markdown]
-# In order to minimize the negative log marginal likelihood, we create a closure to be passed to the optimizer. We also need to specify the variables to train with `m.trainable_variables`, and the number of iterations.
+# In order to train the model, we need to maximize the log marginal likelihood.
+# GPflow models define a `training_loss` that can be passed to the `minimize`
```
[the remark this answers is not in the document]
joelberkeley-pio: GPModel doesn't define a training loss, but models in GPflow provide either training_loss **and** training_loss_closure or neither. I assume they all have both

gpflow/models/gplvm.py
```diff
@@ -75,10 +76,10 @@ def __init__(
         super().__init__(gpr_data, kernel, mean_function=mean_function)
 
 
-class BayesianGPLVM(GPModel):
+class BayesianGPLVM(GPModel, InternalDataTrainingLossMixin):
     def __init__(
         self,
-        data: tf.Tensor,
+        data: OutputData,
```
joelberkeley-pio: what's the difference between input and output data? it feels odd to be passing in output data to a class

gpflow/models/gpmc.py
```diff
@@ -62,7 +63,13 @@ def __init__(
             loc=to_default_float(0.0), scale=to_default_float(1.0)
         )
 
-    def log_likelihood(self, *args, **kwargs) -> tf.Tensor:
+    def log_posterior_density(self) -> tf.Tensor:
+        return self.log_likelihood() + self.log_prior_density()
```
joelberkeley-pio: this appears to be redundant

gpflow/models/gpmc.py
```diff
@@ -62,7 +63,13 @@ def __init__(
...
+
+    def maximum_log_likelihood_objective(self) -> tf.Tensor:
+        return self.log_likelihood()
+
+    def log_likelihood(self) -> tf.Tensor:
```
joelberkeley-pio: in `GPR` below, you've used `log_marginal_likelihood`, but here the `log_likelihood`. why's that? if the technical explanation is complicated, maybe we can have it offline, separately from the PR

gpflow/models/sgpmc.py
```diff
@@ -86,7 +87,13 @@ def __init__(
             loc=to_default_float(0.0), scale=to_default_float(1.0)
         )
 
-    def log_likelihood(self, *args, **kwargs) -> tf.Tensor:
+    def log_posterior_density(self) -> tf.Tensor:
+        return self.log_conditional_likelihood_lower_bound() + self.log_prior_density()
```
joelberkeley-pio: isn't this redundant?

gpflow/monitor.py
```diff
@@ -304,7 +304,7 @@ class Monitor:
         log_dir = "logs"
         model_task = ModelToTensorBoard(log_dir, model)
         image_task = ImageToTensorBoard(log_dir, plot_prediction, "image_samples")
-        lml_task = ScalarToTensorBoard(log_dir, lambda: model.log_likelihood(), "lml")
+        lml_task = ScalarToTensorBoard(log_dir, lambda: model.log_marginal_likelihood(), "lml")
```
joelberkeley-pio: there are `log_marginal_likelihood`s for some but not all models. is that OK for this example?

gpflow/optimizers/mcmc.py
```diff
@@ -22,7 +22,7 @@
 
 
 ModelParameters = Sequence[TypeVar("ModelParameter", tf.Variable, Parameter)]
-LogProbabilityFunction = Callable[[ModelParameters], tf.Tensor]
+LogProbabilityFunction = Callable[[], tf.Tensor]  # implicitly a function of ModelParameters
```
joelberkeley-pio: is this comment needed?

gpflow/models/training_mixins.py
joelberkeley-pio: i thought it was different from the superclass before

gpflow/optimizers/mcmc.py
st--: removed it (the whole line will be removed by #1374)

gpflow/monitor.py
st--: I think it's fine, it's just an example

gpflow/models/sgpmc.py
st--: for the MCMC models I think it's good to be explicit

gpflow/models/gpmc.py
st--: it's intentional! to be explicit for MCMC models

gpflow/models/gplvm.py
st--: Y = f(X) <-- X input, Y output; here we're observing Y but don't know X, so the model only gets the output. (and for point process models, there is no Y, just X.)

doc/source/notebooks/intro_to_gpflow2.pct.py
```diff
@@ -190,14 +191,15 @@ def noisy_sin(x):
 
 
 # %% [markdown]
-# For a more elaborate example of a gradient update we can define an ```optimization_step``` that uses the decorator ```tf.function``` on a closure. A closure is a callable that returns the model objective evaluated at a given dataset when called.
+# For a more elaborate example of a gradient update we can define an ```optimization_step``` that explicitly computes and applies gradients to the model.
+# The `optimization_step` can (and should) be wrapped in `tf.function` to be compiled to a graph if executing it many times.
 
 # %%
 def optimization_step(model: gpflow.models.SVGP, batch: Tuple[tf.Tensor, tf.Tensor]):
```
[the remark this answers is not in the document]
st--: I've showed that above? using minimize() doesn't let you access the loss without re-computing it, so it might be helpful to show that?

doc/source/notebooks/understanding/models.pct.py
st--: done - added a paragraph and a link to the intro_to_gpflow2 notebook which contains more details.

gpflow/models/training_mixins.py
st--: no, it's the same, it's just about in terms of what it is defined

gpflow/models/sgpmc.py
```diff
@@ -86,7 +87,13 @@ def __init__(
...
+
+    def maximum_log_likelihood_objective(self) -> tf.Tensor:
+        return self.log_conditional_likelihood_lower_bound()
+
+    def log_conditional_likelihood_lower_bound(self) -> tf.Tensor:
```
st--: @jameshensman if you have a better suggestion for how to name this, please shout!

gpflow/models/gplvm.py
```diff
@@ -144,14 +145,18 @@ def __init__(
         assert self.X_prior_var.shape[0] == self.num_data
         assert self.X_prior_var.shape[1] == self.num_latent_gps
 
-    def log_likelihood(self) -> tf.Tensor:
+    def maximum_log_likelihood_objective(self) -> tf.Tensor:
```
joelberkeley-pio: these need updating right?
```suggestion
    def _training_loss(self) -> tf.Tensor:
```

tests/gpflow/models/test_variational.py
```diff
@@ -111,7 +111,7 @@ def test_reference_implementation_consistency():
         q_mean.reshape(-1), p_mean.reshape(-1), q_cov.reshape(-1), p_cov.reshape(-1)
     )
 
-    assert_allclose(univariate_KL - multivariate_KL, 0, atol=4)
+    assert_allclose(univariate_KL.squeeze(), multivariate_KL.squeeze(), atol=4)
```
joelberkeley-pio: why squeeze?

tests/integration/test_method_equivalence.py
```diff
@@ -208,13 +189,9 @@ def test_equivalence(approximate_model):
...
     gpr_model = _create_full_gp_model()
-    gpr_likelihood = -gpr_model.log_likelihood()
-    if isinstance(approximate_model, gpflow.models.SVGP):
-        approximate_likelihood = -approximate_model.log_likelihood(Datum.data)
-    else:
-        approximate_likelihood = -approximate_model.log_likelihood()
-
-    assert_allclose(gpr_likelihood, approximate_likelihood, rtol=1e-6)
+    gpr_likelihood = gpr_model.log_marginal_likelihood()
+    approximate_likelihood = model_maximum_log_likelihood_objective(approximate_model, Datum.data)
```
joelberkeley-pio: we know this is a GPR so we don't need to use this function

tests/gpflow/test_base_prior.py
```diff
@@ -102,15 +102,15 @@ def __init__(self, with_transform):
...
         return (self.theta + 5) ** 2
 
 
 def test_map_invariance_to_transform():
     m1 = DummyModel(with_transform=True)
     m2 = DummyModel(with_transform=False)
     assert np.allclose(
-        m1.log_marginal_likelihood().numpy(), m2.log_marginal_likelihood().numpy()
+        m1.log_posterior_density().numpy(), m2.log_posterior_density().numpy()
     ), "MAP objective should not be affected by a transform"
```
joelberkeley-pio: is this message still accurate?

tests/integration/test_method_equivalence.py
st--: this is `approximate_model` which can be SGPR, VGP, or SVGP

doc/source/notebooks/tailor/external-mean-function.pct.py
```diff
@@ -139,7 +139,7 @@ def create_optimization_step(optimizer, model: gpflow.models.GPR):
     def optimization_step():
```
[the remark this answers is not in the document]
st--: yeah probably worth going through some notebooks and cleaning them up separately

gpflow/models/training_mixins.py
```diff
@@ -0,0 +1,125 @@
...
+            tf.data.Dataset)
+        :param jit: whether to wrap training loss in tf.function()
+        :param input_signature: to be passed to tf.function() when jit=True;
+            will be inferred from the iterator if `data` is an iterator on a
+            tf.data.Dataset.
+        """
+        training_loss = self.training_loss
+        if jit:
+            if isinstance(data, DatasetOwnedIterator):
+                input_signature = [data.element_spec]
```
joelberkeley-pio: what happens if I pass a `DatasetOwnedIterator` and an `input_signature` and the latter gets overridden by this line? I reckon that if someone wants to specify their own input signature, it's reasonable to expect them to wrap the closure with `tf.function` explicitly themselves

st--: removed the input_signature argument and reworked the body of training_loss_closure

tests/gpflow/models/test_variational.py
st--: ah, because they had different shapes ((N) vs (N,1)) or something like that; the subtraction before worked because of broadcasting, but it seemed more appropriate to use assert_allclose as intended. note that this is only a self-consistency check of two different reference implementations inside the tests.

doc/source/notebooks/intro_to_gpflow2.pct.py
```diff
@@ -174,42 +175,92 @@ def noisy_sin(x):
...
+# GPflow models come with training_loss and training_loss_closure methods to make it easy to train your models.
+# There is a slight difference between models that own their own data (most of them, e.g. GPR, VGP, ...) and models that do not own the data (SVGP).
 #
-# In TensorFlow 2, we can optimize (trainable) model parameters with TensorFlow optimizers using `tf.GradientTape`. In this simple example, we perform one gradient update of the Adam optimizer to minimize the negative marginal log likelihood (or ELBO) of our model.
+# ### Model-internal data
+# For models that own their own data (inheriting from InternalDataTrainingLossMixin), data is provided at model construction time.
+# In this case, model.training_loss does not take any arguments, and can be directly passed to an optimizer's `minimize()` method:
+
+# %%
+gpr_model = gpflow.models.VGP(data, kernel, likelihood)
```
vdutor: Looks like a typo? gpr_model = VGP ?

doc/source/notebooks/intro_to_gpflow2.pct.py
```diff
@@ -174,42 +175,92 @@ def noisy_sin(x):
...
+
+# %% [markdown]
+# The SVGP model inherits from ExternalDataTrainingLossMixin and expects the data to be passed to training_loss():
+
+# %%
+assert isinstance(model, gpflow.models.SVGP)
+model.training_loss(data)
+
+# %% [markdown]
+# To make optimizing it easy, it has a `training_loss_closure()` method, that takes the data and returns a closure that computes the training loss on this data:
```
vdutor: would be nice to add a sentence saying what `data` is in this example. A `tf.dataset` or `np.ndarray` tuple?

doc/source/notebooks/theory/Sanity_check.pct.py
```diff
@@ -101,13 +102,16 @@
 # %%
 for m in models:
     opt = gpflow.optimizers.Scipy()
-    if isinstance(m, gpflow.models.SVGP):
-        loss_fn = lambda: -m.log_marginal_likelihood(data)
+    if isinstance(m, gpflow.models.ExternalDataTrainingLossMixin):
```
vdutor: can we make a helper function for this utility?

st--: there already is ! should've used it. good catch

gpflow/models/model.py
condnsdmatters: I'm sorry but I think this misses the point of the whole abstraction. The idea of the TrainingLossMixin is that all the logic relevant to training is on it. The model knows nothing about a training loss. Here you are saying, secretly the model does know about a training loss! But what happens if you want to train with different objective function, maybe you want to importance sample or something. 

Before you just implement a new Mixin and you're good to go: after all the model knows nothign about training. Now, if you do that you'll get a different `training_loss` and `training_loss_closure`  which may be completely unrelated to `_training_loss`

condnsdmatters:
```suggestion
    def maximum_log_likelihood_objective_with_priors(self, *args, **kwargs) -> tf.Tensor:
        """
        Training loss definition. To allow MAP (maximum a-posteriori) estimation,
        adds the log density of all priors to maximum_log_likelihood_objective().
        """
        return -(self.maximum_log_likelihood_objective(*args, **kwargs) + self.log_prior_density())
```

condnsdmatters: its possible making this private is a good idea though.

gpflow/models/training_mixins.py
```diff
@@ -0,0 +1,131 @@
...
+    their own data.
+    """
+
+    def training_loss(self, data: Data) -> tf.Tensor:
+        """
+        Returns the training loss for this model.
+        
+        :param data: the data to be used for computing the model objective.
+        """
+        return self._training_loss(data)
```
condnsdmatters: see above for my comment

gpflow/models/training_mixins.py
```diff
@@ -0,0 +1,131 @@
...
+
+        return closure
+
+
+class MCMCTrainingLossMixin(InternalDataTrainingLossMixin):
+    def training_loss(self) -> tf.Tensor:
+        """
+        Training loss for gradient-based relaxation of MCMC models.
+        """
+        return -self.log_posterior_density()
```
condnsdmatters: why are we using log posterior density instead of `_training_loss`? Stuff like this is confusing

condnsdmatters: we're sayin gon this model: x is an MCMCTrainingLossMixin, but the model `has a _training_loss`, but we ignore this, and define a new one, that is actually the same!

condnsdmatters: if we stick with `_training_loss` lets delete this class.

doc/source/notebooks/intro_to_gpflow2.pct.py
st--: thanks, fixed

doc/source/notebooks/intro_to_gpflow2.pct.py
st--: tried adding a sentence, let me know what you think

doc/source/notebooks/theory/Sanity_check.pct.py
st--: there wasn't, just a similar one, but i've added one now

gpflow/models/training_mixins.py
st--: ok, done

gpflow/models/sgpmc.py
st--: yes, it is (and renamed to log_likelihood_lower_bound for SGPMC)

PR #1278, opened 2020-02-24
Rename model.predict_... arguments to Xnew

### LINKED ISSUE #1250: SVGP.predict_f has inconsistent argument names
SVGP.predict_f has a parameter named Xnew, but the inherited method GPModel.predict_f has the parameter named predict_at. We should keep using predict_at. Some other models have the same issue.

Also, in GPModel.predict_f, header comments are missing.

st--:
Currently, the `predict_...` method signatures in gpflow2 are very inconsistent. Some take `Xnew`, others take `predict_at`. I find `predict_at` confusing. Some variant on "X" would be consistent with all the GP literature. `Xnew` would be consistent with both gpflow1 and GPy. Scikit-learn uses simply `X`.
Closes #1250.


awav:
I prefer `predict_at` argument name for a couple of reasons:

* It is more informative to newbie people
* It is a language for "normal" human
* It does not violate pep8 in function signatures. We don't follow pep8 always, but we shouldn't assume that users of GPflow do the same thing.
* It shows the maturity of the project via respecting the rules - pep8 again.

In short, function interfaces must follow pep8 and internals can break pep8 sometimes.

joelberkeley-pio:
@st-- the PR title says you're making the `predict` signature consistent. As a reviewer I'd need more information: what is the proposed chosen signature (you've chosen `Xnew` not `foo` or `bar`)? Where did that come from? Why that over alternatives?

st--:
I think `X` should be familiar notation for input points for anyone using GPs. I dislike `predict_at` because `model.predict_f(predict_at)` looks rather redundant and a bit confusing. How about `X_prediction`?

joelberkeley-pio:
it sounds like there's two tasks here

1. choose a signature
2. implement it

my vote is to do them as separate tasks

st--:
Yes, I'm happy to do it in separate tasks. How about we use this draft-PR to discuss signature, and then once we've agreed on one, can turn it into a full ready-for-review PR afterwards?

awav:
@st--, do you mind if @joelberkeley-pio takes a lead on that and will work on these PRs?

> How about X_prediction?

[As I mentioned before](https://github.com/GPflow/GPflow/pull/1278#pullrequestreview-365628741), in function signatures we probably want to follow pep8 rules . Therefore, arguments should be lower case. I don't mind having an argument `at`, so that signature would look: `predict_f(at: Data, ...)`.

st--:
I'd be very happy for other people to do the work once we've found a consistent name everyone can live with. Personally I think `Xnew` contains more information than `at`. I'm also not convinced that we must follow pep8 in function signatures.

joelberkeley-pio:
@st-- what information in `Xnew` that isn't in `at` would you like to preserve?

as an aside, I'd like to know why the type checker didn't pick up this inconsistency. do we have a type checker?

st--:
> what information in `Xnew` that isn't in `at` would you like to preserve?

That it is about inputs (X), and specifically new test points (new).

Some variant on "X" would be consistent with *all* the GP literature. Xnew would be consistent with both gpflow1 and GPy. Scikit-learn uses simply X.

> as an aside, I'd like to know why the type checker didn't pick up this inconsistency. do we have a type checker?

There is no type checker run on GPflow.

joelberkeley-pio:
Candidates so far:

  - `Xnew` - consistent with both gpflow1 and GPy
  - `predict_at` - PEP8, jargon-free
  - `at` - PEP8, jargon-free
  - `X_predictions`
  - `X` - like scikit-learn

joelberkeley-pio:
> > what information in `Xnew` that isn't in `at` would you like to preserve?
> 
> That it is about inputs (X), and specifically new test points (new).
> 

In the Bayesian optimization toolbox, we evaluate `predict_f` at points for which the model has already been optimized, if we're modelling from noisy data rather than using the observed values themselves. Isn't that a reasonable use case? In light of this, is `new` accurate for noisy observations?

doc/source/notebooks/advanced/coregionalisation.pct.py
```diff
@@ -136,14 +136,14 @@ def plot_gp(x, mu, var, color, label):
 
 def plot(m):
     plt.figure(figsize=(8, 4))
-    xtest = np.linspace(0, 1, 100)[:, None]
+    Xtest = np.linspace(0, 1, 100)[:, None]
```
joelberkeley-pio: I get why you added these to the PR, but I'd much prefer it if they were in a separate PR. it's not necessarily quick to work out what should be `x` and `X`

joelberkeley-pio: would you mind splitting this into two PRs? one for just predict method parameters and one for everything else?

gpflow/models/vgp.py
```diff
@@ -236,23 +228,21 @@ def predict_f(
         if full_output_cov:
             raise NotImplementedError
 
-        x_data, _y_data = self.data
+        X_data, _Y_data = self.data
```
vdutor:
```suggestion
        X_data, _ = self.data
```

gpflow/models/vgp.py
```diff
@@ -104,24 +104,18 @@ def log_likelihood(self):
...
+        var_exp = self.likelihood.variational_expectations(fmean, fvar, Y_data)
 
         return tf.reduce_sum(var_exp) - KL
 
     def predict_f(
-        self, predict_at: DataPoint, full_cov: bool = False, full_output_cov: bool = False
+        self, Xnew: DataPoint, full_cov: bool = False, full_output_cov: bool = False
     ) -> MeanAndVariance:
-        x_data, _y_data = self.data
+        X_data, _Y_data = self.data
```
vdutor:
```suggestion
        X_data, _ = self.data
```

gpflow/models/gplvm.py
```diff
@@ -70,48 +70,48 @@ def __init__(
...
+        :param X_data_var: variance of latent positions ([N, Q]), for the initialisation of the latent space.
         :param kernel: kernel specification, by default Squared Exponential
         :param num_inducing_variables: number of inducing points, M
         :param inducing_variable: matrix of inducing points, size M (inducing points) x Q (latent dimensions). By default
-            random permutation of x_data_mean.
-        :param x_prior_mean: prior mean used in KL term of bound. By default 0. Same size as x_data_mean.
-        :param x_prior_var: pripor variance used in KL term of bound. By default 1.
+            random permutation of X_data_mean.
+        :param X_prior_mean: prior mean used in KL term of bound. By default 0. Same size as X_data_mean.
+        :param X_prior_var: pripor variance used in KL term of bound. By default 1.
```
vdutor:
```suggestion
        :param X_prior_var: prior variance used in KL term of bound. By default 1.
```

doc/source/notebooks/advanced/ordinal_regression.pct.py
```diff
@@ -113,19 +113,19 @@ def pred_log_density(m):
...
 plt.colorbar()
 plt.plot(X, Y, 'kx', mew=2, scalex=False, scaley=False)
 
 # %%
 # Predictive density for a single input x=0.5
 x_new = 0.5
-ys = np.arange(np.max(y_data+1)).reshape([-1, 1])
-x_new_vec = x_new*np.ones_like(ys)
+Y_new = np.arange(np.max(Y_data+1)).reshape([-1, 1])
+X_new = np.full_like(Y_new, x_new)
```
alexisboukouvalas: remove `x_new` as it's only used once?
```
X_new = np.full_like(Y_new, 0.5)
```

PR #1289, opened 2020-02-26
clean up SGPR common __init__

st--:
`gpflow.models.sgpr` contains two sparse GP models, SGPR and GPRFITC. They had a common subclass to provide the upper_bound() method. They also have the same `__init__`, which is unified in this PR.

### DIFF
--- a/gpflow/models/sgpr.py
+++ b/gpflow/models/sgpr.py
@@ -41,12 +41,14 @@ class SGPRBase(GPModel):
                  noise_variance: float = 1.0,
                  ):
         """
-        X is a data matrix, size [N, D]
-        Y is a data matrix, size [N, R]
-        Z is a matrix of pseudo inputs, size [M, D]
-        kernel, mean_function are appropriate GPflow objects
+        `data`:  a tuple of (X, Y), where the inputs X has shape [N, D]
+            and the outputs Y has shape [N, R].
+        `inducing_variable`:  an InducingPoints instance or a matrix of
+            the pseudo inputs Z, of shape [M, D].
+        `kernel`, `mean_function` are appropriate GPflow objects
 
-        This method only works with a Gaussian likelihood.
+        This method only works with a Gaussian likelihood, its variance is
+        initialized to `noise_variance`.
         """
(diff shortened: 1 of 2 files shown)


gpflow/models/sgpr.py
```diff
@@ -26,37 +26,68 @@
 from ..mean_functions import Zero, MeanFunction
 
 
-class SGPRUpperMixin(GPModel):
+class SGPRBase(GPModel):
+    """
+    Common base class for SGPR and GPRFITC that provides the common __init__
+    and upper_bound() methods.
```
joelberkeley-pio: if it's going to remain public, you may want to change this docstring to reflect that

gpflow/models/sgpr.py
```diff
@@ -26,37 +26,68 @@
...
+                 kernel: Kernel,
+                 mean_function: Optional[MeanFunction] = None,
+                 inducing_variable: Optional[InducingPoints] = None,
+                 num_latent: Optional[int] = None,
+                 noise_variance: float = 1.0,
+                 ):
+        """
+        X is a data matrix, size [N, D]
+        Y is a data matrix, size [N, R]
+        Z is a matrix of pseudo inputs, size [M, D]
```
joelberkeley-pio: where are X, Y and Z?

gpflow/models/sgpr.py
```diff
@@ -26,37 +26,68 @@
...
+        kernel, mean_function are appropriate GPflow objects
```
joelberkeley-pio: this sentence doesn't convey much

gpflow/models/sgpr.py
```diff
@@ -26,37 +26,68 @@
...
 
-    ::
+        This method only works with a Gaussian likelihood.
```
joelberkeley-pio: i suggest
```suggestion
        Assumes a Gaussian likelihood.
```
that is, unless the args needs to be compatible with a Gaussian likelihood

gpflow/models/sgpr.py
joelberkeley-pio: perhaps describe what it does rather than how it's used?

PR #1293, opened 2020-03-02
Deepcopy and freeze models

awav:
Credits to @marcoadurno. This PR resolves issues in #1266 - it doesn't handle paths like `"attr[0].attr.attr[-1]"`.


gpflow/utilities/utilities.py
```diff
@@ -246,19 +247,93 @@ def clear_bijector(bijector, _, state):
...
+    """Set `value` by a given path to a nested attribute.
+
+    :param target: Root object that contains a nested attribute.
+    :param attr_path: String type value that represents path to the attribute.
+    :param value: Value to assign to the attribute.
+
+    Example:
+        k = gpflow.kernels.Matern52()
+        m = gpflow.models.GPR(..., kernel=kernel)
+        setattr_by_string(m, "kernel.lengthscale", tf.constant(1.0, dtype=...))
```
```suggestion
        setattr_by_path(m, "kernel.lengthscale", tf.constant(1.0, dtype=...))
```

gpflow/utilities/utilities.py
```diff
@@ -246,19 +247,93 @@ def clear_bijector(bijector, _, state):
...
+def getattr_by_path(target: Any, attr_path: str) -> Any:
+    """Get a value of nested attribute by a string path.
+
+    :param target: Root object that contains nested attribute
+    :param attr_path: String type value that represents path to an attribute.
+
+    Example:
+        k = gpflow.kernels.Matern52()
+        m = gpflow.models.GPR(..., kernel=kernel)
+        lengthscale = getattr_by_string(m, "kernel.lengthscale")
```
```suggestion
        lengthscale = getattr_by_path(m, "kernel.lengthscale")
```

tests/test_deepcopy.py
```diff
@@ -27,19 +25,33 @@ def __call__(self, x):
...
     assert module.var != module_copy.var
+
+
+def test_deepcopy_with_freeze():
+    module = Nested(Nested(A()))
+    module_frozen = deepcopy(module, freeze=True)
+    assert len(module.variables) == 2
+    assert module_frozen.variables == ()
+    assert isinstance(module_frozen.module.module.var, tf.Tensor)
+    assert isinstance(module_frozen.module.module.var, tf.Tensor)
```
awav:
```suggestion
    assert isinstance(module_frozen.module.module.var, tf.Tensor)

```

gpflow/utilities/utilities.py
```diff
@@ -30,6 +30,7 @@
 Path = str
 Accumulator = Tuple[Path, State]
 TraverseUpdateCallable = Callable[[TraverseInput, Path, State], State]
+M = TypeVar('M', bound=tf.Module)
```
st--: What's the advantage of using this as a type annotation instead of `tf.Module` directly? If it *is* necessary, how about a more descriptive name (e.g. `ModuleType`)?

gpflow/utilities/utilities.py
```diff
@@ -246,22 +248,93 @@ def clear_bijector(bijector, _, state):
...
+        except ValueError:
+            raise ValueError(f"Invalid index passed in the path")
+        attr = getattr(parent, attr_str)
+        attr[index] = value
+    else:
+        setattr(parent, attr_str, value)
+
+
+def _get_last_attr_spec(parent: Any, attr_path: str) -> Any:
+    restr = r"((\w+)(\[(-?[\s\d]+)\])?(?<!\.))+"
```
st--: What does this do ?

vdutor: can you add your example you gave me this morning to the docstrings please

gpflow/utilities/utilities.py
joelberkeley-pio: type variables are more specific https://docs.python.org/3/library/typing.html#typing.TypeVar

they're typically single (or few) capital letters. it's not ideal in python since they're global, but that certainly the standard in other languages, and looks so for the docs too. I'm not sure of the benefit of moving it away from the function it's used in, but I don't feel strongly about that

joelberkeley-pio:
is this related to a particular ticket?

gpflow/utilities/utilities.py
awav: @st--, that change wasn't introduced by me. It was already in the code.

st--: I think it'd be much more helpful to have it right next to the function that it is used in (`deepcopy[_components]`), as in that context it's easily understood what it does, whereas having it at the top of the module makes both the definition and its usage more confusing.

st--:
> is this related to a particular ticket?

It (partially) addresses issue #1127, as mentioned by PR #1266, which is superseded by this PR to make it more generally usable.

joelberkeley-pio:
> Credits to @marcoadurno. This PR resolves issues in #1266 - it doesn't handle paths like `"attr[0].attr.attr[-1]"`.

what about that won't it handle? the indexing? the repeated names? the nested attribute access?

gpflow/utilities/utilities.py
st--: @awav I'd be happy to keep this in but I think it should go back next to the function it's used for.

tests/gpflow/utilities/test_deepcopy.py
```diff
@@ -36,9 +40,18 @@ def test_deepcopy_component_clears_bijector_cache_and_deecopy(module):
...
     assert module.var != module_copy.var
+
+
+def test_deepcopy_with_freeze():
+    module = Nested(Nested(A()))
+    module_frozen = deepcopy(module, freeze=True)
+    assert len(module.variables) == 2
+    assert module_frozen.variables == ()
+    assert isinstance(module_frozen.module.module.var, tf.Tensor)
+    assert isinstance(module_frozen.module.module.var, tf.Tensor)
```
st--: why this duplication? was it supposed to test something else?

gpflow/utilities/utilities.py
```diff
@@ -246,22 +248,118 @@ def clear_bijector(bijector, _, state):
...
 
 
-M = TypeVar('M', bound=tf.Module)
+def _get_by_name_index(parent: object, attr_str: str, index_str: str) -> object:
+    attr = getattr(parent, attr_str)
+    if index_str != "":
+        try:
+            index = int(index_str)
+        except ValueError:
+            raise ValueError(f"Invalid index passed in the path'")
```
```suggestion
            raise ValueError(f"Invalid index {index_str} passed in the path")
```

gpflow/utilities/utilities.py
```diff
@@ -246,22 +248,118 @@ def clear_bijector(bijector, _, state):
...
+        return attr[index]
+    return attr
+
+
+def _set_by_name_index(parent: object, value: Any, attr_str: str, index_str: str):
+    if index_str != "":
+        try:
+            index = int(index_str)
+        except ValueError:
+            raise ValueError(f"Invalid index passed in the path")
```
```suggestion
            raise ValueError(f"Invalid index {index_str} passed in the path")
```

gpflow/utilities/utilities.py
```diff
@@ -246,22 +248,118 @@ def clear_bijector(bijector, _, state):
...
-M = TypeVar('M', bound=tf.Module)
+def _get_by_name_index(parent: object, attr_str: str, index_str: str) -> object:
+    attr = getattr(parent, attr_str)
+    if index_str != "":
+        try:
+            index = int(index_str)
+        except ValueError:
+            raise ValueError(f"Invalid index passed in the path'")
+        return attr[index]
+    return attr
```
```suggestion
    else:
        return attr
```

gpflow/utilities/utilities.py
```diff
@@ -246,22 +248,118 @@ def clear_bijector(bijector, _, state):
...
+            index = int(index_str)
+        except ValueError:
+            raise ValueError(f"Invalid index passed in the path")
+        attr = getattr(parent, attr_str)
+        attr[index] = value
+    else:
+        setattr(parent, attr_str, value)
+
+
+def _get_last_attr_spec(parent: Any, attr_path: str) -> Any:
```
```suggestion
def _get_last_attr_spec(parent: object, attr_path: str) -> Tuple[object, str, str]:
```

gpflow/utilities/utilities.py
awav: I don't mind changing it, but this `else` doesn't add clarity and it is more or less a style preference.

st--: it makes it nicely parallel with the `_set_by_name_index` below though

gpflow/utilities/utilities.py
```diff
@@ -30,6 +30,7 @@
 Path = str
 Accumulator = Tuple[Path, State]
 TraverseUpdateCallable = Callable[[TraverseInput, Path, State], State]
+M = TypeVar("M", bound=tf.Module)
```
st--: could you move this back next to the `deepcopy` definition so that the context can make clear what this is & what it's used for ?

awav:
> > Credits to @marcoadurno. This PR resolves issues in #1266 - it doesn't handle paths like `"attr[0].attr.attr[-1]"`.
> 
> what about that won't it handle? the indexing? the repeated names? the nested attribute access?

nested attribute access with indices

st--:
Note that this PR reduces our test coverage (there are a bunch of code branches in the get/set utilities that are not tested by anything).

gpflow/utilities/utilities.py
```diff
@@ -91,11 +91,13 @@ def __init__(self):
...
-                  var_list: List[tf.Variable] = None,
-                  maxiter=1e3,
-                  jit=False):
+def training_loop(
+    closure: Callable[..., tf.Tensor],
+    optimizer: Optional[tf.optimizers.Optimizer] = None,
+    var_list: List[tf.Variable] = None,
+    maxiter=1e3,
+    jit=False,
+):
```
joelberkeley-pio: is this a black formatter change? if so, would you mind getting the changes from develop so that this PR's on topic?

tests/gpflow/utilities/test_deepcopy.py
```diff
@@ -1,17 +1,16 @@
...
 import pytest
+import copy
 
 import tensorflow as tf
 import tensorflow_probability as tfp
-from gpflow.utilities import deepcopy_components
+from gpflow.utilities import deepcopy
 
 
 class A(tf.Module):
```
joelberkeley-pio: `A` and `B` could do with more descriptive names - what are their purposes? what distinguishes them?

tests/gpflow/utilities/test_deepcopy.py
```diff
@@ -26,8 +25,13 @@ def __call__(self, x):
         return self.a(x)
 
 
+class Nested(tf.Module):
```
joelberkeley-pio:
```suggestion
class NestedModule(tf.Module):
```

joelberkeley-pio: and wherever it's used

tests/gpflow/utilities/test_deepcopy.py
```diff
@@ -26,8 +25,13 @@ def __call__(self, x):
...
+    def __init__(self, module: tf.Module):
+        self.module = module
+
+
 @pytest.mark.parametrize("module", [A(), B()])
-def test_deepcopy_component_clears_bijector_cache_and_deecopy(module):
+def test_deepcopy_clears_bijector_cache_and_deecopy(module):
     """
     With each forward pass through a bijector, a cache is stored inside which prohibits the deepcopy of the bijector.
     This is due to the fact that HashableWeakRef objects are not pickle-able, which raises a TypeError. Alternatively,
```
joelberkeley-pio: `deepcopy_component` has been renamed

tests/gpflow/utilities/test_deepcopy.py
```diff
@@ -36,9 +40,18 @@ def test_deepcopy_component_clears_bijector_cache_and_deecopy(module):
     input = 1.0
     _ = module(input)
     with pytest.raises(TypeError):
-        deepcopy(module)
-    module_copy = deepcopy_components(module)
+        copy.deepcopy(module)
+    module_copy = deepcopy(module)
```
joelberkeley-pio: might improve clarity to use `gpflow.utilities.deepcopy`?

tests/gpflow/utilities/test_deepcopy.py
```diff
@@ -36,9 +40,18 @@ def test_deepcopy_component_clears_bijector_cache_and_deecopy(module):
...
     assert module.var == module_copy.var
     assert module.var is not module_copy.var
     module_copy.var.assign([5.0])
     assert module.var != module_copy.var
+
+
+def test_deepcopy_with_freeze():
```
joelberkeley-pio: test what about it?

tests/gpflow/utilities/test_deepcopy.py
```diff
@@ -36,9 +40,18 @@ def test_deepcopy_component_clears_bijector_cache_and_deecopy(module):
...
+    module = Nested(Nested(A()))
+    module_frozen = deepcopy(module, freeze=True)
+    assert len(module.variables) == 2
```
joelberkeley-pio: this looks like a test for the testing utilities - I'm not sure about including that in the test.

gpflow/utilities/utilities.py
```diff
@@ -246,22 +247,123 @@ def clear_bijector(bijector, _, state):
...
+        k = gpflow.kernels.Matern52()
+        m = gpflow.models.GPR(..., kernel=kernel)
+        lengthscale = getattr_by_path(m, "kernel.lengthscale")
+    """
+    try:
+        descendant, attr, index = _get_last_attr_spec(target, attr_path)
+        return _get_by_name_index(descendant, attr, index)
+    except ValueError as error:
+        raise ValueError(f"Cannot get value at path '{attr_path}'\n"
+                         f"Got error: {error}")
```
joelberkeley-pio: why this function? it's not used anywhere

gpflow/utilities/utilities.py
```diff
@@ -246,22 +247,123 @@ def clear_bijector(bijector, _, state):
...
+
+
+def setattr_by_path(target: object, attr_path: str, value: Any):
```
joelberkeley-pio: can be module-private

gpflow/utilities/utilities.py
```diff
@@ -246,22 +247,123 @@ def clear_bijector(bijector, _, state):
...
+    :param value: Value to assign to the attribute.
+
+    Example:
+        k = gpflow.kernels.Matern52()
+        m = gpflow.models.GPR(..., kernel=kernel)
+        setattr_by_path(m, "kernel.lengthscale", tf.constant(1.0, dtype=...))
+    """
+    try:
+        descendant, attr, index = _get_last_attr_spec(target, attr_path)
+        return _set_by_name_index(descendant, value, attr, index)
```
joelberkeley-pio:
```suggestion
        _set_by_name_index(descendant, value, attr, index)
```

gpflow/utilities/utilities.py
joelberkeley-pio:
```suggestion
def setattr_by_path(target: object, attr_path: str, value: Any) -> None:
```

joelberkeley-pio: also you've used `object` for `target`, but `Any` for `value`. why so?

gpflow/utilities/utilities.py
```diff
@@ -246,22 +247,123 @@ def clear_bijector(bijector, _, state):
...
+    except (AttributeError, IndexError, ValueError) as error:
+        raise ValueError(f"Cannot assign value at path '{attr_path}'\n"
+                         f"Got error: {error}")
+
+
+M = TypeVar("M", bound=tf.Module)
+
+
+def deepcopy(input_module: M, freeze: bool = False) -> M:
```
joelberkeley-pio: `copy.deepcopy` [already has an optional argument](https://docs.python.org/3.8/library/copy.html). I suggest making a separate function `freeze` which deepcopy-s and freezes in one go

tests/gpflow/utilities/test_deepcopy.py
```diff
@@ -26,8 +25,13 @@ def __call__(self, x):
...
 
 
+class Nested(tf.Module):
+    def __init__(self, module: tf.Module):
+        self.module = module
+
+
 @pytest.mark.parametrize("module", [A(), B()])
-def test_deepcopy_component_clears_bijector_cache_and_deecopy(module):
+def test_deepcopy_clears_bijector_cache_and_deecopy(module):
```
joelberkeley-pio:
```suggestion
def test_deepcopy_clears_bijector_cache_and_deepcopy(module):
```

gpflow/utilities/utilities.py
```diff
@@ -246,22 +247,123 @@ def clear_bijector(bijector, _, state):
...
+        k = gpflow.kernels.Matern52()
+        m = gpflow.models.GPR(..., kernel=kernel)
+        setattr_by_path(m, "kernel.lengthscale", tf.constant(1.0, dtype=...))
+    """
+    try:
+        descendant, attr, index = _get_last_attr_spec(target, attr_path)
+        return _set_by_name_index(descendant, value, attr, index)
+    except (AttributeError, IndexError, ValueError) as error:
+        raise ValueError(f"Cannot assign value at path '{attr_path}'\n"
+                         f"Got error: {error}")
```
joelberkeley-pio: why `ValueError`? i realise we're assigning values, but that's not what [`ValueError`](https://docs.python.org/3/library/exceptions.html#ValueError) means

> Raised when an operation or function receives an argument that has the right type but an inappropriate value, and the situation is not described by a more precise exception such as IndexError.

gpflow/utilities/utilities.py
joelberkeley-pio:
```suggestion
        _set_by_name_index(descendant, value, attr, index)
```

gpflow/utilities/utilities.py
```diff
@@ -246,22 +247,123 @@ def clear_bijector(bijector, _, state):
...
+def _set_by_name_index(parent: object, value: Any, attr_str: str, index_str: str):
+    if index_str != "":
+        index = int(index_str)
+        attr = getattr(parent, attr_str)
+        attr[index] = value
+    else:
+        setattr(parent, attr_str, value)
+
+
+def _get_last_attr_spec(parent: object, attr_path: str) -> Tuple[object, str, str]:
```
joelberkeley-pio: you mention that it doesn't handle paths like `attr[0].attr.attr[-1]`

  - what are the implications of that?
  - if we don't handle that, can we be explicit about it in the docs?
  - is the docstring example if we don't handle this case?

joelberkeley-pio: more generally, it may be much safer to generate a lambda function that sets the value, then pass that the value

gpflow/utilities/utilities.py
```diff
@@ -246,22 +247,123 @@ def clear_bijector(bijector, _, state):
     return input_module
 
 
-M = TypeVar('M', bound=tf.Module)
+def _get_by_name_index(parent: object, attr_str: str, index_str: str) -> object:
+    attr = getattr(parent, attr_str)
+    if index_str != "":
+        index = int(index_str)
```
joelberkeley-pio: what if the index isn't an index by a dictionary key?

tests/gpflow/utilities/test_deepcopy.py
```diff
@@ -36,9 +40,18 @@ def test_deepcopy_component_clears_bijector_cache_and_deecopy(module):
...
     with pytest.raises(TypeError):
-        deepcopy(module)
-    module_copy = deepcopy_components(module)
+        copy.deepcopy(module)
+    module_copy = deepcopy(module)
     assert module.var == module_copy.var
     assert module.var is not module_copy.var
     module_copy.var.assign([5.0])
     assert module.var != module_copy.var
+
```
joelberkeley-pio: can we have some tests for frozen deepcopy with lists, nested lists, dictionaries etc? I realise that's a lot of tests but there's a lot that can go wrong with this kind of functionality

joelberkeley-pio: perhaps this can be done efficiently with parametrised tests

gpflow/utilities/utilities.py
awav: this is a utility function for users.

gpflow/utilities/utilities.py
awav: Why private? The utility function for a user.

awav: Honestly, I don't like this. If I don't specify return type that means, it doesn't return anything. This is the stupidity of Python.

gpflow/utilities/utilities.py
awav: Attribute path is wrong not the object where you assign to.

gpflow/utilities/utilities.py
awav: It does handle this case. Where did you get it from?

gpflow/utilities/utilities.py
awav: Module path can have only lists or modules.

gpflow/utilities/utilities.py
joelberkeley-pio: what would be a typical use case?

joelberkeley-pio: yeah but GPflow's in python. gonna have to live with it until you decide to rewrite it in swift

gpflow/utilities/utilities.py
joelberkeley-pio: err ... your comment at the top of the PR

gpflow/utilities/utilities.py
joelberkeley-pio: looking at `traverse_module`, which is used indirectly via `parameter_dict`, it appears there is the possibility of dictionaries

gpflow/utilities/utilities.py
joelberkeley-pio: ok. got ya. Can I suggest `raise ValueError(...) from error` rather than `"Got error: {error}"`?

gpflow/utilities/utilities.py
joelberkeley-pio: what would be a typical use case?

gpflow/utilities/utilities.py
joelberkeley-pio: or do you mean you'd just rather skip type hints when they're a nuisance?

doc/source/notebooks/intro_to_gpflow2.pct.py
```diff
@@ -361,4 +364,47 @@ def checkpointing_training_loop(model: gpflow.models.SVGP,
 
 # %%
 params = gpflow.utilities.parameter_dict(model)
-gpflow.utilities.multiple_assign(model, params)
+gpflow.utilities.multiple_assign(model, params)\
```
vdutor:
```suggestion
gpflow.utilities.multiple_assign(model, params)
```

vdutor: not sure why the backslash is there?

doc/source/notebooks/intro_to_gpflow2.pct.py
```diff
@@ -361,4 +364,47 @@ def checkpointing_training_loop(model: gpflow.models.SVGP,
...
+
+# %% [markdown]
+# Load module back as new instance and compare predict results
+
+# %%
+loaded_model = tf.saved_model.load(save_dir)
+loaded_result = loaded_model.predict(samples_input)
+
+np.testing.assert_allclose(loaded_result[0], original_result[0])
+np.testing.assert_allclose(loaded_result[1], original_result[1])
```
vdutor:
```suggestion
np.testing.assert_allclose(loaded_result[1], original_result[1])

```

doc/source/notebooks/intro_to_gpflow2.pct.py
```diff
@@ -361,4 +364,47 @@ def checkpointing_training_loop(model: gpflow.models.SVGP,
...
+tf.saved_model.save(module_to_save, save_dir)
+
+# %% [markdown]
+# Load module back as new instance and compare predict results
+
+# %%
+loaded_model = tf.saved_model.load(save_dir)
+loaded_result = loaded_model.predict(samples_input)
+
+np.testing.assert_allclose(loaded_result[0], original_result[0])
```
vdutor: you can probably group these two asserts into a single using `np.testing.assert_array_allclose`

doc/source/notebooks/intro_to_gpflow2.pct.py
```diff
@@ -361,4 +364,47 @@ def checkpointing_training_loop(model: gpflow.models.SVGP,
...
+
+# %%
+frozen_model = gpflow.utilities.freeze(model)
+
+# %% [markdown]
+# In order to save the model we need to define a `tf.Module` holding the `tf.function`'s that we wish to export, as well as a reference to the underlying model:
+
+# %%
+module_to_save = tf.Module()
+predict_fn = tf.function(frozen_model.predict_f, input_signature=[tf.TensorSpec(shape=[None, 1], dtype=tf.float64)], autograph=False)
```
vdutor: does this approach allow to store multiple functions?

doc/source/notebooks/intro_to_gpflow2.pct.py
```diff
@@ -361,4 +364,47 @@ def checkpointing_training_loop(model: gpflow.models.SVGP,
...
 params = gpflow.utilities.parameter_dict(model)
-gpflow.utilities.multiple_assign(model, params)
+gpflow.utilities.multiple_assign(model, params)\
+
+# %% [markdown]
+# ### TensorFlow saved_model
+#
+# At present tensorflow does not support saving custom variables like instances of the `gpflow.base.Parameter` class, see [here](https://github.com/tensorflow/tensorflow/issues/34908).
+#
+# However, once trained completed, it is possible to clone the model and replace all Parameters with constants holding the same value:
```
vdutor:
```suggestion
# However, once trained completed, it is possible to clone the model and replace all `gpflow.base.Parameter`s with `tf.constant`s holding the same value:
```

gpflow/utilities/utilities.py
awav: Set value for attribute given by a path from `params_dict`.

doc/source/notebooks/intro_to_gpflow2.pct.py
```diff
@@ -361,4 +364,47 @@ def checkpointing_training_loop(model: gpflow.models.SVGP,
 
 # %%
 params = gpflow.utilities.parameter_dict(model)
-gpflow.utilities.multiple_assign(model, params)
+gpflow.utilities.multiple_assign(model, params)\
+
+# %% [markdown]
+# ### TensorFlow saved_model
+#
+# At present tensorflow does not support saving custom variables like instances of the `gpflow.base.Parameter` class, see [here](https://github.com/tensorflow/tensorflow/issues/34908).
```
st--:
```suggestion
# At present tensorflow does not support saving custom variables like instances of the `gpflow.base.Parameter` class, see [this TensorFlow github issue](https://github.com/tensorflow/tensorflow/issues/34908).
```
It's good to name links descriptively:)

doc/source/notebooks/intro_to_gpflow2.pct.py
```diff
@@ -361,4 +364,47 @@ def checkpointing_training_loop(model: gpflow.models.SVGP,
 
 # %%
 params = gpflow.utilities.parameter_dict(model)
-gpflow.utilities.multiple_assign(model, params)
+gpflow.utilities.multiple_assign(model, params)\
+
+# %% [markdown]
+# ### TensorFlow saved_model
```
st--:
```suggestion
# ### TensorFlow `saved_model`
```

gpflow/utilities/utilities.py
```diff
@@ -247,18 +251,133 @@ def clear_bijector(bijector, _, state):
     return input_module
 
 
+def _get_by_name_index(parent: object, attr_str: str, index_str: str) -> object:
+    attr = getattr(parent, attr_str)
+    if index_str is not None:
```
st--: should the type annotation be `Optional[str]` then?

gpflow/utilities/utilities.py
```diff
@@ -247,18 +251,133 @@ def clear_bijector(bijector, _, state):
     return input_module
 
 
+def _get_by_name_index(parent: object, attr_str: str, index_str: str) -> object:
```
```suggestion
def _get_by_name_index(parent: object, attr_str: str, index_str: Optional[str]) -> object:
```

gpflow/utilities/utilities.py
st--: > you mention that it doesn't handle paths like `attr[0].attr.attr[-1]`

I believe that is referring to the original PR (#1266)

gpflow/utilities/utilities.py
st--: @vdutor mentioned wanting to make use of this for monitoring

doc/source/notebooks/intro_to_gpflow2.pct.py
joelberkeley-pio: @vdutor do you mean `tf.Tensor`s? `tf.constant`'s a function not a type right?

doc/source/notebooks/intro_to_gpflow2.pct.py
joelberkeley-pio: to keep you on your toes vincent

gpflow/utilities/utilities.py
joelberkeley-pio: also nested lists are a concern. it's OK if we don't want to support them but we should be clear to users about that

doc/source/notebooks/intro_to_gpflow2.pct.py
awav: You store functions that you want to store. You have to assign them as an attribute. The explanation will come in the following issue (we discussed it offline).

doc/source/notebooks/intro_to_gpflow2.pct.py
awav: it is not a type, but it is a by-product of tf.constant that has a different meaning in TF engine.

gpflow/utilities/utilities.py
awav: Nope, but it should be another type.

doc/source/notebooks/intro_to_gpflow2.pct.py
awav: The answer is "yes you can store multiple functions".

Issue #1240, opened 2020-02-03
How to enforce input dimension on kernels?


st--:
In gpflow2, there is no longer a required `input_dim` argument to kernels (yay!). However, some kernels only support certain input dimensions (e.g. `Changepoint` only 1D, [Haversine distance kernels](https://stackoverflow.com/questions/59512401/custom-haversine-matern52-kernel-in-gpflow-2-0/59515458#59515458) only 2D, ...) - as this is a potential source of errors, what would be the best way of including error-checking and enforcing an input_dim where needed?

PR #1303, opened 2020-03-10
adding the sinc kernel

NicolasDurrande:
### Adding new kernel: the sinc covariance function

The sinc covariance is interesting for modelling quasi-periodic signals. It is a stationary covariance defined as
    k(r) = \sigma^2 sin(r) / r     if r > 0
    k(r) = \sigma^2                  if r = 0

### Details

Tensorflow does not seem to have a sinc function, so it is implemented with a tf.where to handle the case `r=0`. I've added some tests comparing it to the numpy implementation. As I'm a good citizen, I've also added a test to the cosine covariance function on the way ;)

### DIFF
--- a/gpflow/kernels/stationaries.py
+++ b/gpflow/kernels/stationaries.py
@@ -170,6 +170,25 @@ class Matern52(Stationary):
         return self.variance * (1.0 + sqrt5 * r + 5.0 / 3.0 * tf.square(r)) * tf.exp(-sqrt5 * r)
 
 
+class Sinc(Stationary):
+    """
+    The sinc kernel. Functions drawn from a GP with this kernel have a quasi-periodic behaviour.
+    The kernel equation is
+
+        k(r) = σ² sin{r}/r     for r>0
+        k(r) = σ²              for r=0
+
+    where:
+    r  is the Euclidean distance between the input points, scaled by the lengthscale parameter ℓ,
+    σ² is the variance parameter.
+    """
+
+    def K_r(self, r):
+        k0 = tf.ones_like(r)
+        k = tf.divide(tf.sin(r), r)
+        return self.variance * tf.where(tf.equal(r, 0), k0, k)
+
+
 class Cosine(Stationary):
     """
     The Cosine kernel. Functions drawn from a GP with this kernel are sinusoids
--- a/gpflow/kernels/stationaries.py
+++ b/gpflow/kernels/stationaries.py
@@ -186,7 +186,7 @@ class Sinc(Stationary):
     def K_r(self, r):
         k0 = tf.ones_like(r)
         k = tf.sin(r) / r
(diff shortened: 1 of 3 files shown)


gpflow/kernels/stationaries.py
```diff
@@ -170,6 +170,25 @@ def K_r(self, r):
...
+        k(r) = σ²              for r=0
+
+    where:
+    r  is the Euclidean distance between the input points, scaled by the lengthscale parameter ℓ,
+    σ² is the variance parameter.
+    """
+
+    def K_r(self, r):
+        k0 = tf.ones_like(r)
+        k = tf.divide(tf.sin(r), r)
```
```suggestion
        k = tf.sin(r) / r
```

gpflow/kernels/stationaries.py
```diff
@@ -170,6 +170,25 @@ def K_r(self, r):
...
+        return self.variance * tf.where(tf.equal(r, 0), k0, k)
```
```suggestion
        return self.variance * tf.where(tf.equal(r, 0.0), k0, k)
```
could also use `tf.math.is_nan` as a check, but that might hide issues if `r` already has NaNs in...

tests/gpflow/kernels/test_kernels.py
```diff
@@ -65,6 +65,28 @@ def test_rbf_1d(variance, lengthscale):
...
+    assert_allclose(gram_matrix, reference_gram_matrix)
+
+
+@pytest.mark.parametrize("variance, lengthscale", [[2.3, 1.4]])
+def test_sinc_1d(variance, lengthscale):
+    X = rng.randn(3, 1)
+    kernel = gpflow.kernels.Sinc(lengthscale=lengthscale, variance=variance)
+
+    gram_matrix = kernel(X)
+    reference_gram_matrix = variance * np.sinc((X - X.T) / np.pi / lengthscale)
```
st--: why does this one have a `np.pi` in it whereas the `Sinc` implementation doesn't?

NicolasDurrande: There seem to be two alternative definitions for sinc. The one with the $\pi$ is called centered sinc, and is the one implemented in numpy.

NicolasDurrande:
I've done some further numerical checks and the sinc function isn't positive definite for an input dim greater than 3. We know that it is for d=1, but it remains to be proved that it is for d=2 and d=3.

PR is not quite ready to merge then...

st--:
Would be good if we could simply define that this kernel is only defined for input_dim=1... see #1240

st--:
@NicolasDurrande do you believe this kernel would be generally useful? If so, would you want to change this to add a check that it operates only on 1D inputs (on which it is certainly positive definite) to enable us to merge it in?

st--:
@NicolasDurrande please feel free to re-open this PR with answers to the questions above!

PR #1311, opened 2020-03-13
shared interface with inducing variables

condnsdmatters:


condnsdmatters:
#1311 is this one

gpflow/kernels/mo_kernels.py
```diff
@@ -120,6 +131,11 @@ def __init__(self, kernels, name=None):
     def num_latents(self):
         return len(self.kernels)
 
+    @property
+    def kernels(self):
+        """The underlying kernels in the multioutput kernel"""
+        return tuple(self._kernels)
+
```
st--: Is this not already provided by inheriting from `Combination`?

gpflow/kernels/mo_kernels.py
```diff
@@ -164,6 +180,11 @@ def __init__(self, kernels, W, name=None):
     def num_latents(self):
         return self.W.shape[-1]  # L
 
+    @property
+    def kernels(self):
+        """The underlying kernels in the multioutput kernel"""
+        return tuple(self._kernels)
+
```
st--: Again, is this needed? it should be provided by the Combination inheritance already, no?

st--:
Note that before merging this you make sure to merge #1317 into this branch, as notebook tests are otherwise not being run!

gpflow/kernels/base.py
```diff
@@ -206,7 +206,11 @@ def _set_kernels(self, kernels: List[Kernel]):
                 kernels_list.extend(k.kernels)
             else:
                 kernels_list.append(k)
-        self.kernels = kernels_list
+        self._kernels = kernels_list
```
st--: convert this to a tuple then don't have to overwrite it in subclasses anymore

st--: (can also construct a tuple from the start)

gpflow/kernels/mo_kernels.py
```diff
@@ -93,6 +99,11 @@ def num_latent_gps(self):
         # In this case number of latent GPs (L) == output_dim (P)
         return self.output_dim
 
+    @property
+    def kernels(self):
+        """The underlying kernels in the multioutput kernel"""
+        return (self.kernel,)
+
```
st--: Can this be deleted now that it is inherited from `Combinations`?
```suggestion
```

gpflow/kernels/mo_kernels.py
```diff
@@ -120,6 +131,11 @@ def __init__(self, kernels, name=None):
     def num_latent_gps(self):
         return len(self.kernels)
 
+    @property
+    def kernels(self):
+        """The underlying kernels in the multioutput kernel"""
+        return self._kernels
+
```
st--: Can this be deleted now that it's inherited from `Combinations`?
```suggestion
```

gpflow/kernels/mo_kernels.py
```diff
@@ -164,6 +180,11 @@ def __init__(self, kernels, W, name=None):
     def num_latent_gps(self):
         return self.W.shape[-1]  # L
 
+    @property
+    def kernels(self):
+        """The underlying kernels in the multioutput kernel"""
+        return self._kernels
+
```
st--: Same- can this be deleted now ?
```suggestion
```

condnsdmatters: no I dont think so. I'm happy to add the property to the Combination if you'd prefer but we should keep it as a property

condnsdmatters: take it back its in the base! oops!

st--:
Looks great! just need to fix the failing tests (e.g. Changepoints kernel is a Combination kernel and will need updating to set self._kernels instead of self.kernels)

gpflow/kernels/mo_kernels.py
```diff
@@ -107,7 +118,7 @@ def K_diag(self, X, full_output_cov=True):
         return tf.linalg.diag(Ks) if full_output_cov else Ks  # [N, P, P] or [N, P]
 
 
-class SeparateIndependent(MultioutputKernel, Combination):
+class SeparateIndependent(Combination, MultioutputKernel):
```
st--: does the order matter ?

PR #1315, opened 2020-03-13
Make conditionals function signatures consistent

st--:
function -> f


st--:
I went through all files in conditionals/ - I may have missed one but if so we can fix it later...

Issue #951, opened 2019-03-05
SwitchedLikelihood fails in conjunction with likelihoods relying on the quadrature fallback


awayt039:
Are there any tips and trick to using the Student T likelihood? Below is an example of some data that I have... similar to the gpflow example of a coregionalized model with a student T.  When I run the code below, I get this error.

```   File "/opt/conda/lib/python3.6/site-packages/gpflow-1.3.0-py3.6.egg/gpflow/quadrature.py", line 187, in eval_func
    feval = f(*Xs, **Ys)  # f should be elementwise: return shape N x H**Din
  File "/opt/conda/lib/python3.6/site-packages/gpflow-1.3.0-py3.6.egg/gpflow/decors.py", line 67, in tensor_mode_wrapper
    result = method(obj, *args, **kwargs)
  File "/opt/conda/lib/python3.6/site-packages/gpflow-1.3.0-py3.6.egg/gpflow/likelihoods.py", line 217, in logp
    return logdensities.student_t(Y, F, self.scale, self.df)
  File "/opt/conda/lib/python3.6/site-packages/gpflow-1.3.0-py3.6.egg/gpflow/logdensities.py", line 59, in student_t
    tf.log(1. + (1. / df) * (tf.square((x - mean) / scale)))
```

maybe a convergence error...? I have tried playing around with the jitter parameter and the degrees of freedom, but to no avail. Any ideas?

Thanks!

 Below is a minimal bug-reproducing example
```
import gpflow as gpflow
import numpy as np

idx = np.array([0]*12 + [1]*15 + [2]*36)
X_aug = np.c_[np.random.randn(12+15+36, 1), idx]
Y_aug = np.c_[np.random.randn(12+15+36, 1), idx]

k1 = gpflow.kernels.Matern32(1, variance=.1, active_dims=[0]) +gpflow.kernels.White(1)
coreg = gpflow.kernels.Coregion(1, output_dim=3, rank=3, active_dims=[1])
kern = k1 * coreg

# build a variational model. This likelihood switches between Student-T noise with different variances:
lik = gpflow.likelihoods.SwitchedLikelihood([gpflow.likelihoods.StudentT(), gpflow.likelihoods.StudentT(), gpflow.likelihoods.StudentT()])
m = gpflow.models.VGP(X_aug, Y_aug, kern=kern, likelihood=lik)
## optimization errors out
gpflow.train.ScipyOptimizer().minimize(m)

st--:
[...]

PR #1316, opened 2020-03-13
Fix SwitchedLikelihood vs num_latent_gps bug

st--:
Rescues PR #957 which fixed bug #951.


awav:
Could you please link the issue from our project board to that issue? Thanks

gpflow/models/model.py
```diff
@@ -103,6 +104,32 @@ def __init__(
...
+        _, Y = data
+        output_dim = Y.shape[1]
+        return GPModel.num_latent_gps_for_likelihood_and_output_dim(likelihood, output_dim)
+
+    @staticmethod
+    def num_latent_gps_for_likelihood_and_output_dim(likelihood, output_dim) -> int:
+        """
+        Note: It's not nice for `GPModel` to need to be aware of specific
+        likelihoods as here. However, `num_latent_gps` is a bit more broken in
+        general, specifically regarding multioutput kernels. We should fix this
```
st--: check that we've got a separate issue to keep track of this; add issue url to code

st--: @condnsdmatters I merged your suggestion in here. Do you think we can remove this point from the docstring now ?

gpflow/models/model.py
```diff
@@ -103,6 +104,36 @@ def __init__(
...
         self.likelihood = likelihood
 
+    @staticmethod
+    def calc_num_latent_gps_from_data(data, kernel: Kernel, likelihood: Likelihood) -> int:
+        _, Y = data
+        output_dim = Y.shape[-1]
+        return GPModel.calc_num_latent_gps(kernel, likelihood, output_dim)
+
+    @staticmethod
+    def calc_num_latent_gps(kernel: likelihood: Likelihood, Kernel, output_dim: int) -> int:
```
condnsdmatters: looks like some vim tricks went wrong ;)
```suggestion
    def calc_num_latent_gps(kernel: Kernel, likelihood: Likelihood, output_dim: int) -> int:
```

PR #1288, opened 2020-02-26
improve organization of test modules

st--:
Historically, there was just a single flat directory full of test modules. This had several negative implications: a) some test files ended up very long, b) it was occasionally hard to find a good place for new tests, c) no separation between unit and integration tests.

This PR gives the tests a directory structure that mirrors that of the `gpflow` package. Where possible, the code in gpflow/submodule/file.py is tested in tests/gpflow/submodule/test_file.py. Integration tests (e.g. notebooks, method equivalence) are now in tests/integration/.

Most of this PR is simple file moves, plus some splitting up of tests (particularly for models) according to the new structure. The reference kernel implementations are now unified in tests/gpflow/kernels/reference.py.

PR #1317, opened 2020-03-13
Reactivate notebook tests

st--:
Notebook tests were silently deactivated by #1288.
This PR a) fixes the notebook test, b) adds another test to ensure this can't happen again, and c) fixes the notebooks for the PRs that have since made it into `develop`.

### DIFF
--- a/tests/integration/test_notebooks.py
+++ b/tests/integration/test_notebooks.py
@@ -33,7 +33,7 @@ BLACKLISTED_NOTEBOOKS = []
 
 def _nbpath():
     this_dir = os.path.dirname(__file__)
-    return os.path.join(this_dir, "../doc/source/notebooks")
+    return os.path.join(this_dir, "../../doc/source/notebooks")
 
 
 def get_notebooks():
--- a/tests/integration/test_notebooks.py
+++ b/tests/integration/test_notebooks.py
@@ -24,16 +24,22 @@ import pytest
 from nbconvert.preprocessors import ExecutePreprocessor
 from nbconvert.preprocessors.execute import CellExecutionError
 
+
+NOTEBOOK_DIR = "../../doc/source/notebooks"
+
+def test_notebook_dir_exists():
+    assert os.path.isdir(NOTEBOOK_DIR)
+
+
 # To blacklist a notebook, add its full base name (including .ipynb extension,
 # but without any directory component). If there are several notebooks in
 # different directories with the same base name, they will all get blacklisted
 # (change the blacklisting check to something else in that case, if need be!)
 BLACKLISTED_NOTEBOOKS = []
 
-
 def _nbpath():
     this_dir = os.path.dirname(__file__)
-    return os.path.join(this_dir, "../../doc/source/notebooks")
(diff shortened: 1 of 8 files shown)


tests/integration/test_notebooks.py
```diff
@@ -24,18 +24,26 @@
...
+
+def _nbpath():
+    this_dir = os.path.dirname(__file__)
+    return os.path.join(this_dir, NOTEBOOK_DIR)
+
+
+def test_notebook_dir_exists():
+    assert os.path.isdir(_nbpath())
+
+
```
awav:
```suggestion
```

awav: The `test_has_notebooks` tests `_nbpath`.

tests/integration/test_notebooks.py
```diff
@@ -24,18 +24,26 @@
 from nbconvert.preprocessors import ExecutePreprocessor
 from nbconvert.preprocessors.execute import CellExecutionError
 
+
+NOTEBOOK_DIR = "../../doc/source/notebooks"
+
+
+def _nbpath():
+    this_dir = os.path.dirname(__file__)
+    return os.path.join(this_dir, NOTEBOOK_DIR)
```
awav: I think `NOTEBOOK_DIR` variable is unnecessary here. Could you return a string back and remove `NOTEBOOK_DIR`, please?

tests/integration/test_notebooks.py
```diff
@@ -74,3 +82,7 @@ def _exec_notebook(notebook_filename):
 @pytest.mark.parametrize("notebook_file", get_notebooks())
 def test_notebook(notebook_file):
     _exec_notebook(notebook_file)
+
+
+def test_has_notebooks():
+    assert len(get_notebooks()) >= 35, "there are probably some notebooks that were not discovered"
```
joelberkeley-pio: does it need to be this high? would 10 do? i guess it would mean this test will fail less frequently

joelberkeley-pio: also can you make the text more descriptive? I'd find it a little hard to debug that error message

st--: this test would only fail if we delete some notebooks, I think it's suitably obvious why the test would fail in that case...:)

tests/integration/test_notebooks.py
st--: I think it's better to have config-style things like that as global variables at the top of the file rather than hidden inside some helper function...

awav:
@st--, I would appreciate if the next time comments that reviewer makes would be addressed or discussed before merging PRs

awav:
@joelberkeley-pio ^^^

tests/integration/test_notebooks.py
awav: `35` looks like a random number, and that is a random number.

st--:
@awav I replied to your comments, I didn't see anything I needed to change about my PR from your comments, and I decided it important to get the notebook tests reactivated asap. We can always make more changes in new PRs: I can clear up the meaning of 35 separately.

Issue #1318, opened 2020-03-13
Rename keyword argument in kernel's `__call__` from full to full_cov


joelberkeley-pio:
dtt

PR #1319, opened 2020-03-13
rename kernel full to full_cov

joelberkeley-pio:
does the thing for #1318


st--:
Note that "tests passing" includes having merged #1317 into this.

PR #1322, opened 2020-03-13
Clean up TODOs in the code

st--:
There are still some TODOs in the code that we should clean up before releasing 2.0 - either by addressing them, removing them, or turning them into issues.

Also see open issues:
* #1330 
* #1331 
* #1332 
* #1339 (+ #1340 for better test).


awav:
I don't understand how this PR changes and questions are related. Could you make an issue for that, please? Thanks

st--:
> Also, we didn't discuss it in our meeting and it doesn't look like a high priority task.

I think for a release it would be appropriate not to have random TODO's left in the code. I'm sorry I didn't think of it earlier.

doc/source/notebooks/advanced/variational_fourier_features.pct.py
```diff
@@ -336,7 +336,7 @@ def objective():
 opt = gpflow.optimizers.Scipy()
 opt.minimize(objective,
              variables=m.trainable_variables,
-             options=dict(maxiter=5000), method='L-BFGS-B')  # TODO: make work with BFGS
+             options=dict(maxiter=5000))
```
alexisboukouvalas: Question: Is BFGS the default? So this TODO is just out of date?

tests/gpflow/utilities/test_ops.py
```diff
@@ -0,0 +1,45 @@
...
+    return (X - X.mean(0)).dot(W)
+
+
+@pytest.mark.parametrize("N", [3, 7])
+@pytest.mark.parametrize("D", [2, 5, 9])
+@pytest.mark.parametrize("Q", [2, 5, 9])
+def test_numpy_equivalence(N, D, Q):
+    X = np.random.randn(N, D)
+
+    if Q > D:
```
alexisboukouvalas: nit: this looks a bit ugly - how about splitting the exception case (Q > D) to another unittest.

doc/source/notebooks/advanced/variational_fourier_features.pct.py
st--: No, L-BFGS-B is the default, I think the comment was from when we thought we should get things to work with BFGS (the scipy.minimize default) but actually for functions with very parameters such as here L-BFGS-B works much better, and L-BFGS-B is now explicitly the gpflow.Scipy default. I.e., now I think it makes no sense "trying to get it to work with BFGS". L-BFGS-B is the better option for us. If it optimizes with L-BFGS-B we're done.

tests/gpflow/utilities/test_ops.py
st--: true, but makes it harder to parametrize (you can't parametrize conditionally, so then I'd have to write out the whole set of combinations instead of two single lists)

Issue #1184, opened 2019-12-09
gpflow.utilities.set_trainable vs Parameter.trainable


st--:
There are now two different ways for changing the `trainable` state of parameters: e.g. `kernel.variance.trainable = True # False` for `Parameter` instances, and `gpflow.utilities.set_trainable(kernel, False)` for `tf.Module` instances. The danger is setting `kernel.trainable = False` which actually has no effect. Can we somehow catch this and at least throw an error? Alternatively, would it be worth explicitly using `set_trainable()` *everywhere* to show users "this is the way to do it"?

st--:
We should encourage using `set_trainable(param, is_trainable)` instead of `param.trainable = is_trainable` in the notebooks, and perhaps add a warning to the `trainable` attribute setter in `gpflow.Parameter` to raise a warning and encouraging use of `set_trainable()` instead.

PR #1323, opened 2020-03-13
remove trainable setter on Parameter

joelberkeley-pio:
delete `Parameter.trainable` setter, for #1184 , **at least I think this was the intended action. let me know if not**

### DIFF
--- a/gpflow/base.py
+++ b/gpflow/base.py
@@ -139,7 +139,7 @@ class Parameter(tf.Module):
         """
         `True` if this instance is trainable, else `False`. Cannot be set directly. Instead use
 
-            >>> from gpflow.utilities import set_trainable
+            >>> from gpflow import set_trainable
             >>> p = Parameter(...)
             >>> set_trainable(p)
 
--- a/gpflow/base.py
+++ b/gpflow/base.py
@@ -138,10 +138,6 @@ class Parameter(tf.Module):
     def trainable(self):
         return self._unconstrained.trainable
 
-    @trainable.setter
-    def trainable(self, flag: Union[bool, int]):
-        self._unconstrained._trainable = bool(flag)
-
     @property
     def initial_value(self):
         return self._unconstrained.initial_value
--- a/gpflow/base.py
+++ b/gpflow/base.py
@@ -136,6 +136,14 @@ class Parameter(tf.Module):
 
     @property
     def trainable(self):
+        """
+        `True` if this instance is trainable, else `False`. Cannot be set directly. Instead use
+
+            >>> from gpflow.utilities import set_trainable
+            >>> p = Parameter(...)
+            >>> set_trainable(p)
+
+        """
         return self._unconstrained.trainable
 
     @property
--- a/gpflow/base.py
+++ b/gpflow/base.py
@@ -137,12 +137,9 @@ class Parameter(tf.Module):
     @property
     def trainable(self):
         """
-        `True` if this instance is trainable, else `False`. Cannot be set directly. Instead use
+        `True` if this instance is trainable, else `False`.
 
-            >>> from gpflow import set_trainable
-            >>> p = Parameter(...)
-            >>> set_trainable(p)
-
+        This attribute cannot be set directly. Use :func:`gpflow.set_trainable`.
         """
         return self._unconstrained.trainable
 
--- a/gpflow/utilities/utilities.py
+++ b/gpflow/utilities/utilities.py
@@ -49,13 +49,6 @@ def set_trainable(model: tf.Module, flag: bool):
     """
     Set trainable flag for all `tf.Variable`s and `gpflow.Parameter`s in a module.
     """
-    if not tf.executing_eagerly():
-        tf.print(
-            f"{set_trainable.__name__} may exhibit unexpected behaviour when used outside eager"
-            f" mode",
-            output_stream=logging.warning
-        )
-
     for variable in model.variables:
         variable._trainable = flag
 
--- a/gpflow/utilities/utilities.py
+++ b/gpflow/utilities/utilities.py
@@ -6,7 +6,6 @@ from typing import Any, Callable, Dict, List, Optional, Tuple, TypeVar, Union
 import numpy as np
 import tensorflow as tf
 import tensorflow_probability as tfp
-from absl import logging
 from tabulate import tabulate
 
 from ..base import Parameter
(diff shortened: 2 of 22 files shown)


gpflow/base.py
```diff
@@ -138,10 +138,6 @@ def transform(self, new_transform):
     def trainable(self):
         return self._unconstrained.trainable
 
-    @trainable.setter
```
vdutor: Would be possible to throw a more descriptive error message when a user tries to set `trainable`? Something that points them to using `gpflow.utilities.set_trainable` for example.

joelberkeley-pio: hmmm ... not sure. the only way I can think of doing that is to put the setter back and raise an error if it's used, but having the setter there just so it can raise an error sounds like a source for confusion. how about I document it in the getter?

awav: A user will not be able to set a trainable attribute without an error: it is a property. I would like to have a clear message in the docs rather than a setter with a custom error. The exception will be too much honestly.

joelberkeley-pio: i'll add it to the docs

st--: In this case I think a custom setter that raises an error message would significantly help user experience. Otherwise it is kind of confusing - it has a `trainable` attribute like a tf.Variable, but there it is simply settable, the whole point of removing it was to make it *easier* for users not to do it wrong - we shouldn't make it harder for them.

joelberkeley-pio: what do you mean by it's settable on `tf.Variable`
```python3
>>> v = tf.Variable(1)
>>> v.trainable
True
>>> v.trainable = False
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: can't set attribute
```

st--: You're right, sorry my bad... should've tried it out myself first. I've tried looking for tensorflow documentation but could not find any mention of how they would suggest you handle wanting to change the trainable status of a variable! This makes me wonder if we'll run into some subtle bugs with tf.function() later ? does `set_trainable()` need a "cannot run inside tf.function / only supports eager mode" check?

tests/gpflow/kernels/test_coregion.py
```diff
@@ -67,11 +68,11 @@ def _prepare_models():
     )
     # 2. Coregionalized GPR
     kc = gpflow.kernels.SquaredExponential(active_dims=[0, 1])
-    kc.lengthscales.trainable = False
-    kc.variance.trainable = False  # variance is handles by the coregion kernel
+    set_trainable(kc.lengthscales, False)
+    set_trainable(kc.variance, False)  # variance is handles by the coregion kernel
```
st--:
```suggestion
    set_trainable(kc.variance, False)  # variance is handled by the Coregion kernel
```

gpflow/base.py
joelberkeley-pio: good catch. I could add a warning that prints every time `set_trainable` is used inside `tf.function`

man, i'm yet to be convinced by this `tf.function` thing. I reckon you basically need to assume everything that's not a tensor is immutable inside a `tf.function`. so much room for error

awav: TensorFlow never allowed to change trainable status. Although, there are ways of doing it. No setter gives the same behaviour as a `tf.Variable`. That was a reason why _I said before that removing is okay_.

The `tf.function` is not affected by a `trainable` status of the `tf.Variable`, although you may find cases where it can be a case.

awav:
I do not understand the request about "Should set_trainable() be constrained to only run in eager mode?". The whole TF2.0 is the eager mode and doing `set_trainable` doesn't make sense at all - this is just a utility function.

gpflow/base.py
joelberkeley-pio: @awav i don't understand. do you mind rephrasing?

st--: I'm happy to remove the setter outright. What I was concerned about is calling set_trainable() inside a function that will then get wrapped in tf.function(), and this leading to behaviour that perhaps wasn't expected by the caller (because set_trainable() would probably only be called during the tracing step, not each time)... so a warning when executing in not-eager-mode might be helpful.

tests/gpflow/kernels/test_coregion.py
st--: @joelberkeley-pio

joelberkeley-pio: meh, ok

joelberkeley-pio: out of scope tbh, but sure

gpflow/utilities/utilities.py
```diff
@@ -48,6 +49,13 @@ def set_trainable(model: tf.Module, flag: bool):
     """
     Set trainable flag for all `tf.Variable`s and `gpflow.Parameter`s in a module.
     """
+    if not tf.executing_eagerly():
```
awav: Honestly! This is madness :) you can put this message in every single function of GPflow, just because things might change with `tf.function`. I do not know any case where this can happen - none!

awav: @joelberkeley-pio ^^^

gpflow/utilities/utilities.py
```diff
@@ -48,6 +49,13 @@ def set_trainable(model: tf.Module, flag: bool):
...
+        tf.print(
+            f"{set_trainable.__name__} may exhibit unexpected behaviour when used outside eager"
+            f" mode",
+            output_stream=logging.warning
+        )
```
joelberkeley-pio: i spent all morning trying to figure out how to write a unit test for this - it's tricky cos tensorflow checks the `output_stream` value according to some unknown criteria. I can't work out how to satisfy those criteria with a mock, so alas we'll have to live without a test for this

i've checked manually that it works

gpflow/base.py
joelberkeley-pio: @st-- i've added a warning that shows only if you use `set_trainable` inside `tf.function`

awav:
I see what you’re doing here. However, there are lots of functions that don’t work with `tf.function` (and with tf.functioned functions), and it would be unmanageable to add such a check to all of them. I suggest we leave the check out.

gpflow/utilities/utilities.py
joelberkeley-pio: will remove

gpflow/base.py
joelberkeley-pio: and now removed it

joelberkeley-pio:
> If you are interested, I can make a presentation on how TF is fragile!

i'm not sure i can cope with hearing about how things are fragile atm

PR #1325, opened 2020-03-16
training_loss: another proposal

st--:


gpflow/models/model.py
```diff
@@ -43,41 +43,25 @@ def log_prior_density(self) -> tf.Tensor:
...
         This may be the posterior with respect to the hyperparameters (e.g. for
         GPR) or the posterior with respect to the function (e.g. for GPMC and
         SGPMC). It assumes that maximum_likelihood_objective() is defined
         sensibly.
         """
-        return self._call_maximum_likelihood_objective(data) + self.log_prior_density()
+        return self.maximum_likelihood_objective(*args, **kwargs) + self.log_prior_density()
 
-    def training_loss(self, data: Optional[Data] = None) -> tf.Tensor:
+    def _training_loss(self, *args, **kwargs) -> tf.Tensor:
```
alexisboukouvalas: Docstring discussing data input?
`Data will be an accepted input for BayesianModelWithoutData and will fail on BayesianModelWithData as in the latter case the model already has data.`

st--: I will add that and generally clean it up if this draft is a structure that would make everyone happy

st--:
@alexisboukouvalas and if we put that helper function into the BayesianModel base class we're back at my previous proposal? ;)

st--:
Closed in favour of #1362

PR #1326, opened 2020-03-16
update contributing.md

st--:


contributing.md
```diff
@@ -63,6 +63,10 @@ class Foo:
         return session.run(incr_tensor)
 ```
 
+### Naming conventions
+
+Variable names: scalars and vectors start lowercase, but following the notation used in Gaussian process papers, all matrices are denoted with upper case. For example, `lengthscales` denotes a vector, whereas `Xnew` denotes a matrix.
```
alexisboukouvalas: I still get confused with (N,1)  is a matrix and (N,) is a vector. Perhaps worth calling out in the guide what you mean by the difference between vector and matrix.

contributing.md
```diff
@@ -98,16 +101,13 @@ We use the [semantic versioning scheme](https://semver.org/). The semver implies
...
 ### Keeping up with TensorFlow
 
 GPflow tries to keep up with API changes in TensorFlow as far as is reasonable, so that the latest GPflow will work with the latest stable TensorFlow release. Changing the version of TensorFlow that we're compatible with requires a few tasks:
 
-- Update version used on travis via `travis.yml`
-- Update version used on codeship (requires codeship login)
 - Update `README.md`
-- Update version used by readthedocs.org via `docsrequire.txt`
-- Increment the GPflow version (see below).
+- Increment the GPflow version (see above).
```
st--: Does anything else need to happen with the current setup? I think both CircleCI and readthedocs automatically pull in the latest tensorflow ?

contributing.md
```diff
@@ -98,16 +101,13 @@ We use the [semantic versioning scheme](https://semver.org/). The semver implies
 
 When incrementing the version number, the following tasks are required:
```
st--: Are we *actually* using semver schema? I thought at some point in our discussions it was said that we weren't using it?

contributing.md
st--: clarified. better?

doc/source/conf.py
```diff
@@ -94,9 +94,9 @@
 # built documents.
 #
 # The short X.Y version.
-version = u'1.0.0'
+version = u'2.0.0'
```
st--: Is there some way we can pull this in from the `VERSION` file rather than having to remember to edit it in two places?

contributing.md
```diff
@@ -63,6 +63,10 @@ class Foo:
         return session.run(incr_tensor)
 ```
 
+### Naming conventions
+
+Variable names: scalars and vectors start lowercase, but following the notation used in Gaussian process papers, all matrices are denoted with upper case. For example, `lengthscales` denotes a vector (i.e., rank 1, for example a tensor with shape [D]), whereas `Xnew` denotes a matrix (i.e., rank 2, for example a tensor with shape [N, D]; note that a [N, 1] tensor is a matrix, not a vector).
```
alexisboukouvalas: that's perfect - very clear!

contributing.md
st--: @awav ?

contributing.md
st--: @awav ?

doc/source/conf.py
st--: Or, to rephrase, any reason NOT to just replace this with (some kind of ) `open("../../VERSION").read()`?

contributing.md
awav: Yes, they are pulling the latest version.

doc/source/conf.py
awav: Easier to keep it this way, otherwise we will have to update a release somehow as well. We could create a ticket for that.

st--: I'm not sure I understand what you mean. What would we have to update as well ?
Would you be up for opening a ticket for this - you probably have a better understanding what it's got to be about?

st--: See https://github.com/GPflow/GPflow/issues/1371

PR #1327, opened 2020-03-16
Fix predict_f_samples

vdutor:
`model.predict_f_samples` was broken: it did not work for `full_cov` equals `False`. This PR fixes this and adds a test. The code reuses `sample_mvn` which also makes more sense.

### DIFF
--- a/gpflow/models/model.py
+++ b/gpflow/models/model.py
@@ -20,6 +20,7 @@ import numpy as np
 import tensorflow as tf
 
 from ..base import Module
+from ..conditionals.util import sample_mvn
 from ..config import default_float, default_jitter
 from ..kernels import Kernel
 from ..likelihoods import Likelihood
@@ -118,16 +119,43 @@ class GPModel(BayesianModel):
     ) -> tf.Tensor:
         """
         Produce samples from the posterior latent function(s) at the input points.
+
+        :param Xnew: DataPoint
+            Input locations at which to draw samples
+        :param num_samples: int
+            Number of samples to draw.
+            If `None` a one sample is drawn and the return shape is [..., N, P],
+            for any other positive integer the return shape contains an extra batch
+            dimension, [..., S, N, P], with S = num_samples.
+        :param full_cov: bool, default to `True`,
+            Draw correlated samples over the inputs. Computes the cholesky over the
+            dense covariance matrix of size [num_data, num_data].
+        :param full_output_cov: bool, defaults to `False`.
+            Draw correlated samples over the outputs.
+            The method currently only supports `full_output_cov` equals to `False`.
         """
-        mu, var = self.predict_f(Xnew, full_cov=full_cov)  # [N, P], [P, N, N]
-        num_latent_gps = var.shape[0]
-        num_elems = tf.shape(var)[1]
-        var_jitter = ops.add_to_diagonal(var, default_jitter())
-        L = tf.linalg.cholesky(var_jitter)  # [P, N, N]
-        V = tf.random.normal([num_latent_gps, num_elems, num_samples], dtype=mu.dtype)  # [P, N, S]
-        LV = L @ V  # [P, N, S]
-        mu_t = tf.linalg.adjoint(mu)  # [P, N]
-        return tf.transpose(mu_t[..., np.newaxis] + LV)  # [S, N, P]
+        if full_output_cov:
+            raise NotImplementedError(
+                "`full_output_cov` True is currently not supported by `predict_f_samples`"
+            )
+
+        mean, cov = self.predict_f(Xnew, full_cov=full_cov)  # [N, P], [P, N, N] or [N, P]
+        if full_cov:
+            # mean: [..., N, P]
+            # cov: [..., P, N, N]
+            mean_for_sample = tf.linalg.adjoint(mean)  # [..., P, N]
+            samples = sample_mvn(
+                mean_for_sample, cov, "full", num_samples=num_samples
+            )  # [..., (S), P, N]
+            samples = tf.linalg.adjoint(samples)  # [..., (S), N, P]
+        else:
+            # mean: [..., N, P]
+            # cov: [..., N, P] or [..., N, P, P]
+            cov_structure = "full" if full_output_cov else "diag"
+            samples = sample_mvn(
+                mean, cov, cov_structure, num_samples=num_samples
+            )  # [..., (S), N, P]
+        return samples  # [..., (S), N, P]
 
     def predict_y(
         self, Xnew: DataPoint, full_cov: bool = False, full_output_cov: bool = False
--- a/gpflow/models/model.py
+++ b/gpflow/models/model.py
@@ -128,15 +128,18 @@ class GPModel(BayesianModel):
             for any positive integer the return shape contains an extra batch
             dimension, [..., S, N, P], with S = num_samples.
         :param full_cov:
-            Draw correlated samples over the inputs. Computes the cholesky over the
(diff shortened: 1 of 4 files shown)


gpflow/models/model.py
```diff
@@ -118,16 +119,43 @@ def predict_f_samples(
     ) -> tf.Tensor:
         """
         Produce samples from the posterior latent function(s) at the input points.
+
+        :param Xnew: DataPoint
+            Input locations at which to draw samples
+        :param num_samples: int
+            Number of samples to draw.
+            If `None` a one sample is drawn and the return shape is [..., N, P],
```
```suggestion
            If `None`, a single sample is drawn and the return shape is [..., N, P],
```

gpflow/models/model.py
```diff
@@ -118,16 +119,43 @@ def predict_f_samples(
...
+            for any other positive integer the return shape contains an extra batch
```
```suggestion
            for any positive integer the return shape contains an extra batch
```

gpflow/models/model.py
```diff
@@ -118,16 +119,43 @@ def predict_f_samples(
...
-        num_elems = tf.shape(var)[1]
-        var_jitter = ops.add_to_diagonal(var, default_jitter())
-        L = tf.linalg.cholesky(var_jitter)  # [P, N, N]
-        V = tf.random.normal([num_latent_gps, num_elems, num_samples], dtype=mu.dtype)  # [P, N, S]
-        LV = L @ V  # [P, N, S]
-        mu_t = tf.linalg.adjoint(mu)  # [P, N]
-        return tf.transpose(mu_t[..., np.newaxis] + LV)  # [S, N, P]
+        if full_output_cov:
+            raise NotImplementedError(
+                "`full_output_cov` True is currently not supported by `predict_f_samples`"
```
awav:
```suggestion
                "The `full_output_cov=True` is currently not supported by `predict_f_samples`"
```

gpflow/models/model.py
```diff
@@ -118,16 +119,43 @@ def predict_f_samples(
...
         """
         Produce samples from the posterior latent function(s) at the input points.
+
+        :param Xnew: DataPoint
+            Input locations at which to draw samples
+        :param num_samples: int
+            Number of samples to draw.
+            If `None` a one sample is drawn and the return shape is [..., N, P],
+            for any other positive integer the return shape contains an extra batch
+            dimension, [..., S, N, P], with S = num_samples.
```
st--: how about making this consistent with tensorflow_probability ? would that be possible?

gpflow/models/model.py
```diff
@@ -118,16 +119,43 @@ def predict_f_samples(
     ) -> tf.Tensor:
         """
         Produce samples from the posterior latent function(s) at the input points.
+
+        :param Xnew: DataPoint
+            Input locations at which to draw samples
+        :param num_samples: int
```
```suggestion
        :param num_samples:
```
type is in function signature annotation

st--: NB- should it be `num_samples: Optional[int] = 1` or `num_samples: Optional[int] = None` there?
or: what's the value of the `None` option - should we just get rid of it?

gpflow/models/model.py
```diff
@@ -118,16 +119,43 @@ def predict_f_samples(
...
+            Number of samples to draw.
+            If `None` a one sample is drawn and the return shape is [..., N, P],
+            for any other positive integer the return shape contains an extra batch
+            dimension, [..., S, N, P], with S = num_samples.
+        :param full_cov: bool, default to `True`,
```
```suggestion
        :param full_cov:
```
type and default in signature - don't repeat yourself ;)

gpflow/models/model.py
```diff
@@ -118,16 +119,43 @@ def predict_f_samples(
...
+            Draw correlated samples over the inputs. Computes the cholesky over the
```
```suggestion
            If True, draw correlated samples over the inputs. Computes the Cholesky over the
```

gpflow/models/model.py
```diff
@@ -118,16 +119,43 @@ def predict_f_samples(
...
+            dense covariance matrix of size [num_data, num_data].
```
```suggestion
            dense covariance matrix of size [num_data, num_data].
            If False, draw samples that are uncorrelated over the inputs.
```

gpflow/models/model.py
```diff
@@ -118,16 +119,43 @@ def predict_f_samples(
...
+        :param full_output_cov: bool, defaults to `False`.
```
```suggestion
        :param full_output_cov:
```

gpflow/models/model.py
```diff
@@ -118,16 +119,43 @@ def predict_f_samples(
...
+            Draw correlated samples over the outputs.
```
```suggestion
            If True, draw correlated samples over the outputs.
```

gpflow/models/model.py
```diff
@@ -118,16 +119,43 @@ def predict_f_samples(
...
+            The method currently only supports `full_output_cov` equals to `False`.
```
```suggestion
            Currently, only `full_output_cov=False` is supported.
```

gpflow/models/model.py
```suggestion
                "full_output_cov=True is currently not supported by predict_f_samples"
```

gpflow/models/model.py
```diff
@@ -118,16 +119,43 @@ def predict_f_samples(
...
+            # cov: [..., P, N, N]
+            mean_for_sample = tf.linalg.adjoint(mean)  # [..., P, N]
+            samples = sample_mvn(
+                mean_for_sample, cov, "full", num_samples=num_samples
+            )  # [..., (S), P, N]
+            samples = tf.linalg.adjoint(samples)  # [..., (S), N, P]
+        else:
+            # mean: [..., N, P]
+            # cov: [..., N, P] or [..., N, P, P]
+            cov_structure = "full" if full_output_cov else "diag"
```
st--: above it says full_output_cov=True isn't supported - then why bother handling it here? or is it supported when full_cov=False ?

gpflow/models/model.py
vdutor: We could, but it's not the goal of this PR.

gpflow/models/model.py
vdutor: These shapes follow the conventions of our multi-output framework, which closely matches tfp.

gpflow/models/model.py
```diff
@@ -112,22 +113,52 @@ def predict_f(
...
-        V = tf.random.normal([num_latent_gps, num_elems, num_samples], dtype=mu.dtype)  # [P, N, S]
-        LV = L @ V  # [P, N, S]
-        mu_t = tf.linalg.adjoint(mu)  # [P, N]
-        return tf.transpose(mu_t[..., np.newaxis] + LV)  # [S, N, P]
+        if full_cov and full_output_cov:
+            raise NotImplementedError(
+                "The combination of both `full_cov` and `full_output_cov` is not supported."
+            )
+
+        mean, cov = self.predict_f(Xnew, full_cov=full_cov)  # [N, P], [P, N, N] or [N, P]
```
st--: should it pass full_output_cov as well?
```suggestion
        mean, cov = self.predict_f(Xnew, full_cov=full_cov, full_output_cov=full_output_cov)  # [N, P], [P, N, N] or [N, P]
```
?

vdutor:
I think having a default behaviour of `num_samples = None`, and not adding any batch dimensions makes sense. This function is typically used for plotting anyway.

vdutor:
Thanks for your comments - I'll address the last ones. I'll merge once the tests pass.

gpflow/models/model.py
```diff
@@ -112,22 +113,53 @@ def predict_f(
     def predict_f_samples(
         self,
         Xnew: DataPoint,
-        num_samples: int = 1,
+        num_samples: Optional[int] = 1,
```
```suggestion
        num_samples: Optional[int] = None,
```

PR #1329, opened 2020-03-16
log_prior -> log_prior_density

st--:
As discussed


vdutor:
I thought we agreed on `log_prior_pdf`?

st--:
> I thought we agreed on log_prior_pdf?
Checked with @jameshensman who suggested log_prior_density.

PR #1338, opened 2020-03-18
enable pickling of frozen models

ilia-kats:
When simply replacing a Parameter with tf.constant, tensorflow did not update its internal list of tracked variables, which prevented deepcopying and pickling of frozen models. Explicitly deleting the
tracked variable before assigning a constant value forces tensorflow to update its state.

Before:
```python
import gpflow
import tensorflow as tf
import pickle
X = tf.random.uniform((10,2), -10, 10)
Y = tf.random.uniform((10,1), -10, 10)
model = gpflow.models.GPR((X,Y), kernel=gpflow.kernels.SquaredExponential())
frozen = gpflow.utilities.freeze(model)
pickle.dump(frozen, open("test.pkl", "wb"))
```
```python
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-11-0908f58f59ae> in <module>
----> 1 pickle.dump(frozen, open("test.pkl", "wb"))

TypeError: cannot pickle 'HashableWeakRef' object
```

After:
```python
import gpflow
import tensorflow as tf
import pickle
X = tf.random.uniform((10,2), -10, 10)
Y = tf.random.uniform((10,1), -10, 10)
model = gpflow.models.GPR((X,Y), kernel=gpflow.kernels.SquaredExponential())
frozen = gpflow.utilities.freeze(model)
pickle.dump(frozen, open("test.pkl", "wb"))
loaded = pickle.load(open("test.pkl", "rb"))
```

### DIFF
--- a/gpflow/utilities/utilities.py
+++ b/gpflow/utilities/utilities.py
@@ -265,7 +265,7 @@ def _set_by_name_index(parent: object, value: Any, attr_str: str, index_str: Uni
         index = int(index_str)
         attr = getattr(parent, attr_str)
         attr[index] = value
-        # tensorflow's __setattr__ override does not contain a check for the case where
+        # NOTE: tensorflow's __setattr__ override does not contain a check for the case where
         # an attribute holding a trackable object (e.g. a Variable) is being reassigned
         # a non-trackable object (e.g. a constant). Therefore, tensorflow still stores
         # internal references to the trackable object after reassignment, which can lead
--- a/gpflow/utilities/utilities.py
+++ b/gpflow/utilities/utilities.py
@@ -271,6 +271,7 @@ def _set_by_name_index(parent: object, value: Any, attr_str: str, index_str: Uni
         # internal references to the trackable object after reassignment, which can lead
         # to various problems, for example https://github.com/GPflow/GPflow/pull/1338
         # By calling delattr first, we are forcing tensorflow to remove its internal reference.
+        # This is a tensorflow bug, see https://github.com/tensorflow/tensorflow/issues/37806
         delattr(parent, attr_str)
         setattr(parent, attr_str, attr)
     else:
--- a/gpflow/utilities/utilities.py
+++ b/gpflow/utilities/utilities.py
@@ -274,7 +274,7 @@ def _set_by_name_index(parent: object, value: Any, attr_str: str, index_str: Uni
         delattr(parent, attr_str)
         setattr(parent, attr_str, attr)
     else:
-        delattr(parent, attr_str)
+        delattr(parent, attr_str)  # as above
         setattr(parent, attr_str, value)
 
 
--- a/gpflow/utilities/utilities.py
+++ b/gpflow/utilities/utilities.py
@@ -265,7 +265,10 @@ def _set_by_name_index(parent: object, value: Any, attr_str: str, index_str: Uni
         index = int(index_str)
         attr = getattr(parent, attr_str)
         attr[index] = value
+        delattr(parent, attr_str)
+        setattr(parent, attr_str, attr)
     else:
+        delattr(parent, attr_str)
         setattr(parent, attr_str, value)
 
 
--- a/gpflow/utilities/utilities.py
+++ b/gpflow/utilities/utilities.py
@@ -265,6 +265,12 @@ def _set_by_name_index(parent: object, value: Any, attr_str: str, index_str: Uni
         index = int(index_str)
         attr = getattr(parent, attr_str)
         attr[index] = value
+        # tensorflow's __setattr__ override does not contain a check for the case where
+        # an attribute holding a trackable object (e.g. a Variable) is being reassigned
+        # a non-trackable object (e.g. a constant). Therefore, tensorflow still stores
+        # internal references to the trackable object after reassignment, which can lead
+        # to various problems, for example https://github.com/GPflow/GPflow/pull/1338
+        # By calling delattr first we are forcing tensorflow to remote its internal reference.
         delattr(parent, attr_str)
         setattr(parent, attr_str, attr)
     else:
--- a/gpflow/utilities/utilities.py
+++ b/gpflow/utilities/utilities.py
@@ -270,7 +270,7 @@ def _set_by_name_index(parent: object, value: Any, attr_str: str, index_str: Uni
         # a non-trackable object (e.g. a constant). Therefore, tensorflow still stores
         # internal references to the trackable object after reassignment, which can lead
         # to various problems, for example https://github.com/GPflow/GPflow/pull/1338
-        # By calling delattr first we are forcing tensorflow to remote its internal reference.
+        # By calling delattr first, we are forcing tensorflow to remove its internal reference.
         delattr(parent, attr_str)
         setattr(parent, attr_str, attr)
     else:
--- a/tests/gpflow/utilities/test_deepcopy.py
+++ b/tests/gpflow/utilities/test_deepcopy.py
@@ -71,6 +71,9 @@ def test_freeze():
 
 
 def test_pickle_frozen():
+    """
+    Regression test for the bug described in GPflow/GPflow#1338
+    """
     module = D()
     module_frozen = gpflow.utilities.freeze(module)
 
--- a/tests/gpflow/utilities/test_deepcopy.py
+++ b/tests/gpflow/utilities/test_deepcopy.py
@@ -1,4 +1,5 @@
 import copy
+import pickle
 
 import gpflow
 import pytest
@@ -23,6 +24,18 @@ class B(tf.Module):
 
     def __call__(self, x):
         return self.a(x)
+
+
+class C(tf.Module):
+    def __init__(self):
+        self.var = gpflow.Parameter([2.0], transform=gpflow.utilities.positive(lower=1e-6))
+
+
+class D(tf.Module):
+    def __init__(self):
+        self.var = gpflow.Parameter([10.0], transform=gpflow.utilities.positive())
+        self.var2 = gpflow.Parameter([5.0])
+        self.c = C()
 
 
 class NestedModule(tf.Module):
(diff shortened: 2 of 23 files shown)


awav:
Hello @ilia-kats, 

Thanks a lot for your PR. The problem is slightly deeper than that. By resetting the attribute, you neutralize the aftereffect of `tf.Module` checkpointing structure, but there is a danger that something else doesn't work as expected.

Btw, the error that you see is caused by bijector, you could do `gpflow.utilities.reset_bijector_cache` (not sure about the correct name) function to get rid of the issue.

ilia-kats:
reset_cache_bijectors is already being called from gpflow.utilities.deepcopy, which in turn is being called from gpflow.utilities.freeze(), so that is not the solution here. reset_cache_bijectors only clears the _from_x and _from_y attributes, the problem here are the _self_unconditional_checkpoint_dependencies and _self_unconditional_dependency_names attributes of tf.Module subclasses that are created automatically by tensorflow when a non-constant attribute is assigned.

As far as I can tell, all unit tests still pass with the patch, so at least there is no obvious breakage

ilia-kats:
I can also try to add some unit tests specifically for pickling, if that increases the chances of this getting in

awav:
@ilia-kats, I edited my comment shortly after and I hoped you will not see the message about bijectors, sorry about that. Yes, you are right checkpointing attributes have references to TensorFlow objects and that triggers a problem. I would really like to understand the issue before trying to solve it. Could you elaborate on what is happening with these attributes that belong to trackable abstractions, please?

> As far as I can tell, all unit tests still pass with the patch, so at least there is no obvious breakage

Indeed, although that doesn't mean that our tests cover cases where the issue might appear.

ilia-kats:
As far as I can tell, _self_unconditional_checkpoint_dependencies is storing additional references to the bijectors, which remain when the respective parameter is just reassigned a constant. I did a very quick test, and it appears that clearing the caches also on these bijectors solves the issue, but that does not seem like a clean solution to me.

Trying to simply delete _self_unconditional_checkpoint_dependencies and _self_unconditional_dependency_names raises an exception.

Upon deletion of the Parameter, tensorflow removes all references to it and its children (i.e. bijectors) from _self_unconditional_checkpoint_dependencies. Assignment of the constant then does not change anything, since it is not a trackable object.

st--:
@ilia-kats thank you for investigating this ! Could you please add a summary of why your proposed changes are needed (and perhaps a link to this PR so it is possible to find the details again in the future)? I'd be happy to merge this in - if we encounter issues in the future we should be able to simply revert the change...

tests/gpflow/utilities/test_deepcopy.py
```diff
@@ -57,6 +70,18 @@ def test_freeze():
     assert isinstance(module_frozen.module.module.var, tf.Tensor)
 
 
+def test_pickle_frozen():
```
st--:
```suggestion
def test_pickle_frozen():
    """
    Regression test for the bug described in GPflow/GPflow#1338
    """
```

gpflow/utilities/utilities.py
```diff
@@ -265,7 +265,10 @@ def _set_by_name_index(parent: object, value: Any, attr_str: str, index_str: Uni
         index = int(index_str)
         attr = getattr(parent, attr_str)
         attr[index] = value
+        delattr(parent, attr_str)
+        setattr(parent, attr_str, attr)
     else:
+        delattr(parent, attr_str)
```
st--: Could you please add some comments explaining the issue and why the extra delattr is needed? thanks:)

awav:
> As far as I can tell, _self_unconditional_checkpoint_dependencies is storing additional references to the bijectors, which remain when the respective parameter is just reassigned a constant. I did a very quick test, and it appears that clearing the caches also on these bijectors solves the issue, but that does not seem like a clean solution to me.
> 
> Trying to simply delete _self_unconditional_checkpoint_dependencies and _self_unconditional_dependency_names raises an exception.
> 
> Upon deletion of the Parameter, tensorflow removes all references to it and its children (i.e. bijectors) from _self_unconditional_checkpoint_dependencies. Assignment of the constant then does not change anything, since it is not a trackable object.

@ilia-kats, I have an extra question. _"_self_unconditional_checkpoint_dependencies is storing additional references to the bijectors"_ - this is correct, and I'm curious what happens when you call `delattr` and `setattr`, what these methods trigger in tractable object so that it cleans the references. Why I'm asking this question: tensorflow doesn't handle properly these references and fixing it would not require this fishy `delattr` and `setattr`; plus, `_set_by_name_index` method is generic it will del/set for all attribute types, blindly ignoring that this is not necessary for them.

Thanks!

ilia-kats:
`setattr` calls [`AutoTrackable.__setattr__`](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/training/tracking/tracking.py#L77 ), which in turn calls [`data_structures.sticky_attribute_assignment`](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/training/tracking/data_structures.py#L103). The latter in particular does not have a check for the special case that a trackable is being replaced by a non-trackable, it only handles cases where a non-trackable is replaced by a trackable or a trackable is being replaced by another trackable.

`delattr` calls [`AutoTrackable.__setattr__`](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/training/tracking/tracking.py#L91), which removes the trackable from tensorflow's internal state.

I can try to open an issue in tensorflow, but given that they have over 3k open issues and this is a very special case I'm skeptical that this will get attention any time soon.

gpflow/utilities/utilities.py
```diff
@@ -265,7 +265,16 @@ def _set_by_name_index(parent: object, value: Any, attr_str: str, index_str: Uni
         index = int(index_str)
         attr = getattr(parent, attr_str)
         attr[index] = value
+        # tensorflow's __setattr__ override does not contain a check for the case where
+        # an attribute holding a trackable object (e.g. a Variable) is being reassigned
+        # a non-trackable object (e.g. a constant). Therefore, tensorflow still stores
+        # internal references to the trackable object after reassignment, which can lead
+        # to various problems, for example https://github.com/GPflow/GPflow/pull/1338
+        # By calling delattr first we are forcing tensorflow to remote its internal reference.
```
```suggestion
        # By calling delattr first we are forcing tensorflow to remove its internal reference.
```

gpflow/utilities/utilities.py
```diff
@@ -265,7 +265,16 @@ def _set_by_name_index(parent: object, value: Any, attr_str: str, index_str: Uni
...
+        delattr(parent, attr_str)
```
st--: should this not delete getattr(parent, attr_str)[index] instead ?... :thinking:

gpflow/utilities/utilities.py
```suggestion
        # By calling delattr first, we are forcing tensorflow to remove its internal reference.
```

gpflow/utilities/utilities.py
```diff
@@ -265,7 +265,16 @@ def _set_by_name_index(parent: object, value: Any, attr_str: str, index_str: Uni
...
+        setattr(parent, attr_str, attr)
     else:
+        delattr(parent, attr_str)
```
```suggestion
        delattr(parent, attr_str)  # as above
```

gpflow/utilities/utilities.py
ilia-kats: I am a bit concerned about the performance hit of deleting list elements. If we are deleting in the beginning of the list, all the elements after it will be moved, same when inserting into the list. This way, by deleting the entire list, this is (in theory, I do not know about tensorflow internals to be honest) an O(1) operation, as Python attributes are stored in a dict.

awav:
> `setattr` calls [`AutoTrackable.__setattr__`](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/training/tracking/tracking.py#L77), which in turn calls [`data_structures.sticky_attribute_assignment`](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/training/tracking/data_structures.py#L103). The latter in particular does not have a check for the special case that a trackable is being replaced by a non-trackable, it only handles cases where a non-trackable is replaced by a trackable or a trackable is being replaced by another trackable.
> 
> `delattr` calls [`AutoTrackable.__delattr__`](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/training/tracking/tracking.py#L91), which removes the trackable from tensorflow's internal state.
> 
> I can try to open an issue in tensorflow, but given that they have over 3k open issues and this is a very special case I'm skeptical that this will get any attention anytime soon.
> 
> _edit:_ [tensorflow/tensorflow#37806](https://github.com/tensorflow/tensorflow/issues/37806)

Thanks a lot, @ilia-kats!

> I did a very quick test, and it appears that clearing the caches also on these bijectors solves the issue, but that does not seem like a clean solution to me.

Can you post this workaround here?

I don't see a particular issue with `setattr` and `delattr` and ready to approve this change and leave it till the proper fix will appear in TF. However, I would prefer to stick to the workaround solution. What do others think about it?

st--:
I'd be happy with either merging in these changes as they are, or a more specific workaround - I'd just be keen for this to make it into the 2.0 release!

Either way, I think it'd be good to add a link to the tensorflow issue (thanks for opening it @ilia-kats !) to the workaround code (whether it's the more generic `delattr` or the specific workaround)

ilia-kats:
The basic workaround was
```python
class cls(tf.Module):
    pass
    
test = cls()
test.var = gpflow.Parameter(42, transform=gpflow.utilities.positive())
frozen = gpflow.utilities.freeze(test)

frozen._checkpoint_dependencies[0].ref.transform._from_x.clear()
frozen._checkpoint_dependencies[0].ref.transform._from_y.clear()
```

But this gets very messy with bijector chains. There one has to do:
```python
test = cls()
test.var = gpflow.Parameter(42, transform=gpflow.utilities.positive(lower=1))
frozen = gpflow.utilities.freeze(test)

frozen._checkpoint_dependencies[0].ref.transform._from_x.clear()
frozen._checkpoint_dependencies[0].ref.transform._from_y.clear()

for bijector in frozen._checkpoint_dependencies[0].ref.transform.bijectors:
    bijector._from_x.clear()
    bijector._from_y.clear()
```

ilia-kats:
so what's the plan here?

tests/gpflow/utilities/test_deepcopy.py
```diff
@@ -25,6 +26,18 @@ def __call__(self, x):
         return self.a(x)
 
 
+class C(tf.Module):
```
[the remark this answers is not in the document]
st--: Don't you think it'd be valuable to keep it as a regression test?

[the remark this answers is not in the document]
st--: (but +1 on moving the dummy class definitions inside the test_pickle_frozen function)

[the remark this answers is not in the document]
ilia-kats: This doesn't work, I get `AttributeError: Can't pickle local object 'test_pickle_frozen.<locals>.D'`

st--:
@awav it looks like all comments are now addressed. Should we merge it then? do you want to open an issue to keep track of fixing this properly in the future?

PR #1340, opened 2020-03-18
Fix "comprehensiveness" tests

st--:
There is a test in test_likelihoods that checks whether we missed any likelihood in the test - this is so we are reminded to add tests when we add new likelihoods! But the test was broken, and this PR fixes the test. I've also done the same for the kernel broadcasting test.

### DIFF
--- a/gpflow/ci_utils.py
+++ b/gpflow/ci_utils.py
@@ -38,3 +38,12 @@ def ci_range(n: int, test_n: int = 2):
 
 def ci_list(lst: list, test_n=2):
     return lst[:test_n] if is_continuous_integration() else lst
+
+
+def subclasses(cls):
+    """
+    Generator that returns all (not just direct) subclasses of `cls`
+    """
+    for subclass in cls.__subclasses__():
+        yield from subclasses(subclass)
+        yield subclass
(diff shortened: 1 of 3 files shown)


tests/gpflow/kernels/test_broadcasting.py
```diff
@@ -19,38 +19,67 @@
...
+        kernels.Stationary,
+    ]
+    if kernel_class in tested_kernel_classes:
+        return  # tested
+    if kernel_class in skipped_kernel_classes:
+        return  # not tested but currently expected to fail
+    if kernel_class in abstract_base_classes:
+        return  # cannot test abstract base classes
+    if issubclass(kernel_class, kernels.MultioutputKernel):
+        return  # cannot currently test MultioutputKernels
```
alexisboukouvalas: should this comment be a TODO?

gpflow/ci_utils.py
```diff
@@ -38,3 +38,12 @@ def ci_range(n: int, test_n: int = 2):
 
 def ci_list(lst: list, test_n=2):
     return lst[:test_n] if is_continuous_integration() else lst
+
+
+def subclasses(cls):
+    """
+    Generator that returns all (not just direct) subclasses of `cls`
+    """
+    for subclass in cls.__subclasses__():
```
alexisboukouvalas: I like this - very nicely implemented!

tests/gpflow/kernels/test_broadcasting.py
st--:
```suggestion
        return  # TODO: cannot currently test MultioutputKernels - see https://github.com/GPflow/GPflow/issues/1339
```

PR #1344, opened 2020-03-18
Monitoring of optimisation

vdutor:
This PR enables monitoring of optimisation runs. It provides easy methods to write the model's parameters to TensorBoard, plot custom images to TensorBoard and execute callbacks at each step of the optimisation.

The API looks as follows. Notice that we can make different groups of tasks, that each run at a different frequency.
```python
# Create some monitor tasks
log_dir = "logs"
model_task = ModelToTensorBoard(log_dir, model)
image_task = ImageToTensorBoard(log_dir, plot_prediction, "image_samples")
lml_task = ScalarToTensorBoard(log_dir, lambda: model.log_likelihood(), "lml")

# Plotting tasks can be quite slow, so we want to run them less frequently.
# We group them in a `MonitorTaskGroup` and set the period to 5.
slow_tasks = MonitorTaskGroup(image_task, period=5)

# The other tasks are fast. We run them at each iteration of the optimisation.
fast_tasks = MonitorTaskGroup([model_task, lml_task], period=1)

# We pass both groups to the `Monitor`
monitor = Monitor(fast_tasks, slow_tasks)
```

One of the cool things is that the complete monitor task can be autographed, so it can be part of your optimisation loop `tf.function`:
```
monitor = tf.function(monitor)
```


tests/integration/test_monitor.py
```diff
@@ -0,0 +1,120 @@
...
+    ScalarToTensorBoardTask,
+    TasksCollection,
+    ImageToTensorBoardTask,
+)
+
+
+class Data:
+    num_data = 20
+    num_steps = 10
+    log_dir = tempfile.gettempdir()
```
st--: The test might be more robust with using pytest's `tmp_path` fixture, this will also handle eventually deleting them again. This could then be used to also test that a tensorboard was actually written out to disk?

tests/integration/test_monitor.py
```diff
@@ -0,0 +1,120 @@
...
+    opt = tf.optimizers.Adam()
+    for step in range(Data.num_steps):
+        opt.minimize(closure, model.trainable_variables)
+        tasks(step)
+
+    size_after = _get_size_directory(Data.log_dir)
+    assert size_after > size_before
+
+
+def test_ImageToTensorBoardTask():
```
st--: Is there some way we could test the output of this (and the following) tests?
If that's too complex, that's okay, but it might be helpful to make clear in the test name that these are smoke tests, not functionality tests...

gpflow/utilities/monitor.py
```diff
@@ -0,0 +1,231 @@
...
+        self.file_writer = tf.summary.create_file_writer(log_dir)
+
+    def __call__(self, step, **kwargs):
+        with self.file_writer.as_default():
+            super().__call__(step, **kwargs)
+        self.file_writer.flush()
+
+    @abstractmethod
+    def run(self, **kwargs):
+        raise NotImplementedError
```
st--:
```suggestion
```
Same as in parent class, so can delete this

gpflow/utilities/monitor.py
```diff
@@ -0,0 +1,231 @@
...
+        :param kwargs: keyword arguments available to the run method.
+        """
+        raise NotImplementedError
+
+
+class TensorBoardTask(MonitorTask):
+    def __init__(self, log_dir: str, period: int = 1):
+        """
+        :param log_dir: directory in which to store the tensorboard files.
+            Can be a nested. E.g. ./logs/my_run/
```
st--: "can be a nested" sounds like it's missing something?

gpflow/utilities/monitor.py
```diff
@@ -0,0 +1,231 @@
...
+        :param period: interval at which to run the task.
+            For large values of`period` the task will be less frequently ran.
```
st--:
```suggestion
            For large values of `period` the task will be less frequently run.
```
would it be more clear to just write
```
:param period: defines how often to run the task; it will execute every `period`th step.
```
?

gpflow/utilities/monitor.py
```diff
@@ -0,0 +1,231 @@
...
+        self.period = period
+
+    def __call__(self, step: int, **kwargs):
+        """
+        It calls the 'run' function and sets the current step.
+
+        :param step: current step in the optimisation.
+        :param kwargs: additional key-word arguments that can be passed
+            to the `run` method of the task. This is in particular handy for
+            passing keyword argument to the callback of `ScalarToTensorBoardTask`.
```
st--: how would the caller know what kinds of keyword args should be passed ?

gpflow/utilities/monitor.py
```diff
@@ -0,0 +1,231 @@
...
+        raise NotImplementedError
+
+
+class ModelToTensorBoardTask(TensorBoardTask):
+    """
+    Monitoring task that creates a sensible TensorBoard for a model.
+
+    By default, it writes to the TensorBoard all model parameters that are scalars,
+    for parameter arrays (e.g. kernel.lengthscales) the values are sent to the TensorBoard
+    if the array is smaller than 3. This behaviour can be adjusted using `max_size`.
```
st--:
```suggestion
    if the array has no more than `max_size` (default: 3) elements.
```

tests/integration/test_monitor.py
vdutor: done

st--:
> I don't remember why we had run method in the previous gpflow, but here it looks redundant. You could make these objects simpler just moving implementation to __call__ method.

It's not redundant -- `__call__` is executed at each step, `run` is executed at the specified frequency (each `period` steps). I think that separation is very clean actually.

st--:
+1 on the shorter names

tests/integration/test_monitor.py
```diff
@@ -0,0 +1,139 @@
...
+    def closure():
+        return -1.0 * model.log_likelihood()
+
+    opt = tf.optimizers.Adam()
+    for step in range(Data.num_steps):
+        opt.minimize(closure, model.trainable_variables)
+        monitor(step)
+
+    size_after = _get_size_directory(tmp_path)
+    assert size_after > size_before
```
st--: Does tensorflow write any error files when things go wrong? Would it be possible to explicitly check for the existence of the correct tensorboard file ? (I know there's some "randomness" in the path due to the timestamp, but could probably handle that with glob.glob()?)

awav:
> I don't remember why we had run method in the previous gpflow, but here it looks redundant. You could make these objects simpler just moving implementation to __call__ method.

To clarify this point: to make monitoring lightweight, we could split responsibility for each monitoring "task" making interfaces less overloaded with arguments. E.g. a capturing images task could cook only images for  Tensorboard and an aggregation task could decide when to run children tasks.

vdutor:
Hi @awav 

In order:
- *done* - I chose these names for continuity with GPflow1, but your suggestions are also fine for me. I updated them.
- the `__call__` is handled by the monitoring framework and the `run` is the method that the user has to implement for custom behaviour of the task. This is inspired by how `Keras` handles the monitoring.
- *done*
- Multi-dimensional arrays are flattened first and then we print the first `max_size` elements. This imitates the behaviour of `tf.Print(tensor)`
- I re-wrote the `run` method of our monitoring tasks in such a way that we can wrap them in a `tf.function`. This makes them fast - I did some preliminary `timeit` experiments to check this. Also, the `file_writer.as_default` inside the loop is the recommended way - given in the [docs (snippet 2)](https://www.tensorflow.org/api_docs/python/tf/summary).

Hope this address your comments.

gpflow/utilities/monitor.py
vdutor: by checking the matplotlib docs. I'll make this clearer.

gpflow/utilities/monitor.py
vdutor: thanks. done

tests/integration/test_monitor.py
vdutor: This would cost me more time and don't really see the benefit right now. I check them manually and they look fine.

vdutor: I'll indicate that the tests are smoke tests though.

vdutor:
hey @st--  and @awav : I addressed your comments. Please check and approve PR.

tests/integration/test_monitor.py
```diff
@@ -0,0 +1,156 @@
...
+    compiled_tasks = tf.function(tasks.__call__)
+
+    with pytest.raises(TypeError, match=r".*got an unexpected keyword argument 'y'.*"):
+        tasks(0, y=1.0)
+
+    with pytest.raises(TypeError, match=r".*got an unexpected keyword argument 'y'.*"):
+        compiled_tasks(0, y=1.0)
+
+
+def test_ModelToTensboard(model, tmp_path):
```
st--:
```suggestion
def test_ModelToTensorboard(model, tmp_path):
```

tests/integration/test_monitor.py
```diff
@@ -0,0 +1,156 @@
...
+    """ Smoke test `ModelToTensorBoard` in Eager and Compiled mode """
+    tmp_path = str(tmp_path)
+    task = ModelToTensorBoard(tmp_path, model)
+    compiled_task = tf.function(task.__call__)
```
st--: would this work with `tf.function(task)` ? :thinking:

tests/integration/test_monitor.py
```diff
@@ -0,0 +1,156 @@
...
+
+def test_ImageToTensorBoard(tmp_path):
+    """ Smoke test `ImageToTensorBoard` in Eager and Compiled mode """
+    tmp_path = str(tmp_path)
+
+    def plotting_cb(fig, axes):
+        axes[0, 0].plot(np.random.randn(100), np.random.randn(100), ".")
+        axes[1, 0].plot(np.random.randn(100), np.random.randn(100), ".")
+        axes[0, 1].plot(np.random.randn(100), np.random.randn(100), ".")
+        axes[1, 1].plot(np.random.randn(100), np.random.randn(100), ".")
```
st--: I don't want to optimize too prematurely, but plotting can be quite slow (especially with the Agg backend), why not simply plot two or three points per plot?
```suggestion
        axes[0, 0].plot(np.random.randn(2), np.random.randn(2), ".")
        axes[1, 0].plot(np.random.randn(2), np.random.randn(2), ".")
        axes[0, 1].plot(np.random.randn(2), np.random.randn(2), ".")
        axes[1, 1].plot(np.random.randn(2), np.random.randn(2), ".")
```

tests/integration/test_monitor.py
```diff
@@ -0,0 +1,156 @@
...
+
+def test_ScalarToTensorBoard(tmp_path, capfd):
+    """ Smoke test `ScalarToTensorBoard` in Eager and Compiled mode """
+    tmp_path = str(tmp_path)
+
+    def scalar_cb():
+        return 0.0
+
+    task = ScalarToTensorBoard(tmp_path, scalar_cb, "scalar")
+    compiled_task = tf.function(task.__call__)
```
st--:
```suggestion
    compiled_task = tf.function(task)
```
do you need the `.__call__`? it seems to handle it fine in `test_compile_monitor`?

tests/integration/test_monitor.py
```diff
@@ -0,0 +1,156 @@
...
+
+    task(0)
+    compiled_task(0)
+
+
+def test_ScalarToTensorBoard_with_argument(tmp_path):
+    """ Smoke test `ScalarToTensorBoard` in Eager and Compiled mode """
+    tmp_path = str(tmp_path)
+
+    def scalar_cb(x=None):
```
st--:
```suggestion
    def scalar_cb(x):
```

gpflow/monitor.py
```diff
@@ -0,0 +1,245 @@
...
+
+
+class ImageToTensorBoard(ToTensorBoard):
+    def __init__(
+        self,
+        log_dir: str,
+        plotting_function: Callable[[Figure, Axes], Figure],
+        name: Optional[str] = None,
+        period: int = 1,
+        *,
```
st--:
```suggestion
        name: Optional[str] = None,
        *,
        period: int = 1,
```
With this many arguments calling `ImageToTensorBoard("./logs/my_run", plotting_callback, "my_plot", 5)` already seems pretty unintelligible - forcing people to explicitly write `period=5` could make this clearer. (Or `interval=5` :wink: )

gpflow/monitor.py
```diff
@@ -0,0 +1,245 @@
...
+
+class MonitorTask(ABC):
+    """
+    A base class for a monitoring task.
+
+    All monitoring tasks are callable objects.
+    A descendant class must implement the `run` method, which is the body of the monitoring task.
+    """
+
+    def __init__(self, period: int = 1):
```
st--: nit: I think `interval` would make slightly more intuitive sense than `period` - would you be up for a search&replace on that ?

gpflow/monitor.py
```diff
@@ -0,0 +1,245 @@
...
+        """
+        self.current_step = 0
+        self.period = period
+
+    def __call__(self, step: int, **kwargs):
+        """
+        It calls the 'run' function and sets the current step.
+
+        :param step: current step in the optimisation.
+        :param kwargs: additional key-word arguments that can be passed
```
st--:
```suggestion
        :param kwargs: additional keyword arguments that can be passed
```

gpflow/monitor.py
```diff
@@ -0,0 +1,245 @@
...
+        :param kwargs: keyword arguments available to the run method.
+        """
+        raise NotImplementedError
+
+
+class ToTensorBoard(MonitorTask):
+    def __init__(self, log_dir: str, period: int = 1):
+        """
+        :param log_dir: directory in which to store the tensorboard files.
+            Can be a nested. E.g. ./logs/my_run/
```
st--:
```suggestion
            Can be nested, e.g. ./logs/my_run/
```

gpflow/monitor.py
```diff
@@ -0,0 +1,245 @@
...
+
+
+class ModelToTensorBoard(ToTensorBoard):
+    """
+    Monitoring task that creates a sensible TensorBoard for a model.
+
+    By default, it writes to the TensorBoard all model parameters that are scalars,
+    for parameter arrays (e.g. kernel.lengthscales) the values are sent to the TensorBoard
+    if the array has no more than `max_size` (default: 3) elements.
+    This behaviour can be adjusted using `max_size`.
```
st--: This doesn't seem to quite match up with the behaviour coded below (which by default only takes kernel and likelihood parameters, but all of them - except for vectors it will be at most the `max_size` first entries) - is it up to date?

gpflow/monitor.py
```diff
@@ -0,0 +1,245 @@
...
+    """
+
+    def __init__(
+        self,
+        log_dir: str,
+        model: BayesianModel,
```
st--: for a BayesianModel that is not also a GPModel, this ModelToTensorBoard would by default not do anything... do you want to specify this to be a GPModel?

gpflow/monitor.py
```diff
@@ -0,0 +1,245 @@
...
+            Can be a nested: for example, './logs/my_run/'.
+        :param model: model to be monitord.
+        :param max_size: maximum size of arrays (incl.) to store each
+            element of the array independently as a scalar in the TensorBoard.
+        :param period: defines how often to run the task; it will execute every `period`th step.
+            For large values of `period` the task will be less frequently run.
+        :param extra_keywords: pass additional keywords to be monitored.
+            If the parameter's name includes any of the extra keywords specified it
+            will be monitored as well. By default, parameters that match the `kernel` or
+            `likelihood` keyword are monitored.
```
st--: I'm wondering if it might be clearer if instead there was a `keywords` argument that simply had `["kernel", "likelihood"]` as its default value? rather than hardcoding those two?

gpflow/monitor.py
```diff
@@ -0,0 +1,245 @@
...
+                self._summarize_parameter(name, parameter)
+
+    def _summarize_parameter(self, name: str, value: Union[float, np.ndarray]):
+        """
+        :param name: identifier used in tensorboard
+        :param value: value to be stored in tensorboard
+        """
+        value = tf.reshape(value, (-1,))
+        for i in tf.range(tf.minimum(tf.size(value), self.max_size)):
+            tf.summary.scalar(f"{name}[{i}]", value[i], step=self.current_step)
```
st--: does this mean all actually-scalar parameters (e.g. `variance`) would still end up in the tensorboard as `variance[0]` rather than just `variance`? would it be possible to avoid that and only add the index if there is more than one?

gpflow/monitor.py
```diff
@@ -0,0 +1,245 @@
...
+
+
+class ScalarToTensorBoard(ToTensorBoard):
+    """ Stores the returns value of a callback in a TensorBoard. """
```
st--:
```suggestion
    """ Stores the return value of a callback in a TensorBoard. """
```

gpflow/monitor.py
```diff
@@ -0,0 +1,245 @@
...
+    for parameter arrays (e.g. kernel.lengthscales) the values are sent to the TensorBoard
+    if the array has no more than `max_size` (default: 3) elements.
+    This behaviour can be adjusted using `max_size`.
+    """
+
+    def __init__(
+        self,
+        log_dir: str,
+        model: BayesianModel,
+        max_size: int = 3,
```
st--:
```suggestion
        *,
        max_size: int = 3,
```

gpflow/monitor.py
```diff
@@ -0,0 +1,245 @@
...
+            self.axes = self.fig.subplots(**self.subplots_kw)
+        else:
+            self.axes = self.fig.add_subplot(111)
+
+    def _clear_axes(self):
+        try:
+            for ax in self.axes.flatten():
+                ax.clear()
+        except (AttributeError, TypeError):
+            self.axes.clear()
```
st--: nit/thought (happy to keep it as is, too):
```suggestion
        if isinstance(self.axes, np.ndarray):
            for ax in self.axes.flatten():
                ax.clear()
        else:
            self.axes.clear()
```
?

doc/source/notebooks/intro_to_gpflow2.pct.py
```diff
@@ -223,52 +223,68 @@ def simple_training_loop(model: gpflow.models.SVGP, epochs: int = 1, logging_epo
 # %% [markdown]
 # ## Monitoring
 #
-# We can monitor the training procedure using `tf.summary`. First we create a summary writer object through which we can write scalars and images.
+# We can monitor the training procedure using gpflow monitoring utilities.
+# For a more detailed tutorial see [GPflow monitoring notebook](./basics/monitoring.ipynb).
```
st--:
```suggestion
# For a more detailed tutorial see the [monitoring notebook](./basics/monitoring.ipynb).
```

doc/source/notebooks/intro_to_gpflow2.pct.py
```diff
@@ -223,52 +223,68 @@ def simple_training_loop(model: gpflow.models.SVGP, epochs: int = 1, logging_epo
 # %% [markdown]
 # ## Monitoring
 #
-# We can monitor the training procedure using `tf.summary`. First we create a summary writer object through which we can write scalars and images.
+# We can monitor the training procedure using gpflow monitoring utilities.
```
st--:
```suggestion
# `gpflow.monitor` provides a thin wrapper on top of tf.summary that makes it easy to monitor the training procedure.
```

doc/source/notebooks/intro_to_gpflow2.pct.py
```diff
@@ -223,52 +223,68 @@ def simple_training_loop(model: gpflow.models.SVGP, epochs: int = 1, logging_epo
...
+    ax.plot(samples_input, samples[:, :, 0].numpy().T, "C0", linewidth=0.5)
+    ax.set_ylim(-2.0, +2.0)
+    ax.set_xlim(0, 10)
+
 
-samples_input = to_default_float(np.linspace(0, 10, 100).reshape(100, 1))
+output_logdir = enumerated_logdir()
+model_task = ModelToTensorBoard(output_logdir, model)
+image_task = ImageToTensorBoard(output_logdir, plot_model, "samples_image")
+elbo_task = ScalarToTensorBoard(output_logdir, lambda data=None: model.elbo(data).numpy(), "elbo")
```
st--:
```suggestion
elbo_task = ScalarToTensorBoard(output_logdir, lambda data: model.elbo(data).numpy(), "elbo")
```
I don't think elbo() is going to work without data?

doc/source/notebooks/intro_to_gpflow2.pct.py
```diff
@@ -223,52 +223,68 @@ def simple_training_loop(model: gpflow.models.SVGP, epochs: int = 1, logging_epo
...
+monitor = MonitorCollection([model_task, image_task, elbo_task])
 
 
 def monitored_training_loop(
-    model: gpflow.models.SVGP,
-    logdir: str,
-    epochs: int = 1,
-    logging_epoch_freq: int = 10,
-    num_samples: int = 10,
+    epochs: int = 1, logging_epoch_freq: int = 10,
```
st--:
```suggestion
    epochs: int = 1, logging_epoch_interval: int = 10,
```

doc/source/notebooks/intro_to_gpflow2.pct.py
```diff
@@ -223,52 +223,68 @@ def simple_training_loop(model: gpflow.models.SVGP, epochs: int = 1, logging_epo
...
 
-                summary_matplotlib_image(dict(model_samples=fig), step=epoch)
-                tf.summary.scalar("elbo", data=model.elbo(data), step=epoch)
-                tf.summary.scalar("likelihood/variance", data=model.likelihood.variance, step=epoch)
-                tf.summary.scalar("kernel/lengthscales", data=model.kernel.lengthscales, step=epoch)
-                tf.summary.scalar("kernel/variance", data=model.kernel.variance, step=epoch)
+        epoch_id = epoch + 1
+        if epoch_id % logging_epoch_freq == 0:
+            tf.print(f"Epoch {epoch_id}: ELBO (train) {model.elbo(data)}")
+            monitor(epoch, data=batch)
```
st--: why not call it every step (monitor should be handling the only-every-nth-step/epoch)?

st--: also, how about showing how you can compile it ? (or at least mention it here and refer to the main monitoring notebook)?

doc/source/notebooks/basics/monitoring.pct.py
```diff
@@ -0,0 +1,142 @@
...
+# For this tutorial we set up three tasks:
+# - `ModelToTensorBoard`: writes the models hyper-parameters such as `likelihood.variance` and `kernel.lengthscales` to a TensorBoard.
+# - `ImageToTensorBoard`: writes custom matplotlib images to a TensorBoard.
+# - `ScalarToTensorBoard`: writes any scalar value to a TensorBoard. Here, we use it to write the model's `objective`.
+
+# %%
+log_dir = "logs"  # Directory where TensorBoard files will be written.
+model_task = ModelToTensorBoard(log_dir, model)
+image_task = ImageToTensorBoard(log_dir, plot_prediction, "image_samples")
+lml_task = ScalarToTensorBoard(log_dir, lambda: model.log_likelihood().numpy(), "lml")
```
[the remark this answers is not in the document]
vdutor: can be both.

doc/source/notebooks/intro_to_gpflow2.pct.py
```diff
@@ -223,52 +223,68 @@ def simple_training_loop(model: gpflow.models.SVGP, epochs: int = 1, logging_epo
...
+            tf_optimization_step(model, batch)
 
-                summary_matplotlib_image(dict(model_samples=fig), step=epoch)
-                tf.summary.scalar("elbo", data=model.elbo(data), step=epoch)
-                tf.summary.scalar("likelihood/variance", data=model.likelihood.variance, step=epoch)
-                tf.summary.scalar("kernel/lengthscales", data=model.kernel.lengthscales, step=epoch)
-                tf.summary.scalar("kernel/variance", data=model.kernel.variance, step=epoch)
+        epoch_id = epoch + 1
+        if epoch_id % logging_epoch_freq == 0:
+            tf.print(f"Epoch {epoch_id}: ELBO (train) {model.elbo(data)}")
```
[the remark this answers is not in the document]
vdutor: yes this works.

doc/source/notebooks/basics/monitoring.pct.py
```diff
@@ -0,0 +1,142 @@
...
+monitor = MonitorCollection([model_task, lml_task])
+
+
+@tf.function
+def step(i):
+    opt.minimize(lambda: -1.0 * model.log_likelihood(), model.trainable_variables)
+    monitor(i)
+
+
+for i in tf.range(optimisation_steps):
```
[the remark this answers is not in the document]
vdutor: done

doc/source/notebooks/basics/monitoring.pct.py
```diff
@@ -0,0 +1,142 @@
...
+
+# %%
+opt = tf.optimizers.Adam()
+
+model_task = ModelToTensorBoard(log_dir, model)
+lml_task = ScalarToTensorBoard(log_dir, lambda: model.log_likelihood(), "lml")
+# Note that the `ImageToTensorBoard` task can not be compiled, and is omitted from the collection
+monitor = MonitorCollection([model_task, lml_task])
+
+
```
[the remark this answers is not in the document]
vdutor: done

gpflow/monitor.py
vdutor: done. makes it neater indeed.

tests/integration/test_monitor.py
vdutor: yes works indeed. done.

markvdw:
This looks great!

doc/source/notebooks/basics/monitoring.pct.py
```diff
@@ -0,0 +1,146 @@
...
+# Set up model and print
+
+kernel = gpflow.kernels.SquaredExponential(lengthscales=[1.0, 2.0]) + gpflow.kernels.Linear()
+model = gpflow.models.GPR((X, Y), kernel, noise_variance=noise_std ** 2)
+model
+
+
+# %%
+# We define a function that plots the model's prediction (in the form of samples) together with the data.
+# Importantly, this function has no other argument than `fig: matplotlib.figure.Figure` and `axes: matplotlib.figure.Axes`.
```
st--:
```suggestion
# Importantly, this function has no other argument than `fig: matplotlib.figure.Figure` and `ax: matplotlib.figure.Axes`.
# Figure and axes objects are passed by position, not name, so it does not matter what they are called in the `plot_prediction` signature.
```

doc/source/notebooks/intro_to_gpflow2.pct.py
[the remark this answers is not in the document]
st--: Not within tf.function

doc/source/notebooks/intro_to_gpflow2.pct.py
```diff
@@ -223,52 +223,73 @@ def simple_training_loop(model: gpflow.models.SVGP, epochs: int = 1, logging_epo
...
+monitor = MonitorCollection([model_task, image_task, elbo_task])
 
 
 def monitored_training_loop(
-    model: gpflow.models.SVGP,
-    logdir: str,
-    epochs: int = 1,
-    logging_epoch_freq: int = 10,
-    num_samples: int = 10,
+    epochs: int = 1, logging_epoch_freq: int = 10,
```
st--:
```suggestion
    epochs: int = 1, logging_epoch_period: int = 10,
```

gpflow/monitor.py
```diff
@@ -0,0 +1,251 @@
...
+    ):
+        """
+        :param log_dir: directory in which to store the tensorboard files.
+            Can be a nested: for example, './logs/my_run/'.
+        :param model: model to be monitord.
+        :param max_size: maximum size of arrays (incl.) to store each
+            element of the array independently as a scalar in the TensorBoard.
+        :param period: defines how often to run the task; it will execute every `period`th step.
+            For large values of `period` the task will be less frequently run.
+        :param keyword_to_monitor: specifies keywords to be monitored.
```
st--:
```suggestion
        :param keywords_to_monitor: specifies keywords to be monitored.
```

doc/source/notebooks/intro_to_gpflow2.pct.py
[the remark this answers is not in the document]
st--: so I think it'd be good to show something that makes it at least potentially compilable, even if you're not demonstrating it here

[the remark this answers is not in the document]
vdutor: Oh I see - thanks. I didn't know the f-string don't work when autographed.

vdutor:
> Other than that, a key point that I think would be good to address before merging is whether it might be possible to get the ImageToTensorBoard working inside tf.function() as well by using tf.py_function()?

Hi @st-- , I had a look at `tf.py_function` to wrap the plotting callback in `ImageToTensorBoard`, but it's not applicable here. A `tf.py_function` only accepts a list of `tf.Tensor`s.

vdutor:
@awav  and @st-- , could you please review the latest changes. Main differences are more and better functionality tests and the addition of groups of tasks.

gpflow/monitor.py
```diff
@@ -0,0 +1,318 @@
...
+        self.keywords_to_monitor = keywords_to_monitor
+
+    def run(self, **unused_kwargs):
+        for name, parameter in parameter_dict(self.model).items():
+            # check if the parameter name matches any of the specified keywords
+            if any(keyword in name for keyword in self.keywords_to_monitor):
+                name = name.lstrip(".")  # keys are prepended with a '.', which we strip
+                self._summarize_parameter(name, parameter)
+
+    # @tf.function(autograph=False)
```
vdutor:
```suggestion
```

gpflow/monitor.py
vdutor: To show that users can also use the monitor with their own Bayesian model. As long as the model has any `tf.Variable`s or `gpflow.Parameter`s this monitor task will do something sensible.

st--:
Note that you need to fix the broken link that got commented out by #1369.

gpflow/monitor.py
```diff
@@ -0,0 +1,329 @@
...
+        self.tasks = task_or_tasks
+        self._period = period
+
+    @property
+    def tasks(self) -> List[MonitorTask]:
+        return self._tasks
+
+    @tasks.setter
+    def tasks(self, task_or_tasks: Union[List[MonitorTask], MonitorTask]) -> None:
+        """ Ensures the tasks are stored as a list. Even if there is only a single task. """
```
awav:
```suggestion
        """Ensures the tasks are stored as a list. Even if there is only a single task."""
```

gpflow/monitor.py
```diff
@@ -0,0 +1,329 @@
...
+        if not isinstance(task_or_tasks, List):
+            self._tasks = [task_or_tasks]
+        else:
+            self._tasks = task_or_tasks
+
+    def __call__(self, step, **kwargs):
+        """ Call each task in the group """
```
awav:
```suggestion
        """Call each task in the group."""
```

tests/gpflow/test_monitor.py
```diff
@@ -0,0 +1,296 @@
...
+                ScalarToTensorBoard(tmp_path, elbo_callback, "elbo"),
+            ],
+            period=2,
+        ),
+        MonitorTaskGroup(ExecuteCallback(print_callback), period=1,),
+    )
+
+
+def _get_size_directory(dir):
+    """ Calculating the size of a directory (in Bytes) """
```
awav:
```suggestion
    """Calculating the size of a directory (in Bytes)."""
```

tests/gpflow/test_monitor.py
```diff
@@ -0,0 +1,296 @@
...
+        print("ExecuteCallback test")
+
+    task = ExecuteCallback(callback)
+    task(0)
+    compiled_task = tf.function(task)
+    compiled_task(0)
+
+
+def test_ImageToTensorBoard(tmp_path):
+    """ Smoke test `ImageToTensorBoard` in Eager and Compiled mode """
```
awav:
```suggestion
    """Smoke test `ImageToTensorBoard` in Eager and Compiled mode."""
```

tests/gpflow/test_monitor.py
```diff
@@ -0,0 +1,296 @@
...
+        tmp_path, plotting_cb, "image", fig_kw=fig_kwargs, subplots_kw=subplots_kwargs
+    )
+
+    task(0)
+    compiled_task = tf.function(task)
+    compiled_task(0)
+
+
+def test_ScalarToTensorBoard(tmp_path):
+    """ Smoke test `ScalarToTensorBoard` in Eager and Compiled mode """
```
awav:
```suggestion
    """Smoke test `ScalarToTensorBoard` in Eager and Compiled mode."""
```

tests/gpflow/test_monitor.py
```diff
@@ -0,0 +1,296 @@
...
+        return 0.0
+
+    task = ScalarToTensorBoard(tmp_path, scalar_cb, "scalar")
+    task(0)
+    compiled_task = tf.function(task)
+    compiled_task(0)
+
+
+def test_ScalarToTensorBoard_with_argument(tmp_path):
+    """ Smoke test `ScalarToTensorBoard` in Eager and Compiled mode """
```
awav:
```suggestion
    """Smoke test `ScalarToTensorBoard` in Eager and Compiled mode."""
```

tests/gpflow/test_monitor.py
```diff
@@ -0,0 +1,296 @@
...
+
+    with pytest.raises(TypeError, match=r".*got an unexpected keyword argument 'y'.*"):
+        task(0, y=1.0)
+
+    with pytest.raises(TypeError, match=r".*got an unexpected keyword argument 'y'.*"):
+        compiled_task(0, y=1.0)
+
+
+def test_ModelToTensorboard(model, tmp_path):
+    """ Smoke test `ModelToTensorBoard` in Eager and Compiled mode """
```
awav:
```suggestion
    """Smoke test `ModelToTensorBoard` in Eager and Compiled mode."""
```

tests/gpflow/test_monitor.py
```diff
@@ -0,0 +1,296 @@
...
+    # Autograph mode
+    compiled_task = tf.function(task)
+    for i in tf.range(Data.num_steps):
+        compiled_task(i)
+    out, _ = capsys.readouterr()
+    assert out == f"{string_to_print}\n"
+
+
+def test_periodicity_group(capsys):
+    """ Test that groups are called at different periods """
```
awav:
```suggestion
    """Test that groups are called at different periods."""
```

Issue #1252, opened 2020-02-10
Missing documentations (template issue to list things related to documentation that are missing)


yiwei-prowler:
- [x] ~The method GPR.predict_f is missing return type annotation -> MeanAndVariance.~ see #1346 
- [x] The method GPModel.predict_log_density is missing documentation for its parameters, returned value, and is missing the returned type. It also needs shape assertions for its return type.
- [x] The method GPR.log_likelihood is missing returned type annotation.
- [x] ~SVGP.predict_f says its return type is tf.Tensor, but its ancester method says the returned type is MeanAndVariance, which is a tuple. These two types are not compatable.~ see #1346 
- [x] ~SVGP.log_likelihood declares its argument data to have type Tuple[tf.Tensor, tf.Tensor], but there are many places that use the type alias Data, this is confusing. Don't know where should each be used -- I'm working on warping support, and don't know when referencing to data, which type I should use.~ see #1346 
- [x] BayesianModel.log_prior, its comment is missing, the shape of its return value is missing.
- [x] BayesianModel.log_marginal_likelihood, its comment is missing, the shape of its return type is missing..
- [x] BayesianModel.log_likelihood, its comment is missing, the shape of its return type is missing.
- [x] ~GPR.predict_f_samples. The parameter full_cov: bool = True indicates that its value can be either True of False. But try it with False crashes.~ see #1327

st--:
@yiwei-prowler it would be great if you could help us out and contribute this simple fix as a PR. thank you!

joelberkeley:
@yiwei-prowler i can do this if you like. more generally, I don't think we should assert shapes on anything but arguments - that responsibility lies with the tests

st--:
As far as I can tell most of this has actually happened along the way in #1276 and other PRs.

PR #1346, opened 2020-03-18
Clean up type annotations and doc strings

### LINKED ISSUE #1248: The constructor of SGPR is confusing
In the constructor of SGPR class, the parameter inducing_variable: Optional[InducingPoints] = None, which means if not provided, a None value is used. But a None value will cause crash.

Also, the header comment of this __init__ method is out-dated. The comments mentions X and Y, But X and Y are not defined. There is no explanation of inducing_variable, num_latent and noise_variance.

### LINKED ISSUE #1249: Documentation of GPModel.predict_f_samples incomplete
In GPModel.predict_f_samples, the header comment is incomplete. explanation for the parameters are missing. In the implementation, shape indications such as [S, N, P] needs to be explained better. Also consider writing shape assertions too.

st--:
Resolves #1248, resolves #1249, and addresses several points in #1252.

### DIFF
--- a/gpflow/models/sgpr.py
+++ b/gpflow/models/sgpr.py
@@ -358,19 +358,19 @@ class GPRFITC(SGPRBase):
         w = tf.linalg.triangular_solve(Luu, Kus, lower=True)  # [M, N]
 
         tmp = tf.linalg.triangular_solve(tf.transpose(L), gamma, lower=False)
-        mean = tf.linalg.matmul(w, tmp, transpose_a=True) + self.mean_function(X)
+        mean = tf.linalg.matmul(w, tmp, transpose_a=True) + self.mean_function(Xnew)
         intermediateA = tf.linalg.triangular_solve(L, w, lower=True)
 
         if full_cov:
             var = (
-                self.kernel(X)
+                self.kernel(Xnew)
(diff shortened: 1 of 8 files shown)


gpflow/models/sgpr.py
```diff
@@ -346,33 +347,33 @@ def log_likelihood(self):
...
 
-    def predict_f(self, X: tf.Tensor, full_cov=False, full_output_cov=False) -> MeanAndVariance:
+    def predict_f(self, Xnew: InputData, full_cov=False, full_output_cov=False) -> MeanAndVariance:
         """
         Compute the mean and variance of the latent function at some new points
         Xnew.
         """
         _, _, Luu, L, _, _, gamma = self.common_terms()
-        Kus = Kuf(self.inducing_variable, self.kernel, X)  # size  [M, X]new
+        Kus = Kuf(self.inducing_variable, self.kernel, Xnew)  # [M, N]
```
alexisboukouvalas: question: are M, N defined somewhere? Maybe in docstring point to that place. For new user these look a bit mysterious

st--: In GLOSSARY.md

Issue #1282, opened 2020-02-25
training_loss


vdutor:

joelberkeley-pio:
what's the aim of this ticket?

st--:
This is handled by #1276

PR #1347, opened 2020-03-19
Cleaning some todo

awav:
Some todos are necessary and should stay, and there are a couple that converted to notes. Thanks.


gpflow/kullback_leiblers.py
```diff
@@ -130,7 +130,6 @@ def gauss_kl(q_mu, q_sqrt, K=None, *, K_cholesky=None):
             ]  # [M, M] -> [M, 1]
             trace = tf.reduce_sum(K_inv * tf.square(q_sqrt))
         else:
-            # TODO: broadcast instead of tile when tf allows -- tf2.1 segfaults
```
st--: This is handled in #1322 - please don't delete this line here

gpflow/kernels/base.py
```diff
@@ -98,7 +98,6 @@ def slice(self, X: tf.Tensor, X2: Optional[tf.Tensor] = None):
             if X2 is not None:
                 X2 = X2[..., dims]
         elif dims is not None:
-            # TODO(@awav): Convert when TF2.0 will support proper slicing.
```
st--: is there any tensorflow issue that we could link here if they enable proper slicing at some point ? how likely is that to happen?

tests/gpflow/likelihoods/test_likelihoods.py
```diff
@@ -116,7 +116,7 @@ def test_no_missing_likelihoods():
         if likelihood_class is MonteCarloLikelihood:
             continue  # abstract base class
         if issubclass(likelihood_class, MonteCarloLikelihood):
-            continue  # TODO
+            continue
```
st--: This is addressed in #1322 - please don't change it here.

tests/gpflow/likelihoods/test_switched_likelihood.py
```diff
@@ -140,7 +140,6 @@ def test_SwitchedLikelihood_withVGP():
     model = gpflow.models.VGP((X, Y_aug), kernel=kernel, likelihood=likelihood)
     ## optimization errors out
     opt = gpflow.optimizers.Scipy()
-    # TODO fix following line once training_loss PR is merged
```
st--: This will be addressed by #1282 - please don't change it here

gpflow/conditionals/util.py
```diff
@@ -469,7 +469,7 @@ def mix_latent_gp(W, g_mean, g_var, full_cov, full_output_cov):
     if not full_cov:
         shape_constraints.append((g_var, [..., "N", "L"]))
     else:
-        # TODO cannot assert g_var shape here because of the inner "leading"
+        # NOTE(awav) cannot assert g_var shape here because of the inner "leading"
```
st--: this already links to an issue - I would leave this as a todo.

gpflow/base.py
```diff
@@ -303,7 +303,7 @@ def _cast_to_dtype(value: VariableData, dtype: Optional[DType] = None) -> tf.Ten
     if dtype is None:
         dtype = default_float()
     if tf.is_tensor(value):
-        # TODO(awav) TF2.2 resolves issue with cast.
+        # NOTE(awav) TF2.2 resolves issue with cast.
```
st--: Is this not something that we want to do ? why not leave it as a todo?

gpflow/kernels/base.py
awav: Tens of issues and 100% not going to happen.

gpflow/base.py
awav: `TODO:` - intention to do something, because we have control over it. `NOTE:` - we have to live with it, even if this is for a short time.

gpflow/conditionals/util.py
awav: `TODO:` - intention to do something, because we have control over it. `NOTE:` - we have to live with it, even if this is for a short time.

PR #1350, opened 2020-03-20
gast dependency

### LINKED ISSUE #1348: tf2.2-rc1 gast requirement
Hi,

gpflow 2.0.0-rc1 has gast requirement 0.2.2.

TensorFlow has gast requirement 0.3.3 from 2.2-rc1, which is incompatible with gpflow requirement.

Best Regards,

Marco

marcoadurno:
Hi,

This could fix #1348 as:
- gast is only added as a requirement if tensorflow is not already installed;
- we read check what's the latest available version of tensorflow on pypi and pin a version of gast accordingly.

Make sense?

Best Regards,

Marco

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -39,7 +39,8 @@ tf_gpu = 'tensorflow-gpu'
 def latest_version(package_name):
     url = f"https://pypi.python.org/pypi/{package_name}/json"
     data = json.load(request.urlopen(url))
-    # filter out rc and b releases and, more in general, any realeses that do not contain exclusively numbers and dots.
+    # filter out rc and beta releases and, more generally, any releases that
+    # do not contain exclusively numbers and dots.
     versions = [v for v in data["releases"].keys() if re.match('[0-9\.]+$',v)]  
     versions.sort(key=StrictVersion)
     return versions[-1]  # return latest version
--- a/setup.py
+++ b/setup.py
@@ -9,10 +9,6 @@ from pathlib import Path
 
 from pkg_resources import parse_version
 from setuptools import find_packages, setup
-# for latest_version() [see https://github.com/GPflow/GPflow/issues/1348]:
-import json
-from urllib import request
-import re
 
 is_py37 = sys.version_info.major == 3 and sys.version_info.minor == 7
 on_rtd = os.environ.get('READTHEDOCS', None) == 'True'  # copied from the docs
@@ -36,7 +32,12 @@ tf_cpu = 'tensorflow'
 tf_gpu = 'tensorflow-gpu'
 
 
+# for latest_version() [see https://github.com/GPflow/GPflow/issues/1348]:
 def latest_version(package_name):
+    import json
+    from urllib import request
+    import re
+
     url = f"https://pypi.python.org/pypi/{package_name}/json"
     data = json.load(request.urlopen(url))
     # filter out rc and beta releases and, more generally, any releases that
--- a/setup.py
+++ b/setup.py
@@ -37,7 +37,7 @@ tf_gpu = 'tensorflow-gpu'
 
 
 def latest_version(package_name):
-    url = "https://pypi.python.org/pypi/%s/json" % (package_name,)
+    url = f"https://pypi.python.org/pypi/{package_name}/json"
     data = json.load(request.urlopen(url))
     # filter out rc and b releases and, more in general, any realeses that do not contain exclusively numbers and dots.
     versions = [v for v in data["releases"].keys() if re.match('[0-9\.]+$',v)]  
--- a/setup.py
+++ b/setup.py
@@ -11,7 +11,6 @@ from pkg_resources import parse_version
 from setuptools import find_packages, setup
 import json
 from urllib import request
-from distutils.version import StrictVersion
 import re
 
 is_py37 = sys.version_info.major == 3 and sys.version_info.minor == 7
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -18,8 +22,7 @@
     'numpy>=1.10.0',
     'scipy>=0.18.0',
     'multipledispatch>=0.4.9',
-    'tabulate',
-    'gast==0.2.2',
+    'tabulate'
 ]
 
 if not is_py37:
```
st--: NB- this will break with python 3.8. it should instead be a check on `if is_less_than_py37`... (not your change I know, but wanted to flag it up while we're discussing the setup script!)

setup.py
```diff
@@ -32,6 +35,15 @@
 tf_cpu = 'tensorflow'
 tf_gpu = 'tensorflow-gpu'
 
+
+def latest_version(package_name):
+    url = "https://pypi.python.org/pypi/%s/json" % (package_name,)
+    data = json.load(request.urlopen(url))
+    versions = [v for v in data["releases"].keys() if not re.search('[a-z]',v)]
```
st--: what does the filter do?  is it meant to filter out nightly, release candidates, etc? could you add a comment?
Would it be more robust to search on `if re.search("^[0-9.]+$")` instead?

setup.py
```diff
@@ -47,6 +59,9 @@
     if not on_rtd:
         # Do not add TF if we are installing GPflow on readthedocs
         requirements.append(tf_cpu)
+        gast_requirement = 'gast==0.2.2' if parse_version(latest_version('tensorflow')) < parse_version('2.2') else 'gast==0.3.3'
```
st--:
```suggestion
        gast_requirement = 'gast>=0.2.2,<0.3' if parse_version(latest_version('tensorflow')) < parse_version('2.2') else 'gast>=0.3.3'
```

setup.py
marcoadurno: Hi @st-- ,

Which bit will break exactly? Adding tabulate?

setup.py
marcoadurno: Done

setup.py
st--: no, the line I commented on: `if not is_py37`.

st--: don't worry about it here, I've made a separate issue (#1354).

setup.py
```diff
@@ -32,6 +35,16 @@
 tf_cpu = 'tensorflow'
 tf_gpu = 'tensorflow-gpu'
 
+
+def latest_version(package_name):
+    url = "https://pypi.python.org/pypi/%s/json" % (package_name,)
+    data = json.load(request.urlopen(url))
+    # filter out rc and b releases and, more in general, any realeses that do not contain exclusively numbers and dots.
```
st--:
```suggestion
    # filter out rc and beta releases and, more generally, any releases that do not contain exclusively numbers and dots.
```

setup.py
```diff
@@ -32,6 +35,16 @@
...
+    versions = [v for v in data["releases"].keys() if re.match('[0-9\.]+$',v)]
```
st--:
```suggestion
    versions = [v for v in data["releases"].keys() if re.match("[0-9.]+$", v)]  
```
note that in a regexp, a `.` inside `[]` counts as a *literal* period, not as "any character".

setup.py
```diff
@@ -32,6 +35,16 @@
 tf_cpu = 'tensorflow'
 tf_gpu = 'tensorflow-gpu'
 
+
+def latest_version(package_name):
+    url = "https://pypi.python.org/pypi/%s/json" % (package_name,)
+    data = json.load(request.urlopen(url))
+    # filter out rc and b releases and, more in general, any realeses that do not contain exclusively numbers and dots.
+    versions = [v for v in data["releases"].keys() if re.match('[0-9\.]+$',v)]  
+    versions.sort(key=StrictVersion)
```
st--:
```suggestion
    versions.sort()
```
(now handled by the `parse_version` in the previous line, following the other suggestion)

setup.py
```diff
@@ -47,6 +60,9 @@
     if not on_rtd:
         # Do not add TF if we are installing GPflow on readthedocs
         requirements.append(tf_cpu)
+        gast_requirement = 'gast>=0.2.2,<0.3' if parse_version(latest_version('tensorflow')) < parse_version('2.2') else 'gast>=0.3.3'
```
st--:
```suggestion
        gast_requirement = 'gast>=0.2.2,<0.3' if latest_version('tensorflow') < parse_version('2.2') else 'gast>=0.3.3'
```
and following the previous suggestions it's already a Version object so no need to parse it anymore!

setup.py
```diff
@@ -32,6 +35,16 @@
 tf_cpu = 'tensorflow'
 tf_gpu = 'tensorflow-gpu'
 
+
+def latest_version(package_name):
+    url = "https://pypi.python.org/pypi/%s/json" % (package_name,)
```
st--:
```suggestion
    url = f"https://pypi.python.org/pypi/{package_name}/json"
```
python 3.6 brings the lovely format strings :)

setup.py
```diff
@@ -9,6 +9,10 @@
 
 from pkg_resources import parse_version
 from setuptools import find_packages, setup
+import json
+from urllib import request
+from distutils.version import StrictVersion
```
st--:
```suggestion
```

setup.py
```diff
@@ -9,6 +9,10 @@
 
 from pkg_resources import parse_version
 from setuptools import find_packages, setup
+import json
```
st--:
```suggestion
# for latest_version() [see https://github.com/GPflow/GPflow/issues/1348]:
import json
```

st--: or could move these imports into the function body of `latest_version()`.

PR #1351, opened 2020-03-20
make full= and full_cov= to predict_* explicit kwargs

st--:


st--:
Urgh, turns out keyword-only arguments don't play well with tf.function() when passing `input_signature`, as is demonstrated by the intro_to_gpflow2 notebook test failing :(

st--:
Unfortunately, explicit keyword-only arguments are not supported by `tf.function()`, which renders this PR impossible.

PR #1352, opened 2020-03-20
Eric/training loss proposal

condnsdmatters:
Proposal for training loss interface.

### DIFF
--- a/doc/source/notebooks/tailor/mixture_density_network.pct.py
+++ b/doc/source/notebooks/tailor/mixture_density_network.pct.py
@@ -85,7 +85,7 @@ import tensorflow as tf
 
 # %%
 import gpflow
-from gpflow.models import BayesianModel
+from gpflow.models import ExternalDataBayesianModel
 from gpflow.base import Parameter
 
 # %% [markdown]
@@ -97,7 +97,7 @@ from gpflow.base import Parameter
 # %%
 from typing import Callable, Optional, Tuple
 
-class MDN(BayesianModel):
+class MDN(ExternalDataBayesianModel):
 
     def __init__(
         self,
@@ -179,7 +179,7 @@ from gpflow.optimizers import Scipy
 from gpflow.ci_utils import ci_niter
 
 Scipy().minimize(
-    tf.function(model.training_loss_closure(data)),
+    model.training_loss_closure(data, jit=True),
     variables=model.trainable_parameters,
     options=dict(maxiter=ci_niter(1500))
 );
@@ -230,7 +230,7 @@ model = MDN(inner_dims=[100, 100], num_mixtures=5)
 
 # %%
 Scipy().minimize(
-    tf.function(model.training_loss_closure(data)),
+    model.training_loss_closure(data),
     variables=model.trainable_parameters,
     options=dict(maxiter=ci_niter(int(10e3)))
 );
--- a/doc/source/notebooks/tailor/mixture_density_network.pct.py
+++ b/doc/source/notebooks/tailor/mixture_density_network.pct.py
@@ -230,7 +230,7 @@ model = MDN(inner_dims=[100, 100], num_mixtures=5)
 
 # %%
 Scipy().minimize(
-    model.training_loss_closure(data),
+    model.training_loss_closure(data, jit=True),
     variables=model.trainable_parameters,
     options=dict(maxiter=ci_niter(int(10e3)))
 );
--- a/gpflow/models/gpmc.py
+++ b/gpflow/models/gpmc.py
@@ -24,10 +24,10 @@ from ..kernels import Kernel
 from ..likelihoods import Likelihood
 from ..mean_functions import MeanFunction
 from ..utilities import to_default_float
-from .model import RegressionData, GPModel, MeanAndVariance, BayesianModelStoringData
+from .model import RegressionData, MeanAndVariance, InternalDataGPModel
 
 
-class GPMC(GPModel, BayesianModelStoringData):
+class GPMC(InternalDataGPModel):
     def __init__(
         self,
         data: RegressionData,
--- a/gpflow/models/gpmc.py
+++ b/gpflow/models/gpmc.py
@@ -24,10 +24,10 @@ from ..kernels import Kernel
 from ..likelihoods import Likelihood
 from ..mean_functions import MeanFunction
 from ..utilities import to_default_float
-from .model import RegressionData, MeanAndVariance, InternalDataGPModel
+from .model import RegressionData, MeanAndVariance, GPModel
 
 
-class GPMC(InternalDataGPModel):
+class GPMC(GPModel):
     def __init__(
         self,
         data: RegressionData,
@@ -61,9 +61,6 @@ class GPMC(InternalDataGPModel):
         self.V.prior = tfp.distributions.Normal(
             loc=to_default_float(0.0), scale=to_default_float(1.0)
         )
-
-    def training_loss(self) -> tf.Tensor:
-        return -self.log_posterior_density()
 
     def log_posterior_density(self) -> tf.Tensor:
         return self.log_likelihood() + self.log_prior_density()
--- a/gpflow/models/model.py
+++ b/gpflow/models/model.py
@@ -256,7 +256,7 @@ class InternalDataGPModel(GPModel, InternalDataTrainingInterface):
     """
 
     def training_loss(self):
-        return - self.maximum_a_posteriori_objective()
+        return -self.maximum_a_posteriori_objective()
 
 
 class ExternalDataGPModel(GPModel, ExternalDataTrainingInterface):
@@ -266,4 +266,4 @@ class ExternalDataGPModel(GPModel, ExternalDataTrainingInterface):
     """
 
     def training_loss(self, data):
-        return - self.maximum_a_posteriori_objective(data)
+        return -self.maximum_a_posteriori_objective(data)
--- a/gpflow/models/model.py
+++ b/gpflow/models/model.py
@@ -20,6 +20,15 @@ import numpy as np
 import tensorflow as tf
 from tensorflow.python.data.ops import iterator_ops
 
+from .training_interface import (
+    InternalDataTrainingInterface,
+    ExternalDataTrainingInterface,
+    InputData,
+    OutputData,
+    RegressionData,
+    Data,
+)
+
 from ..base import Module
 from ..conditionals.util import sample_mvn
 from ..config import default_float, default_jitter
@@ -28,10 +37,6 @@ from ..likelihoods import Likelihood, SwitchedLikelihood
 from ..mean_functions import MeanFunction, Zero
 from ..utilities import ops, to_default_float
 
-InputData = tf.Tensor
-OutputData = tf.Tensor
-RegressionData = Tuple[InputData, OutputData]
-Data = TypeVar("Data", RegressionData, InputData)
 MeanAndVariance = Tuple[tf.Tensor, tf.Tensor]
 
 
@@ -54,25 +59,21 @@ class BayesianModel(Module, metaclass=abc.ABCMeta):
         SGPMC). It assumes that maximum_likelihood_objective() is defined
         sensibly.
         """
-        return self.maximum_likelihood_objective(*args, **kwargs) + self.log_prior_density()
+        return (
+            self.maximum_likelihood_objective(*args, **kwargs)
+            + self.log_prior_density()
+        )
 
-    def maximum_likelihood_objective_with_priors(self, *args, **kwargs) -> tf.Tensor:
+    def maximum_a_posteriori_objective(self, *args, **kwargs) -> tf.Tensor:
         """
         Minimization objective for TensorFlow optimizers (including
         gpflow.optimizers.Scipy). Includes the log prior density for maximum
         a-posteriori (MAP) estimation.
         """
-        return -(self.maximum_likelihood_objective(*args, **kwargs) + self.log_prior_density())
-
-    @abc.abstractmethod
-    def training_loss(self, *args, **kwargs) -> tf.Tensor:
-        """
-        Overwritten by InternalDataGPModel and ExternalDataGPModel
-        to specify whether or not a model stores data internally and hence
-        whether it should not or should receive data as an argument to the
-        training_loss.
-        """
-        raise NotImplementedError
+        return -(
+            self.maximum_likelihood_objective(*args, **kwargs)
+            + self.log_prior_density()
+        )
 
     @abc.abstractmethod
     def maximum_likelihood_objective(self, *args, **kwargs) -> tf.Tensor:
@@ -83,6 +84,16 @@ class BayesianModel(Module, metaclass=abc.ABCMeta):
         GPs.
         """
         raise NotImplementedError
+
+
+class ExternalDataBayesianModel(BayesianModel, ExternalDataTrainingInterface):
+    """
+    Base class for GP models that do not encapsulate the data; training_loss takes
+    data as argument.
+    """
+
+    def training_loss(self, data):
+        return self.maximum_a_posteriori_objective(data)
 
 
 class GPModel(BayesianModel):
@@ -122,7 +133,9 @@ class GPModel(BayesianModel):
         num_latent_gps: int = None,
     ):
         super().__init__()
-        assert num_latent_gps is not None, "GPModel requires specification of num_latent_gps"
+        assert (
+            num_latent_gps is not None
+        ), "GPModel requires specification of num_latent_gps"
         self.num_latent_gps = num_latent_gps
         # TODO(@awav): Why is this here when MeanFunction does not have a __len__ method
         if mean_function is None:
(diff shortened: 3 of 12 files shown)


gpflow/models/training_interface.py
```diff
@@ -0,0 +1,62 @@
...
+    def training_loss_closure(self, data: Data, jit=True) -> Callable[[], tf.Tensor]:
+        training_loss = self.training_loss
+        if jit:
+            training_loss = tf.function(
+                self.training_loss
+            )  # TODO need to add correct input_signature here to allow for differently sized minibatches
+
+        def closure():
+            batch = next(data) if isinstance(data, collections.abc.Iterator) else data
+            return training_loss(data)
```
vdutor: This looks like a typo?
```suggestion
            return training_loss(batch)
```

gpflow/models/training_interface.py
```diff
@@ -0,0 +1,62 @@
...
+        Specify a training loss for this model
+        """
+        raise NotImplementedError
+
+    def training_loss_closure(self, data: Data, jit=True) -> Callable[[], tf.Tensor]:
+        training_loss = self.training_loss
+        if jit:
+            training_loss = tf.function(
+                self.training_loss
+            )  # TODO need to add correct input_signature here to allow for differently sized minibatches
```
vdutor: how would you do this? By having a user specify `function_kwargs` as argument?

doc/source/notebooks/tailor/mixture_density_network.pct.py
```diff
@@ -230,7 +230,7 @@ def moon_data(N, noise):
 
 # %%
 Scipy().minimize(
-    tf.function(model.training_loss_closure(data)),
+    model.training_loss_closure(data),
```
st--:
```suggestion
    model.training_loss_closure(data, jit=True),
```
if you want to be consistent

gpflow/models/model.py
```diff
@@ -56,24 +61,14 @@ def log_posterior_density(self, *args, **kwargs) -> tf.Tensor:
         """
         return self.maximum_likelihood_objective(*args, **kwargs) + self.log_prior_density()
 
-    def _training_loss(self, *args, **kwargs) -> tf.Tensor:
+    def maximum_a_posteriori_objective(self, *args, **kwargs) -> tf.Tensor:
```
st--: you'd have to flip the sign as well

gpflow/models/model.py
```diff
@@ -283,11 +239,31 @@ def predict_y(
...
+
+
+class InternalDataGPModel(GPModel, InternalDataTrainingInterface):
+    """
+    Base class for models that encapsulate their data and store it as
+    self.data; training_loss does not take any arguments.
+    """
+
+    def training_loss(self):
+        return self.maximum_a_posteriori_objective()
```
st--: and flip the sign here as well as in the other one below, which I still think is a bit messy :/

st--: I think I'd prefer to keep the `_training_loss()` and `_training_loss(data)`...

gpflow/models/model.py
```diff
@@ -85,53 +80,14 @@ def maximum_likelihood_objective(self, *args, **kwargs) -> tf.Tensor:
...
-        if jit:
-            training_loss = tf.function(self.training_loss)  # TODO need to add correct input_signature here to allow for differently sized minibatches
-        else:
-            training_loss = self.training_loss
-
-        def minibatch_training_loss_closure():
-            batch = next(batch_iterator)
-            return training_loss(batch)
-
-        return minibatch_training_loss_closure
```
st--: what about this?

gpflow/models/gpmc.py
```diff
@@ -24,10 +24,10 @@
 from ..likelihoods import Likelihood
 from ..mean_functions import MeanFunction
 from ..utilities import to_default_float
-from .model import RegressionData, GPModel, MeanAndVariance, BayesianModelStoringData
+from .model import RegressionData, MeanAndVariance, InternalDataGPModel
 
 
-class GPMC(GPModel, BayesianModelStoringData):
+class GPMC(InternalDataGPModel):
```
st--: Note that for GPMC (and SGPMC in sgpmc.py) I explicitly re-defined training_loss - not because the code is different but because it's conceptually different and I wanted to highlight that... don't know how that interacts with your changes here?

gpflow/models/training_interface.py
```diff
@@ -0,0 +1,62 @@
...
+        Specify a training loss for this model
+        """
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def training_loss_closure(self, *args, **kwargs):
+        """
+        Return a closure with a training loss for this module
+        """
+        raise NotImplementedError
```
joelberkeley-pio: I've just discussed with eric on slack about APIs with `(*args, **kwargs)`. They present a problem in that subclasses must also accept any number of arguments and keyword arguments to satisfy the contract. In particular, they have to have the signature `(*args: Any, **kwargs: Any)`.

You can think of it like this: Imagine you're working in a function which has signature
```python3
def foo(obj: TrainingInterface) -> None:
```
how do you work with it? I should be able to pass in whatever I want, for example
```
obj.training_loss(4, np.random, values='which', make=not 'sense')
```
This would fail if the runtime object doesn't support those arguments

condnsdmatters: do you mean the last sentence?  TBH am happy to get rid of the base `TrainingInterface`, to abide by liskov principle.

gpflow/models/gpmc.py
condnsdmatters: If you reimplement in the derived it wont interact badly, though i dont know whether we can call the density a `loss`. Its very possible having `training loss` with `GPMC` and `SGPMC` makes no sense, and they should just inherit from `GPModel`

gpflow/models/model.py
condnsdmatters: i think this is much better. i think have a method which is _training_loss is much messier.

condnsdmatters:
```suggestion
        return - self.maximum_a_posteriori_objective()
```

gpflow/models/model.py
```diff
@@ -283,11 +239,31 @@ def predict_y(
...
+
+
+class ExternalDataGPModel(GPModel, ExternalDataTrainingInterface):
+    """
+    Base class for GP models that do not encapsulate the data; training_loss takes
+    data as argument.
+    """
+
+    def training_loss(self, data):
+        return self.maximum_a_posteriori_objective(data)
```
condnsdmatters:
```suggestion
        return - self.maximum_a_posteriori_objective(data)
```

gpflow/models/training_interface.py
condnsdmatters: i think you just specifiy the leading dim size to be None

gpflow/models/gpmc.py
st--: we don't call the density a loss, but you might still want to do some gradient optimization first to relax the model before starting MCMC - that's why I chose to overwrite it

gpflow/models/training_interface.py
st--: Aaaand it feels like we're back to square one. I would very much like to go with the thing that actually works in practice, for users, rather than the thing that makes the type system happy but makes it more cumbersome for users...

gpflow/models/training_interface.py
st--: specify `input_signatures` argument to the tf.function - just have to look up the right TensorSpec ...

gpflow/models/training_interface.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+class ExternalDataTrainingInterface(TrainingInterface):
+    def training_loss(self, data):
+        """
+        Specify a training loss for this model
+        """
+        raise NotImplementedError
+
+    def training_loss_closure(self, data: Data, jit=True) -> Callable[[], tf.Tensor]:
```
st--: type signature doesn't agree with the use in the closure() below...?
```suggestion
    def training_loss_closure(self, data: Union[Data, Iterator], jit=True) -> Callable[[], tf.Tensor]:
```

gpflow/models/training_interface.py
st--: this won't work in a general way for models that have different shapes of data -- e.g. RegressionData (SVGP) with data=(X,Y) vs InputData with data=Y (if there was a minibatching GPLVM) need different (incompatible) input_signatures.

gpflow/models/training_interface.py
joelberkeley-pio: @st-- type systems aren't just there for theoretical aesthetics. I would guess that breaking the type system always results in at least one test that you can write that fails. It translates into code that's buggy in some way, and bugs are cumbersome for users

gpflow/models/gpmc.py
```diff
@@ -24,10 +24,10 @@
 from ..likelihoods import Likelihood
 from ..mean_functions import MeanFunction
 from ..utilities import to_default_float
-from .model import RegressionData, MeanAndVariance, InternalDataGPModel
+from .model import RegressionData, MeanAndVariance, GPModel
 
 
-class GPMC(InternalDataGPModel):
+class GPMC(GPModel):
```
st--: Should this not remain an `InternalDataGPModel`?

gpflow/models/gpmc.py
```diff
@@ -62,9 +62,6 @@ def __init__(
             loc=to_default_float(0.0), scale=to_default_float(1.0)
         )
 
-    def training_loss(self) -> tf.Tensor:
-        return -self.log_posterior_density()
-
```
st--: why did this get removed?

gpflow/models/training_interface.py
st--: Python isn't a strongly-typed language, so from a `*args, **kwargs` signature you can't infer that you should be *allowed* to pass in whatever you want.

gpflow/models/model.py
```diff
@@ -56,23 +61,13 @@ def log_posterior_density(self, *args, **kwargs) -> tf.Tensor:
         """
         return self.maximum_likelihood_objective(*args, **kwargs) + self.log_prior_density()
 
-    def _training_loss(self, *args, **kwargs) -> tf.Tensor:
+    def maximum_a_posteriori_objective(self, *args, **kwargs) -> tf.Tensor:
         """
         Minimization objective for TensorFlow optimizers (including
         gpflow.optimizers.Scipy). Includes the log prior density for maximum
         a-posteriori (MAP) estimation.
```
st--: Docstring inconsistent with function name change

gpflow/models/training_interface.py
st--: Trying to force static typing on Python can result in unpythonic and cumbersome code. I don't think that's the right approach for a research toolbox that's supposed to be flexible and make it easy for GP researchers to implement their ideas.

joelberkeley-pio: > Python isn't a strongly-typed language, so from a `*args, **kwargs` signature you can't infer that you should be _allowed_ to pass in whatever you want.

it's not a matter of it being statically typed (python is strongly typed, but is optionally statically typed), it's a fundamental design principle of OOP (LSP), which applies equally for dynamic and static typing

> 

Trying to force static typing on Python can result in unpythonic and cumbersome code. I don't think that's the right approach for a research toolbox that's supposed to be flexible and make it easy for GP researchers to implement their ideas.

I don't agree it's unpythonic. Static types are after all part of the standard library. Cumbersome? Yes, if overused. But overused and cumbersome are both subjective. Regarding a research toolbox, I'm not really clear yet on what that means. I hope time will tell. Flexible? I don't anticipate this will make it any more or less flexible.

joelberkeley-pio: I suspect the concern here is the conceptual overhead of the design. I agree it's a headache and not something we should keep in the code base. I guess the discussion is what to do until we reach that point. I lean on the side of dealing with it safely until we can converge on a final design. I appreciate that's not the universal opinion.

gpflow/models/model.py
```diff
@@ -85,53 +80,14 @@ def maximum_likelihood_objective(self, *args, **kwargs) -> tf.Tensor:
...
-
-    def training_loss_closure(self, jit=True) -> Callable[[], tf.Tensor]:
-        if jit:
-            return tf.function(self.training_loss)
-        else:
-            return self.training_loss
-
-
-class BayesianModelNotStoringData(BayesianModel):
+class ExternalDataBayesianModel(BayesianModel, ExternalDataTrainingInterface):
```
joelberkeley-pio: this appears only to be used in the notebooks. do we want to have something in the codebase that's only there for a single notebook?

gpflow/models/model.py
```diff
@@ -283,11 +239,31 @@ def predict_y(
...
+
+
+class ExternalDataGPModel(GPModel, ExternalDataTrainingInterface):
+    """
+    Base class for GP models that do not encapsulate the data; training_loss takes
+    data as argument.
+    """
+
+    def training_loss(self, data):
+        return -self.maximum_a_posteriori_objective(data)
```
joelberkeley-pio: you could completely decouple these from the `GPModel`, which may help with keeping the hierarchy simple, using
```python3
class InternalDataGPModel(GPModel, InternalDataTrainingInterface):
    @abstractmethod
    def maximum_a_posteriori_objective(self):
        pass

    def training_loss(self):
        return -self.maximum_a_posteriori_objective()
```

gpflow/models/model.py
```diff
@@ -56,23 +61,13 @@ def log_posterior_density(self, *args, **kwargs) -> tf.Tensor:
         """
         return self.maximum_likelihood_objective(*args, **kwargs) + self.log_prior_density()
 
-    def _training_loss(self, *args, **kwargs) -> tf.Tensor:
+    def maximum_a_posteriori_objective(self, *args, **kwargs) -> tf.Tensor:
```
joelberkeley-pio: i feel the same about `(*args, **kwargs)` here

gpflow/models/model.py
joelberkeley-pio: do we expect users to use it?

gpflow/models/model.py
joelberkeley-pio: this would also make the mixin work with `GPModel` and `BayesianModel` without assuming either

joelberkeley-pio: they could even be merged with `InternalDataTrainingInterface` and `ExternalDataTrainingInterface` then

gpflow/models/training_interface.py
```diff
@@ -0,0 +1,46 @@
...
+import numpy as np
+
+InputData = tf.Tensor
+OutputData = tf.Tensor
+RegressionData = Tuple[InputData, OutputData]
+Data = TypeVar("Data", RegressionData, InputData)
+
+
+class InternalDataTrainingInterface:
+    def training_loss(self):
```
joelberkeley-pio:
```suggestion
class InternalDataTrainingInterface(ABC):
    @abstractmethod
    def training_loss(self):
```
etc?

gpflow/models/model.py
st--: The training_loss / maximum_a_posteriori_objective should go with BayesianModel not GPModel.

gpflow/models/model.py
st--: yes.

gpflow/models/sgpmc.py
```diff
@@ -86,9 +86,6 @@ def __init__(
             loc=to_default_float(0.0), scale=to_default_float(1.0)
         )
 
-    def training_loss(self) -> tf.Tensor:
-        return -self.log_posterior_density()
-
```
st--: can you put this back?

gpflow/models/sgpmc.py
```diff
@@ -24,11 +24,11 @@
 from ..likelihoods import Likelihood
 from ..mean_functions import MeanFunction
 from ..utilities import to_default_float
-from .model import RegressionData, GPModel, MeanAndVariance, BayesianModelStoringData
+from .model import RegressionData, MeanAndVariance, GPModel
 from .util import inducingpoint_wrapper
 
 
-class SGPMC(GPModel, BayesianModelStoringData):
+class SGPMC(GPModel):
```
st--: can you put the training interface back?

gpflow/models/model.py
joelberkeley-pio: ok. that's fine. it's just that we have the option to do either

joelberkeley-pio:
I'd like to see this PR compared to develop, but I don't want to edit your PR. When can we see that?

st--:
> I'd like to see this PR compared to develop, but I don't want to edit your PR. When can we see that?

How about `git diff origin/develop origin/eric/training-loss-proposal gpflow` ?

gpflow/models/model.py
condnsdmatters: im not entirely sure what that buys you, except that once again ppl need to remember to add the mixin for their new derived model (which im not so keen on). lets discuss over zoom

st--:
Closed in favour of #1362

PR #1357, opened 2020-03-21
fix Cosine kernel for multivariate inputs

### LINKED ISSUE #1328: cosine kernel returns indefinite covariance matrices
Consider the following minimal example:
```python
tf.random.set_seed(42)
X = tf.random.uniform((10,2), minval=-10, maxval=10, dtype=tf.float64)

kern = gpflow.kernels.Cosine()
print(tf.linalg.eigvalsh(kern(X)))
```
This gives me
```python
tf.Tensor(
[-2.27306374 -1.58787397 -0.2847927   0.19340605  1.18589017  1.33897599
  1.6473351   2.37452162  2.89101446  4.51458702], shape=(10,), dtype=float64)
```
Because of this, the cosine kernel in GPR or SGPR is unusable, as optimization aborts due to failed cholesky decomposition (I haven't tried other models, yet, but I assume that this is a general issue). As far as I can tell, the cosine kernel was first derived in [1] for the univariate case. There, the cosine kernel is defined as $\sigma^2\cos\left(\frac{2\pi(x - x')}{p}\right)$. Note that it is a function of $x - x'$, not $|x - x'|$, as in the gpflow implementation. Transitioning to the multivariate case and defining $k(x, x') = \sigma^2\cos\left(\frac{2\pi\sum_k (x_k - x'_k)}{p}\right)$, positive semidefinite covariance matrices are obtained up to numerical imprecision:
```python
dist = X[:, tf.newaxis, :] - X[tf.newaxis, :, :]
dist = tf.reduce_sum(dist, axis=-1)
print(tf.linalg.eigvalsh(tf.math.cos(dist)))
```
This gives
```python
tf.Tensor(
[-1.79943150e-15 -6.15775238e-16 -4.15847122e-16 -3.93820396e-17
  9.50847803e-17  2.43329656e-16  6.85377761e-16  1.38407146e-15
  3.22740434e+00  6.77259566e+00], shape=(10,), dtype=float64)
```
The above kernel can be derived from eq. 4.7 in [2] as the limiting case of $v_q^{(p)} = 0 \forall p$

[1] https://dspace.mit.edu/bitstream/handle/1721.1/92869/Tenenbaum_Automatic%20Construction.pdf?sequence=1&isAllowed=y
[2] https://www.cs.cmu.edu/~andrewgw/andrewgwthesis.pdf

sorry for the bad math, I have no idea how to make GitHub render equations...

ilia-kats:
The previous version operating on euclidean distance was returning indefinite covariance matrices on multivariate data. This version, derived from eq. 4.7 of Wilson (2014), is always positive semidefinite.
Closes #1328.

This PR also changes the definition of the cosine kernel slightly, from sigma * cos((x - x') / l) to sigma * cos(2 * pi * (x - x') / l). This makes the lengthscale parameter directly interpretable as period length.

### DIFF
--- a/doc/source/notebooks/tailor/kernel_design.pct.py
+++ b/doc/source/notebooks/tailor/kernel_design.pct.py
@@ -34,7 +34,7 @@ plt.style.use("ggplot")
 
 # %% [markdown]
 # To make this new kernel class, we inherit from the base class `gpflow.kernels.Kernel` and implement the three functions below. **NOTE:** Depending on the kernel to be implemented, other classes can be more adequate. For example, if the kernel to be implemented is isotropic stationary, you can immediately subclass `gpflow.kernels.IsotropicStationary` (at which point you
-# only have to override `K_r` or `K_r2`; see the `IsotropicStationary` class docstring). Anisotropic kernels should subclass `gpflow.kernels.AnisotropicStationary` and override `K_d`.
+# only have to override `K_r` or `K_r2`; see the `IsotropicStationary` class docstring). Stationary but anisotropic kernels should subclass `gpflow.kernels.AnisotropicStationary` and override `K_d`.
 #
 # #### `__init__`
 # In this simple example, the constructor takes no argument (though it could, if that was convenient, for example to pass in an initial value for `variance`). It *must* call the constructor of the superclass with appropriate arguments. Brownian motion is only defined in one dimension, and we'll assume that the `active_dims` are `[0]`, for simplicity.
--- a/doc/source/notebooks/tailor/kernel_design.pct.py
+++ b/doc/source/notebooks/tailor/kernel_design.pct.py
@@ -33,8 +33,8 @@ plt.style.use("ggplot")
 # %matplotlib inline
 
 # %% [markdown]
-# To make this new kernel class, we inherit from the base class `gpflow.kernels.Kernel` and implement the three functions below. **NOTE:** Depending on the kernel to be implemented, other classes can be more adequate. For example, if the kernel to be implemented is stationary, you can immediately subclass `gpflow.kernels.Stationary` (at which point you
-# only have to override `K_r` or `K_r2`; see the `Stationary` class docstring).
+# To make this new kernel class, we inherit from the base class `gpflow.kernels.Kernel` and implement the three functions below. **NOTE:** Depending on the kernel to be implemented, other classes can be more adequate. For example, if the kernel to be implemented is stationary, you can immediately subclass `gpflow.kernels.Stationary` (at which point you 
+# only have to override `K_r`, `K_r2`, or `K_d`; see the `Stationary` class docstring).
 #
 # #### `__init__`
 # In this simple example, the constructor takes no argument (though it could, if that was convenient, for example to pass in an initial value for `variance`). It *must* call the constructor of the superclass with appropriate arguments. Brownian motion is only defined in one dimension, and we'll assume that the `active_dims` are `[0]`, for simplicity.
--- a/doc/source/notebooks/tailor/kernel_design.pct.py
+++ b/doc/source/notebooks/tailor/kernel_design.pct.py
@@ -33,8 +33,8 @@ plt.style.use("ggplot")
 # %matplotlib inline
 
 # %% [markdown]
-# To make this new kernel class, we inherit from the base class `gpflow.kernels.Kernel` and implement the three functions below. **NOTE:** Depending on the kernel to be implemented, other classes can be more adequate. For example, if the kernel to be implemented is stationary, you can immediately subclass `gpflow.kernels.Stationary` (at which point you 
-# only have to override `K_r`, `K_r2`, or `K_d`; see the `Stationary` class docstring).
+# To make this new kernel class, we inherit from the base class `gpflow.kernels.Kernel` and implement the three functions below. **NOTE:** Depending on the kernel to be implemented, other classes can be more adequate. For example, if the kernel to be implemented is isotropic stationary, you can immediately subclass `gpflow.kernels.IsotropicStationary` (at which point you
+# only have to override `K_r` or `K_r2`; see the `IsotropicStationary` class docstring). Anisotropic kernels should subclass `gpflow.kernels.AnisotropicStationary` and override `K_d`.
 #
 # #### `__init__`
 # In this simple example, the constructor takes no argument (though it could, if that was convenient, for example to pass in an initial value for `variance`). It *must* call the constructor of the superclass with appropriate arguments. Brownian motion is only defined in one dimension, and we'll assume that the `active_dims` are `[0]`, for simplicity.
--- a/gpflow/kernels/periodic.py
+++ b/gpflow/kernels/periodic.py
@@ -5,7 +5,7 @@ import tensorflow as tf
 
 from ..base import Parameter
 from ..utilities import positive
-from ..utilities.ops import distance
+from ..utilities.ops import difference_matrix
 from .base import Kernel
 from .stationaries import Stationary, IsotropicStationary
 
@@ -66,7 +66,7 @@ class Periodic(Kernel):
         return self.base_kernel.K_diag(X)
 
     def K(self, X: tf.Tensor, X2: Optional[tf.Tensor] = None) -> tf.Tensor:
-        r = np.pi * (distance(X, X2)) / self.period
+        r = np.pi * (difference_matrix(X, X2)) / self.period
         scaled_sine = tf.sin(r) / self.base_kernel.lengthscales
         if isinstance(self.base_kernel, IsotropicStationary):
             if hasattr(self.base_kernel, "K_r"):
--- a/gpflow/kernels/periodic.py
+++ b/gpflow/kernels/periodic.py
@@ -38,7 +38,7 @@ class Periodic(Kernel):
         the constructor doesn't have it as an argument.
     """
 
-    def __init__(self, base_kernel: Stationary, period: Union[float, List[float]] = 1.0):
+    def __init__(self, base_kernel: IsotropicStationary, period: Union[float, List[float]] = 1.0):
         """
         :param base_kernel: the base kernel to make periodic; must inherit from Stationary
             Note that `active_dims` should be specified in the base kernel.
@@ -46,8 +46,8 @@ class Periodic(Kernel):
             this must be initialized with an array the same length as the number
             of active dimensions e.g. [1., 1., 1.]
         """
-        if not isinstance(base_kernel, Stationary):
-            raise TypeError("Periodic requires a Stationary kernel as the `base_kernel`")
+        if not isinstance(base_kernel, IsotropicStationary):
+            raise TypeError("Periodic requires an IsotropicStationary kernel as the `base_kernel`")
 
         super().__init__()
         self.base_kernel = base_kernel
@@ -68,13 +68,10 @@ class Periodic(Kernel):
     def K(self, X: tf.Tensor, X2: Optional[tf.Tensor] = None) -> tf.Tensor:
         r = np.pi * (difference_matrix(X, X2)) / self.period
         scaled_sine = tf.sin(r) / self.base_kernel.lengthscales
-        if isinstance(self.base_kernel, IsotropicStationary):
-            if hasattr(self.base_kernel, "K_r"):
-                sine_r = tf.reduce_sum(tf.abs(scaled_sine), -1)
-                K = self.base_kernel.K_r(sine_r)
-            else:
-                sine_r2 = tf.reduce_sum(tf.square(scaled_sine), -1)
-                K = self.base_kernel.K_r2(sine_r2)
+        if hasattr(self.base_kernel, "K_r"):
+            sine_r = tf.reduce_sum(tf.abs(scaled_sine), -1)
+            K = self.base_kernel.K_r(sine_r)
         else:
-            K = self.base_kernel.K_d(scaled_sine)
+            sine_r2 = tf.reduce_sum(tf.square(scaled_sine), -1)
+            K = self.base_kernel.K_r2(sine_r2)
         return K
--- a/gpflow/kernels/periodic.py
+++ b/gpflow/kernels/periodic.py
@@ -5,8 +5,9 @@ import tensorflow as tf
 
 from ..base import Parameter
 from ..utilities import positive
+from ..utilities.ops import distance
 from .base import Kernel
-from .stationaries import Stationary
+from .stationaries import Stationary, IsotropicStationary
 
 
 class Periodic(Kernel):
@@ -65,20 +66,15 @@ class Periodic(Kernel):
         return self.base_kernel.K_diag(X)
 
     def K(self, X: tf.Tensor, X2: Optional[tf.Tensor] = None) -> tf.Tensor:
-        if X2 is None:
-            X2 = X
-
-        # Introduce dummy dimension so we can use broadcasting
-        # TODO: does not support kernel broadcasting
-        f = tf.expand_dims(X, 1)  # now [N, 1, D]
-        f2 = tf.expand_dims(X2, 0)  # now [1, M, D]
-
-        r = np.pi * (f - f2) / self.period
+        r = np.pi * (distance(X, X2)) / self.period
         scaled_sine = tf.sin(r) / self.base_kernel.lengthscales
-        if hasattr(self.base_kernel, "K_r"):
-            sine_r = tf.reduce_sum(tf.abs(scaled_sine), -1)
-            K = self.base_kernel.K_r(sine_r)
+        if isinstance(self.base_kernel, IsotropicStationary):
+            if hasattr(self.base_kernel, "K_r"):
+                sine_r = tf.reduce_sum(tf.abs(scaled_sine), -1)
+                K = self.base_kernel.K_r(sine_r)
+            else:
+                sine_r2 = tf.reduce_sum(tf.square(scaled_sine), -1)
+                K = self.base_kernel.K_r2(sine_r2)
         else:
-            sine_r2 = tf.reduce_sum(tf.square(scaled_sine), -1)
-            K = self.base_kernel.K_r2(sine_r2)
+            K = self.base_kernel.K_d(scaled_sine)
         return K
--- a/gpflow/kernels/stationaries.py
+++ b/gpflow/kernels/stationaries.py
@@ -194,7 +194,7 @@ class Cosine(Stationary):
     The Cosine kernel. Functions drawn from a GP with this kernel are sinusoids
     (with a random phase).  The kernel equation is
 
-        k(r) = σ² cos{d}
+        k(r) = σ² cos{2πd}
 
     where:
     d  is the sum of the per-dimension differences between the input points, scaled by the lengthscale parameter ℓ (i.e. Σ_i [(X - X2ᵀ) / ℓ]_i),
@@ -202,4 +202,4 @@ class Cosine(Stationary):
     """
 
     def K_d(self, d):
-        return self.variance * tf.cos(d)
+        return self.variance * tf.cos(2 * np.pi * d)
--- a/gpflow/kernels/stationaries.py
+++ b/gpflow/kernels/stationaries.py
@@ -16,18 +16,6 @@ class Stationary(Kernel):
     This class handles 'ard' behaviour, which stands for 'Automatic Relevance
     Determination'. This means that the kernel has one lengthscale per
     dimension, otherwise the kernel is isotropic (has a single lengthscale).
-
-    Derived classes should implement one of:
-
-        K_r2(self, r2): Returns the kernel evaluated on r² (r2), which is the
-        squared scaled Euclidean distance Should operate element-wise on r².
-
-        K_r(self, r): Returns the kernel evaluated on r, which is the scaled
-        Euclidean distance Should operate element-wise on r².
-
-        K_d(self, d): Returns the kernel evaluated on d, which is the sum of
-        the per-dimension differences between the input points, scaled by the
-        lengthscale parameter ℓ (i.e. Σ_i [(X - X2ᵀ) / ℓ]_i).
     """
 
     def __init__(self, variance=1.0, lengthscales=1.0, **kwargs):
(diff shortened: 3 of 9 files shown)


gpflow/utilities/ops.py
```diff
@@ -102,6 +102,18 @@ def square_distance(X, X2):
     return dist
 
 
+def distance(X, X2):
+    """
+    Returns Σ_i (X - X2ᵀ)_i
+
+    This funciton can deal with leading dimensions in X and X2.
```
st--:
```suggestion
    This function can deal with leading dimensions in X and X2.
```

ilia-kats:
According to Rasmussen and Williams (2006) (section 4.1), a stationary kernel is a function of x - x', not ‖x - x'‖. If you still think that having Cosine inherit directly from Kernel is better, I would be happy to change it.

st--:
I take it back, I think you're right after all! I'm just wondering if we could write it such that we can write K_d = K_r(norm(d)) and K_r = K_r2(r^2) ... (to clean up the implementation) or if that would come at the cost of computational efficiency...

ilia-kats:
We probably can, but that would mean that r2 would be calculated from d and not via the binomial expansion as it is at the moment. I assume that there is a reason for the binomial expansion implementation, especially given the extra bit of documentation for it. Also, I'm not sure if going from r to r2 is the best idea, since calculating r necessarily involves calculating r2, so this would basically be r2 = sqrt(r2)**2. But I can try to think of some way to simplify this.

st--:
> Also, I'm not sure if going from r to r2 is the best idea

No it's not... it's conceptually appealing (d -> r=|d| -> r2=r^2) but of course it's easier to compute r^2 directly and r from that only if needed.

I was mainly thinking the implementation of Stationary.K is fairly complicated now - with just K_r and K_r2 it was fairly easy to handle (a subkernel either overwrites K_r2 directly, or K_r2 handles taking the square-root and passes it to K_r -- if neither K_r nor K_r2 is implemented, this fails; if both are implemented in a subclass, at least you can see that you're duplicating work).

With three options, particularly where one is kind of completely separate, it's getting messy. Maybe a solution would be to add two more base classes: handle `__init__` and `ard` in Stationary, then have  "IsotropicStationary" for the rest of what's currently in Stationary, and add "NonisotropicStationary" (or any better name you can come up with!) that handles the distance vector dependent kernels, and then Cosine can subclass from that. what do you think?

doc/source/notebooks/tailor/kernel_design.pct.py
```diff
@@ -32,8 +32,8 @@
 # %matplotlib inline
 
 # %% [markdown]
-# To make this new kernel class, we inherit from the base class `gpflow.kernels.Kernel` and implement the three functions below. **NOTE:** Depending on the kernel to be implemented, other classes can be more adequate. For example, if the kernel to be implemented is stationary, you can immediately subclass `gpflow.kernels.Stationary` (at which point you 
-# only have to override `K_r` or `K_r2`; see the `Stationary` class docstring).
+# To make this new kernel class, we inherit from the base class `gpflow.kernels.Kernel` and implement the three functions below. **NOTE:** Depending on the kernel to be implemented, other classes can be more adequate. For example, if the kernel to be implemented is isotropic stationary, you can immediately subclass `gpflow.kernels.IsotropicStationary` (at which point you
+# only have to override `K_r` or `K_r2`; see the `IsotropicStationary` class docstring). Anisotropic kernels should subclass `gpflow.kernels.AnisotropicStationary` and override `K_d`.
```
st--:
```suggestion
# only have to override `K_r` or `K_r2`; see the `IsotropicStationary` class docstring). Stationary but anisotropic kernels should subclass `gpflow.kernels.AnisotropicStationary` and override `K_d`.
```

gpflow/kernels/stationaries.py
```diff
@@ -46,35 +46,75 @@ def ard(self) -> bool:
...
+    def K_diag(self, X):
+        return tf.fill(tf.shape(X)[:-1], tf.squeeze(self.variance))
+
+
+class IsotropicStationary(Stationary):
+    """
+    Base class for isotropic stationary kernels, i.e. kernels that only
+    depend on
+
+        d = ‖x - x'‖
```
```suggestion
        r = ‖x - x'‖
```

gpflow/kernels/stationaries.py
```diff
@@ -46,35 +46,75 @@ def ard(self) -> bool:
...
+
+    Derived classes should implement one of:
+
+        K_r2(self, r2): Returns the kernel evaluated on r² (r2), which is the
+        squared scaled Euclidean distance Should operate element-wise on r².
```
```suggestion
        squared scaled Euclidean distance. Should operate element-wise on r2.
```

gpflow/kernels/stationaries.py
```diff
@@ -46,35 +46,75 @@ def ard(self) -> bool:
...
+
+        K_r(self, r): Returns the kernel evaluated on r, which is the scaled
+        Euclidean distance Should operate element-wise on r².
```
```suggestion
        Euclidean distance. Should operate element-wise on r.
```

gpflow/kernels/stationaries.py
```diff
@@ -46,35 +46,75 @@ def ard(self) -> bool:
...
     def K_r2(self, r2):
-        """
-        Returns the kernel evaluated on r² (`r2`), which is the squared scaled Euclidean distance
-        Should operate element-wise on r²
-        """
         if hasattr(self, "K_r"):
             # Clipping around the (single) float precision which is ~1e-45.
             r = tf.sqrt(tf.maximum(r2, 1e-40))
-            return self.K_r(r)  # pylint: disable=no-member
+            return self.K_r(r)
```
```suggestion
            return self.K_r(r)  # pylint: disable=no-member
```

gpflow/kernels/stationaries.py
```diff
@@ -46,35 +46,75 @@ def ard(self) -> bool:
...
+        raise NotImplementedError
+
+    def scaled_squared_euclid_dist(self, X, X2=None):
+        """
+        Returns ||(X - X2ᵀ) / ℓ||² i.e. squared L2-norm.
```
```suggestion
        Returns ‖(X - X2ᵀ) / ℓ‖², i.e. the squared L₂-norm.
```

gpflow/kernels/periodic.py
```diff
@@ -65,20 +66,15 @@ def K_diag(self, X: tf.Tensor) -> tf.Tensor:
         return self.base_kernel.K_diag(X)
 
     def K(self, X: tf.Tensor, X2: Optional[tf.Tensor] = None) -> tf.Tensor:
-        if X2 is None:
-            X2 = X
-
-        # Introduce dummy dimension so we can use broadcasting
-        # TODO: does not support kernel broadcasting
```
st--: does it now handle broadcasting correctly? if so, could you activate the corresponding test (tests/gpflow/kernels/test_broadcasting.py)?

gpflow/kernels/stationaries.py
```diff
@@ -46,35 +46,75 @@ def ard(self) -> bool:
...
+class AnisotropicStationary(Stationary):
+    """
+    Base class for anisotropic stationary kernels, i.e. kernels that only
+    depend on
 
+        d = x - x'
+
+    Derived classes should implement K_d(self, d): Returns the kernel evaluated
+    on d, which is pairwise difference matrix, scaled by the lengthscale
+    parameter ℓ (i.e. Σ_i [(X - X2ᵀ) / ℓ]_i). The last axis corresponds to
```
```suggestion
    parameter ℓ (i.e. Σᵢ [(X - X2ᵀ) / ℓ]ᵢ). The last axis corresponds to
```

gpflow/kernels/stationaries.py
```diff
@@ -46,35 +46,75 @@ def ard(self) -> bool:
...
+    input dimension.
```
```suggestion
    the input dimension.
```

gpflow/kernels/stationaries.py
```diff
@@ -171,17 +211,18 @@ def K_r(self, r):
...
-    r  is the Euclidean distance between the input points, scaled by the lengthscale parameter ℓ,
+    d  is the sum of the per-dimension differences between the input points, scaled by the lengthscale parameter ℓ (i.e. Σ_i [(X - X2ᵀ) / ℓ]_i),
     σ² is the variance parameter.
     """
 
-    def K_r(self, r):
-        return self.variance * tf.cos(r)
+    def K_d(self, d):
+        d = tf.reduce_sum(d, axis=-1)
+        return self.variance * tf.cos(2 * np.pi * d)
```
st--: You're changing the behaviour of Cosine here. Is this intentional? If so, could you make this change and your motivation clear in the PR description please (as it'll be breaking backwards compatibility)?

gpflow/kernels/stationaries.py
```diff
@@ -171,17 +211,18 @@ def K_r(self, r):
...
     where:
-    r  is the Euclidean distance between the input points, scaled by the lengthscale parameter ℓ,
+    d  is the sum of the per-dimension differences between the input points, scaled by the lengthscale parameter ℓ (i.e. Σ_i [(X - X2ᵀ) / ℓ]_i),
     σ² is the variance parameter.
     """
 
-    def K_r(self, r):
-        return self.variance * tf.cos(r)
+    def K_d(self, d):
+        d = tf.reduce_sum(d, axis=-1)
```
st--: Is it correct that the sum is inside the cos ?

gpflow/kernels/stationaries.py
```diff
@@ -171,17 +211,18 @@ def K_r(self, r):
...
     """
     The Cosine kernel. Functions drawn from a GP with this kernel are sinusoids
     (with a random phase).  The kernel equation is
 
-        k(r) = σ² cos{r}
+        k(r) = σ² cos{2πd}
 
     where:
-    r  is the Euclidean distance between the input points, scaled by the lengthscale parameter ℓ,
+    d  is the sum of the per-dimension differences between the input points, scaled by the lengthscale parameter ℓ (i.e. Σ_i [(X - X2ᵀ) / ℓ]_i),
```
```suggestion
    d  is the sum of the per-dimension differences between the input points, scaled by the
    lengthscale parameter ℓ (i.e. Σᵢ [(X - X2ᵀ) / ℓ]ᵢ),
```

gpflow/utilities/ops.py
```diff
@@ -102,6 +102,17 @@ def square_distance(X, X2):
     return dist
 
 
+def distance(X, X2):
```
st--: I would expect a "distance" to be scalar. Would it be more accurate to call this, e.g., `difference_matrix`?

gpflow/utilities/ops.py
```diff
@@ -102,6 +102,17 @@ def square_distance(X, X2):
...
+    """
+    Returns (X - X2ᵀ)
+
+    This function can deal with leading dimensions in X and X2.
```
st--: Could you add what the shape of the return value is, given shapes of the arguments? (Ideally asserted in code using `tf.debugging.assert_shapes` - see #1219 for some examples!)

tests/gpflow/kernels/test_kernels.py
```diff
@@ -172,7 +172,11 @@ def test_periodic_bad_ard_period():
         gpflow.kernels.Periodic(base_kernel, period=[1.0, 1.0, 1.0])
 
 
-kernel_setups = [kernel() for kernel in gpflow.kernels.Stationary.__subclasses__()] + [
+kernel_setups = [
+    kernel()
+    for stationary in gpflow.kernels.Stationary.__subclasses__()
+    for kernel in stationary.__subclasses__()
```
st--:
```suggestion
    for kernel in gpflow.ci_utils.subclasses(gpflow.kernels.Stationary)
```
You will have to add `import gpflow.ci_utils` at the top of the file.

(Sorry, I missed this case in #1340!)

tests/gpflow/kernels/test_kernels.py
```diff
@@ -297,11 +301,11 @@ def test_white(N, D):
...
 
-_kernel_classes_slice = [kernel for kernel in gpflow.kernels.Stationary.__subclasses__()] + [
-    gpflow.kernels.Constant,
-    gpflow.kernels.Linear,
-    gpflow.kernels.Polynomial,
-]
+_kernel_classes_slice = [
+    kernel
+    for stationary in gpflow.kernels.Stationary.__subclasses__()
+    for kernel in stationary.__subclasses__()
```
st--:
```suggestion
    for kernel in gpflow.ci_utils.subclasses(gpflow.kernels.Stationary)
```

tests/gpflow/kernels/test_positive_semidefinite.py
```diff
@@ -0,0 +1,49 @@
...
+
+@pytest.mark.parametrize("kernel", KERNEL_CLASSES)
+def test_positive_semidefinite(kernel):
+    N, D = 100, 5
+    X = rng.randn(N, D)
+    kern = kernel()
+
+    cov = kern(X)
+    eig = tf.linalg.eigvalsh(cov).numpy()
+    assert_array_less(-np.finfo(eig.dtype).eps * 1e4, eig)
```
st--:
```suggestion
    assert_array_less(-np.finfo(eig.dtype).eps * 1e4, eig)
```
Would it be clearer what is compared against to write the expected tolerance explicitly, e.g. `-1e-9`?

tests/gpflow/kernels/test_positive_semidefinite.py
```diff
@@ -0,0 +1,49 @@
...
+    kernels.SquaredExponential,
+    kernels.RationalQuadratic,
+    kernels.Exponential,
+    kernels.Matern12,
+    kernels.Matern32,
+    kernels.Matern52,
+    kernels.Cosine,
+    kernels.Linear,
+    kernels.Polynomial,
+]
```
st--: Could we replace this explicit list with another `gpflow.ci_utils.subclasses(gpflow.kernels...)` call?
How about using similar code to test_broadcasting to ensure all kernel classes are tested (except for known exceptions) ?

tests/gpflow/kernels/test_positive_semidefinite.py
```diff
@@ -0,0 +1,49 @@
...
+
+rng = np.random.RandomState(42)
+
+
+@pytest.mark.parametrize("kernel", KERNEL_CLASSES)
+def test_positive_semidefinite(kernel):
```
st--: thanks for adding this test! you could add the link to the corresponding issue: https://github.com/GPflow/GPflow/issues/1328

tests/gpflow/kernels/test_positive_semidefinite.py
```diff
@@ -0,0 +1,49 @@
...
+    kernels.Matern52,
+    kernels.Cosine,
+    kernels.Linear,
+    kernels.Polynomial,
+]
+
+rng = np.random.RandomState(42)
+
+
+@pytest.mark.parametrize("kernel", KERNEL_CLASSES)
```
st--:
```suggestion
@pytest.mark.parametrize("kernel_class", KERNEL_CLASSES)
```

tests/gpflow/kernels/test_positive_semidefinite.py
st--:
```suggestion
def test_positive_semidefinite(kernel_class):
```

tests/gpflow/kernels/test_positive_semidefinite.py
```diff
@@ -0,0 +1,49 @@
...
+def test_positive_semidefinite(kernel):
+    N, D = 100, 5
+    X = rng.randn(N, D)
+    kern = kernel()
+
+    cov = kern(X)
```
st--:
```suggestion
    kernel = kernel_class()

    cov = kernel(X)
```

gpflow/kernels/stationaries.py
```diff
@@ -46,35 +46,75 @@ def ard(self) -> bool:
...
+    Derived classes should implement K_d(self, d): Returns the kernel evaluated
+    on d, which is pairwise difference matrix, scaled by the lengthscale
+    parameter ℓ (i.e. Σ_i [(X - X2ᵀ) / ℓ]_i). The last axis corresponds to
+    input dimension.
+    """
+
+    def K(self, X, X2=None):
+        return self.K_d(self.scaled_distance(X, X2))
+
+    def scaled_distance(self, X, X2=None):
```
st--: as in the utilities module: "distance" would let me expect a scalar - how about calling this `scaled_difference_matrix`?

what would its return shape be?

gpflow/kernels/stationaries.py
```diff
@@ -46,35 +46,75 @@ def ard(self) -> bool:
...
+
+class AnisotropicStationary(Stationary):
+    """
+    Base class for anisotropic stationary kernels, i.e. kernels that only
+    depend on
 
+        d = x - x'
+
+    Derived classes should implement K_d(self, d): Returns the kernel evaluated
+    on d, which is pairwise difference matrix, scaled by the lengthscale
```
```suggestion
    on d, which is the pairwise difference matrix, scaled by the lengthscale
```

gpflow/kernels/stationaries.py
st--: Though I think this should not include the sum here ?

gpflow/kernels/stationaries.py
ilia-kats: I'm not. In the previous version of this pull request, K_d would get the already summed matrix. Since we're now making an extra Anisotropic subclass, one could imagin that other kernels need to do something different with the distance, so now the scaled_difference_matrix returns a matrix of distance vectors, and it's up to the kernel to to whatever it wants with it.

st--:
@ilia-kats if you want to chat more synchronously about GPflow, join the gpflow slack workspace! link is in the README (https://github.com/GPflow/GPflow/#slack-workspace) :)

gpflow/kernels/periodic.py
```diff
@@ -65,20 +66,15 @@ def K_diag(self, X: tf.Tensor) -> tf.Tensor:
...
+            if hasattr(self.base_kernel, "K_r"):
+                sine_r = tf.reduce_sum(tf.abs(scaled_sine), -1)
+                K = self.base_kernel.K_r(sine_r)
+            else:
+                sine_r2 = tf.reduce_sum(tf.square(scaled_sine), -1)
+                K = self.base_kernel.K_r2(sine_r2)
         else:
-            sine_r2 = tf.reduce_sum(tf.square(scaled_sine), -1)
-            K = self.base_kernel.K_r2(sine_r2)
+            K = self.base_kernel.K_d(scaled_sine)
```
st--: will this always form a positive-definite kernel ?

ilia-kats: I suppose not, it would be up to the user to make sure that this is a valid kernel. For now, it looks like it's not positive semidefinite with the cosine kernel. I can add some documentation for this

st--: The Periodic kernel is positive-definite for IsotropicStationary kernels, so perhaps it'd be better for now to simply say `Periodic` requires an IsotropicStationary (not generally Stationary) subclass as its base_kernel, and take out the special-casing for AnisotropicStationary ?

gpflow/kernels/stationaries.py
st--: I meant the factor 2π between introduced by this change, not the sum!

gpflow/utilities/ops.py
```diff
@@ -103,6 +103,23 @@ def square_distance(X, X2):
     return dist
 
 
+def difference_matrix(X, X2):
+    """
+    Returns (X - X2ᵀ)
+
+    This function can deal with leading dimensions in X and X2.
+    If X has shape [..., N, D] and X2 has shape [..., M, D], the
+    output will have shape [..., N, M, D].
```
st--: Note that this is different broadcasting behaviour from `square_distance`. @vdutor do you have any thoughts on what we might want to do here ?

gpflow/kernels/stationaries.py
ilia-kats: Right, sorry. Yes, this is intentional, since this makes the lenghtscale parameter directly interpretable as period length. Given that gpflow 2 is, as far as I can tell, anyway not backwards compatible to gpflow 1 (i.e. models from 1 can not be loaded into 2) and that the Cosine kernel seems to not be widely used (otherwise someone would have noticed that it doesn't work with multivariate data much sooner), I don't think that this is a big issue.

gpflow/kernels/stationaries.py
ilia-kats: yes, see https://github.com/GPflow/GPflow/pull/1357#discussion_r396715177

tests/gpflow/kernels/test_broadcasting.py
```diff
@@ -62,26 +69,34 @@ def test_no_kernels_missed(kernel_class):
         gpflow.kernels.base.ReducingCombination,
         kernels.Static,
         kernels.Stationary,
+        kernels.IsotropicStationary,
+        kernels.AnisotropicStationary,
     ]
+    needs_constructor_parameters = [kernels.Periodic]
     if kernel_class in tested_kernel_classes:
         return  # tested
```
st--:
```suggestion
        return  # tested by test_broadcast_no_active_dims
```

tests/gpflow/kernels/test_broadcasting.py
```diff
@@ -62,26 +69,34 @@ def test_no_kernels_missed(kernel_class):
...
     if kernel_class in skipped_kernel_classes:
         return  # not tested but currently expected to fail
     if kernel_class in abstract_base_classes:
         return  # cannot test abstract base classes
+    if kernel_class in needs_constructor_parameters:
+        return  # this has a separate test
```
st--:
```suggestion
        return  # this has a separate test, see test_broadcast_no_active_dims_periodic
```

tests/gpflow/kernels/test_broadcasting.py
```diff
@@ -44,12 +44,19 @@
     # Following kernels do not broadcast: see https://github.com/GPflow/GPflow/issues/1339
     pytest.param(kernels.ArcCosine, marks=pytest.mark.xfail),  # broadcasting not implemented
     pytest.param(kernels.Coregion, marks=pytest.mark.xfail),  # broadcasting not implemented
-    pytest.param(kernels.Periodic, marks=pytest.mark.xfail),  # broadcasting not implemented
     pytest.param(kernels.ChangePoints, marks=pytest.mark.xfail),  # broadcasting not implemented
     pytest.param(kernels.Convolutional, marks=pytest.mark.xfail),  # broadcasting not implemented
 ]
 
 
+def broadcast_no_no_active_dims(kernel):
```
st--:
```suggestion
def check_broadcasting(kernel):
```
might make it clearer what's happening?

gpflow/utilities/ops.py
ilia-kats: This should be fixed now.

gpflow/utilities/ops.py
```diff
@@ -103,6 +103,29 @@ def square_distance(X, X2):
...
 
+def difference_matrix(X, X2):
+    """
+    Returns (X - X2ᵀ)
+
+    This function can deal with leading dimensions in X and X2.
+    For example, If X has shape [M, D] and X2 has shape [N, D],
+    the output will have shape [M, N, D]. If X has shape [I, J, M, D]
+    and X2 has shape [K, L, N, D], the output will have shape
+    [I, J, M, K,L, N, D].
```
st--:
```suggestion
    [I, J, M, K, L, N, D].
```

gpflow/utilities/ops.py
```diff
@@ -103,6 +103,29 @@ def square_distance(X, X2):
     return dist
 
 
+def difference_matrix(X, X2):
```
st--: Would you be willing to add a test for this? e.g. along the lines of
```python
def test_difference_matrix_broadcasting_symmetric():
    X = np.random.randn(5, 4, 3, 2)
    d = difference_matrix(X, None)
    assert d.shape == (5, 4, 3, 2)

def test_difference_matrix_broadcasting_cross():
    X = np.random.randn(2, 3, 4, 5)
    X2 = np.random.randn(8, 7, 6, 5)
    d = difference_matrix(X, X2)
    assert d.shape == (2, 3, 4, 8, 7, 6, 5)
```

ilia-kats:
Sure, I can add a test, but I'm not sure how useful it is. With this implementation, the shapes will always match due to the reshape operation at the end.

gpflow/utilities/ops.py
```diff
@@ -103,6 +103,29 @@ def square_distance(X, X2):
...
+    the output will have shape [M, N, D]. If X has shape [I, J, M, D]
+    and X2 has shape [K, L, N, D], the output will have shape
+    [I, J, M, K, L, N, D].
+    """
+    if X2 is None:
+        X2 = X
+        diff = X[..., :, tf.newaxis, :] - X2[..., tf.newaxis, :, :]
+        return diff
+    Xshape = X.shape
+    X2shape = X2.shape
```
st--: Actually, this needs to be
```suggestion
    Xshape = tf.shape(X)
    X2shape = tf.shape(X2)
```
to work inside a tf.function() with potentially unknown static tensor shapes.

st--:
> Sure, I can add a test, but I'm not sure how useful it is. With this implementation, the shapes will always match due to the reshape operation at the end.

I think it's useful to have a test that the reshaping does work the way you expected it to. I've left you a version that you can mostly copy & paste (subject to `make format` and adding imports). :)

PR #1362, opened 2020-03-24
mixin proposal

condnsdmatters:


gpflow/models/training_interface.py
```diff
@@ -12,25 +12,25 @@
...
 
 
-class InternalDataTrainingInterface:
+class InternalDataTrainingLossMixin:
     def training_loss(self):
         """
         Specify a training loss for this model
         """
-        raise NotImplementedError
+        return -self.maximum_a_posteriori_objective(data)
```
st--: I presume you mean
```suggestion
        return -self.maximum_a_posteriori_objective()
```
?

gpflow/models/mixins.py
```diff
@@ -44,3 +44,11 @@ def closure():
             return training_loss(batch)
 
         return closure
+
+
+class MCMCTrainingLossMixin(InternalDataTrainingLossMixin):
```
joelberkeley-pio: i don't know what MCMC is. clearly there's something about MCMC that means it has a different training loss. Should this also be conveyed in the names of the classes `InternalDataTrainingLossMixin` and the other one? Why, for example, don't they use the `log_posterior_density`? what's special about them?

gpflow/models/model.py
```diff
@@ -46,41 +52,25 @@ def log_prior_density(self) -> tf.Tensor:
...
         """
-        return self._call_maximum_likelihood_objective(data) + self.log_prior_density()
+        return self.maximum_likelihood_objective(*args, **kwargs) + self.log_prior_density()
 
-    def training_loss(self, data: Optional[Data] = None) -> tf.Tensor:
+    def maximum_a_posteriori_objective(self, *args, **kwargs) -> tf.Tensor:
         """
         Minimization objective for TensorFlow optimizers (including
         gpflow.optimizers.Scipy). Includes the log prior density for maximum
         a-posteriori (MAP) estimation.
```
```suggestion
        Maximum a-posteriori objective, assuming that maximum_likelihood_objective() is the log marginal likelihood.
```

gpflow/models/mixins.py
```diff
@@ -0,0 +1,54 @@
...
+            batch = next(data) if isinstance(data, collections.abc.Iterator) else data
+            return training_loss(batch)
+
+        return closure
+
+
+class MCMCTrainingLossMixin(InternalDataTrainingLossMixin):
+    def training_loss(self):
+        """
+        Specify a training loss for this model
```
st--:
```suggestion
        Training loss for gradient-based relaxation of MCMC models.
```

gpflow/models/mixins.py
```diff
@@ -0,0 +1,54 @@
...
+    def training_loss_closure(self, jit=True) -> Callable[[], tf.Tensor]:
+        if jit:
+            return tf.function(self.training_loss)
+        return self.training_loss
+
+
+class ExternalDataTrainingLossMixin:
+    def training_loss(self, data):
+        """
+        Specify a training loss for this model
```
st--:
```suggestion
        Training loss for models that do not encapsulate the data.
        
        :param data: the data to be passed to the model objective.
```

gpflow/models/mixins.py
```diff
@@ -0,0 +1,54 @@
+import abc
```
st--:
```suggestion
"""
This module defines training interfaces (realized as mix-in classes) that provide the minimization objective for TensorFlow optimizers (including gpflow.optimizers.Scipy). This `training_loss`  includes the log prior density for maximum a-posteriori (MAP) estimation.
"""

import abc
```

st--: Going to rename this module to training_interface.py as it provides the data types etc.

doc/source/notebooks/tailor/mixture_density_network.pct.py
```diff
@@ -85,7 +85,7 @@ def sinusoidal_data(N, noise):
 
 # %%
 import gpflow
-from gpflow.models import BayesianModel
+from gpflow.models import ExternalDataBayesianModel
```
st--:
```suggestion
from gpflow.models import BayesianModel, ExternalDataTrainingLossMixin
```

doc/source/notebooks/tailor/mixture_density_network.pct.py
```diff
@@ -97,7 +97,7 @@ def sinusoidal_data(N, noise):
 # %%
 from typing import Callable, Optional, Tuple
 
-class MDN(BayesianModel):
+class MDN(ExternalDataBayesianModel):
```
st--:
```suggestion
class MDN(BayesianModel, ExternalDataTrainingLossMixin):
```

gpflow/models/__init__.py
```diff
@@ -18,6 +18,11 @@
 from .gpmc import GPMC
 from .gpr import GPR
 from .model import BayesianModel, GPModel
+from .mixins import (
```
st--:
```suggestion
from .training_interface import (
```

gpflow/models/mixins.py
```diff
@@ -0,0 +1,54 @@
...
+InputData = tf.Tensor
+OutputData = tf.Tensor
+RegressionData = Tuple[InputData, OutputData]
+Data = TypeVar("Data", RegressionData, InputData)
+
+
+class InternalDataTrainingLossMixin:
+    def training_loss(self):
+        """
+        Specify a training loss for this model
```
st--:
```suggestion
        Training loss for models that encapsulate their data.
```

gpflow/models/__init__.py
st--: (as it defines all the types I think "mixin" is rather confusing; it does all relate to 'data' so "training_interface" seems appropriate)

gpflow/models/model.py
```diff
@@ -18,6 +18,16 @@
 
 import numpy as np
 import tensorflow as tf
+from tensorflow.python.data.ops import iterator_ops
+
+from .training_interface import (
+    InternalDataTrainingLossMixin,
+    ExternalDataTrainingLossMixin,
```
st--: I think these aren't actually needed here ?
```suggestion
```
Or if you want them in here to expose them to people who want to build models, we should add the mcmc one:
```suggestion
    InternalDataTrainingLossMixin,
    ExternalDataTrainingLossMixin,
    MCMCTrainingLossMixin,
```

gpflow/models/model.py
```diff
@@ -18,6 +18,16 @@
 
 import numpy as np
 import tensorflow as tf
+from tensorflow.python.data.ops import iterator_ops
```
```suggestion
```

gpflow/models/mixins.py
```diff
@@ -0,0 +1,54 @@
...
+from typing import Callable, Optional, Tuple, TypeVar
+
+
+import tensorflow as tf
+import numpy as np
+
+InputData = tf.Tensor
+OutputData = tf.Tensor
+RegressionData = Tuple[InputData, OutputData]
+Data = TypeVar("Data", RegressionData, InputData)
```
joelberkeley-pio:
```suggestion
Data = Union[RegressionData, InputData]
```

gpflow/models/mixins.py
```diff
@@ -0,0 +1,54 @@
...
+
+    def training_loss_closure(self, data: Data, jit=True) -> Callable[[], tf.Tensor]:
+        training_loss = self.training_loss
+        if jit:
+            training_loss = tf.function(
+                self.training_loss
+            )  # TODO need to add correct input_signature here to allow for differently sized minibatches
+
+        def closure():
+            batch = next(data) if isinstance(data, collections.abc.Iterator) else data
```
joelberkeley-pio: when is a `tf.Tensor` an `Iterator`?

joelberkeley-pio: what happens when `next` raises `StopIteration`? Will the optimizer handle that?

gpflow/models/mixins.py
```diff
@@ -0,0 +1,54 @@
...
+        """
+        Specify a training loss for this model
+        """
+        return -self.maximum_a_posteriori_objective(data)
+
+    def training_loss_closure(self, data: Data, jit=True) -> Callable[[], tf.Tensor]:
+        training_loss = self.training_loss
+        if jit:
+            training_loss = tf.function(
+                self.training_loss
```
joelberkeley-pio:
```suggestion
               training_loss
```

gpflow/models/__init__.py
condnsdmatters: i think mixins keeps it clearer that the pattern we are going for is a mixin pattern

gpflow/models/mixins.py
condnsdmatters: good question: i dont know. normally you would use a tf.data.Datatset and set it to loop so you never hit `StopIteration`

joelberkeley-pio: i think we'd need to be clear on that before we can be confident we support batch processing

gpflow/models/mixins.py
st--: what's the difference ?

gpflow/models/mixins.py
st--: log_posterior_density is the right thing to compute for an MCMC model. maximum likelihood / map objective is the right thing to compute for the other models.

gpflow/models/mixins.py
st--: agreed with @condnsdmatters to move types into `gpflow.base` and rename this to `gpflow.models.training_mixins`

gpflow/models/mixins.py
st--: a tensor isn't an iterator, it's an iterator if it's a dataset to iterate over (e.g. minibatching)

st--: you need to call `.repeat()` on your dataset (otherwise you can't train properly!) and then it loops forever

gpflow/models/mixins.py
```diff
@@ -0,0 +1,56 @@
...
+class ExternalDataTrainingLossMixin:
+    def training_loss(self, data):
+        """
+        Training loss for models that do not encapsulate the data.
+        
+        :param data: the data to be used for computing the model objective.
+        """
+        return -self.maximum_a_posteriori_objective(data)
+
+    def training_loss_closure(self, data: Data, jit=True) -> Callable[[], tf.Tensor]:
```
st--:
```suggestion
    def training_loss_closure(self, data: Union[Data, collections.abc.Iterator], jit=True) -> Callable[[], tf.Tensor]:
```

gpflow/models/mixins.py
st--: see type signature amendment above

st--: I think it is going to be fine as it is for now. If you exhaust your training set but insist on keep having more optimisation steps your code crashes :woman_shrugging:

joelberkeley-pio: how easy is that to do? isn't it just a case of specifying too high a number of iterations? that sounds horrendously fragile

gpflow/models/mixins.py
joelberkeley-pio: what i wonder is if we can be a bit more be a bit more descriptive in the names of the other mixins

Issue #1291, opened 2020-02-26
Allow our Scipy optimizer to store the jitted version of what gets passed to scipy's minimize


st--:
gpflow.optimizers.Scipy() takes a `jit` argument, which if True will wrap the objective-and-gradient computation in tf.function() because that will be executed by scipy.optimize.minimize() many many times. However, the tf.function() wrap comes with an overhead, so it would be good if we could access (and directly use) the jitted version of objective-and-gradients when repeatedly minimizing the same model (e.g. with the data being `tf.Variable`s and assigning different values).

In the gpflow codebase itself, this directly affects FITCvsVFE.py, which repeatedly calls the minimizer to optimize as hard as possible.

PR #1364, opened 2020-03-24
fix typing in gpflow.optimizers.scipy

st--:


gpflow/optimizers/scipy.py
```diff
@@ -8,7 +8,7 @@
 __all__ = ["Scipy"]
 
 Loss = tf.Tensor
-Variables = Tuple[tf.Variable]
+Variables = Iterable[tf.Variable]
```
joelberkeley-pio:
```suggestion
Variables = Sequence[tf.Variable]
```
They're ordered right? I believe an `Iterable` doesn't have to be ordered, for example a set

gpflow/optimizers/scipy.py
```diff
@@ -102,13 +107,13 @@ def _callback(x):
         return _callback
 
     @staticmethod
-    def pack_tensors(tensors: Iterator[tf.Tensor]) -> tf.Tensor:
+    def pack_tensors(tensors: Iterable[tf.Tensor]) -> tf.Tensor:
```
joelberkeley-pio:
```suggestion
    def pack_tensors(tensors: Variables) -> tf.Tensor:
```
perhaps

joelberkeley-pio: as well as similar cases below

gpflow/optimizers/scipy.py
```diff
@@ -68,11 +69,13 @@ def minimize(
         )
 
     @classmethod
-    def initial_parameters(cls, variables):
+    def initial_parameters(cls, variables: Variables) -> tf.Tensor:
```
joelberkeley-pio: separate question from this PR, but why are all these methods public? they don't appear to be used anywhere

gpflow/optimizers/scipy.py
joelberkeley-pio: does the ordering matter for the variables? does it need to be preserved in any way?

gpflow/optimizers/scipy.py
```diff
@@ -53,6 +53,7 @@ def minimize(
         """
         if not callable(closure):
             raise TypeError("Callable object expected.")  # pragma: no cover
+        variables = tuple(variables)
```
joelberkeley-pio: why convert to tuple?

gpflow/optimizers/scipy.py
```diff
@@ -8,7 +8,7 @@
 __all__ = ["Scipy"]
 
 Loss = tf.Tensor
-Variables = Tuple[tf.Variable]
+Variables = Iterable[tf.Variable]
 StepCallback = Callable[[int, Variables, List[tf.Tensor]], None]
 LossClosure = Callable[..., tf.Tensor]
```
joelberkeley-pio: while you're here, is this right? should it not be `Callable[[], tf.Tensor]`?

gpflow/optimizers/scipy.py
st--: in case it IS a single-pass iterator because we need it afterwards...

gpflow/optimizers/scipy.py
st--: the function itself just needs something it can iterate over - that it'll be used for the variables is up to the caller, not the callee, no?

joelberkeley-pio:
remind me what the stance is on running mypy for gpflow. even when I'm being careful with types there are sometimes some that I get wrong that mypy picks up so it's hard to be confident reviewing a PR like this

gpflow/optimizers/scipy.py
st--: @awav would have to be the one answering that.

gpflow/optimizers/scipy.py
st--: No. That's why it's turned into a tuple inside the minimize()

gpflow/optimizers/scipy.py
joelberkeley-pio: yeah i get what you mean. i think either's fine. happy to leave as is

st--:
> remind me what the stance is on running mypy for gpflow.

You can run it to check it, but I would not want to enforce it on the whole codebase.

Even if there's any mistakes still left, I think this PR is a step in the right direction. I looked at the types thinking about a separate issue (#1291) and thought these definitely don't look right so let's go and fix them...

gpflow/optimizers/scipy.py
joelberkeley-pio: good point

gpflow/optimizers/scipy.py
```diff
@@ -121,12 +126,14 @@ def unpack_tensors(to_tensors: Iterator[tf.Tensor], from_vector: tf.Tensor) -> L
...
-    def assign_tensors(to_tensors: Iterator[tf.Variable], values: Iterator[tf.Tensor]):
+    def assign_tensors(to_tensors: Iterable[tf.Variable], values: Iterable[tf.Tensor]) -> None:
         for tensor, tensor_vector in zip(to_tensors, values):
             tensor.assign(tensor_vector)
 
 
-def _compute_loss_and_gradients(loss_cb: LossClosure, variables: Variables):
+def _compute_loss_and_gradients(
+    loss_cb: LossClosure, variables: Variables
+) -> Tuple[tf.Tensor, tf.Tensor]:
```
joelberkeley-pio: the return type of [`tf.GradientTape.gradient`](https://www.tensorflow.org/api_docs/python/tf/GradientTape#gradient) is

> a list or nested structure of Tensors (or IndexedSlices, or None), one for each element in sources. Returned structure is the same as the structure of sources.

how do we know which one we're getting?

joelberkeley-pio: the return type of [`tf.GradientTape.gradient`](https://www.tensorflow.org/api_docs/python/tf/GradientTape#gradient) is

> a list or nested structure of Tensors (or IndexedSlices, or None), one for each element in sources. Returned structure is the same as the structure of sources.

Since `variables` is an `Iterable[tf.Tensor]`, how do we know what the type of `grads` is? We know it's the same as `variables`, but do we know more? Would it not be

```python3
V = TypeVar("V", bound=Variables)

def _compute_loss_and_gradients(
    loss_cb: LossClosure, variables: Variables
) -> Tuple[tf.Tensor, V]:
	...
```
though that seems overkill here. Since it's only used with a `Tuple[tf.Tensor, ...]`, we could be more restrictive in the optimizer and use that instead of `Iterable[tf.Tensor]`. Then we wouldn't need the TypeVar and can just have
```python3
def _compute_loss_and_gradients(
    loss_cb: LossClosure, variables: Tuple[tf.Tensor, ...]
) -> Tuple[tf.Tensor, Tuple[tf.Tensor, ...]]:
```

gpflow/optimizers/scipy.py
```diff
@@ -83,17 +86,19 @@ def _tf_eval(x: tf.Tensor) -> Tuple[tf.Tensor, tf.Tensor]:
         if jit:
             _tf_eval = tf.function(_tf_eval)
 
-        def _eval(x):
+        def _eval(x: np.ndarray) -> Tuple[np.ndarray, np.ndarray]:
```
joelberkeley-pio: looks like the result of `eval_func` is passed as the first arg to the scipy minimize function, which requires

>     fun : callable
        The objective function to be minimized.

            ``fun(x, *args) -> float``

        where x is an 1-D array with shape (n,) and `args`
        is a tuple of the fixed parameters needed to completely
        specify the function.

which doesn't appear to match with this signature. have i missed sth?

gpflow/optimizers/scipy.py
st--: I'd prefer the TypeVar, I find your second suggestion hard to read, and when it gets to the point that it gets in the way of understanding what's going on for the human I'd rather remove the type annotation altogether...

gpflow/optimizers/scipy.py
st--: I don't understand? there are no fixed parameters.

gpflow/optimizers/scipy.py
awav: @joelberkeley-pio, "They don't appear anywhere" doesn't mean that methods cannot be useful for a user. Feel free to address questions of testing and making methods protected/private in _separate_ PR.

awav:
I would prefer a centralized fix for types with one style along with the project. I'm surprised by how many times people tried to change types in scipy file. Sorry, but I must stress it again - this is not a high priority issue for the release 2.0, there are too many places where we could make some adjustments with types.

@joelberkeley-pio, do you (because you have experience with types in python, mypy and other related "things") want to make a task of adding types to the gpflow and take a lead on that?

gpflow/optimizers/scipy.py
joelberkeley-pio: the return type of this function is `(np.ndarray, np.ndarray)`, whereas the return type expected is `float`

st--: That's a flaw in the scipy documentation.

PR #1365, opened 2020-03-24
Move TODO into NotImplementedError message

vdutor:


gpflow/conditionals/uncertain_conditionals.py
```diff
@@ -43,10 +43,9 @@ def uncertain_conditional(
         raise NotImplementedError
 
     if full_cov:
-        # TODO(VD): ``full_cov`` True would return a ``fvar`` of shape [N, N, D, D],
-        # encoding the covariance between input datapoints as well.
-        # This is not implemented as this inducing_variable is only used for plotting purposes.
-        raise NotImplementedError
+        raise NotImplementedError(
+            "The method `uncertain_conditional` currently does not support `full_cov=True`"
```
st--: nit: it's not a method, and I think in error messages markdown tends to be distracting rather than helpful...
```suggestion
            "uncertain_conditional() currently does not support full_cov=True"
```

PR #1374, opened 2020-03-26
MCMC cleanup

st--:
Currently, the SamplingHelper in gpflow.optimizers.mcmc accepts both gpflow.Parameter and tf.Variable instances in its `model_parameters` constructor argument. Following the pattern of all gradient-based optimizers, it is easy to accidentally pass in `model.trainable_variables`, which *works* (it directly returns all unconstrained variables) but will compute the wrong thing when any of the parameters with prior have a transform set (e.g. positive() for kernel hyperparameters), as the log-det of the Jacobian of the transform is only corrected for Parameter instances.


gpflow/optimizers/mcmc.py
```diff
@@ -21,13 +21,10 @@
 __all__ = ["SamplingHelper"]
 
 
-ModelParameters = Sequence[TypeVar("ModelParameter", tf.Variable, Parameter)]
-LogProbabilityFunction = Callable[[ModelParameters], tf.Tensor]
```
st--: Both types were inaccurate and were only used in a single place.

gpflow/optimizers/mcmc.py
```diff
@@ -47,25 +44,21 @@ def run_chain_fn():
...
-        self, target_log_prob_fn: LogProbabilityFunction, model_parameters: ModelParameters
+        self, target_log_prob_fn: Callable[[], tf.Tensor], model_parameters: Sequence[Parameter]
     ):
-        assert all([isinstance(p, (Parameter, tf.Variable)) for p in model_parameters])
+        assert all(isinstance(p, Parameter) for p in model_parameters)
 
         self._model_parameters = model_parameters
         self._target_log_prob_fn = target_log_prob_fn
 
-        self._parameters = []
```
st--: was the same as `self._model_parameters`, and only used in tests

gpflow/optimizers/mcmc.py
```diff
@@ -47,25 +44,24 @@ def run_chain_fn():
...
 
     def __init__(
-        self, target_log_prob_fn: LogProbabilityFunction, model_parameters: ModelParameters
+        self, target_log_prob_fn: Callable[[], tf.Tensor], model_parameters: Sequence[Parameter]
     ):
-        assert all([isinstance(p, (Parameter, tf.Variable)) for p in model_parameters])
+        if not all(isinstance(p, Parameter) for p in model_parameters):
+            raise TypeError(
+                "SamplingHelper only accepts `gpflow.Parameter`s (wrap tf.Variable if required)"
+            )
```
frgsimpson:
```suggestion
        if not all(p.prior is not None for p in model_parameters):
            logger.warning('Priors not set on all parameters')
```

frgsimpson: Now that we know they are all parameters, it might also be worth checking that priors have been assigned to each of them.

gpflow/optimizers/mcmc.py
```diff
@@ -21,13 +21,10 @@
 __all__ = ["SamplingHelper"]
 
 
-Parameters = Sequence[Parameter]
-LogProbabilityFunction = Callable[[], tf.Tensor]
```
st--: was only used once - seemed cleaner to simply write the type in the function signature (especially given how simple these are)

gpflow/optimizers/mcmc.py
st--: would you prefer a warning for parameter but no prior set? or are you fine with just raising a ValueError as soon as anything isn't a parameter-with-prior?

gpflow/optimizers/mcmc.py
```diff
@@ -46,15 +43,18 @@ def run_chain_fn():
...
         hmc_samples = run_chain_fn()
         parameter_samples = hmc_helper.convert_samples_to_parameter_values(hmc_samples)
-
-    :param target_log_prob_fn: a callable which represents log-density under
-        the target distribution.
-    :param parameters: List of :class:`gpflow.Parameter` used as a state of the Markov chain.
     """
 
-    def __init__(self, target_log_prob_fn: LogProbabilityFunction, parameters: Parameters):
-        if not all([isinstance(p, Parameter) and p.prior is not None for p in parameters]):
```
st--: as it's wrapped in `all`, can simply use the generator form instead of the list comprehension

tests/gpflow/optimizers/test_mcmc.py
```diff
@@ -45,11 +45,8 @@ def test_mcmc_helper_parameters():
 
     for i in range(len(model.trainable_parameters)):
         assert model.trainable_parameters[i].shape == hmc_helper.current_state[i].shape
-        assert model.trainable_parameters[i] == hmc_helper._parameters[i]
-        if isinstance(model.trainable_parameters[i], gpflow.Parameter):
```
st--: no longer needed as they are all required to be parameters

tests/gpflow/optimizers/test_mcmc.py
```diff
@@ -199,7 +196,7 @@ def run_chain_fn():
         assert hmc_helper._parameters[i].numpy() == parameter_samples[i][-1]
 
 
-@pytest.mark.xfail(raises=ValueError)
```
st--: `xfail` is meant for tests that you aren't expecting to pass - e.g. for tests of not-yet-implemented features. To test that code raises an exception as expected (e.g. error handling), use `with pytest.raises()` (see below) - this also allows regexp-matching to check it's got the right error message!

gpflow/optimizers/mcmc.py
```diff
@@ -46,15 +43,18 @@ def run_chain_fn():
 
         hmc_samples = run_chain_fn()
         parameter_samples = hmc_helper.convert_samples_to_parameter_values(hmc_samples)
-
-    :param target_log_prob_fn: a callable which represents log-density under
```
st--: moved to the constructor docstring

doc/source/notebooks/advanced/mcmc.pct.py
```diff
@@ -340,6 +340,10 @@ def plot_joint_marginals(samples, parameters_dict, y_axis_label):
 
 gpflow.utilities.print_summary(model)
 
+# %%
+# TODO confirm whether we should allow MCMC on a parameter without prior, add a prior, or set to non-trainable.
```
st--: SGPMC contains the inducing points Z. Should these be included in the MCMC sampling, despite not having a prior (implicit uniform prior)? Should they be excluded from MCMC (set as non-trainable)? Should we give them a prior (which one)? @jameshensman

gpflow/optimizers/mcmc.py
```diff
@@ -103,15 +108,16 @@ def grad_fn(dy, variables: Optional[tf.Tensor] = None):
...
-        for hmc_variable, param in zip(hmc_samples, self._parameters):
+        for hmc_value, param in zip(hmc_samples, self._parameters):
             if param.transform is not None:
-                value = param.transform.forward(hmc_variable)
+                value = param.transform.forward(hmc_value)
             else:
-                value = hmc_variable
-            values.append(value.numpy())
+                value = hmc_value
+            values.append(value)
```
st--: Likely to be used as a tensor again, remove implicit numpy() conversion.

PR #1375, opened 2020-03-26
make num_latent_gps a non-optional keyword-only argument without a default for the GPModel base class

st--:


cdmatters (approved):

condnsdmatters:
Just need tests to pass, I approve the idea, I think it should be kw too

condnsdmatters:
been assigned this - happy to merge!

PR #1382, opened 2020-03-27
Match original experimental setup for "metalearning with GPs" notebook.

christabella:
Replicate the original experimental setup published in Fortuin and Ratsch (2019) and heavily comment the code with references to the paper.

Changes were made after confirming via an email correspondence with the author of the [reference paper](https://arxiv.org/abs/1901.08098), Vincent Fortuin.

### Minimal working example
I have tested these changes on a notebook I downloaded from https://github.com/GPflow/docs/pull/2.

However, the problem is that now each epoch takes 7 minutes instead of 7 seconds. With 5 epochs (instead of the original 100) this is around half an hour.

## Code quality requirements

For details, look at the code quality requirements section in [contributing.md](../contributing.md). The *main* points are repeated here:

* _New feature's_ code must be accompanied with [detailed](../contributing.md) docstrings.
* _New features_ must be covered with tests.
* _Fixed bug_ must be proved by test suites.
* Use type annotations.
* Use `.pylintrc` for code formatting.

### DIFF
--- a/doc/source/notebooks/tailor/external-mean-function.pct.py
+++ b/doc/source/notebooks/tailor/external-mean-function.pct.py
@@ -46,7 +46,14 @@ tf.random.set_seed(2)
 # To generate the meta and test tasks, we sample from a Gaussian process with an Squared Exponential covariance function and a sinusoidal mean function. Each task is a realization of this process.
 
 # %%
-def generate_data(num_functions=10, N=1000):
+def generate_data(num_functions=10, N=50):
+    """
+    For each function, sample the value at 50 equally spaced
+    points in the [−5, 5] interval (Fortuin and Rätsch, 2019).
+
+    Returns:
+        Tuple of np.arrays of size (N, 1) and (N, num_functions).
+    """
     jitter = 1e-6
     Xs = np.linspace(-5.0, 5.0, N)[:, None]
     kernel = RBF(lengthscales=1.0)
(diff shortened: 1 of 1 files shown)


st--:
Hi @christabella, you will need to run `make format` from the directory in which you have checked out GPflow before the rest of the tests can run.

Can you please clarify, are you saying your proposed changes in this PR make this notebook significantly *slower*?

christabella:
Hi @st-- thank you for the tip, I've done that!

Yes, because the original experiment in Fortuin and Rasch actually use the values of `(1000, 200, 50)` for `(num_datapoints, num_meta_tasks, num_test_tasks)` respectively, for 100 epochs (`num_iters`).

This implementation was previously wrongly using `(20, 5, 1000)`, hence it was much faster at 7s/epoch.

I tried lowering the number of tasks from the paper, and increasing N, to `(100, 10, 250)` and now it takes 45s/epoch and achieves comparable results to the reported performance (0.52 ± 0.03):
> The mean MSE over all 200 test tasks is 0.760 +/- 0.020

christabella:
With (50, 10, 500) each epoch is 28s

| |`num_meta_tasks` | `num_test_tasks`| `N` (sampled points/curve) | time (s/epoch) | MSE |
|--|--|--|--|--|--|
|Original experiment from paper| 1000|200|50|?|0.52 ± 0.03 (100 epochs)|
|Previous notebook settings | 20 | 5 | 1000 | 7s | 0.68 +/- 0.14 (5 epochs) |
|Notebook with actual settings| 1000|200|50| 420s | ? (too slow) |
|Settings #2|100 |10| 250| 45s|0.760 +/- 0.020|
|Settings #3|50 | 10 |500| 28s|0.716 +/- 0.015|

image

doc/source/notebooks/tailor/external-mean-function.pct.py
```diff
@@ -227,16 +268,17 @@ def mean_squared_error(y, y_pred):
     mean_squared_errors.append(mse)
     plt.title(f"Test Task {i + 1} | MSE = {mse}")
     plt.legend()
+    plt.show()
```
christabella: If we don't close the `plt.figure` explicitly, if we try to plot too many, we get this warning:
```<ipython-input-38-510b3796bcb3>:6: RuntimeWarning: More than 20 figures have been opened. Figures created through the pyplot interface (`matplotlib.pyplot.figure`) are retained until explicitly closed and may consume too much memory. (To control this warning, see the rcParam `figure.max_open_warning`).
```

christabella:
Re. commit fe6f58f: I also thought it would be nice to plot the target training set so that we can see that `GPR` is fitted onto that data (noisy observations of the ground truth function):

image

christabella:
Re. the latest commit, I also added the prediction variance.

image

doc/source/notebooks/tailor/external-mean-function.pct.py
```diff
@@ -39,14 +39,23 @@
...
 # %matplotlib inline
 
 # %% [markdown]
 # ## Generate the tasks
 # To generate the meta and test tasks, we sample from a Gaussian process with an Squared Exponential covariance function and a sinusoidal mean function. Each task is a realization of this process.
 
+
 # %%
-def generate_data(num_functions=10, N=1000):
+def generate_data(num_functions=10, N=50):
```
st--, line 51: I thought you suggested to go with N=500 ?
would it be worth simply having a cell towards the top of the notebook that sets the three :
num_meta, num_test, N ?
Maybe we can give them more explicit names? num_meta_tasks, num_test_tasks, num_data_per_task (please correct me if I misinterpreted any!)?

christabella: Thanks for taking a look! I've changed the name, there is a cell in the notebook that sets the three with those names now

doc/source/notebooks/tailor/external-mean-function.pct.py
```diff
@@ -39,14 +39,23 @@
...
 # %% [markdown]
 # ## Generate the tasks
 # To generate the meta and test tasks, we sample from a Gaussian process with an Squared Exponential covariance function and a sinusoidal mean function. Each task is a realization of this process.
 
+
 # %%
-def generate_data(num_functions=10, N=1000):
+def generate_data(num_functions=10, N=500):
+    """
+    For each function, sample the value at 50 equally spaced
```
st--, line 53:
```suggestion
    For each function, sample the value at `N` equally spaced
```

doc/source/notebooks/tailor/external-mean-function.pct.py
```diff
@@ -66,32 +75,66 @@ def generate_data(num_functions=10, N=1000):
...
 
-
 # %% [markdown]
 # We generate the meta and test tasks.
 
+
 # %%
-def generate_meta_and_test_tasks(num_datapoints, num_meta, num_test):
-    N = 1000
+def generate_meta_and_test_tasks(num_datapoints, num_meta=1000, num_test=200, N=50):
```
st--, line 83:
```suggestion
def generate_meta_and_test_tasks(num_datapoints, num_meta, num_test, N):
```
as you pass in all arguments explicitly anyways it's safer not to risk having some inconsistency between them.

doc/source/notebooks/tailor/external-mean-function.pct.py
```diff
@@ -66,32 +75,66 @@ def generate_data(num_functions=10, N=1000):
...
+        noise = sd * np.random.randn(num_datapoints, 1)
+        train_X = Xs[train_i]
+        train_Y = F[train_i, num_meta + i][:, None] + noise
+        # Form target test set, \tilde{D}_{test}, which is disjoint from the
+        # target traininig set as it is in the original implementation.
+        test_X = Xs[test_i]
+        test_Y = F[test_i, num_meta + i][:, None]
+        # Form target tasks' datasets \tilde{D} as a pair of pairs,
+        # (\tilde{X}, \tilde{Y}) and (\tilde{X}*, \tilde{Y}*).
+        test.append(((train_X, train_Y), (test_X, test_Y)))
```
st--, line 124: Could you briefly explain what you changed here? Did you just clean up *how* things are generated (but the returned values will be the same as before, if the random seed was the same), or does this change the logic of *what* is generated? (To make it easier to review your changes!)

st--:
Hi @christabella, just wanted to ping in case you hadn't seen my comments - I'm looking forward to merging your contribution into develop once we've resolved these!

christabella:
@st-- thank you for the ping and for the nice review, I missed it somehow!

doc/source/notebooks/tailor/external-mean-function.pct.py
christabella: Thanks for the question! 
1. Previously each meta-training episode was done on only `num_datapoints` points, but it should be all `N` points (so no need for any `meta_indices` anymore where `num_datapoints` points are picked randomly).
2. The target datasets (called `test` for meta-testing) are split into train and test (`train_X, train_Y, test_X, test_Y`) . Previously, the train was a subset of the test, but after clarifying it they are actually disjoint.

So the logic has changed a little bit, although they are probably not that important. Hope this clears things up, let me know if you see more areas for improvement!

PR #1388, opened 2020-03-30
Increased tolerance for sqrt in kernels K_r2.

markvdw:
I remember once I ran into numerical problems when the tolerance was 1e-40. So I changed it to 1e-36. This change was in GPflow 1, but wasn't ported to GPflow 2, hence why we're fixing it now.


st--:
Hi @markvdw, ideally we would have a test that breaks with 1e-40 and passes with 1e-36 ... could you please add a short description to the pull request why this change is necessary?

PR #1391, opened 2020-03-30
Improve doc generation

st--:
Adds documentation to how docs are actually built. Fixes the instructions for local documentation build. Adds module-level docstrings (albeit in a slightly weird, if logical, place).


doc/source/generate_module_rst.py
```diff
@@ -64,6 +64,7 @@
 SPHINX_INCLUDE_MODULE_STRING = """
 {name}
 {level}
+.. automodule:: {name}
```
st--: This is what was missing to pull module-level docstring in at all.

doc/source/generate_module_rst.py
```diff
@@ -134,7 +135,7 @@ def get_component_rst_string(module: ModuleType, component: Callable, level: int
     object_name = f"{module.__name__}.{component.__name__}"
 
     rst_documentation = ""
-    level_underline = RST_LEVEL_SYMBOLS[level] * 6
+    level_underline = RST_LEVEL_SYMBOLS[level] * len(object_name)
```
st--: This (and the related changes below) fixes a bunch of warnings about wrong length of underlines when running sphinx.

doc/README.md
```diff
@@ -1,34 +1,49 @@
...
-```
-make html
-```
-
-5. Check documentation locally by opening (in a browser) doc/build/html/index.html
+1. Change to this directory (e.g. `cd doc` if you are in the GPflow git repository's base directory)
+
+2. Install doc dependencies
+   ```
+   pip install sphinx sphinx_rtd_theme numpydoc nbsphinx sphinx_autodoc_typehints ipython
```
st--: I was tempted to change this to
```suggestion
   pip install -r ./requirements.txt
```
but I don't like that because it has pinned versions of everything, and I don't want it to install an old version of IPython for me just so I can build docs.

Any reason we can't just have >= in the doc/requirements.txt file ?

vdutor: I guess it depends on what readthedocs does with it. Maybe worth a try, but low priority my opinion.
You could also just advice to create a different python env.

doc/README.md
```diff
@@ -1,34 +1,49 @@
...
+6. Check documentation locally by opening (in a browser) build/html/index.html
+
+
+## Setup for automatic documentation generation
+
+Upon each merge to the `develop` branch, this repository's [CircleCI configuration](../.circleci/config.yml) runs the `trigger-docs-generation` step
+which triggers a CircleCI build on the [GPflow/docs repository](https://github.com/GPflow/docs).
+This clones the latest GPflow develop branch and compiles all notebooks from jupytext to .ipynb
+(setting the `DOCS` environment variable so that notebook optimisations are run to convergence)
+and runs the `generate_module_rst.py` script as above to generate the .rst files for API documentation.
```
vdutor: The python script `generate_module_rst.py` runs in circleci containersa (and *not* on readthedocs containers) because it requires `GPflow`, `TensorFlow` and `TensorFlow Probability` to be installed. ReadTheDocs containers, responsible for generating the docs using Sphinx, are not big enough to support these libraries.

doc/source/generate_module_rst.py
vdutor: So it wasn't the auto generation part that caused the missing docstrings, right?

doc/README.md
st--: yeah, exactly, that's why I haven't changed it yet

doc/README.md
st--: added an extra sentence to capture this explanation - thanks!

doc/source/generate_module_rst.py
st--: Yes, correct!

doc/README.md
vdutor: cool, so let's leave it as it for now.

doc/source/generate_module_rst.py
condnsdmatters: great spot gang

PR #1394, opened 2020-03-31
Add a section about custom user config

awav:
subj.

### DIFF
--- a/doc/source/notebooks/intro_to_gpflow2.pct.py
+++ b/doc/source/notebooks/intro_to_gpflow2.pct.py
@@ -491,7 +491,9 @@ global_str = "Global config\t"
 
 with gpflow.config.as_context(user_config):
     print(f"{user_str} gpflow.config.default_float = {gpflow.config.default_float()}")
-    print(f"{user_str} gpflow.config.positive_bijector = {gpflow.config.default_positive_bijector()}")
+    print(
+        f"{user_str} gpflow.config.positive_bijector = {gpflow.config.default_positive_bijector()}"
+    )
 
 print(f"{global_str} gpflow.config.default_float = {gpflow.config.default_float()}")
 print(f"{global_str} gpflow.config.positive_bijector = {gpflow.config.default_positive_bijector()}")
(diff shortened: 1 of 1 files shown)


doc/source/notebooks/intro_to_gpflow2.pct.py
```diff
@@ -477,3 +477,31 @@ def checkpointing_training_loop(
...
+# %% [markdown]
+# ## User config update
+#
+# In this notebook, we used a lot `gpflow.config` methods for setting and getting default attributes from global configuration. However, GPflow provides a way for local config modification without updating values in global. As you can see below, using `gpflow.config.as_context` replaces temporarily global config with your instance. At creation time, custom config instance uses standard values from the global config:
+
+# %%
+user_config = gpflow.config.Config(float=tf.float32, positive_bijector="exp")
+
+user_str = "User config\t"
+global_str = "Global config\t"
```
st--: Nit: I got a bit confused why we'd need a \tab in here - I might have moved that into the print() call itself...

doc/source/notebooks/intro_to_gpflow2.pct.py
```diff
@@ -477,3 +477,31 @@ def checkpointing_training_loop(
...
+
+with gpflow.config.as_context(user_config):
+    print(f"{user_str}gpflow.config.default_float = {gpflow.config.default_float()}")
+    print(f"{user_str}gpflow.config.positive_bijector = {gpflow.config.default_positive_bijector()}")
+
+print(f"{global_str}gpflow.config.default_float = {gpflow.config.default_float()}")
+print(f"{global_str}gpflow.config.positive_bijector = {gpflow.config.default_positive_bijector()}")
+
+# %%
+user_config = gpflow.config.Config(float=tf.float32, positive_bijector="exp")
```
st--: nit: this is the same as in the previous cell, is that intentional?

doc/source/notebooks/intro_to_gpflow2.pct.py
awav: I didn't want to pollute prints with tab symbol

doc/source/notebooks/intro_to_gpflow2.pct.py
awav: Yes, it shows that a user config has global scale impact.

PR #1395, opened 2020-03-31
Bump version to 2.0.0

awav:


VERSION
```diff
@@ -1 +1 @@
-2.0.0-rc1
\ No newline at end of file
+2.0.0
```
vdutor: You don't want a line break?
```suggestion
2.0.0

```

PR #1396, opened 2020-03-31
Release 2.0.0

awav:


.circleci/config.yml
```diff
@@ -33,6 +33,38 @@ jobs:
         environment:
           PYTEST_FILTER: notebooks
 
+  format-checker:
+    docker:
+      - image: awav/tensorflow:2.0-preview
```
vdutor: just noticed we are still using the old docker images?
```suggestion
      - image: awav/tensorflow:2.1
```

.circleci/config.yml
```diff
@@ -33,6 +33,38 @@ jobs:
...
+          command: |
+            pip install black==19.10b0
+      - run:
+          name: Run format checker
+          command: |
+            black --check -t py36 -l 100 gpflow tests doc setup.py
+
+  trigger-docs-generation:
+    docker:
+      - image: awav/tensorflow:2.0-preview
```
vdutor:
```suggestion
      - image: awav/tensorflow:2.1
```

.circleci/config.yml
awav: @vdutor, I cannot apply it here, could you create a PR to develop? Btw, the docker will pull It new tensorflow, it doesn't have pre-installed tensorflow.

PR #1399, opened 2020-03-31
split up monitor into submodules to allow for missing matplotlib import

st--:
This PR splits up the monitor.py module into submodules to extract out the dependency on matplotlib by the ImageToTensorBoard class. Other than the reorganisation, the only logical change is the `try: except ImportError:` in `__init__.py` to replace ImageToTensorBoard with a stub, if matplotlib fails to import.


joelberkeley-pio:
how about just moving the import into the function that uses it? or if that's not performant, storing it in the `__init__`?

joelberkeley-pio:
or if there are issues with importing a module multiple times (as there sometimes is with python), you could
```python3
try:
    from matplotlib.backends.backend_agg import FigureCanvasAgg
except:
    class FigureCanvasAgg:
        def __init__(self, *_, **__):
            raise DependencyError("can't find matplotlib")
```

joelberkeley-pio:
i suppose my feeling is that a user should be informed at import time that the class they're trying to import doesn't exist

another option is to
```
try:
    from matplotlib.backends.backend_agg import FigureCanvasAgg
    from matplotlib.figure import Axes, Figure
except ImportError:
    logging.warn("ImageToTensorBoard class not available due to missing optional dependency matplotlib")
else:
    class ImageToTensorBoard(ToTensorBoard):
        # actual class
```

vdutor:
I agree with @joelberkeley-pio latest suggestion, which I feel is more appropriate to deal with this small import matplotlib issue, compared to creating six new files. Let's try to keep things as simple as possible.

st--:
> i suppose my feeling is that a user should be informed at import time that the class they're trying to import doesn't exist

I would agree with this except for that we have structured GPflow so that you *don't* need to import anything explicitly - `import gpflow` already handles all the imports - and I value that because I don't want to have to explicitly import everything all the time - but it also means we can't see at import time whether a user actually cares about a specific class (such as ImageToTensorBoard) or not.

>  compared to creating six new files.

There is some value in splitting up large modules into chunks that are more individually manageable, and of course the other extreme of one file per class can also be rather cumbersome. I guess `monitor.py` is borderline - it's currently 330 lines...

joelberkeley-pio:
I don't understand how importing everything automatically could be compatible with optional dependencies

st--:
> I don't understand how importing everything automatically could be compatible with optional dependencies

Well that's what this PR in its current form achieves? Everything is imported upon `import gpflow`. ImageToTensorBoard has an optional dependency so it only actually works if you've got that dependency installed.

joelberkeley-pio:
You've not imported `ImageToTensorBoard` as it's defined in the docs, you've imported a different class with the same name

st--:
Yes. That was the point.

st--:
How do you suggest we move forward with this ?

joelberkeley-pio:
I suppose as a user I'd like to fail fast - as soon as I try to import the class without matplotlib installed, I'd like it to error (this is assuming the class doesn't do anything meaningful without matplotlib).

As a gpflow maintainer, I'd like the smallest maintainable change possible

st--:
> as soon as I try to import the class without matplotlib installed, I'd like it to error

How would this work together with "you do `import gpflow` and can access all parts without requiring explicit submodule imports" (which is something I'd like to keep for gpflow to be easy to use)?

joelberkeley-pio:
> > as soon as I try to import the class without matplotlib installed, I'd like it to error
> 
> How would this work together with "you do `import gpflow` and can access all parts without requiring explicit submodule imports" (which is something I'd like to keep for gpflow to be easy to use)?

I propose we only do that for functionality that doesn't require extra dependencies. But if you did want to do it for everything, you could perhaps [customise `__getattr__` for the module](https://stackoverflow.com/a/48916205/5986907).

On this note, I like the rust idea of defining a `prelude` package in libraries where it's standard to (equivalently) do `from gpflow.prelude import *`. I know that's different to what we're doing here, but it might offer some ideas

st--:
> But if you did want to do it for everything, you could perhaps customise __getattr__ for the module.

Looks nice, but only 3.7+ -- we're currently supporting Python 3.6+.

vdutor:
I think we can solve this problem slightly more elegant / less code / easier-to-maintain using something similar to this: (I imagine @joelberkeley-pio was suggesting this as well):  
```python
class ImageToTensorBoard(ToTensorBoard):
    _import_matplotlib_flag = False:

    def __init__(self):
        if not _import_matplotlib_flag:
            self._import_matplotlib()

    def _import_matplotlib(self):
            """Ensures matplotlib dependencies are only imported if an instance is created."""
            from matplotlib.backends.backend_agg import FigureCanvasAgg
            from matplotlib.figure import Axes, Figure
            ImageToTensorBoard._import_matplotlib_flag = True
```

Regarding splitting up the monitoring into several files is something I think that should have been raised during the PR review process, if people were unhappy with a 330 line file. I see value in keeping modules grouped together for easy CTRL-F'ing. If we go for a minimal solution, I see no need to continue with this split. It is also not super cool to see your code file being deleted and copy-pasted into several other modules (losing all git history) a week after merge - especially after a thorough review process.

joelberkeley-pio:
> I think we can solve this problem slightly more elegant / less code / easier-to-maintain using something similar to this: (I imagine @joelberkeley-pio was suggesting this as well):

sort of, but more like
```
class ImageToTensorBoard(ToTensorBoard):
    _imports = None

    def run(self, ...):
        if not self._imports:
            from matplotlib.backends.backend_agg import FigureCanvasAgg
            from matplotlib.figure import Axes, Figure
            ImageToTensorBoard._imports = FigureCanvasAgg, Axes, Figure
        else:
            FigureCanvasAgg, Axes, Figure = ImageToTensorBoard._imports
```
but I this feels like a hack, and I wouldn't be surprised if there's a bug hiding in there somewhere

joelberkeley-pio:
I think you can do both with this:
file containing class (and other classes)
```python3
try:
    import matplotlib
except ImportError:
    print("no matplotlib: Foo not available")
else:
    class Foo:
        pass
```
`__init__.py`
```python3
try:
    import matplotlib
except ImportError:
    pass
else:
    from .tensorboard_mod import Foo
```

st--:
I'd like both of the following:
* When `matplotlib` *is* installed: `import gpflow` and immediately being able to use `gpflow.monitor.ImageToTensorboard` or `from gpflow.monitor import ImageToTensorboard`
* When `matplotlib` is *not* installed: `import gpflow` not to complain about matplotlib missing (because it's not needed for anything beyond ImageToTensorboard, so always printing a warning would be annoying/confusing), and trying to use `gpflow.monitor.ImageToTensorboard` to give a helpful error message that I need to install matplotlib to be able to use this class.

I can see two ways of doing so:
a) replacing with a stub class when matplotlib is not available, as originally proposed in this PR,
b) moving the matplotlib imports *inside* the ImageToTensorboard class as suggested above, though I think it could be even simpler:
```python
class ImageToTensorboard:
    def __init__(self, ...):
        try:
            import matplotlib
        except ImportError:
            raise RuntimeError("You need to install the matplotlib package to be able to use ImageToTensorboard")
        ...

    def run(self, ...):
        from matplotlib import ...
        ...
```
How about that? Re-importing already-imported modules has zero cost, and it seems cleaner than keeping track of extra state on the class.

Issue #1407, opened 2020-04-02
Small noise_variance value leads to weird InvalidArgumentError exception


jjramsey:
Here's what should be a minimal, reproducible example:

```python
import numpy as np
import gpflow

x = np.linspace(0.0, 0.5)
y = np.sin(x) + 1e-2*np.random.standard_normal(x.shape)

# var_val = 1e-2
var_val = 1e-7

model = gpflow.models.GPR(data = (x,y),
                          kernel = gpflow.kernels.RBF(),
                          noise_variance = var_val)

gpflow.utilities.print_summary(model)
```

If the `noise_variance` argument in the above code is set to 1e-2, things work as expected, and I get the following result:

```
╒═════════════════════════╤═══════════╤══════════════════╤═════════╤═════════════╤═════════╤═════════╤═════════╕
│ name                    │ class     │ transform        │ prior   │ trainable   │ shape   │ dtype   │   value │
╞═════════════════════════╪═══════════╪══════════════════╪═════════╪═════════════╪═════════╪═════════╪═════════╡
│ GPR.kernel.variance     │ Parameter │ Softplus         │         │ True        │ ()      │ float64 │    1    │
├─────────────────────────┼───────────┼──────────────────┼─────────┼─────────────┼─────────┼─────────┼─────────┤
│ GPR.kernel.lengthscales │ Parameter │ Softplus         │         │ True        │ ()      │ float64 │    1    │
├─────────────────────────┼───────────┼──────────────────┼─────────┼─────────────┼─────────┼─────────┼─────────┤
│ GPR.likelihood.variance │ Parameter │ Softplus + Shift │         │ True        │ ()      │ float64 │    0.01 │
╘═════════════════════════╧═══════════╧══════════════════╧═════════╧═════════════╧═════════╧═════════╧═════════╛
```

However, if it is set to 1e-7, then I get the following traceback:

```
Traceback (most recent call last):
  File "./test.py", line 14, in <module>
    noise_variance = var_val)
  File "/Users/jjramsey/venvs/gpflow2/lib/python3.7/site-packages/gpflow/models/gpr.py", line 49, in __init__
    likelihood = gpflow.likelihoods.Gaussian(noise_variance)
  File "/Users/jjramsey/venvs/gpflow2/lib/python3.7/site-packages/gpflow/likelihoods/likelihoods.py", line 404, in __init__
    self.variance = Parameter(variance, transform=positive(lower=variance_lower_bound))
[...]

PR #1408, opened 2020-04-02
Improve user experience on invalid assignments to constrained Parameters

st--:
Addresses #1407.

* attempt to improve the error message in the gpflow.Parameter check on assigning a new value that is incompatible with the parameter's transform (e.g. a non-positive value to a parameter with a positive() transform)

* Gaussian likelihood: add explicit `__init__`-time check that variance > variance_lower_bound, add `__init__` docstring, move "default variance lower bound" magic number into class-level constant

* change gpflow.config's positive_minimum to always be a float (initialized to 0.0 by default)


st--:
TODO: should add some explicit tests for the new constructor check in `Gaussian`

gpflow/base.py
```diff
@@ -159,8 +159,9 @@ def validate_unconstrained_value(self, value: tf.Tensor, dtype: DType) -> tf.Ten
         value = _cast_to_dtype(value, dtype)
         unconstrained_value = _to_unconstrained(value, self.transform)
         message = (
-            "gpflow.Parameter: unconstrained value of passed value "
-            "has NaN or Inf and cannot be assigned."
+            "gpflow.Parameter: the value to be assigned is incompatible with this parameter's "
+            "transform (the corresponding unconstrained value has NaN or Inf) and hence cannot be "
+            "assigned."
```
st--, line 164: Could be extra explicit and add `(e.g. positively constrained parameters must be strictly greater than zero)`? :thinking:

gpflow/config/__config__.py
```diff
@@ -76,7 +76,7 @@ class _Values(enum.Enum):
     INT = np.int32
     FLOAT = np.float64
     POSITIVE_BIJECTOR = "softplus"
-    POSITIVE_MINIMUM = None
+    POSITIVE_MINIMUM = 0.0
```
jameshensman, line 79: was this a bug? it ssems like a bit thing to change from None to 0.

st--: I don't know what the motivation behind initializing it as None was originally - @awav do you remember? Setting it to `None` had the same effect as 0.0 - not adding a shift to the exp/softplus transform. But it meant having to do extra queries on checking whether default_positive_minimum() was None or a float. And the setter only allowed you to set it to a float, not back to None. So it seemed more sensible to tidy it up altogether.

awav: None is easier to check and it is safer "label". Comparing with `0.0` is always a gamble in the world of multiple float types. I would stick to the `None`. But, please be careful with `0.0` if you decide to leave it.

@jameshensman, @st--

gpflow/base.py
awav: `positive` is too specific for this message.

gpflow/likelihoods/likelihoods.py
```diff
@@ -391,16 +391,31 @@ def integrand(*X):
...
-    optimization process. A lower bound of 1e-6 is therefore imposed on the likelihood variance
-    by default.
+    optimization process. A lower bound of 1e-6 is therefore imposed on the
+    likelihood variance by default.
     """
 
-    def __init__(self, variance=1.0, variance_lower_bound=1e-6, **kwargs):
+    DEFAULT_VARIANCE_LOWER_BOUND = 1e-6
+
+    def __init__(self, variance=1.0, variance_lower_bound=DEFAULT_VARIANCE_LOWER_BOUND, **kwargs):
```
awav, line 405: This is error prone approach. Changing `DEFAULT_VARIANCE_LOWER_BOUND` in the class doesn't change the default value of `variance_lower_bound` argument. Therefore, I would advise you to move this assignment to the init with `if else` statement, whereas the default value would be `None`.

st--: I added this following a (very valuable) suggestion in #1407, to allows downstream code to check what the default value *is*. It's not meant to be changed, which is why it's all-caps (the python convention for constants that you should not go around and change).

gpflow/base.py
st--: it's the most common use-case for transforms, and examples can be helpful no?

gpflow/likelihoods/likelihoods.py
st--: I think it might be more confusing for `None` to mean "default value" rather than "no lower bound" :thinking:

Issue #1405, opened 2020-04-01
Proposal: reorganize likelihoods directory/submodule structure


st--:
Currently, there are two files in the likelihoods/ directory, `likelihoods.py` for almost all of them (almost 1000 lines) and `robustmax.py` (80 lines, only the RobustMax helper class for the MultiClass likelihood). I think re-grouping likelihood classes and splitting up the big `likelihoods.py` file would make it more maintainable and easier to add new likelihoods (e.g. we're already thinking about adding built-in support for heteroskedastic likelihoods).

This is the proposal I've come up with, I would appreciate further input and discussion before actually going and shuffling code around!

### Base classes: `base.py`
- `Likelihood`
- `ScalarLikelihood`
- `MonteCarloLikelihood`
- Q: should `SwitchedLikelihood` also go in here, as it only wraps other likelihoods?

#### Option 1: Subclasses of `ScalarLikelihood`: these could all go in one file `scalar_likelihoods.py`:
- `Gaussian`
- `Exponential`
- `StudentT`
- `Gamma`
- `Beta`
- `Poisson`
- Q: `Bernoulli` - here or into `classification.py` (see below)? It is classification but in terms of code logic/structure it is much more similar to the other ScalarLikelihoods (it is only parametrized by a single GP per observation dim and can easily broadcast over multiple outputs)

#### Option 2: Group by "type of likelihood", e.g.
`regression.py`:
- `Gaussian`
- `Exponential`
- `StudentT`
- `Gamma`
- `Beta`

Q: should `Poisson` and `Ordinal` go in here or into a different module `counts.py` / `discrete.py` ?

### Multi-class classification
`classification.py`? or be explicit and call it `multiclass_classification.py` (and definitely keep `Bernoulli` in another file)?
- `MultiClass` as well as the `RobustMax` helper
- `Softmax`
- Q: should `Bernoulli` go in here? this would make sense in terms of "classification", but in terms of code logic/structure it is much more similar to the other ScalarLikelihood subclasses (see above, Option 1)

### Discussion:
What other ways of structuring / grouping the different likelihoods can you think of?
What would be the advantages / disadvantages of the different options?
What (if anything) do you personally prefer (and how much do you care)?

jameshensman:
Yes! This has needed sorting for a while. Happy to help.

I definite agree that we should have the base-classes separately. I’m not sure how to organise the rest but I kind of like the idea of regression/classification/count. I think it might help new users understand what the likelihood’s are for.

[...]

PR #1334, opened 2020-03-17
improve shape robustness in likelihoods

jameshensman:
#### Preamble
GPflow likelihoods could do with a stronger contract re what input shapes are expected and what shapes are returned. In particular, we should obey something akin to tfp’s event-shape/batch-shape/sample-shape. Our proposal will mean that very little will change for most users, except that some shapes will be asserted. Advanced users will benefit from more shape checks and better defined return shapes for methods attached to likelihoods.

#### Main proposal:
Likelihood needs to define:
 - Self.num_y_dims: what is the last dimension of Y supposed to be?
 - Self.num_latent_dims: what is the last dimension of F, F_mean and F_var expected to be?
We’ll check that the dimensions of tensors passed in match up.
Return shapes for all methods will be the broadcast-shape of the tensors passed in, with the last dimension removed.
Example: lik.variational_expectations(F_mu, F_var, Y) might take tensors of dimensions [..., 2], [..., 2], [..., 2] and return a tensor of shape [...]

#### Standard likelihoods
Most likelihoods in GPflow are univariate, and treat columns of Y and F as independent variables. For these we propose that num_y_dims and num_latent_dims are None, and that we shape check F and Y on-the-fly for matching last-column dimensions. Note that the return shape contract will change as per that above: this should be easily handled in vgp/svgp/sgpmc.

#### Fancy likelihoods
Likelihoods that depart from the univariate standard include:
SwitchedLikelihood [we’ll check that num_latent_dims = num_y_dims -1]
MultiClass/Softmax [num_y_dims = 1, num_latent_dims = (e.g. 10 for mnist)]
HeteroskedasticGaussian e.g. see GPflow notebook [num_y_dims = 1, num_latent_dims = 2]

PR #1416, opened 2020-04-07
Improve structure of likelihoods subdirectory

st--:
Following on from #1405.

* New structure underneath `gpflow/likelihoods/`:
  * `base.py`: all base classes (Likelihood, MonteCarloLikelihood, ScalarLikelihood) and SwitchedLikelihood
  * `multiclass.py`: multi-class classification (Softmax, MultiClass + RobustMax)
  * `scalar_continuous.py`: continuous-Y subclasses of ScalarLikelihood (Gaussian, StudentT, Exponential, Beta, Gamma)
  * `scalar_discrete.py`: discrete-Y subclasses of ScalarLikelihood (Bernoulli, Poisson, Ordinal)
  * `utils.py`: the `inv_probit` link function used by Bernoulli and Beta likelihoods
  * `misc.py`: GaussianMC - used for demonstration/tests only.
* Tests for multi-class classification likelihoods moved out into their own test module (including stubs for the missing MultiClass quadrature tests of #1091)
* Re-activates the quadrature tests for ScalarLikelihood subclasses with analytic variational_expectations/predict_log_density/predict_mean_and_var that inadvertently got disabled by #1334 
* Fixes a bug in Bernoulli._predict_log_density that was uncovered by these tests


tests/gpflow/optimizers/test_natural_gradient.py
```diff
@@ -59,7 +63,7 @@ def sgpr_and_svgp(data, inducing_variable, kernel, likelihood):
     return sgpr, svgp
 
 
-def assert_different(value1, value2, rtol=0.1):
+def assert_different(value1, value2, rtol=0.07):
```
jameshensman, line 66: what causes the change in rtol?

tests/gpflow/optimizers/test_natural_gradient.py
```diff
@@ -12,19 +12,23 @@
 class Setup:
```
jameshensman, line 12: This is a bit out of place in the refactor / reorganization, but in the spirit of "tidy as you go", I think it stays.

jameshensman:
Yikes, what was the bug in predict_log_density? That could have affected a lot of things...

st--:
> Yikes, what was the bug in predict_log_density? That could have affected a lot of things...

#1334 changed the behaviour of predict_log_density to sum over outputs (instead of one number per output per data row), and the Bernoulli likelihood has an analytic _predict_log_density, but hadn't been summing over it...

tests/gpflow/optimizers/test_natural_gradient.py
st--: It's a stochastic test; reshuffling the tests changed the random number generator, and I gave up on trying to find a seed that makes all of them pass...

tests/gpflow/optimizers/test_natural_gradient.py
st--: yeah, part of this was not wanting to have to think too much about how to consistently get tensorflow random numbers - I know there's tf.random.set_seed() but I trusted a RandomState object more...

PR #1419, opened 2020-04-08
make set_trainable return its argument

st--:
This would allow it to clean up some model setup code, e.g.
```python
model = VGP((X, Y), set_trainable(kernel, False), likelihood)
```

### DIFF
--- a/gpflow/utilities/utilities.py
+++ b/gpflow/utilities/utilities.py
@@ -46,12 +46,13 @@ def to_default_float(x):
     return cast(x, dtype=default_float())
 
 
-def set_trainable(model: tf.Module, flag: bool):
+def set_trainable(module: tf.Module, flag: bool):
     """
     Set trainable flag for all `tf.Variable`s and `gpflow.Parameter`s in a module.
     """
-    for variable in model.variables:
+    for variable in module.variables:
         variable._trainable = flag
+    return module
 
 
 def multiple_assign(module: tf.Module, parameters: Dict[str, tf.Tensor]):


gpflow/utilities/utilities.py
```diff
@@ -46,12 +46,13 @@ def to_default_float(x):
...
 
-def set_trainable(model: tf.Module, flag: bool):
+def set_trainable(module: tf.Module, flag: bool):
     """
     Set trainable flag for all `tf.Variable`s and `gpflow.Parameter`s in a module.
     """
-    for variable in model.variables:
+    for variable in module.variables:
         variable._trainable = flag
+    return module
```
condnsdmatters, line 55: It's a nice idea, but the problem with having functions that have side effects that return the object being effected is that the api become unclear. You have to go to check the source to work out what is happening, rather than just the function signature. Returning None is much clearer in this case!

```
x = UnsortedThing()
y = SortMyThing(x)
# is x sorted or unsorted?
```

gpflow/utilities/utilities.py
```diff
@@ -46,12 +46,13 @@ def to_default_float(x):
     return cast(x, dtype=default_float())
 
 
-def set_trainable(model: tf.Module, flag: bool):
+def set_trainable(module: tf.Module, flag: bool):
```
condnsdmatters, line 49:
```suggestion
def set_trainable(module: tf.Module, flag: bool)->None:
```

condnsdmatters:
Can we close this - I hadnt seen @awav 's review but I agree

Issue #1368, opened 2020-03-25
Full type hint coverage


joelberkeley-secondmind:
**What?**

Full type hint coverage. That is, types should be specified everywhere they can't be inferred by a type checker.

@joelberkeley-pio would like mypy to be part of the CI run so that we have some visibility of degradation. We need to discuss whether it makes sense to require mypy passes for merge.

**Why?**

As a user (or GPflow maintainer), I want full type hint coverage so that I know what types I'm expected to provide and which I should expect to receive back.

As a production user, I want full type hint coverage so that I can gain confidence my code is integrated correctly with the library.

As a GPflow maintainer, I want full type hint coverage so that I can be confident my changes are sound.

joelberkeley-pio:
**How?**

I propose we apply this a sub-package at a time.

joelberkeley-pio:
Before we do this, I recommend we reconsider the current approach of aliasing types. I find type aliases can be problematic for the following reasons:

  * they hide the actual meaning behind library-local names and therefore
    * are potentially less clear or accurate
    * require an indirection to find out what type I'm actually looking at
  * they give the user the unwarranted impression of type safety for different aliases of the same type. For example a user seeing `InputData` and `OutputData` may interpret these as different types, and thus feel that the type system (in the form of IDE highlighting/mypy) will warn them when there's a mistake. However, they are both aliases for `tf.Tensor` so they are actually the same type

There are cases where aliases make sense, for example when shortening lengthy types which are used repeatedly in a module

st--:
I'd be happy to go with a submodule-at-a-time approach and see how comfortable we all end up being with it. How about you & others who are keen to add types make a start with a couple PRs so we can see what that would look like?

joelberkeley-pio:
Things learnt from adding types to gpflow/base.py

  - it's looking promising that we can leave nothing without a type or suitable comment/pragma
  - use `# type unknown` where a type is unknown
  - use `if TYPE_CHECKING:` to avoid importing optional dependencies at runtime
  - use `# type: ignore` where the correct solution would require too much boilerplate

joelberkeley-pio:
I'd like to clarify that I think it's really important that we don't consider casually choose that omitting a type is a good final state, simply because we then don't know what's been looked at and what hasn't. Alternatives include comments or an alias of `Any`

st--:
Before we spend more effort on this, can we try adding mypy to the build and ask it to only check the submodules we've typified so far ?

PR #1420, opened 2020-04-08
add mypy to test requirements; type ci utils module

joelberkeley-pio:
starts #1368

### DIFF
--- a/gpflow/ci_utils.py
+++ b/gpflow/ci_utils.py
@@ -15,9 +15,10 @@
 # pylint: skip-file
 
 import os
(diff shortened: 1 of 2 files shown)


st--:
I am happy with the added types to ci_utils - I'm not sure though about making mypy a part of tests_requirements.txt while it's not actually required anywhere?

joelberkeley-pio:
> I am happy with the added types to ci_utils - I'm not sure though about making mypy a part of tests_requirements.txt while it's not actually required anywhere?

good point. i suppose I should add it to the build at some point, even just for the sake of logging. until then, i'll leave it out of test_requirements.txt

PR #1421, opened 2020-04-08
add type hints for probability distributions

joelberkeley-pio:
does #1368 for probability distributions

required a fix for `TensorLike` problem for multiple dispatch

### DIFF
--- a/gpflow/base.py
+++ b/gpflow/base.py
@@ -311,7 +311,7 @@ Type alias for tensor-like types that are supported by most TensorFlow, NumPy an
 NOTE: Does not work with multipledispatch dispatchers. See `TensorLikeTypes`.
 """
 
-TensorLikeTypes = (tf.Tensor, tf.Variable, np.ndarray, Parameter)
+TensorLikeTypes = (np.ndarray, tf.Tensor, tf.Variable, Parameter)
 """ Collection of tensor-like types for use with multipledispatch dispatchers.  """
 
 
--- a/gpflow/base.py
+++ b/gpflow/base.py
@@ -11,9 +11,8 @@ from .config import default_float
 
 DType = Union[np.dtype, tf.DType]
 VariableData = Union[List, Tuple, np.ndarray, int, float]
-TensorLike = (
-    object  # Union[tf.Tensor, tf.Variable, np.ndarray], but doesn't work with multipledispatch
-)
+TensorLike = Union[np.ndarray, tf.Tensor, tf.Variable]
+TensorLikeTypes = (tf.Tensor, tf.Variable, np.ndarray)
 Transform = Union[tfp.bijectors.Bijector]
 Prior = Union[tfp.distributions.Distribution]
 
--- a/gpflow/base.py
+++ b/gpflow/base.py
@@ -306,7 +306,7 @@ Parameter._OverloadAllOperators()
 tf.register_tensor_conversion_function(Parameter, lambda x, *args, **kwds: x.read_value())
 
 
-TensorLike = Union[np.ndarray, tf.Tensor, tf.Variable, Parameter]
+TensorType = Union[np.ndarray, tf.Tensor, tf.Variable, Parameter]
 """
 Type alias for tensor-like types that are supported by most TensorFlow, NumPy and GPflow operations.
 
@@ -317,9 +317,9 @@ NOTE: Union types like this do not work with the `register` method of multipledi
 
 # We've left this as object until we've tested the performance consequences of using the full set
 # (np.ndarray, tf.Tensor, tf.Variable, Parameter), see https://github.com/GPflow/GPflow/issues/1434
-TensorLikeTypes: Final[Tuple[type, ...]] = (object,)
+TensorLike: Final[Tuple[type, ...]] = (object,)
 """
-:var TensorLikeTypes: Collection of tensor-like types for registering implementations with
+:var TensorLike: Collection of tensor-like types for registering implementations with
     `multipledispatch` dispatchers.
 """
 
--- a/gpflow/base.py
+++ b/gpflow/base.py
@@ -308,7 +308,8 @@ TensorLike = Union[np.ndarray, tf.Tensor, tf.Variable, Parameter]
 """
 Type alias for tensor-like types that are supported by most TensorFlow, NumPy and GPflow operations.
 
-NOTE: Does not work with multipledispatch dispatchers. See `TensorLikeTypes`.
+NOTE: Union types like this do not work with the `register` method of multipledispatch's
+`Dispatcher` class. Instead use `TensorLikeTypes` for dispatching on tensor-like types.
 """
 
 # We've left this as object until we've tested the performance consequences of using the full set
--- a/gpflow/base.py
+++ b/gpflow/base.py
@@ -11,7 +11,6 @@ from .config import default_float
 
 DType = Union[np.dtype, tf.DType]
 VariableData = Union[List, Tuple, np.ndarray, int, float]
-TensorLike = Union[np.ndarray, tf.Tensor, tf.Variable]
 Transform = Union[tfp.bijectors.Bijector]
 Prior = Union[tfp.distributions.Distribution]
 
@@ -305,6 +304,9 @@ class Parameter(tf.Module):
 Parameter._OverloadAllOperators()
 tf.register_tensor_conversion_function(Parameter, lambda x, *args, **kwds: x.read_value())
 
+TensorLike = Union[np.ndarray, tf.Tensor, tf.Variable, Parameter]
+TensorLikeTypes = (tf.Tensor, tf.Variable, np.ndarray, Parameter)
+
 
 def _cast_to_dtype(value: VariableData, dtype: Optional[DType] = None) -> tf.Tensor:
     if dtype is None:
@@ -328,6 +330,3 @@ def _to_unconstrained(value: VariableData, transform: Transform) -> tf.Tensor:
     if transform is not None:
         return transform.inverse(value)
     return value
-
-
-TensorLikeTypes = (tf.Tensor, tf.Variable, np.ndarray, Parameter)
--- a/gpflow/base.py
+++ b/gpflow/base.py
@@ -305,7 +305,14 @@ Parameter._OverloadAllOperators()
 tf.register_tensor_conversion_function(Parameter, lambda x, *args, **kwds: x.read_value())
 
 TensorLike = Union[np.ndarray, tf.Tensor, tf.Variable, Parameter]
+"""
+Type alias for tensor-like types that are supported by most TensorFlow, NumPy and GPflow operations.
+
+NOTE: Does not work with multipledispatch dispatchers. See `TensorLikeTypes`.
+"""
+
 TensorLikeTypes = (tf.Tensor, tf.Variable, np.ndarray, Parameter)
+""" Collection of tensor-like types for use with multipledispatch dispatchers.  """
 
 
 def _cast_to_dtype(value: VariableData, dtype: Optional[DType] = None) -> tf.Tensor:
--- a/gpflow/base.py
+++ b/gpflow/base.py
@@ -328,3 +328,6 @@ def _to_unconstrained(value: VariableData, transform: Transform) -> tf.Tensor:
     if transform is not None:
         return transform.inverse(value)
     return value
+
+
+TensorLikeTypes = (tf.Tensor, tf.Variable, np.ndarray, Parameter)
--- a/gpflow/base.py
+++ b/gpflow/base.py
@@ -101,9 +101,7 @@ class Parameter(tf.Module):
 
             if self.transform is not None:
                 # need to include log|Jacobian| to account for coordinate transform
-                log_det_jacobian = self.transform.inverse_log_det_jacobian(
-                    y, y.shape.ndims
-                )
+                log_det_jacobian = self.transform.inverse_log_det_jacobian(y, y.shape.ndims)
                 log_p += tf.reduce_sum(log_det_jacobian)
 
             return log_p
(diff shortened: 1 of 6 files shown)


st--:
typing.Union[], unfortunately, does not play together with the multipledispatch. The multipledispatch package does take tuples; however, if this is used for several arguments this results in generating (number of explicit types in union-type) ^ (number of times this is used) elements. This may slow down dispatch. If we do decide to go for this I think we should include some dispatch timing tests to ensure this does not affect performance significantly.

joelberkeley-pio:
> typing.Union[], unfortunately, does not play together with the multipledispatch, so as it stands this PR is not ready for merging. The multipledispatch package does take tuples; however, if this is used for several arguments this results in generating (number of explicit types in union-type) ^ (number of times this is used) elements. This may slow down dispatch. If we do decide to go for this I think we should include some dispatch timing tests to ensure this does not affect performance significantly.

true. multipledispatch doesn't work with `typing.Union`, but that's not why the build is failing. Line 8 in kufs.py
```
@Kuf.register(InducingPoints, Kernel, TensorLike)
```
worked before for `Parameter` because `TensorLike` was defined as `object`. Had it had been defined using the more specific `(tf.Tensor, tf.Variable, np.ndarray)` as suggested by the comments we'd see the above. In addition to your question about performance, there's also then the question: should `TensorLikeTypes` include `Parameter`, and if not what do we do with `Kuf`?

Regarding performance, would we expect multiple dispatch to be used in an optimization loop?

st--:
> should `TensorLikeTypes` include `Parameter`?

Yes.

> Regarding performance, would we expect multiple dispatch to be used in an optimization loop?

Multiple dispatch is crucial for conditionals() & co which is at the core of every GPflow model... so yes. It may be fine, but I'd want to see a speed comparison before approving these changes.

joelberkeley-pio:
How about I just do
```
TensorLikeTypes = (object,) # should be (tf.Tensor, tf.Variable, np.ndarray, Parameter) but waiting on performance test
```
for now? I'd rather not get bogged down in more details than is needed for now. We'd retain the same level of functionality as before, just with a few extra types

gpflow/base.py
```diff
@@ -307,6 +304,9 @@ def _run_op(a, *args, **kwargs):
 Parameter._OverloadAllOperators()
 tf.register_tensor_conversion_function(Parameter, lambda x, *args, **kwds: x.read_value())
 
+TensorLike = Union[np.ndarray, tf.Tensor, tf.Variable, Parameter]
+TensorLikeTypes = (tf.Tensor, tf.Variable, np.ndarray, Parameter)
```
st--, line 308: Would it not make more sense to have the types together at the top of the file (as it was before)?
Alternatively, it might be helpful to collate them in a `gpflow.types` module (but perhaps do that in a separate PR?)... :thinking:

st--: It might also be helpful to add a comment explaining why we've got two things here that are almost exactly the same but not quite

st--:
> How about I just do `TensorLikeTypes = (object,) # should be (tf.Tensor, tf.Variable, np.ndarray, Parameter) but waiting on performance test` for now?

I guess that could work. It's a bit unfortunate that it doesn't work as a type signature...!

gpflow/base.py
joelberkeley-pio: at the top of the file would be nice, but they need to be after the `Parameter` class to work

gpflow.types module? perhaps

comment? yes. will do

joelberkeley-pio:
> > How about I just do `TensorLikeTypes = (object,) # should be (tf.Tensor, tf.Variable, np.ndarray, Parameter) but waiting on performance test` for now?
> 
> I guess that could work. It's a bit unfortunate that it doesn't work as a type signature...!

This is the risk that one has to accept by taking using third party utilities over the stdlib

gpflow/base.py
```diff
@@ -307,6 +304,16 @@ def _run_op(a, *args, **kwargs):
 Parameter._OverloadAllOperators()
 tf.register_tensor_conversion_function(Parameter, lambda x, *args, **kwds: x.read_value())
 
+TensorLike = Union[np.ndarray, tf.Tensor, tf.Variable, Parameter]
+"""
+Type alias for tensor-like types that are supported by most TensorFlow, NumPy and GPflow operations.
+
+NOTE: Does not work with multipledispatch dispatchers. See `TensorLikeTypes`.
+"""
```
st--, line 312:
```suggestion
# Type alias for tensor-like types that are supported by most TensorFlow, NumPy and GPflow operations.
#
# NOTE: The multipledispatch package uses tuples, not typing.Union, for dispatching on multiple types.
# For calling a Dispatcher's register() method (e.g. the `@conditional.register` decorator), use `TensorLikeTypes` instead
```

gpflow/base.py
```diff
@@ -307,6 +304,16 @@ def _run_op(a, *args, **kwargs):
...
+
+TensorLikeTypes = (np.ndarray, tf.Tensor, tf.Variable, Parameter)
+""" Collection of tensor-like types for use with multipledispatch dispatchers.  """
```
st--, line 315:
```suggestion
# Collection of tensor-like types for registering implementations with multipledispatch dispatchers.
```

gpflow/base.py
joelberkeley-pio: i chose docstrings so that it would appear in the docs. users may want to use these types in their own code

st--: Will we have to add `autodata` or some such to the API documentation for this to actually get pulled in?
I just looked it up as I didn't know it handled module-constant docstrings like this; Sphinx also supports `#:` instead which I think might be more readable when just reading the source code?

gpflow/base.py
```diff
@@ -307,6 +304,18 @@ def _run_op(a, *args, **kwargs):
...
+"""
+Type alias for tensor-like types that are supported by most TensorFlow, NumPy and GPflow operations.
+
+NOTE: Does not work with multipledispatch dispatchers. See `TensorLikeTypes`.
+"""
+
+# We've left this as object until we've tested the performance consequences of using the full set
+# see github issue #1434
+TensorLikeTypes = (object,)  # (np.ndarray, tf.Tensor, tf.Variable, Parameter)
+""" Collection of tensor-like types for use with multipledispatch dispatchers.  """
```
st--, line 317:
```suggestion
#: Collection of tensor-like types for registering implementations with multipledispatch dispatchers.
TensorLikeTypes = (object,)  # (np.ndarray, tf.Tensor, tf.Variable, Parameter)
# We've left this as object until we've tested the performance consequences of using the full set
# see https://github.com/GPflow/GPflow/issues/1434
```

gpflow/base.py
```diff
@@ -307,6 +304,18 @@ def _run_op(a, *args, **kwargs):
 Parameter._OverloadAllOperators()
 tf.register_tensor_conversion_function(Parameter, lambda x, *args, **kwds: x.read_value())
 
+TensorLike = Union[np.ndarray, tf.Tensor, tf.Variable, Parameter]
+"""
+Type alias for tensor-like types that are supported by most TensorFlow, NumPy and GPflow operations.
+
+NOTE: Does not work with multipledispatch dispatchers. See `TensorLikeTypes`.
+"""
```
st--, line 312:
```suggestion
#: Type alias for tensor-like types that are supported by most TensorFlow, NumPy and GPflow operations.
#:
#: NOTE: The multipledispatch package uses tuples, not typing.Union, for dispatching on multiple types.
#: For calling a Dispatcher's register() method (e.g. the `@conditional.register` decorator), use `TensorLikeTypes` instead
TensorLike = Union[np.ndarray, tf.Tensor, tf.Variable, Parameter]
```

gpflow/base.py
joelberkeley-pio: meh - is that needed here?

gpflow/base.py
joelberkeley-pio: docstrings are usually below the item, no?

st--: strings after, comments before (or same line): https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#directive-autodata

joelberkeley-pio: ok. either way, is there a problem with using """ xyz """? I prefer that

joelberkeley-pio: pycharm renders `"""` differently to `#:`, and I am (and I would wager others are) more used to `#` meaning comment not docstring, even if it has a `:` after it.

also multiline docstrings are common and easier with `"""`

gpflow/base.py
```diff
@@ -307,6 +304,19 @@ def _run_op(a, *args, **kwargs):
...
+TensorLike = Union[np.ndarray, tf.Tensor, tf.Variable, Parameter]
+"""
+Type alias for tensor-like types that are supported by most TensorFlow, NumPy and GPflow operations.
+
+NOTE: Union types like this do not work with the `register` method of multipledispatch's
+`Dispatcher` class. Instead use `TensorLikeTypes` for dispatching on tensor-like types.
+"""
+
+# We've left this as object until we've tested the performance consequences of using the full set
+# see github issue #1434
```
st--, line 316:
```suggestion
# see https://github.com/GPflow/GPflow/issues/1434
```
I think it's generally more helpful to have the full link there - one click instead of a bunch of clicking & typing

gpflow/base.py
st--: I prefer the comments but I don't mind very much :shrug:
I've left the content change as a new separate suggestion

st--:
> This is the risk that one has to accept by using third party utilities over the stdlib

I wish Python would provide multiple dispatch by default (in the stdlib)...

Did you already check whether sphinx pulls in the module level docstrings even without `autodata`?

gpflow/base.py
```diff
@@ -307,6 +304,19 @@ def _run_op(a, *args, **kwargs):
 Parameter._OverloadAllOperators()
 tf.register_tensor_conversion_function(Parameter, lambda x, *args, **kwds: x.read_value())
 
+TensorLike = Union[np.ndarray, tf.Tensor, tf.Variable, Parameter]
```
st--, line 307:
```suggestion

TensorLike = Union[np.ndarray, tf.Tensor, tf.Variable, Parameter]
```
(to visually set it more apart from the Parameter class details; black is happy with two-line blanks in this context)

gpflow/base.py
joelberkeley-pio: ok. will do

joelberkeley-pio:
> I just realised that this is breaking backwards compatibility, if anyone actually used `TensorLike` to mean `object` before (e.g. in registering their own implementations for multipledispatch)... how do you think we should handle this ?

We can either:

  * class it as a bug fix (which is reasonable), and just take the flak for changing it
  * use `TensorLike` to refer to the tuple of `(np.ndarray, ...)`, so the change wouldn't be less likely to impact users, and have a different name for the `Union`

Other ideas?

gpflow/base.py
joelberkeley-pio: will do

joelberkeley-pio:
e.g.
```
TensorType = Union[np.ndarray, tf.Tensor, tf.Variable, Parameter]
TensorLike = (object,)  # (np.ndarray, tf.Tensor, tf.Variable, Parameter)
```
or instead of `TensorType`

gpflow/base.py
joelberkeley-pio: will do

gpflow/base.py
```diff
@@ -308,6 +306,24 @@ def _run_op(a, *args, **kwargs):
 tf.register_tensor_conversion_function(Parameter, lambda x, *args, **kwds: x.read_value())
 
 
+TensorType = Union[np.ndarray, tf.Tensor, tf.Variable, Parameter]
+"""
+Type alias for tensor-like types that are supported by most TensorFlow, NumPy and GPflow operations.
+
+NOTE: Union types like this do not work with the `register` method of multipledispatch's
+`Dispatcher` class. Instead use `TensorLikeTypes` for dispatching on tensor-like types.
```
st--, line 314:
```suggestion
`Dispatcher` class. Instead use `TensorLike` for dispatching on tensor-like types.
```

gpflow/base.py
```diff
@@ -308,6 +306,24 @@ def _run_op(a, *args, **kwargs):
...
+"""
+
+
+# We've left this as object until we've tested the performance consequences of using the full set
+# (np.ndarray, tf.Tensor, tf.Variable, Parameter), see https://github.com/GPflow/GPflow/issues/1434
+TensorLike: Final[Tuple[type, ...]] = (object,)
+"""
+:var TensorLike: Collection of tensor-like types for registering implementations with
```
st--, line 322:
```suggestion
:const TensorLike: Collection of tensor-like types for registering implementations with
```
?

setup.py
```diff
@@ -18,7 +18,13 @@
 
 
 # Dependencies of GPflow
-requirements = ["numpy>=1.10.0", "scipy>=0.18.0", "multipledispatch>=0.6", "tabulate"]
+requirements = [
+    "numpy>=1.10.0",
+    "scipy>=0.18.0",
+    "multipledispatch>=0.6",
+    "tabulate",
+    "typing_extensions",
```
st--, line 26:
```suggestion
    "typing_extensions",  # for python 3.6 compatibility
```

gpflow/base.py
joelberkeley-pio: nice. didn't know about that one

gpflow/base.py
joelberkeley-pio: good catch

gpflow/base.py
joelberkeley-pio: oh, I think it's only for _referencing_ not declaring ...

for const
>     Reference a “defined” constant. This may be a Python variable that is not intended to be changed.

for var
> var, ivar, cvar: Description of a variable.

though it's not particularly clear

joelberkeley-pio: i noticed cos pycharm doesn't highlight `const` like it highlights `var`

PR #1427, opened 2020-04-14
add type hints: base

joelberkeley-pio:
does #1368 for base

### DIFF
--- a/gpflow/base.py
+++ b/gpflow/base.py
@@ -1,6 +1,6 @@
 import functools
 from enum import Enum
-from typing import List, Optional, Tuple, Union
+from typing import Any, List, Optional, Sequence, TYPE_CHECKING, Tuple, Union
 
 import numpy as np
 import tensorflow as tf
@@ -10,30 +10,56 @@ from typing_extensions import Final
 
 from .config import default_float, default_summary_fmt
 
+if TYPE_CHECKING:
+    from IPython.lib import pretty
+
 DType = Union[np.dtype, tf.DType]
-VariableData = Union[List, Tuple, np.ndarray, int, float]
+VariableData = Union[List, Tuple, np.ndarray, int, float]  # deprecated
 Transform = Union[tfp.bijectors.Bijector]
 Prior = Union[tfp.distributions.Distribution]
 
 
-def _IS_PARAMETER(o):
+TensorType = Union[tf.Tensor, tf.Variable, "Parameter"]
+"""
+Type alias for tensor-like types that are supported by most TensorFlow and GPflow operations.
+
+NOTE: Union types like this do not work with the `register` method of `multipledispatch`'s
+`Dispatcher` class. Instead use `TensorLike`.
+"""
+
+
+# We've left this as object until we've tested the performance consequences of using the full set
+# (np.ndarray, tf.Tensor, tf.Variable, Parameter), see https://github.com/GPflow/GPflow/issues/1434
+TensorLike: Final[Tuple[type, ...]] = (object,)
+"""
+:var TensorLike: Collection of tensor-like types for registering implementations with
+    `multipledispatch` dispatchers.
+"""
+
+
+_NativeScalar = Union[int, float]
+_Array = Sequence[Any]  # a nested array of int, float, bool etc. kept simple for readability
+TensorData = Union[_NativeScalar, _Array, TensorType]
+
+
+def _IS_PARAMETER(o: Any) -> bool:
     return isinstance(o, Parameter)
 
 
-def _IS_TRAINABLE_PARAMETER(o):
+def _IS_TRAINABLE_PARAMETER(o: Any) -> bool:
     return _IS_PARAMETER(o) and o.trainable
 
 
 class Module(tf.Module):
     @property
-    def parameters(self):
+    def parameters(self) -> Tuple["Parameter", ...]:
         return tuple(self._flatten(predicate=_IS_PARAMETER))
 
     @property
-    def trainable_parameters(self):
+    def trainable_parameters(self) -> Tuple["Parameter", ...]:
         return tuple(self._flatten(predicate=_IS_TRAINABLE_PARAMETER))
 
-    def _representation_table(self, object_name, tablefmt):
+    def _representation_table(self, object_name: str, tablefmt: Optional[str]) -> str:
         from .utilities import leaf_components, tabulate_module_summary
 
         repr_components = [object_name]
@@ -41,13 +67,13 @@ class Module(tf.Module):
             repr_components.append(tabulate_module_summary(self, tablefmt=tablefmt))
         return "\n".join(repr_components)
 
-    def _repr_html_(self):
+    def _repr_html_(self) -> str:
         """ Nice representation of GPflow objects in IPython/Jupyter notebooks """
         from html import escape
 
         return self._representation_table(escape(repr(self)), "html")
 
-    def _repr_pretty_(self, p, cycle):
+    def _repr_pretty_(self, p: "pretty.RepresentationPrinter", cycle: bool) -> None:
         """ Nice representation of GPflow objects in the IPython shell """
         repr_str = self._representation_table(repr(self), default_summary_fmt())
         p.text(repr_str)
@@ -61,7 +87,7 @@ class PriorOn(Enum):
 class Parameter(tf.Module):
     def __init__(
         self,
-        value,
+        value: TensorData,
         *,
         transform: Optional[Transform] = None,
         prior: Optional[Prior] = None,
@@ -91,7 +117,7 @@ class Parameter(tf.Module):
                 unconstrained_value, dtype=dtype, name=name, trainable=trainable
             )
 
-    def log_prior_density(self):
+    def log_prior_density(self) -> tf.Tensor:
         """ Log of the prior probability density of the constrained variable. """
 
         if self.prior is None:
@@ -116,41 +142,41 @@ class Parameter(tf.Module):
             return log_p
 
     @property
-    def prior_on(self):
+    def prior_on(self) -> PriorOn:
         return self._prior_on
 
     @prior_on.setter
-    def prior_on(self, value: Union[str, PriorOn]):
+    def prior_on(self, value: Union[str, PriorOn]) -> None:
         self._prior_on = PriorOn(value)
 
-    def value(self):
-        return _to_constrained(self._unconstrained.value(), self.transform)
+    def value(self) -> tf.Tensor:
+        return _to_constrained(self._unconstrained.value(), self.transform)  # type: ignore  # assumes _to_constrained returns a tf.Tensor
 
-    def read_value(self):
-        return _to_constrained(self._unconstrained.read_value(), self.transform)
+    def read_value(self) -> tf.Tensor:
+        return _to_constrained(self._unconstrained.read_value(), self.transform)  # type: ignore  # assumes _to_constrained returns a tf.Tensor
 
-    def experimental_ref(self):
+    def experimental_ref(self) -> "Parameter":
         return self
 
-    def deref(self):
+    def deref(self) -> "Parameter":
         return self
 
     @property
-    def unconstrained_variable(self):
+    def unconstrained_variable(self) -> tf.Variable:
         return self._unconstrained
 
     @property
-    def transform(self):
+    def transform(self) -> Optional[Transform]:
         return self._transform
 
     @transform.setter
-    def transform(self, new_transform):
+    def transform(self, new_transform: Optional[Transform]) -> None:
         constrained_value = self.read_value()
         self._transform = new_transform
         self.assign(constrained_value)
 
     @property
-    def trainable(self):
+    def trainable(self) -> bool:
         """
         `True` if this instance is trainable, else `False`.
 
@@ -159,10 +185,10 @@ class Parameter(tf.Module):
         return self._unconstrained.trainable
 
     @property
-    def initial_value(self):
+    def initial_value(self) -> tf.Tensor:
         return self._unconstrained.initial_value
 
-    def validate_unconstrained_value(self, value: tf.Tensor, dtype: DType) -> tf.Tensor:
+    def validate_unconstrained_value(self, value: TensorData, dtype: DType) -> tf.Tensor:
         value = _cast_to_dtype(value, dtype)
         unconstrained_value = _to_unconstrained(value, self.transform)
         message = (
@@ -173,8 +199,12 @@ class Parameter(tf.Module):
         return tf.debugging.assert_all_finite(unconstrained_value, message=message)
 
     def assign(
-        self, value: tf.Tensor, use_locking=False, name=None, read_value=True
-    ) -> tf.Variable:
+        self,
+        value: TensorData,
+        use_locking: bool = False,
+        name: Optional[str] = None,
+        read_value: bool = True,
+    ) -> tf.Tensor:
         """
         Assigns constrained `value` to the unconstrained parameter's variable.
         It passes constrained value through parameter's transform first.
@@ -201,7 +231,7 @@ class Parameter(tf.Module):
         )
 
     @property
-    def is_tensor_like(self):
+    def is_tensor_like(self) -> bool:
         """
         This method means that TensorFlow's `tensor_util.is_tensor` function
         will return `True`
@@ -209,47 +239,47 @@ class Parameter(tf.Module):
         return True
 
     @property
-    def name(self):
+    def name(self) -> str:
         return self._unconstrained.name
 
     @property
-    def initializer(self):
+    def initializer(self):  # type unknown
         return self._unconstrained.initializer
 
     @property
-    def device(self):
+    def device(self) -> Optional[str]:
         return self._unconstrained.device
 
     @property
-    def dtype(self):
+    def dtype(self) -> tf.DType:
         return self._unconstrained.dtype
 
     @property
-    def op(self):
+    def op(self) -> tf.Operation:
         return self._unconstrained.op
 
     @property
-    def shape(self):
+    def shape(self) -> tf.TensorShape:
         if self.transform is not None:
             return self.transform.forward_event_shape(self._unconstrained.shape)
         return self._unconstrained.shape
 
-    def numpy(self):
+    def numpy(self) -> np.ndarray:
         return self.read_value().numpy()
 
-    def get_shape(self):
+    def get_shape(self) -> tf.TensorShape:
         return self.shape
 
-    def _should_act_as_resource_variable(self):
+    def _should_act_as_resource_variable(self):  # type unknown
         # needed so that Parameters are correctly identified by TensorFlow's
         # is_resource_variable() in resource_variable_ops.py
         pass  # only checked by TensorFlow using hasattr()
 
     @property
-    def handle(self):
+    def handle(self):  # type unknown
         return self._unconstrained.handle
 
-    def __repr__(self):
+    def __repr__(self) -> str:
         unconstrained = self.unconstrained_variable
         constrained = self.read_value()
         if tf.executing_eagerly():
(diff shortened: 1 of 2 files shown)


gpflow/base.py
```diff
@@ -74,7 +64,7 @@ def __init__(
 
         self._transform = transform
         self.prior = prior
-        self.prior_on = prior_on
+        self._prior_on = prior_on
```
st--, line 67: this is changing behaviour, it will break passing in a string as `prior_on`.

joelberkeley-pio: are you sure? we have the getters and setters defined below (which is why I made this change)

st--: Try it out for yourself! If the `__init__` is called with, e.g., `prior_on="unconstrained"`, then this line assigns `self._prior_on` to be that string, bypassing the setter - which would have transformed it from a string to the corresponding PriorOn enum instance.

gpflow/base.py
```diff
@@ -1,51 +1,41 @@
 import functools
 from enum import Enum
-from typing import List, Optional, Tuple, Union
+from typing import Any, List, Optional, Sequence, Tuple, TypeVar, Union
 
 import numpy as np
 import tensorflow as tf
 import tensorflow_probability as tfp
+from IPython.lib import pretty
```
st--, line 8: Can we please not add IPython as a strict dependency?

gpflow/base.py
```diff
@@ -307,10 +305,36 @@ def _run_op(a, *args, **kwargs):
...
+    @property
+    def trainable_parameters(self) -> Tuple[Parameter, ...]:
+        return tuple(self._flatten(predicate=_IS_TRAINABLE_PARAMETER))
+
+    def _repr_html_(self) -> str:
+        from .utilities import tabulate_module_summary
+
+        return tabulate_module_summary(self, tablefmt="html")
+
+    def _repr_pretty_(self, p: pretty.RepresentationPrinter, cycle: bool) -> None:
```
st--, line 325: I'm generally happy to add types but in this instance I don't think it's worth adding the IPython dependency and import. Could mypy handle it as
```suggestion
    def _repr_pretty_(self, p: "IPython.lib.pretty.RepresentationPrinter", cycle: bool) -> None:
```
?
If not, I would forgo the type
```suggestion
    def _repr_pretty_(self, p, cycle: bool) -> None:
```
you can still add it in a docstring if you want.

gpflow/base.py
```diff
@@ -320,13 +344,15 @@ def _cast_to_dtype(value: VariableData, dtype: Optional[DType] = None) -> tf.Ten
         return tf.convert_to_tensor(value, dtype=dtype)
 
 
-def _to_constrained(value: VariableData, transform: Transform) -> tf.Tensor:
+def _to_constrained(value: tf.Tensor, transform: Optional[Transform]) -> tf.Tensor:
```
st--, line 347: TensorLike not Tensor

gpflow/base.py
```diff
@@ -320,13 +344,15 @@ def _cast_to_dtype(value: VariableData, dtype: Optional[DType] = None) -> tf.Ten
...
     if transform is not None:
         return transform.forward(value)
     return value
 
 
-def _to_unconstrained(value: VariableData, transform: Transform) -> tf.Tensor:
+def _to_unconstrained(
+        value: Union[tf.Tensor, tf.Variable], transform: Optional[Transform]
```
st--, line 354: can take any TensorLike

gpflow/base.py
joelberkeley-pio: ah, i didn't realise that this would call the setter. i saw that `_prior_on` isn't defined on explicitly in `__init__`. i wonder if it might be safer to use `self._prior_on = Prior(prior_on)` in `__init__`. Either way, it's not needed in this PR, so I'll change it back

gpflow/base.py
joelberkeley-pio: does this mean `_repr_pretty_` is an optional feature? if so, we should mark it as such

gpflow/base.py
joelberkeley-pio: this is a tricky one. i thought very carefully about these types. if we make this anything other than `tf.Tensor`, there's not guarantee `value` and `read_value` on `Parameter` will return the types required for a `tf.Variable`

gpflow/base.py
joelberkeley-pio: (here's how forward references work)[https://www.python.org/dev/peps/pep-0484/#forward-references]. in particular, putting types as strings only works if the expression makes sense once the file is loaded - i.e. we'd still need to import that class. The answer, if we're going to omit it, is to use `Any` and comment why. We can discuss that in the thread above

gpflow/base.py
joelberkeley-pio: yes, this one can take a `TensorLike`. odd - i thought it wouldn't work in `Parameter`, but hey. I've also corrected `validate_unconstrained_value` to return just `tf.Tensor`

gpflow/base.py
joelberkeley-pio: that is, whatever calls `_repr_pretty_` must have IPython as a dependency to be able to construct the `p` argument, so IPython is a dependency even if we don't import it explicitly. So I've taken 'strict' to mean that not all users will use `_repr_pretty_` and shouldn't require the dependency if they don't use it. Is that what you mean?

joelberkeley-pio: if by strict you mean what I've detailed, I'd like to have a wider discussion on what the GPflow standard is wrt optional dependencies. I'm wary of adding lots as it sounds like a serious maintenance headache

st--: The only thing that calls `_repr_pretty_` is IPython itself, and it's marked private, so I can't see any issues in just leaving it as it is. If you don't have IPython, there is no reason for it being called. If you do have IPython, you transparently get much nicer object-prints. I agree that if we had lots and lots of optional dependencies, we should reconsider, but at this point I think it's fine as it is.

gpflow/base.py
st--: It's not Any, though; I agree that adding types can help, but adding type hints that aren't actually correct I think is worse than not having any type whatsoever.

st--: Though in any case, it's a private method that will only internally be called by IPython, so we might as well point to the IPython docs and omit type hints here altogether imo.

gpflow/base.py
```diff
@@ -1,51 +1,41 @@
...
 
+S = TypeVar("S")
+_NestedSeq = Union[
+    Sequence[S],
+    Sequence[Sequence[S]],
+    Sequence[Sequence[Sequence[S]]],
+    Sequence[Sequence[Sequence[Sequence[Any]]]]  # catch all for rank > 3
+]
+_NativeScalar = Union[int, float]
+_ArrayOrScalar = Union[_NativeScalar, _NestedSeq[_NativeScalar]]
```
st--, line 28: This is a case where I find the added typing significantly reduces code readability.

joelberkeley-pio: I can add a docstring for `_ArrayOrScalar` explaining what it is

gpflow/base.py
joelberkeley-pio: one main benefit of typing it is that we gain type checking for our implementation of `_repr_pretty_`. I feel that omitting types cos something's an optional dependency would be somewhat of an antipattern. If it's an optional dependency, there are other ways of allowing for that

gpflow/base.py
joelberkeley-pio: you're right that typing it explicitly as `Any` isn't great cos it gives the impression that's what it is, when it's not. What I'm trying to avoid is leaving types out cos then it's not clear if that's been done out of omission or some other reason. See my other comment for my latest thought

gpflow/base.py
joelberkeley-pio: I could allow `TensorLike` by using overloads, but that's overkill for a private, essentially one-line, function

gpflow/base.py
joelberkeley-pio: I think I've solved it: conditionally import IPython only for type checking

gpflow/base.py
joelberkeley-pio: solved: see above

gpflow/base.py
joelberkeley-pio: added comment. how's that?

gpflow/base.py
st--: It's not correct to say that this function requires `value` to be of type `tf.Tensor`. I'd rather not have a type hint than one that's incorrect - so given that it's private, maybe just remove it ?

gpflow/base.py
```diff
@@ -10,38 +10,42 @@
...
+# represents an `int`, `float`, or nested `Sequence` of `int` or `float`, such as
+#
+# 1, 1.1, [1.1], or [[1.1], [1, 1.1]]
+#
+# The ranks must be consistent, but the lengths are unconstrained
 
-def _IS_TRAINABLE_PARAMETER(o):
-    return _IS_PARAMETER(o) and o.trainable
 
+_TensorTypeExternal = Union[tf.Tensor, tf.Variable, np.ndarray]
```
st--, line 40: Can also be int and float for scalar tensors

gpflow/base.py
```diff
@@ -192,55 +202,55 @@ def assign(
...
-    def numpy(self):
+    def numpy(self) -> np.ndarray:
         return self.read_value().numpy()
 
-    def get_shape(self):
+    def get_shape(self) -> tf.TensorShape:
         return self.shape
 
-    def _should_act_as_resource_variable(self):
+    def _should_act_as_resource_variable(self) -> UNKNOWN:
```
st--, line 244: It's not unknown, it's simply not returning anything.
```suggestion
    def _should_act_as_resource_variable(self):
```
Blame it on tensorflow internals

gpflow/base.py
joelberkeley-pio: keenness get the better of you ti?

gpflow/base.py
joelberkeley-pio: I've tried to type all these methods to be as they are on Tensor and Variable, but I couldn't find anything on this method

what do you mean by "not returning anything"?

gpflow/base.py
joelberkeley-pio: i don't think it's true to say it's not correct. it's perfectly OK to ask for a stricter type than the function implementation supports. then again, if you don't want us to do that in this case, that's a separate valid question

gpflow/base.py
joelberkeley-pio: btw by `UNKNOWN` I mean types that I just can't figure out, not that the types are inherently unknown. It's so that everything has an explicit type, which makes it much easier to find missing type hints

joelberkeley-pio: alternatively, I could add comments to each explaining we don't know the type

gpflow/base.py
joelberkeley-pio: this is a good point. I'm just trying to think if this is right generally

gpflow/base.py
```diff
@@ -52,7 +56,7 @@ class PriorOn(Enum):
 class Parameter(tf.Module):
     def __init__(
         self,
-        value,
+        value: Union[_ArrayOrScalar, _TensorTypeExternal, "Parameter"],
```
condnsdmatters, line 59: is it worth having an alias for this union. Its used a lot

gpflow/base.py
```diff
@@ -192,55 +202,55 @@ def assign(
         )
 
     @property
-    def is_tensor_like(self):
+    def is_tensor_like(self) -> UNKNOWN:
```
condnsdmatters, line 205: `UNKNOWN` not `bool`?

gpflow/base.py
condnsdmatters: `->None`?

gpflow/base.py
```diff
@@ -324,9 +334,32 @@ def _run_op(a, *args, **kwargs):
 """
 
 
-def _cast_to_dtype(value: VariableData, dtype: Optional[DType] = None) -> tf.Tensor:
+class Module(tf.Module):
+    @property
+    def parameters(self) -> Tuple[Parameter, ...]:
```
condnsdmatters, line 339: does this syntax mean ... a tuple of Parameters? or one Parameter, and unkown stuff?

gpflow/base.py
condnsdmatters: i think typehints specify the api further and agree we shouldn't tighten if we intend people to be using tensor like. i agree with ti's second suggestion, since private.

gpflow/base.py
joelberkeley-pio: i don't know what the actual API asks for. I did it this way to avoid working out types from what was in this file, rather work them out from the tf docs

if this method doesn't normally return bool, it's a misnomer, but I wanted to be sure

joelberkeley-pio: changed to bool

gpflow/base.py
joelberkeley-pio: agreed, that's the type as it is here, but as mentioned above, I've tried not to type things based on how they're implemented here. For example, I'd expect `_should_act_as_resource_variable` to actually return `bool` given the name, but I don't really know

gpflow/base.py
joelberkeley-pio: by remove it, do you mean remove the function or remove the types?

joelberkeley-pio: if you mean remove the types, my concern is we're losing out on this part of the original ticket #1368 
> As a GPflow maintainer, I want full type hint coverage so that I can be confident my changes are sound.

gpflow/base.py
joelberkeley-pio: a tuple of parameters

joelberkeley-pio: i think

joelberkeley-pio: https://docs.python.org/3/library/typing.html#typing.Tuple

gpflow/base.py
joelberkeley-pio: probs

joelberkeley-pio: though I'm running out of tensor-related names

gpflow/base.py
joelberkeley-pio: that point is of course up for debate, but I want to draw attention to it

joelberkeley-pio: how about I use `tf.convert_to_tensor` at the call site instead?

gpflow/base.py
```diff
@@ -107,41 +146,41 @@ def log_prior_density(self):
             return log_p
 
     @property
-    def prior_on(self):
+    def prior_on(self) -> Union[str, PriorOn]:
```
st--, line 149:
```suggestion
    def prior_on(self) -> PriorOn:
```

gpflow/base.py
```diff
@@ -164,7 +205,11 @@ def validate_unconstrained_value(self, value: tf.Tensor, dtype: DType) -> tf.Ten
         return tf.debugging.assert_all_finite(unconstrained_value, message=message)
 
     def assign(
-        self, value: tf.Tensor, use_locking=False, name=None, read_value=True
+        self,
+        value: Union[_Array, TensorType],
+        use_locking: bool = False,
+        name: Optional[str] = None,
+        read_value: bool = True,
     ) -> tf.Variable:
```
st--, line 213:
```suggestion
    ) -> tf.Tensor:
```

gpflow/base.py
```diff
@@ -192,55 +237,55 @@ def assign(
...
         return True
 
     @property
-    def name(self):
+    def name(self) -> str:
         return self._unconstrained.name
 
     @property
-    def initializer(self):
+    def initializer(self) -> UNKNOWN:
```
st--, line 252: What's the value of this over leaving it as
```suggestion
    def initializer(self):
```
?

gpflow/base.py
st--: > As a GPflow maintainer, I want full type hint coverage so that I can be confident my changes are sound.

As a GPflow maintainer, I don't want "full type hint coverage" to come at arbitrary cost (w.r.t. type restrictions, code readability, and others).

> how about I use tf.convert_to_tensor at the call site and use TensorType here?

Please don't change the behaviour just to add types.

gpflow/base.py
st--: > alternatively, I could add comments to each explaining we don't know the type, and leave off the type hint

yes please, thanks

gpflow/base.py
```diff
@@ -10,35 +10,74 @@
...
+]
+_NativeScalar = Union[int, float]
+
+
+_Array = _NestedSeq[_NativeScalar]
+# represents a nested `Sequence` of `int` or `float`, such as
+#
+# 1, 1.1, [1.1], or [[1.1], [1, 1.1]]
+#
+# The ranks must be consistent, but the lengths are unconstrained
```
st--, line 37: I find this to be very hard-to-comprehend, I think it's reducing the readability of the source file significantly. I also think the type is wrong (but it's so hard to comprehend it's hard to be sure). What is the value we get out of it in return? I'd say use `Sequence[Union[int, float]]` and be done with it...

joelberkeley-pio: > I find this to be very hard-to-comprehend, I think it's reducing the readability of the source file significantly.

ok. that's a concern. I personally find it makes it easier to read cos I then know what's meant to do what. I suppose my thought would be that it's easier to debug a mypy error than a runtime error. We're trading one for the probability of the other

> I also think the type is wrong (but it's so hard to comprehend it's hard to be sure).

I'd be interested to know in what way you think it's wrong, even just from an academic point of view

> What is the value we get out of it in return?

It's a bit hard to answer this in general. Most of the advantage is at call site: when someone uses the relevant methods with something that's not array-shaped, mypy would say

> I'd say use Sequence[Union[int, float]] and be done with it...

It would have to be `Sequence[Any]`. I could do that

gpflow/base.py
joelberkeley-pio: good point. as an aside to this PR, are users going to get confused if they do `param.prior_on = "constrained"` and then find that `param.prior_on != "constrained"`?

gpflow/base.py
joelberkeley-pio: whoops, yes

gpflow/base.py
joelberkeley-pio: ok. sounds like "full type coverage" isn't quite what we're aiming for. I guess it's not really doable in python anyway. I'll have a think about what we can aim for.

my assumption (ok, more a hope) was that it wouldn't change the functionality since it's returning a Tensor anyway. but I accept there's a risk it might, so I'll rethink that

joelberkeley-pio: I've gone with typing it to `TensorType` cos I think that's better than no type. I've put `# type: ignore` at call site

gpflow/base.py
st--: I think that's an acceptable risk compared to the value of only ever having to compare against the actual enum in our code. There's already a bunch of implicit conversions from numpy to tensor...

gpflow/base.py
st--: How is it different from `TensorLike` (or `TensorType`, whichever one it is now)?

gpflow/base.py
joelberkeley-pio: re :/ i thought planned to hand this PR over to eric since your week was up

gpflow/base.py
joelberkeley-pio: it's currently `Union[_Array, TensorType]`

st--: yeah but I can do that anywhere else that expects a tensor-like thing, too.

joelberkeley-pio: you can't do `[1, 2, 3] / 2` but you can do `tf.constant([1, 2, 3]) / 2`

gpflow/base.py
joelberkeley-pio: wrt ints and floats, they don't have `__matmul__` i.e. you can't do `1 @ 1`. I think this means they can't always be treated in quite the same way as `Tensor`s

gpflow/base.py
st--: urgh. that's the part I really dislike about adding types, I think this is getting to the point where it makes the code less clear and less readable, just to satisfy mypy under all possible conditions...

gpflow/base.py
```diff
@@ -10,35 +10,58 @@
...
 DType = Union[np.dtype, tf.DType]
-VariableData = Union[List, Tuple, np.ndarray, int, float]
+VariableData = Union[List, Tuple, np.ndarray, int, float]  # deprecated
 Transform = Union[tfp.bijectors.Bijector]
 Prior = Union[tfp.distributions.Distribution]
 
+_NativeScalar = Union[int, float]
+_Array = Sequence[Any]  # a nested array of int, float, bool etc. kept simple for readability
+
+TensorType = Union[_NativeScalar, tf.Tensor, tf.Variable, np.ndarray, "Parameter"]
```
st--, line 24: Why include _NativeScalar but not _Array?

gpflow/base.py
joelberkeley-pio: i anticipate this is one of the harder, if not the hardest part of adding types to gpflow - sorting out the tensors, but I hope that once we've done it once, we won't need to do it again. I've just changed the type aliases a little, and aside from the `TensorLike` which is needed for multipledispatch, there's only two standard tensor types: `Union[tf.Tensor, tf.Variable, np.ndarray, "Parameter"]` and that plus `Union[int, float, Sequence[Any]]` (though in some cases we use `tf.Tensor`, `Parameter` or `tf.Variable` explicitly)

joelberkeley-pio: I hope I can demonstrate to you this is doable in a neat way

gpflow/base.py
```diff
@@ -10,35 +10,61 @@
...
+
 DType = Union[np.dtype, tf.DType]
-VariableData = Union[List, Tuple, np.ndarray, int, float]
+VariableData = Union[List, Tuple, np.ndarray, int, float]  # deprecated
 Transform = Union[tfp.bijectors.Bijector]
 Prior = Union[tfp.distributions.Distribution]
 
 
-def _IS_PARAMETER(o):
+TensorType = Union[tf.Tensor, tf.Variable, np.ndarray, "Parameter"]
```
st--, line 22:
```suggestion
TensorType = Union[tf.Tensor, tf.Variable, "Parameter"]
```
`np.ndarray` may work in practice but there are occasional differences (see e.g. #1439)

joelberkeley-pio: was the suggestion to remove `ndarray` entirely from these types?

PR #1430, opened 2020-04-14
fix kernel construction in multioutput notebook

### LINKED ISSUE #1417: Some documentation still treats kernels as if they have an input_dim parameter
In the notebook ["Multi-output Gaussian processes in GPflow"](https://gpflow.readthedocs.io/en/develop/notebooks/advanced/multioutput.html), there are Python statements like the following, where the initializers of kernels each have an integer argument:

```python
kernel = gpf.kernels.SharedIndependent(gpf.kernels.RBF(1) + gpf.kernels.Linear(1), output_dim=P)
kern_list = [gpf.kernels.RBF(D) + gpf.kernels.Linear(1) for _ in range(P)]
kern_list = [gpf.kernels.RBF(D) + gpf.kernels.Linear(D) for _ in range(L)]
```

Judging from the definition of `D`, it looks like those integer arguments are meant to be the `input_dim` arguments for the kernel initializers, but those arguments [no longer exist](https://gpflow.readthedocs.io/en/develop/notebooks/gpflow2_upgrade_guide.html#Kernel-Input-Dims).

I don't know if this problem occurs in other notebooks, but it wouldn't surprise me.

st--:
Closes #1417

### DIFF
--- a/doc/source/notebooks/advanced/multioutput.pct.py
+++ b/doc/source/notebooks/advanced/multioutput.pct.py
@@ -168,7 +168,9 @@ def plot_model(m, lower=-8.0, upper=8.0):
 
 # %%
 # create multi-output kernel
-kernel = gpf.kernels.SharedIndependent(gpf.kernels.SquaredExponential(lengthscales=np.ones(D)) + gpf.kernels.Linear(), output_dim=P)
+kernel = gpf.kernels.SharedIndependent(
+    gpf.kernels.SquaredExponential(lengthscales=np.ones(D)) + gpf.kernels.Linear(), output_dim=P
+)
 # initialization of inducing input locations (M random points from the training inputs)
 Z = Zinit.copy()
 # create multi-output inducing variables from Z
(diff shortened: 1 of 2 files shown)


doc/source/notebooks/advanced/multioutput.pct.py
```diff
@@ -168,7 +168,9 @@ def plot_model(m, lower=-8.0, upper=8.0):
 
 # %%
 # create multi-output kernel
-kernel = gpf.kernels.SharedIndependent(gpf.kernels.RBF(1) + gpf.kernels.Linear(1), output_dim=P)
+kernel = gpf.kernels.SharedIndependent(
```
vdutor, line 171: we used to have two non ARD kernels, now we have one ARD (the RBF one) and one non-ARD (the linear one). Is that what you intended to do?

st--: Ah, somehow I misremembered and assumed the default had been ARD=True. I'm happy to take out the `lengthscales` argument - not that it matters in this concrete example as D=1 ...

PR #1438, opened 2020-04-15
add long_description and project_urls to setup.py

st--:
Will give us a much nicer landing page on https://pypi.org/project/gpflow/

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -72,12 +72,12 @@ except (ImportError, DeprecationWarning):
 
 
 def read_file(filename):
-    with open(str(Path(".", filename).absolute())) as f:
+    with open(filename, encoding="utf-8") as f:
         return f.read().strip()
 
 
 version = read_file("VERSION")
-readme = read_file("README.md")
+readme_text = read_file("README.md")
 
 packages = find_packages(".", exclude=["tests"])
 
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -71,8 +71,13 @@ def latest_version(package_name):
         requirements.append(gast_requirement)
 
 
-with open(str(Path(".", "VERSION").absolute())) as version_file:
-    version = version_file.read().strip()
+def read_file(filename):
+    with open(str(Path(".", filename).absolute())) as f:
+        return f.read().strip()
```
awav, line 76:
```suggestion
    readme_file = Path(".", filename).absolute()
    with open(readme_file, encoding="utf-8") as fd:
        return fd.read().strip()
```

setup.py
```diff
@@ -71,8 +71,13 @@ def latest_version(package_name):
...
+
+
+version = read_file("VERSION")
+readme = read_file("README.md")
```
awav, line 80:
```suggestion
readme_text = read_file("README.md")
```

setup.py
```diff
@@ -82,9 +87,15 @@ def latest_version(package_name):
     author="James Hensman, Alex Matthews",
     author_email="james.hensman@gmail.com",
     description="Gaussian process methods in TensorFlow",
+    long_description=readme,
```
awav, line 90:
```suggestion
    long_description=readme_text,
```

PR #1443, opened 2020-04-22
refactor natgrads to be more efficient

st--:
Previously, GPflow's NaturalGradient optimizer would call the loss_function once for each (q_mu, q_sqrt) set in the var_list. This is a light refactor that separates out applying the natural gradient step from computing the gradients (`_natgrad_apply_gradients`), and changes `_natgrad_steps` to only evaluate the loss function once, computing the gradients for all (q_mu, q_sqrt) tuples passed in the var_list.

Other changes:
- The no-longer-used `_natgrad_step` method got removed.
- NaturalGradient now takes a `xi_transform` argument that is used for all parameter sets without explicitly specified xi transform (i.e. tuples rather than triplets).
- XiTransform has been changed to have staticmethods.

None of this should affect any downstream code; this PR is backwards-compatible.

### DIFF
--- a/gpflow/optimizers/natgrad.py
+++ b/gpflow/optimizers/natgrad.py
@@ -19,7 +19,7 @@ from typing import Callable, Optional, Sequence, Tuple, Union
 import numpy as np
 import tensorflow as tf
 
-from ..base import Parameter
+from ..base import Parameter, _to_constrained
 
 Scalar = Union[float, tf.Tensor, np.ndarray]
 LossClosure = Callable[[], tf.Tensor]
@@ -179,12 +179,15 @@ class NaturalGradient(tf.optimizers.Optimizer):
         self._natgrad_steps(loss_fn, parameters)
 
     def _natgrad_steps(
-        self, loss_fn: LossClosure, parameters: Sequence[Tuple[Parameter, Parameter, XiTransform]]
+        self, loss_fn: LossClosure, parameters: Sequence[Tuple[Parameter, Parameter, Optional[XiTransform]]]
     ):
         """
         Computes gradients of loss_fn() w.r.t. q_mu and q_sqrt, and updates
         these parameters using the natgrad backwards step, for all sets of
-        variational parameters.
+        variational parameters passed in.
+
+        :param loss_fn: Loss function.
+        :param parameters: List of tuples (q_mu, q_sqrt, xi_transform)
         """
         q_mus, q_sqrts, xis = zip(*parameters)
         unconstrained_variables = [
@@ -196,6 +199,7 @@ class NaturalGradient(tf.optimizers.Optimizer):
             loss = loss_fn()
 
         q_mu_grads, q_sqrt_grads = tape.gradient(loss, [q_mus, q_sqrts])
+        # NOTE that these are the gradients in *unconstrained* space
 
         with tf.name_scope(f"{self._name}/natural_gradient_steps"):
             for q_mu_grad, q_sqrt_grad, q_mu, q_sqrt, xi_transform in zip(
@@ -213,12 +217,14 @@ class NaturalGradient(tf.optimizers.Optimizer):
     ):
         """
         This function does the backward step on the q_mu and q_sqrt parameters,
-        given the gradients of the loss function with respect to them. I.e.,
-        it expects the arguments to come from
+        given the gradients of the loss function with respect to their unconstrained
+        variables. I.e., it expects the arguments to come from
 
             with tf.GradientTape() as tape:
                 loss = loss_function()
             q_mu_grad, q_mu_sqrt = tape.gradient(loss, [q_mu, q_sqrt])
+
+        (Note that tape.gradient() returns the gradients in *unconstrained* space!)
 
         Implements equation [10] from
 
@@ -239,13 +245,19 @@ class NaturalGradient(tf.optimizers.Optimizer):
 
         Note that if ξ = θ (i.e. [q_mu, q_sqrt]) some of these calculations are the identity.
         In the code η = eta, ξ = xi, θ = nat.
+
+        :param q_mu_grad: gradient of loss w.r.t. q_mu (in unconstrained space)
+        :param q_sqrt_grad: gradient of loss w.r.t. q_sqrt (in unconstrained space)
+        :param q_mu: parameter for the mean of q(u)
+        :param q_sqrt: parameter for the square root of the covariance of q(u)
+        :param xi_transform: the ξ transform to use (self.xi_transform if not specified)
         """
         if xi_transform is None:
             xi_transform = self.xi_transform
 
         # 1) the ordinary gpflow gradient
-        dL_dmean = q_mu_grad
-        dL_dvarsqrt = q_sqrt.transform.forward(q_sqrt_grad)
+        dL_dmean = _to_constrained(q_mu_grad, q_mu.transform)
+        dL_dvarsqrt = _to_constrained(q_sqrt_grad, q_sqrt.transform)
 
         with tf.GradientTape(persistent=True, watch_accessed_variables=False) as tape:
             tape.watch([q_mu.unconstrained_variable, q_sqrt.unconstrained_variable])
--- a/gpflow/optimizers/natgrad.py
+++ b/gpflow/optimizers/natgrad.py
@@ -33,6 +33,12 @@ __all__ = [
 ]
 
 
+#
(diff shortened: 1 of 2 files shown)


gpflow/optimizers/natgrad.py
```diff
@@ -70,18 +69,22 @@ def naturals_to_xi(self, nat1, nat2):
         :param nat2: the θ₂ parameter
         :return: tuple `xi1`, `xi2`
         """
-        pass  # pragma: no cover
 
 
 class NaturalGradient(tf.optimizers.Optimizer):
-    def __init__(self, gamma: Scalar, name=None):
+    def __init__(self, gamma: Scalar, xi_transform: XiTransform = XiNat(), name=None):
```
awav, line 75: Custom transformations go next to parameters and the can be different for a set of parameters. You cannot pass it in `init` method.

gpflow/optimizers/natgrad.py
```diff
@@ -132,21 +178,29 @@ def _natgrad_step(
             year={2018}
 
         In addition, for convenience with the rest of GPflow, this code computes ∂L/∂η using
-        the chain rule:
+        the chain rule (the following assumes a numerator layout where the gradient is a row
+        vector; note that TensorFlow actually returns a column vector), where L is the loss:
 
-        ∂L/∂η = (∂L / ∂[q_μ, q_sqrt])(∂[q_μ, q_sqrt] / ∂η)
+        ∂L/∂η = (∂L / ∂[q_mu, q_sqrt])(∂[q_mu, q_sqrt] / ∂η)
```
awav, line 184: Why is that changed?

gpflow/optimizers/natgrad.py
condnsdmatters: this is a default, it seems

condnsdmatters: see line 101

st--: This is simply changing the default when for a parameter you haven't passed in a transformation.

gpflow/optimizers/natgrad.py
st--: to be consistent with the code

condnsdmatters: i think given this is the name of the variables we are using here, its kinda helpful to have it as q_mu. in general agree being against superfluous changes to docs for formatting but yeah

gpflow/optimizers/natgrad.py
condnsdmatters: Warning, this line will instantiate a static variable `XiNat` which will be shared by *all* instances of `NaturalGradient`.

I suggest changing to a class and instantiating.

```suggestion
    def __init__(self, gamma: Scalar, default_xi_transform: XiTransform = XiNat, name=None):
```
I know that currently the XiTransforms are stateless, but the abstractmethod APIs are not (they take self/arent classmethods)! So either we should say: you can define your own XiTransform which can take state (in which case, we should change this API), or you can not define your own stateful XiTransform  (and that should be reflected in API).

I suggest former.

gpflow/optimizers/natgrad.py
st--: you are against cleaning up the docs ? O.o

gpflow/optimizers/natgrad.py
st--: the XiNat transform does not have a state so it makes no difference if its instance is shared.

st--: We can change the XiTransforms to be classmethods

st--:
@condnsdmatters I've changed the XiTransform methods to all be static.

> _natgrad_step isn't used anywhere now?

no, the public API is just `minimize()`.

gpflow/optimizers/natgrad.py
condnsdmatters:
```suggestion
    def __init__(self, gamma: Scalar, xi_transform: XiTransform = XiNat, name=None):
```

gpflow/optimizers/natgrad.py
awav: I'm not against cleaning the docs. The cleaning could be done in different direction. I like q_μ more, and it doesn't contradicts the code.

gpflow/optimizers/natgrad.py
```diff
@@ -39,114 +47,212 @@ class XiTransform(metaclass=abc.ABCMeta):
...
         Applies the transform so that `nat1`, `nat2` is mapped to `xi1`, `xi2`
 
         :param nat1: the θ₁ parameter
         :param nat2: the θ₂ parameter
         :return: tuple `xi1`, `xi2`
         """
-        pass  # pragma: no cover
+
+
+class XiNat(XiTransform):
```
awav, line 84: I'm not sure. What happened here? :)

gpflow/optimizers/natgrad.py
awav: I'm sorry, but I still don't understand what transform does in `init`. It is "per parameter set" attribute. The default is `XiNat`, and if somebody wants to change it, he/she would need to specify it explicitly via parameters settings.

stefanosele: What Artem want's to point out is that you may pass in let's say variational distributions of 3 layers (at least you could do that in the old code I think), and for some reason you want different transformation to be used for some of the parameters. You don't seem to be able to do that now. In general, the transformation should be tied to the parameter and not the optimizer

gpflow/optimizers/natgrad.py
```diff
@@ -39,114 +47,223 @@ class XiTransform(metaclass=abc.ABCMeta):
...
         self.gamma = gamma
+        self.xi_transform = xi_transform
 
     def minimize(
-        self,
-        loss_fn: Callable,
-        var_list: List[
-            Union[Tuple[Parameter, Parameter], Tuple[Parameter, Parameter, XiTransform]]
-        ],
+        self, loss_fn: LossClosure, var_list: Sequence[NatGradParameters],
```
joelberkeley-pio, line 156: i'm going to chime in about one thing cos I think it's important. This method signature isn't compatible with that of `tf.optimizer.Optimizer`. If that doesn't affect this PR, we can come back to it another time

st--: Yes, very true! When reworking the code for this PR that is something that I also thought of. That is why I tried to make it much clearer in the docstrings - let me know if you have any further suggestions for clarifying that! I'm not sure what a good solution for this could be, I'd appreciate suggestions. It's beyond the scope of this PR though - I haven't actually changed the interface of this method, it behaves exactly as before.

gpflow/optimizers/natgrad.py
st--: I moved it to the top of the file so it could be used within the NaturalGradient class, and I changed the methods to be staticmethods as discussed with @condnsdmatters.

gpflow/optimizers/natgrad.py
st--: q_μ is consistent with neither paper notation (where it is just μ) nor the code (where it is q_mu), it took me extra mental effort to parse this as referring to q_mu...

gpflow/optimizers/natgrad.py
st--: You can still do that, just as before. I've not changed the interface (just minorly added to it, as I needed that for downstream code).

gpflow/optimizers/natgrad.py
joelberkeley-pio: perhaps it doesn't need to be a `tf.optimizers.Optimizer` and could just be its own class

to make it an actual `tf.optimizers.Optimizer`, I don't know how we'd do that. It may very well be possible but I couldn't say just yet

gpflow/optimizers/natgrad.py
awav: @st--, I think what I wanted to know is "why are you changing it"?

gpflow/optimizers/natgrad.py
awav: As you can see I had an opposite view. I would argue that it is better to keep consistency within one thing, rather trying to find consistency among everything. In this case, it was a documentation.

st--: Sorry, did I not change it consistently - did I miss any occurrences? what is it that you think it's not consistent with, now?

PR #1451, opened 2020-05-02
Jitter

markvdw:
Following some discussion on the Slack channel, here is a version of SGPR which has jitter as a parameter. By default, the parameter is not trainable. Moreover, `set_trainable()` has been adapted to not change the trainability of this parameter.

It is still possible for the parameter to become trainable, by a user directly setting the `_trainable` flag in the `tf.Variable` object.

Originally, I thought this solution would be a bit dangerous, in that the jitter could accidentally be set to be trainable. However, given that a _private_ method needs to be adapted to do this, I now don't feel so bad about this.

I believe that *some* kind of model-level jitter adaptation is very important in the longer term, for some work on making the training of SGPR models more robust.

We can also think about adding a similar property to `SVGP`.

### DIFF
--- a/gpflow/models/sgpr.py
+++ b/gpflow/models/sgpr.py
@@ -103,7 +103,7 @@ class SGPRBase(GPModel, InternalDataTrainingLossMixin):
         num_data = to_default_float(tf.shape(Y_data)[0])
 
         Kdiag = self.kernel(X_data, full_cov=False)
-        kuu = Kuu(self.inducing_variable, self.kernel, jitter=default_jitter())
+        kuu = Kuu(self.inducing_variable, self.kernel, jitter=self.jitter_variance)
         kuf = Kuf(self.inducing_variable, self.kernel, X_data)
 
         I = tf.eye(tf.shape(kuu)[0], dtype=default_float())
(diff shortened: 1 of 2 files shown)


jameshensman:
Looks sensible. What would it take to do the same for the other models?

st--:
It'd be great to get more robust model optimisation! It seems a bit premature to completely change the API before having spent some time exploring how to actually make it work, though. I don't want you to have to maintain your own copy of SGPR though, so how about, for now, just introducing a private `_jitter_level` method to SGPR (and the other models, if you want): this could keep the current behaviour the default:
```python
    def _jitter_level(self):
        return gpflow.config.default_jitter()
```
and then it is very easy to create subclassed models and investigate different ways of making it work with optimizers:
```python
class RobustSGPR(gpflow.models.SGPR):
    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
        self.jitter_variance = Parameter(
            default_jitter(), transform=positive(lower=0.0), trainable=False, name="jitter"
        )

    def _jitter_level(self):
        return self.jitter_variance
```
Having this change as a private method would mean we can easily change it in the future.

This doesn't answer your concern about the jitter variable accidentally being made trainable by `set_trainable(model, True)`, but until the whole robust optimisation is in gpflow-core perhaps it would be ok for you to just use your own set_trainable that includes the name-check (in the longer run it'd be good to find a more robust solution than that...)?

vdutor:
Hi @markvdw, how do you want to proceed with this?

st--:
@markvdw please feel free to re-open this PR if you still want to merge it into gpflow-core and discuss how we can turn it into a more robust solution. Thanks!

Issue #1439, opened 2020-04-16
"ValueError: Cannot create a tensor proto whose content is larger than 2GB." after upgrade from 2.0.0rc1 to 2.0.0 release


clwgg:
Hi,
I've been running standard GPR on a fairly large dataset using the `2.0.0rc1` version of GPflow. Since the upgrade to `2.0.0`, I'm running into the error
```
ValueError: Cannot create a tensor proto whose content is larger than 2GB.
```
on a dataset which worked fine before.

Here is the full error message:
```
Traceback (most recent call last):
  File "main.py", line 177, in <module>
    main(args)
  File "main.py", line 64, in main
    build_allele(args)
  File "/path/to/1_model_sim/drivers.py", line 226, in build_allele
    opt_model_list(m)
  File "/path/to/1_model_sim/model.py", line 355, in opt_model_list
    m.trainable_variables)
  File "/path/to/1_model_sim/venv/lib/python3.6/site-packages/gpflow/optimizers/scipy.py", line 73, in minimize
    func, initial_params, jac=True, method=method, **scipy_kwargs
  File "/path/to/1_model_sim/venv/lib/python3.6/site-packages/scipy/optimize/_minimize.py", line 610, in minimize
    callback=callback, **options)
  File "/path/to/1_model_sim/venv/lib/python3.6/site-packages/scipy/optimize/lbfgsb.py", line 345, in _minimize_lbfgsb
    f, g = func_and_grad(x)
  File "/path/to/1_model_sim/venv/lib/python3.6/site-packages/scipy/optimize/lbfgsb.py", line 295, in func_and_grad
    f = fun(x, *args)
  File "/path/to/1_model_sim/venv/lib/python3.6/site-packages/scipy/optimize/optimize.py", line 327, in function_wrapper
    return function(*(wrapper_args + args))
  File "/path/to/1_model_sim/venv/lib/python3.6/site-packages/scipy/optimize/optimize.py", line 65, in __call__
    fg = self.fun(x, *args)
  File "/path/to/1_model_sim/venv/lib/python3.6/site-packages/gpflow/optimizers/scipy.py", line 95, in _eval
    loss, grad = _tf_eval(tf.convert_to_tensor(x))
  File "/path/to/1_model_sim/venv/lib/python3.6/site-packages/tensorflow_core/python/eager/def_function.py", line 568, in __call__
    result = self._call(*args, **kwds)
[...]

PR #1452, opened 2020-05-04
Convert data structures to tensor in model init method

awav:
Addresses https://github.com/GPflow/GPflow/issues/1439

### DIFF
--- a/gpflow/models/sgpr.py
+++ b/gpflow/models/sgpr.py
@@ -18,15 +18,16 @@ import numpy as np
 import tensorflow as tf
 
 from gpflow.kernels import Kernel
+
 from .. import likelihoods
 from ..config import default_float, default_jitter
 from ..covariances.dispatch import Kuf, Kuu
 from ..inducing_variables import InducingPoints
-from ..mean_functions import Zero, MeanFunction
+from ..mean_functions import MeanFunction, Zero
 from ..utilities import to_default_float
-from .model import GPModel, InputData, RegressionData, MeanAndVariance
+from .model import GPModel, InputData, MeanAndVariance, RegressionData
 from .training_mixins import InternalDataTrainingLossMixin
-from .util import inducingpoint_wrapper
+from .util import data_input_to_tensor, inducingpoint_wrapper
 
 
 class SGPRBase(GPModel, InternalDataTrainingLossMixin):
@@ -56,11 +57,11 @@ class SGPRBase(GPModel, InternalDataTrainingLossMixin):
         initialized to `noise_variance`.
         """
         likelihood = likelihoods.Gaussian(noise_variance)
-        X_data, Y_data = data
+        X_data, Y_data = data_input_to_tensor(data)
         num_latent_gps = Y_data.shape[-1] if num_latent_gps is None else num_latent_gps
         super().__init__(kernel, likelihood, mean_function, num_latent_gps=num_latent_gps)
 
-        self.data = data
+        self.data = X_data, Y_data
         self.num_data = X_data.shape[0]
 
         self.inducing_variable = inducingpoint_wrapper(inducing_variable)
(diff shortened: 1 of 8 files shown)


gpflow/models/util.py
```diff
@@ -52,3 +54,18 @@ def maximum_log_likelihood_objective(model: BayesianModel, data: Data) -> tf.Ten
...
+    """
+    Converts non-tensor elements of a structure to TensorFlow tensors retaining the structure itself.
+    The function doesn't keep original element's dtype and forcefully converts
+    them to GPflow's default float type.
+    """
+
+    def convert_to_tensor(elem):
+        if tf.is_tensor(elem) or isinstance(elem, Iterator):
+            return elem
+        return tf.convert_to_tensor(elem, dtype=default_float())
```
st--, line 69: Why not keep the original dtype?

gpflow/models/sgpr.py
```diff
@@ -56,11 +57,11 @@ def __init__(
         initialized to `noise_variance`.
         """
         likelihood = likelihoods.Gaussian(noise_variance)
-        X_data, Y_data = data
+        X_data, Y_data = data_input_to_tensor(data)
```
st--, line 60:
```suggestion
        self.data = X_data, Y_data = data_input_to_tensor(data)
```

gpflow/models/sgpr.py
```diff
@@ -56,11 +57,11 @@ def __init__(
...
         num_latent_gps = Y_data.shape[-1] if num_latent_gps is None else num_latent_gps
         super().__init__(kernel, likelihood, mean_function, num_latent_gps=num_latent_gps)
 
-        self.data = data
+        self.data = X_data, Y_data
```
st--, line 64:
```suggestion
```

gpflow/models/vgp.py
```diff
@@ -66,10 +67,10 @@ def __init__(
             num_latent_gps = self.calc_num_latent_gps_from_data(data, kernel, likelihood)
         super().__init__(kernel, likelihood, mean_function, num_latent_gps)
 
-        X_data, Y_data = data
+        self.data = data_input_to_tensor(data)
+        X_data, Y_data = self.data
```
st--, line 71:
```suggestion
        self.data = X_data, Y_data = data_input_to_tensor(data)
```

gpflow/models/vgp.py
```diff
@@ -163,8 +164,8 @@ def __init__(
             num_latent_gps = self.calc_num_latent_gps_from_data(data, kernel, likelihood)
         super().__init__(kernel, likelihood, mean_function, num_latent_gps)
 
-        X_data, Y_data = data
-        self.data = data
+        self.data = data_input_to_tensor(data)
+        X_data, Y_data = self.data
```
st--, line 168:
```suggestion
        self.data = X_data, Y_data = data_input_to_tensor(data)
```

gpflow/models/util.py
awav: Because it is not possible to deduce the original type in python for sequence structures.

gpflow/models/sgpr.py
awav: Cannot accept it, as self.data should not exist before `super.__init__`

gpflow/models/sgpr.py
awav: Same reason as above

gpflow/models/vgp.py
awav: Double assign is dirty style and unambiguous with comma operator. This approach is prone to mistakes.

gpflow/models/vgp.py
awav: Same comment as above.

gpflow/models/vgp.py
st--:
```suggestion
        self.data = (X_data, Y_data) = data_input_to_tensor(data)
```
?

awav: Still, having two separate lines is preferable.

PR #1454, opened 2020-05-04
clean up Scipy types & code

st--:
Fixes some typos in the Scipy docstring and adds a note on scipy minimize()'s `callback` argument vs GPflow Scipy().minimize()'s `step_callback`, and cleans up types in the module. Changes the arguments to a step_callback to be passed by position, not name, in line with the StepCallback type (Python's typing.Callable documentation suggests that callbacks should take arguments preferentially by position, not name).


gpflow/optimizers/scipy.py
```diff
@@ -99,34 +100,37 @@ def _eval(x: np.ndarray) -> Tuple[np.ndarray, np.ndarray]:
...
             values = cls.unpack_tensors(variables, x)
-            step_callback(step=step, variables=variables, values=values)
+            step_callback(step, variables, values)
             step += 1
 
         return _callback
 
     @staticmethod
-    def pack_tensors(tensors: Iterable[tf.Tensor]) -> tf.Tensor:
+    def pack_tensors(tensors: Iterable[Union[tf.Tensor, tf.Variable]]) -> tf.Tensor:
```
st--, line 116: the type was wrong, we actually pass in the variables, not tensors (we *only* pass in variables, but the code also works for tensors, so didn't want to restrict needlessly)

gpflow/optimizers/scipy.py
```diff
@@ -99,34 +100,37 @@ def _eval(x: np.ndarray) -> Tuple[np.ndarray, np.ndarray]:
...
         flats = [tf.reshape(tensor, (-1,)) for tensor in tensors]
         tensors_vector = tf.concat(flats, axis=0)
         return tensors_vector
 
     @staticmethod
-    def unpack_tensors(to_tensors: Iterable[tf.Tensor], from_vector: tf.Tensor) -> List[tf.Tensor]:
+    def unpack_tensors(
+        to_tensors: Iterable[Union[tf.Tensor, tf.Variable]], from_vector: tf.Tensor
```
st--, line 123: again, we actually pass in a List[tf.Variable], not tensors

gpflow/optimizers/scipy.py
```diff
@@ -136,10 +140,9 @@ def assign_tensors(to_tensors: Iterable[tf.Variable], values: Iterable[tf.Tensor
             tensor.assign(tensor_vector)
 
 
-V = TypeVar("V", bound=Variables)
-
-
-def _compute_loss_and_gradients(loss_closure: LossClosure, variables: V) -> Tuple[tf.Tensor, V]:
```
st--, line 142: This type says we return a list of variables as the second return value, but actually it's a list of tensors

st--:
> I'm for fixing the typos, but leaving the code as is.
> 
>     * Type variables help skipping unnecessary mental parsing in function arguments.
> 
>     * Type aliases help to avoid issues in types. This is important for python (we don't have mypy setup)

I removed the `Variables = Iterable[tf.Variable]` because looking at the code I found it more mental effort to have to remember what it's defined as. The other type changes are to bring them in line with what's actually being passed around.

gpflow/optimizers/scipy.py
awav: tf.Variable is a tensor.

joelberkeley-pio: are you sure artem?
```
>>> isinstance(tf.Variable(0), tf.Tensor)
False
```

gpflow/optimizers/scipy.py
```diff
@@ -7,16 +7,15 @@
 
 __all__ = ["Scipy"]
 
-Variables = Iterable[tf.Variable]
```
joelberkeley-pio, line 10: not backwards-compatible ... perhaps deprecate instead?

gpflow/optimizers/scipy.py
```diff
@@ -7,16 +7,15 @@
...
-StepCallback = Callable[[int, Variables, List[tf.Tensor]], None]
+StepCallback = Callable[[int, List[tf.Variable], List[tf.Tensor]], None]
```
joelberkeley-pio, line 10: you've replaced `Variables` with `List[tf.Variable]` in a number of places. This is an API change (though admittedly a small one). was `Variables` a bug?

gpflow/optimizers/scipy.py
```diff
@@ -99,34 +100,37 @@ def _eval(x: np.ndarray) -> Tuple[np.ndarray, np.ndarray]:
...
+        for target_tensor in to_tensors:
+            shape = tf.shape(target_tensor)
+            dtype = target_tensor.dtype
             tensor_size = tf.reduce_prod(shape)
-            tensor_vector = tf.cast(from_vector[s : s + tensor_size], tensor.dtype)
-            tensor_vector = tf.reshape(tensor_vector, shape)
-            values.append(tensor_vector)
+            tensor_vector = from_vector[s : s + tensor_size]
+            tensor = tf.reshape(tf.cast(tensor_vector, dtype), shape)
+            values.append(tensor)
```
joelberkeley-pio, line 133: does this change any functionality?

gpflow/optimizers/scipy.py
joelberkeley-pio: re

> didn't want to restrict needlessly

I'd say simplicity is a good argument for ... simplicity. Also a more constrained type can be made more flexible later

gpflow/optimizers/scipy.py
st--: I'd argue it's not part of the "interface" as it's not included in `__all__` ... how would you deprecate a type ?

gpflow/optimizers/scipy.py
st--: It depends which way around you see it: the step_callback will be passed lists - does the type signature need to reflect what the types of the arguments are going to be, or what types of arguments it can accept? If the latter, happy to change this to
```suggestion
StepCallback = Callable[[int, Iterable[tf.Variable], Iterable[tf.Tensor]], None]
```
it should be consistent for both second and third argument though!

gpflow/optimizers/scipy.py
st--: tf.Variable is, unfortunately, NOT a tensor... that's why we've had so much discussion around `TensorType` in #1427 (and I would replace all these unions with `TensorType` if #1427 was merged).

gpflow/optimizers/scipy.py
joelberkeley-pio: `__all__` doesn't define the public API - it defines how `from foo import *` works

oh, I don't mean "officially", just for our sake, with e.g. `# deprecated`

gpflow/optimizers/scipy.py
awav: @joelberkeley-pio, Interesting. I believe it now :) Although, this was not true in `2.0.0` version.

st--:
@joelberkeley-pio @awav please let me know concretely what it would take for you to approve this PR! thanks :)

gpflow/optimizers/scipy.py
st--: for our sake it had only been used in this module though

gpflow/optimizers/scipy.py
joelberkeley-pio: types _usually_ reflect what the types need to be for the functionality to work. There are exceptions, such as where there's a good change you'll need to constrain the types in the future and don't want to prematurely make compatibility promises that you'll need to retract later.

joelberkeley-pio: in this case, if it's valid for it to be `Iterable`, it should be `Iterable` for backwards compatibility

gpflow/optimizers/scipy.py
joelberkeley-pio: we can't know that - Joe Bloggs might have imported in when using 2.0.1

gpflow/optimizers/scipy.py
st--: Let me rephrase: The step_callback will only ever get called with a list. Is the type signature meant to specify what this function *will get* or *what it can take*?

st--: Or we can make it a tuple...

joelberkeley-pio: ah, so my first statement was a mistake (my last one was accurate but gets inverted here due to functions being contravariant in their argument type), apologies for the confusion!

Here we should use `List`, as you have done, as that gives the most flexibility to the user

joelberkeley-pio: it does constrain _us_ to pass it a `List`, but I don't envisage that being a problem

gpflow/optimizers/scipy.py
```diff
@@ -74,12 +75,12 @@ def minimize(
...
 
     @classmethod
-    def initial_parameters(cls, variables: Variables) -> tf.Tensor:
+    def initial_parameters(cls, variables: Iterable[tf.Variable]) -> tf.Tensor:
         return cls.pack_tensors(variables)
 
     @classmethod
     def eval_func(
-        cls, closure: LossClosure, variables: Variables, compile: bool = True
+        cls, closure: LossClosure, variables: List[tf.Variable], compile: bool = True
```
joelberkeley-pio, line 83: unlike `step_callback`, this _is_ backwards incompatible

gpflow/optimizers/scipy.py
st--: Ah. Maybe we should define it to be a tuple (which signals some sort of immutability) - actually changing the list that gets passed in would probably lead to some very weird effects??

gpflow/optimizers/scipy.py
```diff
@@ -74,12 +75,12 @@ def minimize(
         )
 
     @classmethod
-    def initial_parameters(cls, variables: Variables) -> tf.Tensor:
+    def initial_parameters(cls, variables: Iterable[tf.Variable]) -> tf.Tensor:
```
awav, line 78: I still think a type alias is the right thing to use in this case. Could you please reuse variables here and in other places?

gpflow/optimizers/scipy.py
```diff
@@ -136,10 +140,9 @@ def assign_tensors(to_tensors: Iterable[tf.Variable], values: Iterable[tf.Tensor
             tensor.assign(tensor_vector)
 
 
-V = TypeVar("V", bound=Variables)
-
-
-def _compute_loss_and_gradients(loss_closure: LossClosure, variables: V) -> Tuple[tf.Tensor, V]:
+def _compute_loss_and_gradients(
+    loss_closure: LossClosure, variables: List[tf.Variable]
+) -> Tuple[tf.Tensor, List[tf.Tensor]]:
```
awav, line 145: Alias could stay as `List[tf.Tensor]` is repeated multiple times in the file.

gpflow/optimizers/scipy.py
joelberkeley-pio: yeah i guess tuple would be safer

gpflow/optimizers/scipy.py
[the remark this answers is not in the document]
joelberkeley-pio:
```
        s = 0
        v = []
        for t in to_tensors:
            s_ = tf.shape(t)
            d = t.dtype
            s__ = tf.reduce_prod(s_)
            t_ = from_vector[s : s + s__]
            t__ = tf.reshape(tf.cast(t_, d) s_)
            v.append(t__)
            s += s__
        return v
```

gpflow/optimizers/scipy.py
st--: urgh. alright

joelberkeley-pio: lol

gpflow/optimizers/scipy.py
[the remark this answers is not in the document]
st--: I found the original code hard to parse, particularly because of the re-assignment to the same variable. I've tried to clarify what is happening here by separating the splicing of the right subvector from turning that vector into the fully shaped&typed tensor again. If you have suggestions for further improvements I'd be happy to incorporate them!

st--:
Unit tests fail due to unrelated code in test_freeze which now runs into a recursion error!?

gpflow/optimizers/scipy.py
```diff
@@ -74,12 +76,12 @@ def minimize(
...
 
     @classmethod
-    def initial_parameters(cls, variables: Variables) -> tf.Tensor:
+    def initial_parameters(cls, variables: Iterable[tf.Variable]) -> tf.Tensor:
         return cls.pack_tensors(variables)
 
     @classmethod
     def eval_func(
-        cls, closure: LossClosure, variables: Variables, compile: bool = True
+        cls, closure: LossClosure, variables: Tuple[tf.Variable], compile: bool = True
```
joelberkeley-pio, line 84:
```suggestion
        cls, closure: LossClosure, variables: Tuple[tf.Variable, ...], compile: bool = True
```
I believe `Tuple[tf.Variable]` can only be `(tf.Variable(0.),)`

joelberkeley-pio: and for the others

gpflow/optimizers/scipy.py
```diff
@@ -7,16 +7,16 @@
 
 __all__ = ["Scipy"]
 
-Variables = Iterable[tf.Variable]
-StepCallback = Callable[[int, Variables, List[tf.Tensor]], None]
+Variables = Iterable[tf.Variable]  # deprecated
```
awav, line 10: I'm sorry for being late with a reply. Why is that deprecated instead of replacing `Iterable` with `Sequence`. You could still reuse `Variables` alias in the code and minimize the change. I think it is a sensible thing to do, could you please leave `Variables` alias in the code below (where Sequence[tf.Variable] is used). Thanks!

joelberkeley-pio: i'm also not in favour of changing it now. i've already reviewed it weeks ago and don't want to have to look at it again.

PR #1460, opened 2020-05-08
Update setup.py for tensorflow 2(.2)

st--:
We had some complicated workarounds for TensorFlow CPU vs GPU and gast 0.2.2 vs 0.3.3 (#1348, #1350). Unfortunately, they turned out not to be compatible with `pip install gpflow` now that TensorFlow 2.2 is released - I don't understand WHY the logic isn't working, but it's not; pip seems to believe that gpflow requires gast 0.2.2 (despite the workaround logic!). In any case, the CPU vs GPU workaround is no longer required for TensorFlow 2.x as there is no separate GPU package. The gast workaround should no longer required either because TensorFlow 2.2 is compatible with the latest gast release so the CI should be happy. This PR removes both (and adds black to tests_requirements, which had been missing).

### DIFF
--- a/tests_requirements.txt
+++ b/tests_requirements.txt
@@ -1,3 +1,4 @@
+black
 pytest>=3.5.0
 pytest-random-order
 
--- a/setup.py
+++ b/setup.py
@@ -31,9 +31,7 @@ if sys.version_info < (3, 7):
     requirements.append("dataclasses")
 
 if not on_readthedocs:
-    requirements.extend(
-        ["tensorflow>=2.1.0", "tensorflow-probability>=0.9"]
-    )
+    requirements.extend(["tensorflow>=2.1.0", "tensorflow-probability>=0.9"])
 
 
 def read_file(filename):
--- a/setup.py
+++ b/setup.py
@@ -31,50 +31,9 @@ if sys.version_info < (3, 7):
     requirements.append("dataclasses")
 
 if not on_readthedocs:
-    requirements.append("tensorflow-probability>=0.9")
-
-min_tf_version = "2.1.0"
-tf_cpu = "tensorflow"
-tf_gpu = "tensorflow-gpu"
-
-
-# for latest_version() [see https://github.com/GPflow/GPflow/issues/1348]:
-def latest_version(package_name):
-    import json
-    from urllib import request
-    import re
-
-    url = f"https://pypi.python.org/pypi/{package_name}/json"
-    data = json.load(request.urlopen(url))
-    # filter out rc and beta releases and, more generally, any releases that
-    # do not contain exclusively numbers and dots.
-    versions = [parse_version(v) for v in data["releases"].keys() if re.match("^[0-9.]+$", v)]
-    versions.sort()
-    return versions[-1]  # return latest version
-
-
-# Only detect TF if not installed or outdated. If not, do not do not list as
-# requirement to avoid installing over e.g. tensorflow-gpu
-# To avoid this, rely on importing rather than the package name (like pip).
-
-try:
-    # If tf not installed, import raises ImportError
-    import tensorflow as tf
-
-    if parse_version(tf.__version__) < parse_version(min_tf_version):
-        # TF pre-installed, but below the minimum required version
-        raise DeprecationWarning("TensorFlow version below minimum requirement")
-except (ImportError, DeprecationWarning):
-    # Add TensorFlow to dependencies to trigger installation/update
-    if not on_readthedocs:
-        # Do not add TF if we are installing GPflow on readthedocs
-        requirements.append(tf_cpu)
-        gast_requirement = (
-            "gast>=0.2.2,<0.3"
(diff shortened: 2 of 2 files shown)


tests_requirements.txt
```diff
@@ -1,3 +1,4 @@
+black
```
awav, line 1: Tests per se don't require black. The CI job requires it yes, this is not a test, but format checking.

st--: If you want people to be able to check that tests run locally, they'll need black.

st--:
> What was the problem with gast before? We didn't need that workaround?

Like I wrote before, I don't completely understand the technical details, but there's some difference to whether you do `pip install .` or `python setup.py` in a git clone (for which the workarounds did work) from doing `pip install gpflow` from PyPI (which seems to determine dependencies in some weird way that bypasses the workaround)...

tests_requirements.txt
st--: Or even to run `make format` before creating PRs

st--:
> Do I understand it correctly that this PR enables the installation of gpflow with the latest tensorflow (2.2.0) and gast (0.3.3) automatically using python setup.py develop or pip install -e .? GPflow is still compatible with TF 2.1.0 (and gast 0.2.2) but a user will have to manually install the correct versions of these libraries?

Yes, this PR should in no way affect GPflow's compatibility with TensorFlow 2.1. It will be up to the user to install TensorFlow 2.1 (and gast 0.2.2) correctly if that is what they want. As TensorFlow 2.2 has now been released, which works with the latest version of gast, the standard ">=2.1.0" should work fine. (Obviously we can only confirm this once we've pushed a new release to PyPI...)

tests_requirements.txt
awav: Right, why is that in test dependency?

awav:
I don't see how it helps users of TF2.1 and TF2.2. We allow both versions, but for installing GPflow with TF2.1, the user will have to perform some manual steps. Is this PR fixes one "bug" and introduces another?

tests_requirements.txt
joelberkeley-pio: how about renaming the file to not be specific to tests?

st--: the format-check is a test...

st--: it's required to run the tests (including format check) locally

st--:
> I don't see how it helps users of TF2.1 and TF2.2.

It helps users of TF2.2 because with this fix they can do `pip install gpflow` and get the latest version of TF2.2 and no gast issues.

> We allow both versions, but for installing GPflow with TF2.1, the user will have to perform some manual steps. Is this PR fixes one "bug" and introduces another?

GPflow does not have a direct gast dependency. If a GPflow user runs into any gast dependency issues with tensorflow 2.1, it is their problem. We should not pin gast version in GPflow.

> Also, I cannot reproduce the issue.

I can reproduce the issue by creating a new conda environment:
```bash
conda create -n pip_install_test python=3.8
conda activate pip_install_test
pip install gpflow
```

awav:
> It helps users of TF2.2 because with this fix they can do pip install gpflow and get the latest version of TF2.2 and no gast issues.

As I mentioned before, this PR solves the problem for TF2.2, but creates a problem for TF2.1 users. I would prefer that the problem would go away for both versions. Some users (e.g. prowler.io) that cannot afford to switch TF2.1 as quick as we do in GPflow.

> GPflow does not have a direct gast dependency. If a GPflow user runs into any gast dependency issues with tensorflow 2.1, it is their problem. We should not pin gast version in GPflow.

The same reasoning applies to TF2.2. This dependency has no consequences for our tests - they pass. User can continue work with TF2.2 (e.g. I have been using TF2.2rc1 since its release and had no problems) without bumping to any issue.

I still cannot reproduce the issue. Can it be a side effect of python 3.8?

st--:
>  but creates a problem for TF2.1 users.

@awav please explain *what* problem you anticipate this PR creating for TF 2.1 users.

> The same reasoning applies to TF2.2.

I do not understand what point you are trying to make. Yes, we do not pin TF version, only say >= 2.1 (because of the PR that removed compatibility with 2.0 specifically) - that's as it should be, no?

> (e.g. I have been using TF2.2rc1 since its release and had no problems)

The issue is in how you install gpflow. If you do `pip install -e .` in a git-clone that does not have any issues. It's with installing directly from PyPI.

> I still cannot reproduce the issue. Can it be a side effect of python 3.8?

I have not tried with conda+python3.7, but I can also reproduce the issue with python3.6 and virtualenv:
```bash
python3 -m venv test_venv
source ./test_venv/bin/activate
pip install -U pip setuptools  # the system versions on my Ubuntu 18.04 are too old
pip install gpflow
```
end up installing tensorflow==2.2.0 and gast==0.2.2 which is *incompatible* (e.g., gpflow tests fail).

PR #1464, opened 2020-05-12
bump version to 2.0.3

st--:
We found out that unfortunately `pip install gpflow` did not play well with the tensorflow 2.2 release, as addressed by #1460. Bumping the version for a 2.0.3 minor release.


doc/source/conf.py
```diff
@@ -92,9 +92,9 @@
 # built documents.
 #
 # The short X.Y version.
-version = "2.0.2"
+version = "2.0.3"
 # The full version, including alpha/beta/rc tags.
-release = "2.0.2"
+release = "2.0.3"
```
st--, line 97: The [Sphinx documentation on `version` and `release`](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-version) makes me think these should both be `2.0` (instead of tracking minor version)...

PR #1467, opened 2020-05-13
Sync develop with master after release

vdutor:


st--:
I think you need to do it via a third branch, in which you resolve the conflicts (see https://github.com/GPflow/GPflow/pull/1437)

PR #1470, opened 2020-05-14
Added code to clear cache for the Invert bijector.

yiwei-prowler:
To properly clear cache for the Invert bijector, we need to go into its contained bijector and clear that as well.


vdutor:
Hi @yiwei-prowler many thanks for the PR! How much work would it be to add a unit test to your PR that covers the failure you are fixing here?

yiwei-prowler:
> Hi @yiwei-prowler many thanks for the PR! How much work would it be to add a unit test to your PR that covers the failure you are fixing here?

I'm working on the test.

PR #1471, opened 2020-05-15
enable mypy on CI

joelberkeley-pio:
does [this](https://github.com/GPflow/GPflow/issues/1368#issuecomment-628704285) so as to avoid regressions in types. Mypy will not block merge (for now at least). It should however show a warning

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -19,6 +19,21 @@ runtest: &runtest
 version: 2.1
 
 jobs:
+  type-check:
+    docker:
+      - image: awav/tensorflow:2.1
+
+    steps:
+      - checkout
+      - run:
+          name: Install mypy
+          command: |
+            pip install mypy
+      - run:
+          name: Run type checker
+          command: |
+            mypy .
+
   unit-test:
     <<: *runtest
     docker:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -26,13 +26,11 @@ jobs:
     steps:
       - checkout
       - run:
-          name: Install mypy
-          command: |
-            pip install mypy
+          name: Install type checker
+          command: pip install mypy
       - run:
           name: Run type checker
-          command: |
-            mypy .
+          command: mypy .
 
   unit-test:
     <<: *runtest
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -27,10 +27,12 @@ jobs:
       - checkout
       - run:
           name: Install type checker
-          command: pip install mypy
+          command: |
+            pip install mypy
       - run:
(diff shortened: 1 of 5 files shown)


.circleci/config.yml
```diff
@@ -19,6 +19,21 @@ runtest: &runtest
...
+  type-check:
+    docker:
+      - image: awav/tensorflow:2.1
+
+    steps:
+      - checkout
+      - run:
+          name: Install mypy
+          command: |
+            pip install mypy
```
st--, line 31: Should this not be part of `tests_requirements.txt` instead?

joelberkeley-pio: debatable - I don't consider type checking to be part of the tests, but then again neither would I consider black to be part of the tests and that's in there

won't it will slow down the type checking job to install the other stuff?

st--: I guess it should be "this should _also_ be part of `tests_requirements.txt`". We should make it easy for contributors to run tests locally!

joelberkeley-pio: curiously, black's in test_requirements.txt _and_ installed separately like i've done for mypy

joelberkeley-pio: but the whole thing's moot until I can get the type checking to appear on the checks below

gpflow/base.py
```diff
@@ -107,7 +107,7 @@ def __init__(
 
         self._transform = transform
         self.prior = prior
-        self.prior_on = prior_on
+        self._prior_on = PriorOn(value)
```
joelberkeley-pio, line 110: mypy currently complains about the former code. it's an open ticket in mypy to allow it. for now, i've just changed it

joelberkeley-pio: the alternative is a `# type: ignore`

mypy.ini
```diff
@@ -1,2 +1,14 @@
...
+ignore_errors = True
+
+[mypy-gpflow.kernels.*,gpflow.models.*,gpflow.monitor.*,gpflow.optimizers.*,gpflow.utilities.*]
+ignore_errors = True
+
+[mypy-gpflow.mean_functions]
+ignore_errors = True
+
+[mypy-tests.*]
+ignore_errors = True
```
st--, line 14: these are the modules that haven't been type-fixed yet ?

st--:
Would be nice to have the Makefile contain targets for easily replicating running CI tests locally (`make format-check` etc) without contributors having to delve into configs to figure out what settings to use.

mypy.ini
joelberkeley-pio: exactly

PR #1472, opened 2020-05-15
Temporary fix for TFP version

awav:


setup.py
```diff
@@ -31,7 +31,7 @@
     requirements.append("dataclasses")
 
 if not on_readthedocs:
-    requirements.append("tensorflow-probability>=0.9")
+    requirements.append("tensorflow-probability==0.9")
```
joelberkeley-pio, line 34: does this allow for e.g. 0.9.1?

st--: No. You can use `~=` for that though.

joelberkeley-pio: that'd be good

st--:
```suggestion
    requirements.append("tensorflow-probability~=0.9")
```

awav: There is no 0.9.1 and there will not be such thing :)

st--:
Can we please fix it properly instead? I've found the issue - see #1473

st--:
Closing in favour of actually fixing the issue: #1473 (unit tests pass!)

PR #1473, opened 2020-05-15
Workaround for tensorflow_probability==0.10 in gpflow.utilities.traverse_module

st--:
This fixes a test failure in `test_freeze()` with tensorflow_probability==0.10. Note that this only showed up because of the test using a tf.Module() that *directly* contained a tfp.bijectors.Bijector instance - gpflow.Parameter instances that contain a bijector inside weren't affected as traverse_module() didn't try to peek inside their internals.

Looks like we'll need a 2.0.4 release next week...!

### DIFF
--- a/gpflow/utilities/utilities.py
+++ b/gpflow/utilities/utilities.py
@@ -421,10 +421,10 @@ def traverse_module(
             new_state = traverse_module(subterm, new_acc, update_cb, target_types)
     elif isinstance(m, tf.Module):
         items = vars(m)
-        if isinstance(m, tfp.bijectors.Bijector) and '_parameters' in items:
-            assert items['_parameters']['self'] is m
-            del items['_parameters']['self']  # avoid infinite recursion
-            del items['_parameters']  # fix other bug
+        if isinstance(m, tfp.bijectors.Bijector) and "_parameters" in items:
+            assert items["_parameters"]["self"] is m
+            del items["_parameters"]["self"]  # avoid infinite recursion
+            del items["_parameters"]  # fix other bug
         for name, submodule in vars(m).items():
             if name in tf.Module._TF_MODULE_IGNORED_PROPERTIES:
                 continue
--- a/gpflow/utilities/utilities.py
+++ b/gpflow/utilities/utilities.py
@@ -406,19 +406,17 @@ def traverse_module(
     """
     path, state = acc
 
-    new_state = state
-
     if isinstance(m, target_types):
         return update_cb(m, path, state)
 
     if isinstance(m, (list, tuple)):
         for term_idx, subterm in enumerate(m):
-            new_acc = (f"{path}[{term_idx}]", new_state)
-            new_state = traverse_module(subterm, new_acc, update_cb, target_types)
+            new_acc = (f"{path}[{term_idx}]", state)
+            state = traverse_module(subterm, new_acc, update_cb, target_types)
     elif isinstance(m, dict):
         for term_idx, subterm in m.items():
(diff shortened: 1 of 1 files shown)


gpflow/utilities/utilities.py
```diff
@@ -420,6 +420,11 @@ def traverse_module(
             new_acc = (f"{path}['{term_idx}']", new_state)
             new_state = traverse_module(subterm, new_acc, update_cb, target_types)
     elif isinstance(m, tf.Module):
+        items = vars(m)
+        if isinstance(m, tfp.bijectors.Bijector) and "_parameters" in items:
+            assert items["_parameters"]["self"] is m
+            del items["_parameters"]["self"]  # avoid infinite recursion
```
st--, line 426: This fixes the issue with `leaf_components(bijector)` where `bijector` is any tfp.bijectors.Bijector instance.

gpflow/utilities/utilities.py
```diff
@@ -420,6 +420,11 @@ def traverse_module(
...
+            del items["_parameters"]  # fix other bug
```
st--, line 427: This is needed specifically to make the `test_freeze` happy. We could potentially move this into the `freeze()` below.

gpflow/utilities/utilities.py
```diff
@@ -420,6 +420,11 @@ def traverse_module(
             new_acc = (f"{path}['{term_idx}']", new_state)
             new_state = traverse_module(subterm, new_acc, update_cb, target_types)
     elif isinstance(m, tf.Module):
+        items = vars(m)
+        if isinstance(m, tfp.bijectors.Bijector) and "_parameters" in items:
```
st--, line 424: `_parameters` was added in tensorflow_probability 0.10 release.

st--: https://github.com/tensorflow/probability/commit/a95c160abc2c8ab9f8502817a9cea7062a8103e8

gpflow/utilities/utilities.py
st--: More specifically, the set_attr_by_path doesn't work with the `_parameters` attribute. Of course it's redundant to delete just the `"self"` entry and the whole dictionary - my intent was to highlight that there are two slightly separate issues. We can resolve both in one go by deleting `_parameters` entirely, or we could only delete its `"self"` entry and separately fix the other issue.

condnsdmatters: since `vars` returns the __dict__ of an object and as this isnt used in the line below - we can see its modifying the object. Lets take a copy of the vars object and delete from that instead!

gpflow/utilities/utilities.py
```diff
@@ -420,7 +420,10 @@ def traverse_module(
             new_acc = (f"{path}['{term_idx}']", new_state)
             new_state = traverse_module(subterm, new_acc, update_cb, target_types)
     elif isinstance(m, tf.Module):
-        for name, submodule in vars(m).items():
+        module_vars = vars(m)
+        if isinstance(m, tfp.bijectors.Bijector):
+            module_vars = {name: module_vars[name] for name in module_vars if name != "_parameters"}
```
condnsdmatters, line 425: nit:
I feel it's more idiomatic to use the items iterator, so you dont need to check the dict with the value you are looking up is the same one as your are iterating through

 ```suggestion
             module_vars = {name: var for name, var in module_vars.items() if name != "_parameters"}
```

st--:
Superseded by the decidedly less hacky #1476 :)

PR #1476, opened 2020-05-18
Fix freeze and traverse utility methods

awav:
TensorFlow Probability 0.10 introduces a new `_parameters` attribute in tfp.bijectors.Bijector instances that contains a self-reference, which breaks GPflow's module-traversion utilities (leaf_components, print_summary, deepcopy, freeze). This PR changes `traverse_module` to ignore the `_parameters` attribute and proposes a new deepcopy approach that simply uses copy.deepcopy()'s `memo` argument.

@st-- has this solution #1473. This PR provides a less hacky and improved approach.


doc/source/notebooks/intro_to_gpflow2.pct.py
```diff
@@ -564,3 +563,7 @@ def checkpointing_training_loop(
 
 p = gpflow.Parameter(1.1, transform=gpflow.utilities.positive())
 print(f"{global_str}{p}")
+
+# %%
+
+# %%
```
st--, line 569:
```suggestion
```

gpflow/utilities/utilities.py
```diff
@@ -382,12 +383,9 @@ def freeze(input_module: M) -> M:
...
     """
-    module_copy = deepcopy(input_module)
-    for name, value in parameter_dict(module_copy).items():
-        tensor = tf.convert_to_tensor(value, dtype=value.dtype)
-        const_value = tf.constant(tensor)
-        attr_path = name.lstrip(".")
-        setattr_by_path(module_copy, attr_path, const_value)
+    objects_to_freeze = _get_leaf_components(input_module)
+    memo_tensors = {id(v): tf.constant(tf.convert_to_tensor(v)) for v in objects_to_freeze.values()}
+    module_copy = deepcopy(input_module, memo_tensors)
```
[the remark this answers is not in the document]
awav: Yes, existing tests fit well. This method is how we were supposed to do `deepcopy` in the first place.

doc/source/notebooks/intro_to_gpflow2.pct.py
```diff
@@ -510,17 +510,16 @@ def checkpointing_training_loop(
...
+    frozen_model.predict_f, input_signature=[tf.TensorSpec(shape=[None, 1], dtype=tf.float64)]
 )
 module_to_save.predict = predict_fn
 
 # %% [markdown]
 # Save original result for futher comparison
 
 # %%
-original_result = module_to_save.predict(samples_input)
+sample_input_tensor = tf.convert_to_tensor(samples_input, dtype=default_float())
```
st--, line 521: Could you add a comment explaining why you convert to tensor first ?

gpflow/utilities/utilities.py
```diff
@@ -421,7 +419,13 @@ def traverse_module(
             new_state = traverse_module(subterm, new_acc, update_cb, target_types)
     elif isinstance(m, tf.Module):
         for name, submodule in vars(m).items():
-            if name in tf.Module._TF_MODULE_IGNORED_PROPERTIES:
+            ignore = m._TF_MODULE_IGNORED_PROPERTIES
```
st--, line 422:
```suggestion
            ignored_attributes = m._TF_MODULE_IGNORED_PROPERTIES
```

gpflow/utilities/utilities.py
```diff
@@ -421,7 +419,13 @@ def traverse_module(
...
+            # NOTE(awav): since tfp version 0.10.0, tfp.bijectors.Bijector instances have
+            # `_parameters` dictionary with `self` references that cause
+            # infinite recursive loop.
+            if isinstance(m, tfp.bijectors.Bijector):
+                ignore = ignore.union(["_parameters"])
```
st--, line 427:
```suggestion
                ignored_attributes = ignored_attributes.union({"_parameters"})
```

gpflow/utilities/utilities.py
```diff
@@ -421,7 +419,13 @@ def traverse_module(
...
+            if name in ignore:
```
st--, line 428:
```suggestion
            if name in ignored_attributes:
```

gpflow/utilities/utilities.py
```diff
@@ -421,7 +419,13 @@ def traverse_module(
             new_state = traverse_module(subterm, new_acc, update_cb, target_types)
     elif isinstance(m, tf.Module):
         for name, submodule in vars(m).items():
-            if name in tf.Module._TF_MODULE_IGNORED_PROPERTIES:
+            ignore = m._TF_MODULE_IGNORED_PROPERTIES
+            # NOTE(awav): since tfp version 0.10.0, tfp.bijectors.Bijector instances have
+            # `_parameters` dictionary with `self` references that cause
```
st--, line 424:
```suggestion
            # `_parameters` dictionary with "self" references that cause
```

gpflow/utilities/utilities.py
```diff
@@ -382,12 +383,9 @@ def freeze(input_module: M) -> M:
...
     :return: Returns a frozen deepcopy of an input object.
     """
-    module_copy = deepcopy(input_module)
-    for name, value in parameter_dict(module_copy).items():
-        tensor = tf.convert_to_tensor(value, dtype=value.dtype)
-        const_value = tf.constant(tensor)
-        attr_path = name.lstrip(".")
-        setattr_by_path(module_copy, attr_path, const_value)
+    objects_to_freeze = _get_leaf_components(input_module)
+    memo_tensors = {id(v): tf.constant(tf.convert_to_tensor(v)) for v in objects_to_freeze.values()}
```
st--, line 387:
```suggestion
    memo_tensors = {id(v): tf.convert_to_tensor(v) for v in objects_to_freeze.values()}
```
the tf.constant() doesn't do anything:
```python
a = tf.convert_to_tensor(1.0)
b = tf.constant(a)
assert a is b
```

gpflow/utilities/utilities.py
```diff
@@ -364,15 +364,16 @@ def setattr_by_path(target: object, attr_path: str, value: Any):
...
 
 
-def deepcopy(input_module: M) -> M:
+def deepcopy(input_module: M, memo: Optional[Dict[int, Any]] = None) -> M:
     """
     Returns a deepcopy of the input tf.Module. To do that first resets the caches stored inside each
     tfp.bijectors.Bijector to allow the deepcopy of the tf.Module.
 
     :param input_module: tf.Module including keras.Model, keras.layers.Layer and gpflow.Module.
+    :param memo: Check details for func:`copy.deepcopy` `memo` argument.
```
st--, line 373:
```suggestion
    :param memo: passed through to func:`copy.deepcopy` (see https://docs.python.org/3/library/copy.html).
```
The docstring of the `copy` module and `copy.deepcopy` itself are actually rather unhelpful...

st--:
Can we delete the whole `set_attr_by_path` set of functions now ?

awav:
> Can we delete the whole `set_attr_by_path` set of functions now ?

Yes, I did.

st--:
@awav do we also want to get rid of the helper functions such as `_get_by_name_index`?

gpflow/utilities/utilities.py
```diff
@@ -382,12 +343,9 @@ def freeze(input_module: M) -> M:
     :param input_module: tf.Module or gpflow.Module.
     :return: Returns a frozen deepcopy of an input object.
     """
-    module_copy = deepcopy(input_module)
-    for name, value in parameter_dict(module_copy).items():
-        tensor = tf.convert_to_tensor(value, dtype=value.dtype)
-        const_value = tf.constant(tensor)
-        attr_path = name.lstrip(".")
-        setattr_by_path(module_copy, attr_path, const_value)
+    objects_to_freeze = _get_leaf_components(input_module)
```
st--, line 346:
```suggestion
    objects_to_freeze = leaf_components(input_module)
```
why are there two ways of calling it, anyways?

awav:
> @awav do we also want to get rid of the helper functions such as `_get_by_name_index`?

There is no much use in this method. It is too generic for gpflow after upgrading the deepcopy.

gpflow/utilities/utilities.py
awav: I prefer to leave this call to a private module method. This method was never supposed to be exposed. For some reason it was. Let's discuss it outside of this PR.

st--:
> There is no much use in this method. It is too generic for gpflow after upgrading the deepcopy.

Exactly, but they're still in the code, that's why I asked - to clarify, as you already removed `[gs]et_attr_by_path`, please also remove the now-no-longer-used helper functions :)

doc/source/notebooks/intro_to_gpflow2.pct.py
```diff
@@ -510,17 +510,16 @@ def checkpointing_training_loop(
...
+original_result = module_to_save.predict(sample_input_tensor)
```
awav, line 522:
```suggestion
original_result = module_to_save.predict(samples_input)
```

awav:
> > There is no much use in this method. It is too generic for gpflow after upgrading the deepcopy.
> 
> Exactly, but they're still in the code, that's why I asked - to clarify, as you already removed `[gs]et_attr_by_path`, please also remove the now-no-longer-used helper functions :)

if you have a list, please share it with me.

awav:
> > There is no much use in this method. It is too generic for gpflow after upgrading the deepcopy.
> 
> Exactly, but they're still in the code, that's why I asked - to clarify, as you already removed `[gs]et_attr_by_path`, please also remove the now-no-longer-used helper functions :)

Apologies, I was confused by naming. It is done.

doc/source/notebooks/intro_to_gpflow2.pct.py
awav: I changed it during debugging and found out that using `np.array` with `tf.function` here is a bad example and I left it.

PR #1482, opened 2020-05-21
Makefile quickfix

st--:
#1471 unfortunately introduced a bug to the Makefile (spaces instead of tab for indenting) which makes it unusable. I've also added an extra test for formatting & turned `make test` into a target for all tests.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,6 @@
+BLACK_CONFIG=-t py36 -l 100
+BLACK_TARGETS=gpflow tests doc setup.py
+
 .PHONY: help clean dev-install install package format test
 
 help:
@@ -21,10 +24,15 @@ package:
 	python setup.py bdist
 
 format:
-	black -t py36 -l 100 gpflow tests doc setup.py
+	black $(BLACK_CONFIG) $(BLACK_TARGETS)
+
+format-check:
+	black --check $(BLACK_CONFIG) $(BLACK_TARGETS)
 
 type-check:
-    mypy .
+	mypy .
 
-test:
+pytest:
 	pytest -v --durations=10 tests/
+
+test: format-check type-check pytest
--- a/Makefile
+++ b/Makefile
@@ -1,15 +1,19 @@
 BLACK_CONFIG=-t py36 -l 100
 BLACK_TARGETS=gpflow tests doc setup.py
 
-.PHONY: help clean dev-install install package format test
+.PHONY: help clean dev-install install package format format-check type-check test check-all
 
 help:
 	@echo "The following make targets are available:"
 	@echo "	dev-install		install all dependencies for dev environment and sets a egg link to the project sources"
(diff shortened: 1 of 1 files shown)


joelberkeley-pio:
this can be two PRs - one for the fix, one for the other changes

st--:
> this can be two PRs - one for the fix, one for the other changes

Yes, it could be, but given that it changes 11 lines in total is that really worth the extra effort? How much time does it take to ask for it to be split up, splitting it up, and reviewing two PRs (where one is literally changing a couple of spaces to a single tab) vs just reviewing this one ?

joelberkeley-pio:
> > this can be two PRs - one for the fix, one for the other changes
> 
> Yes, it could be, but given that it changes 11 lines in total is that really worth the extra effort? How much time does it take to ask for it to be split up, splitting it up, and reviewing two PRs (where one is literally changing a couple of spaces to a single tab) vs just reviewing this one ?

yes, yes it is. we can get the fix in basically immediately. I have other things to attend to before I look at the other parts of this PR

Makefile
```diff
@@ -21,10 +24,15 @@ package:
...
+
+format-check:
+	black --check $(BLACK_CONFIG) $(BLACK_TARGETS)
 
 type-check:
-    mypy .
+	mypy .
 
-test:
+pytest:
```
joelberkeley-pio, line 35: i think `test` is better. it expresses what it does, whilst not commit to a particular test runner

joelberkeley-pio: oh, I just saw you've redefined `test` ... one sec

joelberkeley-pio: ok. my opinion, which I hold lightly, is that type checking isn't part of testing. I also don't think formatting is part of testing

Makefile
```diff
@@ -21,10 +24,15 @@ package:
 	python setup.py bdist
 
 format:
-	black -t py36 -l 100 gpflow tests doc setup.py
+	black $(BLACK_CONFIG) $(BLACK_TARGETS)
+
+format-check:
+	black --check $(BLACK_CONFIG) $(BLACK_TARGETS)
```
joelberkeley-pio, line 30:
```suggestion
	$(BLACK_COMMAND)

format-check:
	$(BLACK_COMMAND) --check
```
with
```
BLACK_COMMAND=black -t py36 -l 100 gpflow tests doc setup.py
```

joelberkeley-pio: ^^^ that's an optional suggestion

Makefile
joelberkeley-pio: e.g. `check-all`

Makefile
```diff
@@ -21,10 +24,15 @@ package:
...
 
 type-check:
-    mypy .
+	mypy .
```
joelberkeley-pio, line 33: thanks for fixing this

PR #1485, opened 2020-05-26
Alternative automatic API ref doc generation

JamesALeedham:
### Description

Enhancement. There are pros and cos to this new method. See https://docs.google.com/document/d/1AIV8-nGOAzqzrzcA8mxpcgdGKDKyQ9wpjVweCyFbl1g/edit for a comparison with the current method.

### Minimal working example

To build the docs locally, don't run `generate_module_rst.py`. Instead, follow these instructions: https://github.com/JamesALeedham/GPflow/blob/33a89cb9b4ea23d57160492f27836d29691a4178/doc/README.md

### DIFF
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -50,6 +50,8 @@ clean:
 
 .PHONY: html
 html:
+	rm -rf build
+	rm -rf source/_autosummary
 	$(SPHINXBUILD) -b html $(ALLSPHINXOPTS) $(BUILDDIR)/html
 	@echo
 	@echo "Build finished. The HTML pages are in $(BUILDDIR)/html."
--- a/doc/README.md
+++ b/doc/README.md
@@ -18,7 +18,7 @@ To compile the GPflow documentation locally:
 
 2. Install doc dependencies
    ```
-   pip install sphinx sphinx_rtd_theme numpydoc nbsphinx sphinx_autodoc_typehints ipython jupytext jupyter_client ipywidgets
+   pip install sphinx sphinx_rtd_theme numpydoc nbsphinx sphinx-autodoc-typehints sphinx-autopackagesummary ipython jupytext jupyter_client ipywidgets matplotlib sklearn tensorflow_datasets
    ```
 
 3. Install pandoc
@@ -27,9 +27,7 @@ To compile the GPflow documentation locally:
    ```
    If pandoc does not install via pip, or step 5 does not work, go to pandoc.org/installing.html (the PyPI package depends on the external system-wide installation of pandoc executables)
 
-4. Generate auto-generated files
-   * Notebooks (.ipynb): run `make -C source/notebooks -j 4` (here with 4 parallel threads)
-   * API documentation (.rst): run `python source/generate_module_rst.py`
+4. Generate Jupyter Notebooks (.ipynb): run `make -C source/notebooks -j 4` (here with 4 parallel threads)
 
 5. Compile the documentation
    ```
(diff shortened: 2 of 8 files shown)


st--:
Overall this looks very good - I've noticed that it doesn't show the type annotations in the signature anymore, but we can fix that by moving to proper Sphinx docstrings and explicitly having the `:param argument_name: description` everywhere.

More problematic is that as you point out it does not play nicely with the multiple dispatch. Would it be possible to make use of your approach for most of the modules, and have our custom multipledispatch-handling script for where it's needed?

doc/Makefile
```diff
@@ -50,6 +50,8 @@ clean:
 
 .PHONY: html
 html:
+	rm -rf build
```
condnsdmatters, line 53: do we really need this here? how was it working before?

doc/README.md
```diff
@@ -18,7 +18,7 @@ To compile the GPflow documentation locally:
 
 2. Install doc dependencies
    ```
-   pip install sphinx sphinx_rtd_theme numpydoc nbsphinx sphinx_autodoc_typehints ipython jupytext jupyter_client ipywidgets
+   pip install sphinx sphinx_rtd_theme numpydoc nbsphinx sphinx-autodoc-typehints sphinx-autopackagesummary ipython jupytext jupyter_client ipywidgets matplotlib sklearn tensorflow_datasets
```
condnsdmatters, line 21: do we also need to install all these doc dependencies (`matplotlib`, `tensorflow_datasets`)? isn't there a docs_require.txt or smth we can put here instead? What does RTD use?

doc/source/conf.py
```diff
@@ -46,29 +44,79 @@
...
+    "nbsphinx",  # Integrate Jupyter Notebooks and Sphinx
     "IPython.sphinxext.ipython_console_highlighting",
 ]
 
-set_type_checking_flag = True
-typehints_fully_qualified = False
-always_document_param_types = True
-# autoclass_content = 'both'
+# Mappings for sphinx.ext.intersphinx. Projects have to have Sphinx-generated doc (.inv file)
+intersphinx_mapping = {
```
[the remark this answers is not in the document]
JamesALeedham: Intersphinx seems quite useful to me - it means that when you reference a third party object from Python or Numpy then the documentation automatically links through to the Python/Numpy documentation for that object. It's a nice to have, I guess, rather than essential - but it makes for a nice user experience.

Unfortunately, the .inv file for Tensorflow, the library that gets referenced a lot, is non-functional at the moment, and the original developer is unresponsive. It may be possible to build and host the .inv file for TF ourselves, or we could just remove the TF mapping.

doc/Makefile
JamesALeedham: Don't need this, no... it was just a safety check for me when building locally that each new build was completely splatting the old...

doc/README.md
JamesALeedham: Building the doc locally (even now, irrespective of this PR) fails unless `matplotlib`, `sklearn` and `tensorflow_datasets` are installed - I guess these are required by the Notebooks in the `doc/source/notebooks` directory. So I added them here as it was pretty frustrating.

doc/source/conf.py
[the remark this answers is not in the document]
JamesALeedham: > Personally, I'm not convinced that the individual multiple dispatch functions are really considered public API, since many have leading underscore and the whole point of multiple dispatch is to have one API to get one thing done, even if the implementations differ.

I would agree with this. My original plan was to write some custom words in the `__init__.py` file for each of the modules that uses multiple dispatch summarising what they were actually doing and include it in this PR, but I ran out of time.

It seems a shame to write custom summaries to document an API when a combination of Sphinx and the docstrings should be doing it automatically for us, but I would argue that even now, the current documentation doesn't do a great job of explaining what's going on in the dispatch modules. There's a lot of noise and it's hard to consume.

PR #1486, opened 2020-05-26
Allow users to pass an iterable to set_trainable

joelberkeley-pio:
Allows users to specify an iterable of `tf.Module`s on which to set `tf.Variable`s as trainable, like
```python3
set_trainable([model.q_mu, model.q_sqrt], False)
```

### DIFF
--- a/gpflow/utilities/utilities.py
+++ b/gpflow/utilities/utilities.py
@@ -1,7 +1,7 @@
 import copy
 import re
 from functools import lru_cache
-from typing import Any, Callable, Dict, List, Optional, Tuple, TypeVar, Union
+from typing import Any, Callable, Dict, List, Optional, Tuple, TypeVar, Union, Iterable
 
 import numpy as np
 import tensorflow as tf
@@ -44,12 +44,16 @@ def to_default_float(x):
     return cast(x, dtype=default_float())
 
 
-def set_trainable(model: tf.Module, flag: bool):
+def set_trainable(model: Union[tf.Module, Iterable[tf.Module]], flag: bool) -> None:
     """
-    Set trainable flag for all `tf.Variable`s and `gpflow.Parameter`s in a module.
+    Set trainable flag for all `tf.Variable`s and `gpflow.Parameter`s in a `tf.Module` or collection of
+    of `tf.Module`s.
     """
-    for variable in model.variables:
-        variable._trainable = flag
+    modules = [model] if isinstance(model, tf.Module) else model
+
+    for mod in modules:
+        for variable in mod.variables:
+            variable._trainable = flag
 
 
 def multiple_assign(module: tf.Module, parameters: Dict[str, tf.Tensor]):
--- a/gpflow/utilities/utilities.py
+++ b/gpflow/utilities/utilities.py
@@ -46,7 +46,7 @@ def to_default_float(x):
 
 def set_trainable(model: Union[tf.Module, Iterable[tf.Module]], flag: bool) -> None:
     """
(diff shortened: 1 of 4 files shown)


gpflow/utilities/utilities.py
```diff
@@ -44,12 +44,16 @@ def to_default_float(x):
     return cast(x, dtype=default_float())
 
 
-def set_trainable(model: tf.Module, flag: bool):
+def set_trainable(model: Union[tf.Module, Iterable[tf.Module]], flag: bool) -> None:
```
st--, line 47: One note while I'm looking at this: `flag` seems like a pretty bad argument name (it doesn't say anything), whereas `model` is too specific (it works on any mod*ule*, e.g. a kernel by itself)...

tests/gpflow/utilities/test_utilities.py
```diff
@@ -4,12 +4,52 @@
...
 import gpflow
+from gpflow import Parameter
 from gpflow.config import Config, as_context
 from gpflow.models.util import data_input_to_tensor
-from gpflow.utilities import positive, triangular
+from gpflow.utilities import positive, triangular, set_trainable
 from gpflow.utilities.ops import difference_matrix
 
 
+def _module() -> tf.Module:
```
st--, line 14: the `utilities` structure is too packed as it is - if you're already moving tests & adding more, how about putting them in a `test_set_trainable` file in this directory ?

tests/gpflow/utilities/test_utilities.py
```diff
@@ -4,12 +4,52 @@
...
+    class _Mod(tf.Module):
+        def __init__(self):
+            super().__init__()
+            self.var = tf.Variable(0.0)
+            self.param = Parameter(0.0)
+
+    assert len(_Mod().trainable_variables) == 2
+    assert len(_Mod().variables) == 2
+
+    return _Mod()
```
st--, line 24:
```suggestion
    module = _Mod()
    assert len(module.trainable_variables) == 2
    assert len(module.variables) == 2

    return module
```

gpflow/utilities/utilities.py
joelberkeley-pio: changing names isn't backwards compatible. if you're ok with that I'll change them

tests/gpflow/utilities/test_utilities.py
joelberkeley-pio: doesn't that make it harder to find tests, if they don't follow the same directory structure as the code?

tests/gpflow/utilities/test_utilities.py
joelberkeley-pio: will do

tests/gpflow/utilities/test_utilities.py
st--: I think a file called "test_set_trainable" makes them easier to find than having to traverse through a long file (note that e.g. test_deepcopy is already separate). The key issue is that gpflow.utilities.utilities has too much stuff and should be split - but that's more effort and probably better-suited for a separate PR.

gpflow/utilities/utilities.py
st--: I'd be happy with changing the names, but you don't have to change them in this PR.

I'm just thinking these are awful names to pass by name.

tests/gpflow/utilities/test_utilities.py
joelberkeley-pio: done. btw there are only 8 tests in the test_utilities module

PR #1487, opened 2020-05-26
add type hints: inducing_variables

joelberkeley-pio:
does #1368 for `inducing_variables`

### DIFF
--- a/gpflow/inducing_variables/inducing_variables.py
+++ b/gpflow/inducing_variables/inducing_variables.py
@@ -80,7 +80,7 @@ class Multiscale(InducingPointsBase):
             )  # pragma: no cover
 
     @staticmethod
-    def _cust_square_dist(A: TensorData, B: TensorData, sc: TensorData) -> tf.Tensor:
+    def _cust_square_dist(A: TensorType, B: TensorType, sc: TensorType) -> tf.Tensor:
         """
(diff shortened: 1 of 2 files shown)


gpflow/inducing_variables/inducing_variables.py
```diff
@@ -36,14 +37,14 @@ def __len__(self) -> int:
 
 
 class InducingPointsBase(InducingVariables):
-    def __init__(self, Z, name=None):
+    def __init__(self, Z: TensorData, name: Optional[str] = None):
```
st--, line 40: Why is this one TensorData but in `Multiscale` below it's TensorType ?

gpflow/inducing_variables/inducing_variables.py
```diff
@@ -69,7 +70,7 @@ class Multiscale(InducingPointsBase):
       }
     """
 
-    def __init__(self, Z, scales):
+    def __init__(self, Z: TensorType, scales: TensorType):
```
st--, line 73:
```suggestion
    def __init__(self, Z: TensorData, scales: TensorData):
```
?

joelberkeley-pio: it was because of the `.shape` access on `Z` and `scales`, but I realise I can work round that

PR #1488, opened 2020-05-27
Improve contributing.md

### LINKED ISSUE #1474: The code example in contributing.md uses GPflow 1
The title says it all really. It would be better if it was up to date

st--:
Resolves #1474 and some other minor improvements to contributing.md.

### DIFF
--- a/contributing.md
+++ b/contributing.md
@@ -64,7 +64,7 @@ GPflow is ~97% covered by the testing suite. We expect changes to code to pass t
 
 ## Documentation
 
-GPflow's documentation is not comprehensive, but covers enough to get users started. We expect that new features have documentation that can help others get up to speed. The docs are mostly IPython notebooks (stored in the git repository in jupytext format) that compile into HTML via Sphinx, using nbsphinx.
+GPflow's documentation is not comprehensive, but covers enough to get users started. We expect that new features have documentation that can help others get up to speed. The docs are mostly IPython notebooks (stored in the git repository in Jupytext format) that compile into HTML via Sphinx, using nbsphinx.
 
 ## Version numbering
 
--- a/contributing.md
+++ b/contributing.md
@@ -17,7 +17,7 @@ This file contains notes for potential contributors to GPflow, as well as some n
 
 With GPflow, we aim to make an extensible library for Gaussian processes which makes building complex models easy. In order to do this, we aim to make GPflow a complete library for doing inference and prediction in sophisticated ways (focussing mainly on variational inference) for *single layer models* only. In order to allow more complicated models to be implemented, we also provide functionality for latent / uncertain inputs. Other models, like deep GPs, can be implemented in their own repository by using GPflow as a dependency. We choose to limit the scope deliberately in order to ensure a high-quality codebase.
 
-We welcome contributions to GPflow. If you would like to contribute a feature, please raise discussion via a GitHub issue, to discuss the suitability of the feature within GPflow. If the feature is outside the envisaged scope, we can still link to a separate project in our Readme. <!-- Large features also make it onto the [roadmap](roadmap.md). -->
+We welcome contributions to GPflow. If you would like to contribute a feature, please raise discussion via a GitHub issue, to discuss the suitability of the feature within GPflow. If the feature is outside the envisaged scope, we can still link to a separate project in our Readme.
 
 ### I have this big feature/extension I would like to add...
 
--- a/contributing.md
+++ b/contributing.md
@@ -15,9 +15,9 @@ This file contains notes for potential contributors to GPflow, as well as some n
 
 ## Project scope
 
-With GPflow, we aim to make an extensible library for Gaussian processes which makes building complex models easy. In order to do this, we aim to make GPflow a complete library for doing inference and prediction in sophisticated ways (focussing mainly on variational inference) for *single layer models* only. In order to allow more complicated models to be implemented, we also provide functionality for latent / uncertain inputs. Other models, like deep GPs, can be implemented in their own repository by using GPflow as a dependency. We choose to limit the scope deliberately in order to ensure a high quality codebase.
+With GPflow, we aim to make an extensible library for Gaussian processes which makes building complex models easy. In order to do this, we aim to make GPflow a complete library for doing inference and prediction in sophisticated ways (focussing mainly on variational inference) for *single layer models* only. In order to allow more complicated models to be implemented, we also provide functionality for latent / uncertain inputs. Other models, like deep GPs, can be implemented in their own repository by using GPflow as a dependency. We choose to limit the scope deliberately in order to ensure a high-quality codebase.
 
-We welcome contributions to GPflow. If you would like to contribute a feature, please raise discussion via a GitHub issue, to discuss the suitability of the feature within GPflow. If the feature is outside the envisaged scope, we can still link to a separate project in our Readme. Large features also make it onto the [roadmap](roadmap.md).
+We welcome contributions to GPflow. If you would like to contribute a feature, please raise discussion via a GitHub issue, to discuss the suitability of the feature within GPflow. If the feature is outside the envisaged scope, we can still link to a separate project in our Readme. <!-- Large features also make it onto the [roadmap](roadmap.md). -->
 
 ### I have this big feature/extension I would like to add...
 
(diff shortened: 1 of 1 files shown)


contributing.md
```diff
@@ -15,17 +15,17 @@ This file contains notes for potential contributors to GPflow, as well as some n
...
 
-We welcome contributions to GPflow. If you would like to contribute a feature, please raise discussion via a GitHub issue, to discuss the suitability of the feature within GPflow. If the feature is outside the envisaged scope, we can still link to a separate project in our Readme. Large features also make it onto the [roadmap](roadmap.md).
+We welcome contributions to GPflow. If you would like to contribute a feature, please raise discussion via a GitHub issue, to discuss the suitability of the feature within GPflow. If the feature is outside the envisaged scope, we can still link to a separate project in our Readme. <!-- Large features also make it onto the [roadmap](roadmap.md). -->
```
st--, line 20: We don't currently have a roadmap.md file. Should we have one ?

contributing.md
```diff
@@ -36,74 +36,43 @@ If the code which you are working on does not yet have any documentation, we wou
...
-            the method. It can be either a scalar or NumPy array.
-        """
-        default_name = self.__class__.__name__
-        name = op_name if op_name is not None else default_name
-        session = gpflow.get_default_session()
-        with tf.name_scope(name):
-            incr_tensor = tensor + 1
-        return session.run(incr_tensor)
-```
-
```
st--, line 65: I don't think this example adds enough to warrant the effort of updating it to tensorflow2. Might be better to point to some actual code example inside gpflow that we think showcases our standard particularly well.

contributing.md
```diff
@@ -36,74 +36,43 @@ If the code which you are working on does not yet have any documentation, we wou
...
 ## Documentation
 
-GPflow's documentation is not comprehensive, but covers enough to get users started. We expect that new features have documentation that can help others get up to speed. The docs are mostly IPython notebooks that compile into HTML via Sphinx, using nbsphinx.
+GPflow's documentation is not comprehensive, but covers enough to get users started. We expect that new features have documentation that can help others get up to speed. The docs are mostly IPython notebooks (stored in the git repository in jupytext format) that compile into HTML via Sphinx, using nbsphinx.
 
 ## Version numbering
 
 The main purpose of versioning GPflow is user convenience.
 
 We use the [semantic versioning scheme](https://semver.org/). The semver implies `MAJOR.MINOR.PATCH` version scheme, where `MAJOR` changes when there are incompatibilities in API, `MINOR` means adding functionality without breaking existing API and `PATCH` presumes the code update has backward compatible bug fixes.
```
st--, line 73: I don't think we do (and I don't think we should)...

contributing.md
joelberkeley-pio: i think github's doing just fine for shorter term ideas. would a roadmap be for longer term ideas? I reckon they could be captured in github tickets for now. But I don't know what the longer term research ideas are. What do the others think?

joelberkeley-pio: if we don't go with a roadmap.md, we should remove it from this doc

contributing.md
```diff
@@ -15,17 +15,17 @@ This file contains notes for potential contributors to GPflow, as well as some n
...
 
 ### ...but it won't work without changes to GPflow core?
 
-We aim to have the GPflow core infrastructure be sufficiently extensible and modular to enable a wide range of third-party extensions without having to touch the core of GPflow. The `inducing_variables` module is an example of this to enable interdomain approximations (multiscale inducing features, Fourier features, etc.). If your feature/extension does not work outside of GPflow-core because something is hard-coded, please open an issue to discuss this with us!
+We aim to have the GPflow core infrastructure be sufficiently extensible and modular to enable a wide range of third-party extensions without having to touch the core of GPflow. The `inducing_variables` module is an example of this to enable interdomain approximations (multiscale inducing features, Fourier features, etc.). If your feature/extension does not work outside of GPflow-core because something is hard-coded, please open an issue to discuss this with us! We are happy to discuss and implement changes to the core code that make it easier for you to extend GPflow with a separate package.
```
joelberkeley-pio, line 28: > We are happy to discuss and implement changes to the core code that make it easier for you to extend GPflow with a separate package.

are we definitely happy to implement the changes?

contributing.md
```diff
@@ -36,74 +36,43 @@ If the code which you are working on does not yet have any documentation, we wou
...
 
-
 ## Tests and continuous integration
 
 GPflow is ~97% covered by the testing suite. We expect changes to code to pass these tests, and for new code to be covered by new tests. Currently, tests are run by CircleCI and coverage is reported by codecov. Pull requests should aim to have >97% *patch* coverage (i.e., all the lines that are *changing* should be covered by tests).
 
 ## Documentation
 
-GPflow's documentation is not comprehensive, but covers enough to get users started. We expect that new features have documentation that can help others get up to speed. The docs are mostly IPython notebooks that compile into HTML via Sphinx, using nbsphinx.
+GPflow's documentation is not comprehensive, but covers enough to get users started. We expect that new features have documentation that can help others get up to speed. The docs are mostly IPython notebooks (stored in the git repository in jupytext format) that compile into HTML via Sphinx, using nbsphinx.
```
joelberkeley-pio, line 67:
```suggestion
GPflow's documentation is not comprehensive, but covers enough to get users started. We expect that new features have documentation that can help others get up to speed. The docs are mostly IPython notebooks (stored in the git repository in Jupytext format) that compile into HTML via Sphinx, using nbsphinx.
```

contributing.md
joelberkeley-pio: support semver?

contributing.md
joelberkeley-pio: I agree with removing this example. The code as a whole may speak louder than pointing to a particular example. If we want to champion a certain level of quality, we'd be best off walking our talk.

contributing.md
st--: if we discuss and agree we should do it, yes - I don't want it to seem like it'd be up to the external person to fix gpflow-core to make it more modular&extendible.

contributing.md
st--: What I mean is, it's still code for research, we may find things we want to change that won't exactly be backwards compatible - if we strictly increase the major version, we'll quickly be at gpflow 10.3. Or keep lots of cruft around to maintain backwards compatibility. I'd rather go with something like Django - in general, aim to add a deprecation warning e.g. in GPflow 2.1, and then remove the feature in the following minor version, and occasionally to just say "backwards incompatibilities when upgrading:" in the release notes when it's not worth the effort to support both old and new way.

contributing.md
st--: I think a roadmap would be useful for putting things into perspective, instead of an unsorted list of issues... which is why I didn't delete it outright

contributing.md
joelberkeley-pio: ok. fair enough. i'm happy with either. definitely sth to ask the others about

contributing.md
joelberkeley-pio: oooh ... i read it backwards. I thought you were removing comments

contributing.md
joelberkeley-pio: > if we discuss and agree we should do it

exactly. but this isn't mentioned in the text (the fact that we'd need to agree to it)

contributing.md
joelberkeley-pio: i'm going to resolve this - can put a ticket in the new board: "Do we want to support semver?"

contributing.md
joelberkeley-pio: resolving comment and adding a ticket to the new board

st--: I've removed it from here for now

contributing.md
st--: Then, please, leave a suggestion as to how you'd word it. I don't see the issue you seem to have with this change.

joelberkeley-pio:
```suggestion
We aim to have the GPflow core infrastructure be sufficiently extensible and modular to enable a wide range of third-party extensions without having to touch the core of GPflow. The `inducing_variables` module is an example of this to enable interdomain approximations (multiscale inducing features, Fourier features, etc.). If your feature/extension does not work outside of GPflow-core because something is hard-coded, please open an issue to discuss this with us! We are happy to discuss changes to the core code that make it easier for you to extend GPflow with a separate package.
```

joelberkeley-pio: suggestion: remove "and implement"

PR #1490, opened 2020-06-01
quickfix to enable GPR to be used with variable data

st--:
GPR used the `.shape` attribute instead of calling `tf.shape()` on the X tensor; this prevents using it with Variable data, as demonstrated in the accompanying test.

### DIFF
--- a/tests/gpflow/models/test_gpr.py
+++ b/tests/gpflow/models/test_gpr.py
@@ -15,6 +15,7 @@
 import gpflow
 import numpy as np
 import pytest
+import tensorflow as tf
 from gpflow import set_trainable
 
 rng = np.random.RandomState(0)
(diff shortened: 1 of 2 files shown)


tests/gpflow/models/test_gpr.py
```diff
@@ -45,3 +46,20 @@ def test_non_trainable_model_objective():
...
+    var_data = (tf.Variable(X, shape=[None, input_dim]), tf.Variable(Y, shape=[None, output_dim]))
+    m = gpflow.models.GPR(var_data, gpflow.kernels.SquaredExponential())
+    lml_func = tf.function(m.log_marginal_likelihood)
+    old_lml = lml_func()
+    new_N = 7
+    new_X, new_Y = rng.randn(new_N, input_dim), rng.randn(new_N, output_dim)
+    var_data[0].assign(new_X)
+    var_data[1].assign(new_Y)
+    new_lml = lml_func()
+    assert np.abs((old_lml - new_lml) / (new_lml + old_lml)) > 0.1
```
thevincentadam, line 65: what does this .1 mean?

st--: it means "their relative difference is larger than 10%"

st--: (well, 5% technically, there's a factor of 2 missing)

st--: I'll add a comment clarifying that this is asserting they're *different*

PR #1495, opened 2020-06-04
Supporting preslicing of multi-output combination kernels

vdutor:
A fix and test for an unsupported feature that people (cc @NicolasDurrande) rightfully thought to work out-of-the-box.

### DIFF
--- a/gpflow/kernels/multioutput/kernels.py
+++ b/gpflow/kernels/multioutput/kernels.py
@@ -73,7 +73,9 @@ class MultioutputKernel(Kernel):
         """
         raise NotImplementedError
 
-    def __call__(self, X, X2=None, *, full_cov=False, full_output_cov=True):
+    def __call__(self, X, X2=None, *, full_cov=False, full_output_cov=True, presliced=False):
+        if not presliced:
+            X, X2 = self.slice(X, X2)
         if not full_cov and X2 is not None:
             raise ValueError("Ambiguous inputs: `diagonal` and `y` are not compatible.")
         if not full_cov:
--- a/tests/gpflow/kernels/test_kernels.py
+++ b/tests/gpflow/kernels/test_kernels.py
@@ -541,3 +541,18 @@ def test_latent_kernels():
     assert len(multioutput_kernel_list[0].latent_kernels) == 1
     assert multioutput_kernel_list[1].latent_kernels == tuple(kernel_list)
     assert multioutput_kernel_list[2].latent_kernels == tuple(kernel_list)
+
+
+def test_combination_LMC_kernels():
+    N, D = 100, 3
+    L1, L2, P = 2, 3, 2
+    kernel_list1 = [Linear(active_dims=[1]), SquaredExponential()]
+    assert len(kernel_list1) == L1
+    kernel_list2 = [SquaredExponential(), Linear(), Linear()]
+    assert len(kernel_list2) == L2
+    kernel = (
+        LinearCoregionalization(kernel_list1, np.random.randn(P, L1)) +
+        LinearCoregionalization(kernel_list2, np.random.randn(P, L2))
+    )
+    X = np.random.randn(N, D)
+    assert kernel(X, full_cov=True).shape == [N, P, N, P]


gpflow/kernels/multioutput/kernels.py
```diff
@@ -73,7 +73,9 @@ def K_diag(self, X, full_output_cov=True):
         """
         raise NotImplementedError
 
-    def __call__(self, X, X2=None, *, full_cov=False, full_output_cov=True):
+    def __call__(self, X, X2=None, *, full_cov=False, full_output_cov=True, presliced=False):
```
awav, line 76: Can you make sure that the parent class uses `presliced` option as well?

vdutor: What do you mean? The parent class is `Kernel` which supports preslicing.

gpflow/kernels/multioutput/kernels.py
```diff
@@ -73,7 +73,9 @@ def K_diag(self, X, full_output_cov=True):
...
+        if not presliced:
+            X, X2 = self.slice(X, X2)
         if not full_cov and X2 is not None:
             raise ValueError("Ambiguous inputs: `diagonal` and `y` are not compatible.")
```
st--, line 80: Would you mind updating the error message while we're looking at this?
```suggestion
            raise ValueError("Ambiguous inputs: passing in `X2` is not compatible with `full_cov=False`.")
```

gpflow/kernels/multioutput/kernels.py
st--, line 76: The change from the `full_cov=True` default in the `Kernel` base class to `full_cov=False` here is another reason that makes me think it shouldn't inherit from `Kernel` at all... (out of scope of this PR of course)

tests/gpflow/kernels/test_kernels.py
```diff
@@ -541,3 +541,17 @@ def test_latent_kernels():
...
     assert multioutput_kernel_list[2].latent_kernels == tuple(kernel_list)
+
+
+def test_combination_LMC_kernels():
+    N, D = 100, 3
+    L1, L2, P = 2, 3, 2
+    kernel_list1 = [Linear(active_dims=[1]), SquaredExponential()]
+    assert len(kernel_list1) == L1
+    kernel_list2 = [SquaredExponential(), Linear(), Linear()]
+    assert len(kernel_list2) == L2
```
st--, line 552: A bit [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself)er:
```suggestion
    N, D, P = 100, 3, 2
    kernel_list1 = [Linear(active_dims=[1]), SquaredExponential()]
    L1 = len(kernel_list1)
    kernel_list2 = [SquaredExponential(), Linear(), Linear()]
    L2 = len(kernel_list2)
```

tests/gpflow/kernels/test_kernels.py
```diff
@@ -541,3 +541,17 @@ def test_latent_kernels():
...
+    kernel = LinearCoregionalization(
+        kernel_list1, np.random.randn(P, L1)
+    ) + LinearCoregionalization(kernel_list2, np.random.randn(P, L2))
+    X = np.random.randn(N, D)
+    assert kernel(X, full_cov=True).shape == [N, P, N, P]
```
st--, line 557: Does the sum actually work correctly? Could we add a quick check along the lines of
```suggestion
    k1 = LinearCoregionalization(kernel_list1, np.random.randn(P, L1))
    k2 = LinearCoregionalization(kernel_list2, np.random.randn(P, L2))
    kernel = k1 + k2
    X = np.random.randn(N, D)
    K1 = k1(X, full_cov=True)
    K2 = k2(X, full_cov=True)
    K = kernel(X, full_cov=True)
    assert K.shape == [N, P, N, P]
    np.testing.assert_allclose(K, K1 + K2)
```
?

gpflow/kernels/multioutput/kernels.py
st--: As noted separately, makes me think perhaps one shouldn't be a parent class of the other in the first place...

PR #1496, opened 2020-06-04
give each multipledispatch-registered implementation a unique name

### LINKED ISSUE #1494: Documentation for `_conditional` function incorrect
Check out [this page containing API doc](https://gpflow.readthedocs.io/en/master/gpflow/conditionals/index.html#gpflow-conditionals-conditional) for the `gpflow.conditionals.conditionals._conditional` function (screenshot of key section below). As I understand it, this doc is supposed to say (something like):

- "If you call the `gpflow.conditionals.conditional` function and pass in `object,InducingVariables,Kernel,object` then we'll forward these things to `_conditional`. Below is the doc for this version of `_conditional` ..."
- "If on the other hand you call the `gpflow.conditionals.conditional` function and pass in `object,object,kernel,object` then we'll forward **these** things to `_conditional`. Below is the doc for **this** version of `_conditional` ..."

But the docs for both versions of `_conditional` are the same (it's the second, `object,object,kernel,object`, version). Sphinx has found two methods with the same name and pasted the documentation it's found last for both of them.

The [doc for the 'InducingVariables' version](https://gpflow.readthedocs.io/en/master/_modules/gpflow/conditionals/conditionals.html) is never shown (starts "Single-output GP conditional...")

image

PS. The `_sample_conditional` function is implemented differently, and on the surface at least, seems to generate documentation that makes more sense...

st--:
so that docs are generated correctly. resolves #1494


vdutor (commented):
Naming is unique but not really consistent across different multiple-dispatch methods. Can we follow one convention like, for example, `_name__Type1__Type2`?

awav (commented):
I recommend reconsidering names. At least remove double underscores. Often in names, you cannot encode all combinations.

st--:
@awav would you be satisfied by "simply replace all double-underscores with single-underscores"?
@vdutor what's the minimum that'd satisfy you to approve this PR?

insysion:
I'll be addressing the concerns of @awav and @vdutor week beginning 28th October

vdutor:
@insysion, are you still planning to do this?

insysion (commented):
Leaving a comment to remove me from requested reviews

PR #1498, opened 2020-06-08
Update link to distill article on GPs

agaier:
in the source (for auto generation of documentation)


st--:
Thanks for bringing the broken link to our attention! Interestingly, the old link is _not_ the same as the one you replaced it with - we might want to check what happened to the other one.

PR #1505, opened 2020-06-11
Quadrature Refactoring

gustavocmv:
# Goals
This PR refactors the `ndiagquad` function responsible for performing the quadrature needed to compute expectations related to non-conjugate likelihoods.

The new implementation uses classes to implement quadrature methods. This enables the class to store the points and weights used for quadrature. It also opens the possibility of componentizing the quadrature class inside the likelihood using it, allowing for operations such as training with some quadrature algorithm (e.g. Gauss-Hermite) and predicting with another (e.g. Monte Carlo).

It also clears the quadrature function signature. Instead of expecting lists Fmu and Fvar representing multiple latent GP's, it now takes the tensors as arguments, with the latent GPs stacked in the last dimension.

# Design
The new quadrature class design can be summarized as follows:
- The `GaussianQuadrature` abstract base class, which provides the methods for computing expectations (or log-expections-exp) given points X and weights W, which must be provided by the inheriting class n
- The `NDiagGHQuadrature` class, which inherits from `GaussianQuadrature`, and implements the construction of the points X and weights W with the method
This separation aims to modularize these concepts and provide better extension capacity.

# Backward Compatibility
Due to the migration to a class architecture, the old `gpflow/quadrature.py` file was converted to a submodule in the folder `gpflow/quadrature`, with it's contents moved to `gpflow/quadrature/deprecated.py`. 

For backward compatibility, The `ndiagquad` function was refactored to wrap the new `NDiagGHQuadrature` class. 

The old code was not deleted -- it was moved to `gpflow/quadrature/old.py` and can be accessed with `gpflow.quadrature.ndiagquad_old`. @st-- suggested moving it to the `tests` and build tests to compare both implementations.


gpflow/quadrature/base.py
```diff
@@ -0,0 +1,93 @@
...
+        :return: Array/Tensor with shape [b1, b2, ...., bf, N_quad_points, d'],
+            usually [N, d] or [N, 1]
+        """
+
+        X, W = self._build_X_W(mean, var)
+        if isinstance(fun, Iterable):
+            # Maybe this can be better implemented by stacking [f1(X), f2(X), ...]
+            # and sum-reducing all at once
+            # The problem: there is no garantee that f1(X), f2(X), ...
+            # have comaptible shapes
```
vdutor, line 55:
```suggestion
```
I think doing the reduce_sum per f is perfectly fine.

gpflow/quadrature/base.py
```diff
@@ -0,0 +1,93 @@
...
+import abc
+import tensorflow as tf
+
+from ..base import TensorType
+
+
+class GaussianQuadrature:
+    """
+    Abstract class implementing quadrature methods to compute Gaussian Expectations.  
+    Inhering classes must provide the method _build_X_W to create points and weights
```
vdutor, line 12:
```suggestion
    Inheriting classes must provide the method _build_X_W to create points and weights
```

gpflow/quadrature/gauss_hermite.py
```diff
@@ -0,0 +1,110 @@
...
+    dz = dz / np.sqrt(np.pi)
+    z, dz = z.astype(default_float()), dz.astype(default_float())
+    return tf.convert_to_tensor(z), tf.convert_to_tensor(dz)
+
+
+def list_to_flat_grid(xs: List[TensorType]):
+    """
+    :param xs: List with d rank-1 Tensors, with shapes N1, N2, ..., Nd
+    :return: Tensor with shape [N1*N2*...*Nd, dim] representing the flattened
+        D-dimensional grid built from the input tensors xs
```
vdutor, line 36:
```suggestion
    :return: Tensor with shape [N1*N2*...*Nd, d] representing the flattened
        d-dimensional grid built from the input tensors xs
```

gpflow/quadrature/gauss_hermite.py
```diff
@@ -0,0 +1,110 @@
...
+    """
+    return tf.reshape(tf.stack(tf.meshgrid(*xs), axis=-1), (-1, len(xs)))
+
+
+def reshape_Z_dZ(zs: List[TensorType], dzs: List[TensorType]):
+    """
+    :param zs: List with d rank-1 Tensors, with shapes N1, N2, ..., Nd
```
vdutor, line 43: are N1, ..., Nd all equal to n_gh here? I would assume so given the shape of the return value?

gpflow/quadrature/gauss_hermite.py
```diff
@@ -0,0 +1,110 @@
...
+    return Z, dZ
+
+
+def repeat_as_list(x: TensorType, n: int):
+    """
+    :param x: Array/Tensor to be repeated
+    :param n: Integer with the number of repetitions
+    :return: List of n repetitions of Tensor x
+    """
+    return tf.unstack(tf.repeat(tf.expand_dims(x, axis=0), n, axis=0), axis=0)
```
vdutor, line 59: shouldn't this be equivalent, and a bit easier to read?
```suggestion
    return [x for _ in range(n)]
```

gpflow/quadrature/gauss_hermite.py
```diff
@@ -0,0 +1,110 @@
...
+    def __init__(self, dim: int, n_gh: int):
+        """
+        :param n_gh: number of Gauss-Hermite points, integer
+        :param dim: dimension of the multivariate normal, integer
+        """
+        Z, dZ = ndgh_points_and_weights(dim, n_gh)
+        self.n_gh_total = n_gh ** dim
+        self.Z = tf.convert_to_tensor(Z)
+        self.dZ = tf.convert_to_tensor(dZ)
+        #  Z: [n_gh_total, dim]
```
vdutor, line 85: you could add a `tf.ensure_shape` to enforce this check.

gpflow/quadrature/gauss_hermite.py
gustavocmv: It is, indeed!
I think I tried  something like `[x] * n` when I wrote it for the first time, but it did not work with tensorflow's `tf.function`.
I don't know about the details of it, so I tried to re-do it with only tf-specific code. If your code works with `tf.function` (can't test it right now), the substitution looks good for me.

gpflow/quadrature/gauss_hermite.py
gustavocmv: In the current implementation, yes, they are equal.
However, `reshape_Z_dZ` can also covers grids with different resolution for each latent function, hence why specifying the shapes that way.
In fact, I will fix the docsting for the return type. Instead of:
```
    :returns: points Z, Tensor with shape [n_gh**dim, dim],
        and weights dZ, Tensor with shape [n_gh**dim, 1]
```
It should be
```
    :returns: points Z, Tensor with shape [N1*N2*...*Nd, d],
        and weights dZ, Tensor with shape [N1*N2*...*Nd, 1]
```

vdutor: yes - perfect.

gustavocmv: fixed in a69acda

gpflow/quadrature/gauss_hermite.py
gustavocmv: Done in 2524561

PR #1507, opened 2020-06-12
Allow conditionals to be computed based on pre-computed Kmm and Lm

nfergu:
## PR content:

* [Title](#title)
* [Description](#description)

### Title

Allow conditionals to be computed based on pre-computed `Kmm` and `Lm`

### Description

This PR allows conditionals to be computed based on pre-computed values for `Kmm` and `Lm`. This allows the performance of, for example, prediction to be improved by passing these pre-computed values into the conditionals calculation, instead of calculating these each time. 

I can add some tests if you think this is a reasonable approach.

### DIFF
--- a/gpflow/conditionals/conditionals.py
+++ b/gpflow/conditionals/conditionals.py
@@ -23,7 +23,7 @@ def _conditional(
     q_sqrt=None,
     white=False,
     Kmm=None,
-    Lm=None
+    Lm=None,
 ):
     """
     Single-output GP conditional.
@@ -82,7 +82,7 @@ def _conditional(
     q_sqrt=None,
     white=False,
     Kmm=None,
-    Lm=None
+    Lm=None,
 ):
     """
     Given f, representing the GP at the points X, produce the mean and
@@ -126,7 +126,8 @@ def _conditional(
         Kmm = kernel(X) + eye(tf.shape(X)[-2], value=default_jitter(), dtype=X.dtype)  # [..., M, M]
     Kmn = kernel(X, Xnew)  # [M, ..., N]
     Knn = kernel(Xnew, full_cov=full_cov)  # [..., N] (full_cov = False) or [..., N, N] (True)
-    mean, var = base_conditional(Kmn, Kmm, Knn, f, full_cov=full_cov, q_sqrt=q_sqrt,
-                                 white=white, Lm=Lm)
+    mean, var = base_conditional(
+        Kmn, Kmm, Knn, f, full_cov=full_cov, q_sqrt=q_sqrt, white=white, Lm=Lm
+    )
 
     return mean, var  # [N, R], [N, R] or [R, N, N]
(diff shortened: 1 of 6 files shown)


st--:
Hi @nfergu , you need to run the black formatter (`make format`). I'm mildly concerned about the option of passing in both `Kmm` *and* `Lm` that aren't consistent with each other. I'm a bit more concerned about this only being available in some specific conditionals, not as a whole.

What would it take to actually include this potential speed-up in the code?
Lack of "caching" of tensors is actually a bigger issue.... :/

nfergu:
Thanks for your comments @st-- 

> Hi @nfergu , you need to run the black formatter (`make format`). 

Done!

> I'm mildly concerned about the option of passing in both `Kmm` _and_ `Lm` that aren't consistent with each other. 

This occurred to me also, but I couldn't think of a way of enforcing this without a significant refactor of the conditionals code.

This also relies on the caller of these functions knowing the correct way to calculate `Kmm` and `Lm` (and therefore knowing which function the call will actually be dispatched-to in the end), which also seems not ideal, but again I can't think of a way round this without a major refactor. 

> I'm a bit more concerned about this only being available in some specific conditionals, not as a whole.

I have hopefully addressed this now. I hadn't realised there were so many functions to calculate conditionals! 
 
> What would it take to actually include this potential speed-up in the code?

I was thinking that my model would look something like this:

```
class MySVGP(SVGP):

    def train(self):
        # Do training here...

        q_mu = self.q_mu
        q_sqrt = self.q_sqrt
        mu, var = conditional(
            Xnew,
            self.inducing_variable,
            self.kernel,
            q_mu,
            q_sqrt=q_sqrt,
            full_cov=full_cov,
            white=self.whiten,
            full_output_cov=full_output_cov,
            Kmm=self._Kmm_for_prediction,
            Lm=self._Lm_for_prediction
        )
        # tf.debugging.assert_positive(var)  # We really should make the tests pass with this here
```

I realise all of this is a bit kludgy, and I'm not particularly happy with it, but we need _some_ way of improving the prediction performance and I don't have any better ideas.

> Lack of "caching" of tensors is actually a bigger issue.... :/

Well, ideally TensorFlow would figure out that it doesn't actually need to compute these values each time and just cache them. I know @awav didn't looked into using TensorFlow's graph optimisations but didn't have any luck so far. Hence this PR.

markvdw:
An alternative would be to rely on the multiple dispatch to do this. You could define a custom feature `PrecomputedInducingPoints` or something, and then have a custom conditional that takes precomputed matrices. You can make the conditional do whatever you want, and you wouldn't have to modify GPflow at all.

nfergu:
> An alternative would be to rely on the multiple dispatch to do this. You could define a custom feature `PrecomputedInducingPoints` or something, and then have a custom conditional that takes precomputed matrices. You can make the conditional do whatever you want, and you wouldn't have to modify GPflow at all.

Thanks for your comments @markvdw. If I understand you correctly, I think this would work, but I'm concerned that I'd end up duplicating lots of code.

Thinking about it though, if I changed only the functions in `conditionals/util.py` to accept `Lm`, this would be a very minimal change to GPflow, and would enable me to create a custom conditional without duplicating too much code. I'll give that a try and see how it looks.

nfergu:
I've opened #1514 as a more minimal version of this. Closing this PR

PR #1455, opened 2020-05-05
Heteroskedastic likelihood (WIP)

st--:
Add generic heteroskedastic likelihoods to GPflow, based on top of tensorflow_probability distributions. Work in progress - join the discussion in the #likelihood-design channel on gpflow.slack.com (invite link in README.md).

PR #1508, opened 2020-06-16
Adding heteroskedastic tests

joshuacoales-pio:
This adds some tests to @st-- 's heteroskedastic branch.
These tests ensure that heteroskedastic likelihood with a constant variance, will give the same results as a Gaussian likelihood with the same variance.

3 of these tests are failing and currently skipped.
- `conditional_mean()` and `conditional_variance()` tests are skipped because they are not implemented in Heteroskedastic likelihood, which cause the tests to fail. I've [asked in GPFlow slack](https://gpflow.slack.com/archives/C0118UMH14N/p1591888632015800) as to whether anyone uses these directly.
- `predict_log_density()` test is skipped because it does not return the same result, as it returns the sum over outputs when given multiple outputs. I have also [asked about this in GPFlow slack](https://gpflow.slack.com/archives/C0118UMH14N/p1592230688019100).

### DIFF
--- a/tests/gpflow/likelihoods/test_heteroskedastic_constant_variance.py
+++ b/tests/gpflow/likelihoods/test_heteroskedastic_constant_variance.py
@@ -72,7 +72,7 @@ def test_predict_mean_and_var():
     )
 
 
-@pytest.skip("Conditional mean is not implemented in heteroskedastic likelihood")
+@pytest.mark.skip("Conditional mean is not implemented in heteroskedastic likelihood")
 def test_conditional_mean():
     l1 = gpflow.likelihoods.Gaussian(variance=Data.g_var)
     l2 = HeteroskedasticTFPDistribution(tfp.distributions.Normal)
@@ -82,7 +82,7 @@ def test_conditional_mean():
     )
 
 
-@pytest.skip("Conditional variance is not implemented in heteroskedastic likelihood")
+@pytest.mark.skip("Conditional variance is not implemented in heteroskedastic likelihood")
 def test_conditional_variance():
     l1 = gpflow.likelihoods.Gaussian(variance=Data.g_var)
     l2 = HeteroskedasticTFPDistribution(tfp.distributions.Normal)
@@ -92,7 +92,7 @@ def test_conditional_variance():
     )
 
 
-@pytest.skip("Currently broken as it returns the sum over outputs when given multiple outputs")
+@pytest.mark.skip("Currently broken as it returns the sum over outputs when given multiple outputs")
 def test_predict_log_density():
     l1 = gpflow.likelihoods.Gaussian(variance=Data.g_var)
     l2 = HeteroskedasticTFPDistribution(tfp.distributions.Normal)
--- a/tests/gpflow/likelihoods/test_heteroskedastic_constant_variance.py
+++ b/tests/gpflow/likelihoods/test_heteroskedastic_constant_variance.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import numpy as np
+import pytest
 import tensorflow as tf
 import tensorflow_probability as tfp
 
@@ -71,6 +72,7 @@ def test_predict_mean_and_var():
     )
 
 
+@pytest.skip("Conditional mean is not implemented in heteroskedastic likelihood")
 def test_conditional_mean():
     l1 = gpflow.likelihoods.Gaussian(variance=Data.g_var)
     l2 = HeteroskedasticTFPDistribution(tfp.distributions.Normal)
@@ -80,6 +82,7 @@ def test_conditional_mean():
     )
 
 
+@pytest.skip("Conditional variance is not implemented in heteroskedastic likelihood")
 def test_conditional_variance():
     l1 = gpflow.likelihoods.Gaussian(variance=Data.g_var)
     l2 = HeteroskedasticTFPDistribution(tfp.distributions.Normal)
@@ -89,6 +92,7 @@ def test_conditional_variance():
     )
 
 
+@pytest.skip("Currently broken as it returns the sum over outputs when given multiple outputs")
 def test_predict_log_density():
     l1 = gpflow.likelihoods.Gaussian(variance=Data.g_var)
     l2 = HeteroskedasticTFPDistribution(tfp.distributions.Normal)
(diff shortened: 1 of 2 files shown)


tests/gpflow/likelihoods/test_heteroskedastic_constant_variance.py
```diff
@@ -0,0 +1,102 @@
...
+
+class Data:
+    g_var = 0.345
+    # toData.Y data: shape [N, 1]
+    Y = np.c_[0.5, 0.4, 1.0].T
+    N, _ = Y.shape
+    # single "GP" (for the mean):
+    f_mean = np.c_[0.4, 0.7, 0.9].T
+    f_var = np.c_[0.7, 0.2, 1.3].T
+    equivalent_f2 = np.log(np.sqrt(0.345))
```
st--, line 34:
```suggestion
    equivalent_f2 = np.log(np.sqrt(g_var))
```
We can re-use `g_var` which makes clearer what's happening (and more robust to changing the value above...)

tests/gpflow/likelihoods/test_heteroskedastic_constant_variance.py
```diff
@@ -0,0 +1,102 @@
...
+from gpflow.likelihoods.heteroskedastic import HeteroskedasticTFPDistribution
+
+tf.random.set_seed(99012)
+
+
+class Data:
+    g_var = 0.345
+    # toData.Y data: shape [N, 1]
+    Y = np.c_[0.5, 0.4, 1.0].T
+    N, _ = Y.shape
```
st--, line 30: It doesn't actually matter what the values are (as long as they aren't some "special" number like 0, which could mask bugs):
```suggestion
    rng = np.random.RandomState(123)
    N = 5
    Y = rng.randn(N, 1)
```
this simplifies the code and makes it immediately obvious what the shape is.

tests/gpflow/likelihoods/test_heteroskedastic_constant_variance.py
```diff
@@ -0,0 +1,102 @@
...
+    # single "GP" (for the mean):
+    f_mean = np.c_[0.4, 0.7, 0.9].T
+    f_var = np.c_[0.7, 0.2, 1.3].T
```
st--, line 33: Same as for Y:
```suggestion
    f_mean = rng.randn(N, 1)
    f_var = rng.uniform(0.1, 1.0, (N, 1))  # must be positive
```
already makes clear the shape, only f_var must be positive

tests/gpflow/likelihoods/test_heteroskedastic_constant_variance.py
```diff
@@ -0,0 +1,102 @@
...
+    equivalent_f2 = np.log(np.sqrt(0.345))
+    F2_mean = np.full((N, 1), equivalent_f2)
+    f2_var = np.zeros((N, 1))
+    # stacked N x 2 arraData.Ys
+    F2_mean = np.c_[f_mean, F2_mean]
```
st--, line 38:
```suggestion
    f2_mean = np.full((N, 1), equivalent_f2)
    f2_var = np.zeros((N, 1))
    F2_mean = np.c_[f_mean, f2_mean]  # [N, 2]
```
to avoid overwriting name

tests/gpflow/likelihoods/test_heteroskedastic_constant_variance.py
```diff
@@ -0,0 +1,102 @@
...
+    F2_var = np.c_[f_var, f2_var]
+
+
+def test_log_prob():
+    """
+    heteroskedastic likelihood where the variance parameter is alwaData.Ys constant
```
st--, line 44:
```suggestion
    heteroskedastic likelihood where the variance parameter is always constant
```

tests/gpflow/likelihoods/test_heteroskedastic_constant_variance.py
```diff
@@ -0,0 +1,102 @@
...
+def test_conditional_variance():
+    l1 = gpflow.likelihoods.Gaussian(variance=Data.g_var)
+    l2 = HeteroskedasticTFPDistribution(tfp.distributions.Normal)
+    np.testing.assert_allclose(
+        l1.conditional_variance(Data.f_mean),
+        l2.conditional_variance(Data.F2_mean),
+    )
+
+
+@pytest.mark.skip("Currently broken as it returns the sum over outputs when given multiple outputs")
```
st--, line 95: Do you want to fix the code to make this test pass or merge it as is and I can change the code (and reactivate the test)?

tests/gpflow/likelihoods/test_heteroskedastic_constant_variance.py
joshuacoales-pio: Whoops, that was why I pulled out the g_var constant, and then forgot to use it. Thanks!

tests/gpflow/likelihoods/test_heteroskedastic_constant_variance.py
joshuacoales-pio: Whoops

tests/gpflow/likelihoods/test_heteroskedastic_constant_variance.py
joshuacoales-pio: I think this would be easier for you to fix, than for me to fix. I would rather merge it as is, such that you can change the code and reactivate the test

tests/gpflow/likelihoods/test_heteroskedastic_constant_variance.py
joshuacoales-pio: Sounds good, fixed in 3927efd5cac5cad6d32279e331b640e7142db9d4

tests/gpflow/likelihoods/test_heteroskedastic_constant_variance.py
joshuacoales-pio: Sounds good, fixed in 3927efd5cac5cad6d32279e331b640e7142db9d4

tests/gpflow/likelihoods/test_heteroskedastic_constant_variance.py
joshuacoales-pio: Fixed in 3927efd5cac5cad6d32279e331b640e7142db9d4

tests/gpflow/likelihoods/test_heteroskedastic_constant_variance.py
joshuacoales-pio: Sounds good, done that in 3927efd5cac5cad6d32279e331b640e7142db9d4

joshuacoales-pio:
The type-check issues in circleci are the same issues as seen in the source branch PR: #1455

PR #1511, opened 2020-06-22
updated models.gpr docstring

polaschwoebel:
Updated documentation for models.gpr according to my suggestion in this [issue](https://github.com/GPflow/docs/issues/4).


markvdw:
I made a small edit
- Removed the K in the description of the likelihood. This is mathematically incorrect.
- Added an additional description of the marginal likelihood. Now we have both, which should be extra clear.

If nobody has any comments within the next day or so, I'll squash-merge.

markvdw:
Oops, we found another bug in this. The mean of the likelihood should be f, not zero.

PR #1514, opened 2020-06-23
Allow Lm to be passed-in to conditionals/util.py

nfergu:
## PR content:

* [Title](#title)
* [Description](#description)

### Title

Allow `Lm` to be passed into utility functions for computation of conditionals. 

### Description

This PR allows the utility methods in `conditionals/util.py` to use pre-computed values for `Lm`. This allows the performance of, for example, prediction to be improved by passing these pre-computed values into the conditionals calculation, instead of calculating these each time.

This is a more minimal version of #1507, requiring many fewer changes to GPflow (at the expense of slightly more code in my model -- see the example below).

### Minimal working example

Here is a rough example of how I would use the new `base_conditional` in my model. This improves prediction time by about 10x.

```
class MySVGP(SVGP):

    def train(self):
        # Do training here...

        mu, var = self.compute_conditional(Xnew, full_cov, full_output_cov)

    def compute_conditional(self, Xnew, full_cov, full_output_cov):
        Kmn = Kuf(self.inducing_variable, self.kernel, Xnew)  # [M, N]
        Knn = self.kernel(Xnew, full_cov=full_cov)
        fmean, fvar = base_conditional(
            Kmn, self.Kmm, Knn, self.q_mu, full_cov=full_cov, q_sqrt=self.q_sqrt, white=self.whiten
        )  # [N, R],  [R, N, N] or [N, R]
        return fmean, expand_independent_outputs(fvar, full_cov, full_output_cov)
```


nfergu:
@st-- I've opened this as a more minimal alternative to #1507. This involves very few changes GPflow. What do you think?

st--:
Hi @nfergu, I thought about this and discussed it with @markvdw. In some sense this is a band-aid for a larger underlying issue (and also related to a previous discussion about introducing a separate, decoupled posterior object that allows more efficient predictions.

The code you changed doesn't actually use `Kmm` directly other than to compute the Cholesky: to avoid having to pass in both `Kmm` _and_ `Lm`, with the former not getting used at all, and wondering whether we should ensure consistency between the two, we were thinking it might make more sense, instead, to introduce an extra flag `Kmm_is_cholesky`, defaulting to False, and then
```python
    Lm = Kmm if Kmm_is_cholesky else tf.linalg.cholesky(Kmm)
```

What do you think about that?

nfergu:
Thanks for your comments @st-- 

> Hi @nfergu, I thought about this and discussed it with @markvdw. In some sense this is a band-aid for a larger underlying issue (and also related to a previous discussion about introducing a separate, decoupled posterior object that allows more efficient predictions.

I agree that it is a band-aid. I'm also vaguely aware of the discussion around a separate posterior object. We need a solution for this in the fairly near future though, so it may be that a band-aid is the only realistic solution.  

> 
> The code you changed doesn't actually use `Kmm` directly other than to compute the Cholesky: to avoid having to pass in both `Kmm` _and_ `Lm`, with the former not getting used at all, and wondering whether we should ensure consistency between the two, we were thinking it might make more sense, instead, to introduce an extra flag `Kmm_is_cholesky`, defaulting to False, and then
> 
> ```python
>     Lm = Kmm if Kmm_is_cholesky else tf.linalg.cholesky(Kmm)
> ```
> 
> What do you think about that?

I'm happy with that solution if you are. I guess an alternative could be for the conditional computation function to take a "cholesky calculation function" argument, which is a function that takes `Kmm` as its argument and defaults to `tf.linalg.cholesky(Kmm)`, but which the user could override to just return `Kmm`? I guess this wouldn't actually enforce that `Lm` is consistent with `Kmm` though, since the caller could return whatever they like from this function. 

I don't feel strongly about either approach, so feel free to advise as to which you prefer.

marcoadurno:
Hi @nfergu,

I've been thinking of doing something similar because I have similar requirements (fast predictions). Are you running predictions from a TF saved_model or via aot/tfcompile? 

I'm doing the latter and I'm observing something really odd in term of runtime performance. Basically even if `Lm` is pre-computed the prediction time doesn't drop. I'll try and share a minimum working example this week so that you can have a look as well, just to double check that I'm not doing anything silly. Unless I'm doing something really silly it's most likely a TF issue.

Best Regards,

Marco

nfergu:
Hi @marcoadurno.

> I've been thinking of doing something similar because I have similar requirements (fast predictions). Are you running predictions from a TF saved_model or via aot/tfcompile?

I'm using a TF savedmodel. With this I get around a 10x speed-up by computing `Kmm` and `Lm`. I haven't tried via aot/tfcompile I'm afraid. I'd be happy to look at your example and see if I can spot any problems though.

nfergu:
Superceded by #1528

marcoadurno:
> Hi @marcoadurno.
> 
> > I've been thinking of doing something similar because I have similar requirements (fast predictions). Are you running predictions from a TF saved_model or via aot/tfcompile?
> 
> I'm using a TF savedmodel. With this I get around a 10x speed-up by computing `Kmm` and `Lm`. I haven't tried via aot/tfcompile I'm afraid. I'd be happy to look at your example and see if I can spot any problems though.

Hi @nfergu ,

We have done some further work with aot and identified two possible issues with TensorFlow. 

One is a possible bug in the code-gen step: [41938](https://github.com/tensorflow/tensorflow/issues/41938)
The other is related to the performance of the generated code for `triangular_solve`: [41940](https://github.com/tensorflow/tensorflow/issues/41940)

Regarding [41938](https://github.com/tensorflow/tensorflow/issues/41938) I don't understand how on earth the whole computational graph generated by GpFlow works (we routinely compile it via aot and it works), but we'll first try to get to the bottom of this issue and then we'll see.

Best Regards,

Marco

PR #1518, opened 2020-06-30
Update parameter to be savable

awav:
An attempt to make the custom tensor Parameter savable (loadable) with `tf.saved_model`.


vdutor:
Hey @awav, how would you like to proceed with this?

gpflow/base.py
```diff
@@ -225,125 +190,11 @@ def assign(
...
-    # mechanism, which allows more control over how Variables interact
-    # with ndarrays.
-    __array_priority__ = 100
-
-
-Parameter._OverloadAllOperators()
-tf.register_tensor_conversion_function(Parameter, lambda x, *args, **kwds: x.read_value())
-
 
 def _cast_to_dtype(
```
vdutor, line 199: Didn't TF2.2 resolve the issue with cast - can we simply use tf.case now?

gpflow/base.py
```diff
@@ -103,35 +103,32 @@ def __init__(
         therefore we need a positive constraint and it is natural to use constrained values.
         A prior can be imposed either on the constrained version (default) or on the unconstrained version of the parameter.
         """
-        super().__init__()
+        if transform is None:
+            transform = tfp.bijectors.Identity()
+
+        value = _cast_to_dtype(value, dtype)
+        value_dtype = value.dtype
```
vdutor, line 110: variable `value_dtype` is not used anywhere - maybe remove? Are we certain that `dtype` == `value_dtype` as a result of the `_cast_to_dtype`?

joelberkeley-pio:
these types can be removed if they're not in use any more https://github.com/GPflow/GPflow/pull/1518/files#diff-e2281e00a6a807545fc8c393630c5c36L17-L19

gpflow/base.py
```diff
@@ -225,125 +190,11 @@ def assign(
         :param read_value: if True, will return something which evaluates to the new
             value of the variable; if False will return the assign op.
         """
-        unconstrained_value = self.validate_unconstrained_value(value, self.dtype)
-        return self._unconstrained.assign(
+        unconstrained_value = _validate_unconstrained_value(value, self.transform, self.dtype)
+        return self.unconstrained_variable.assign(
```
vdutor, line 194: The super class already has this implemented. Making use of it makes it more clear what is happening: 
1. validate that the unconstrained value exists.
2. assign it.

```suggestion
        return super().assign(value, use_locking, name, read_value)
```

gpflow/base.py
```diff
@@ -149,31 +146,13 @@ def prior_on(self) -> PriorOn:
...
-    def deref(self) -> "Parameter":
-        return self
-
     @property
     def unconstrained_variable(self) -> tf.Variable:
-        return self._unconstrained
+        return self._pretransformed_input
 
     @property
     def transform(self) -> Optional[Transform]:
```
vdutor, line 154: I understand this is mainly because of legacy, but why are we still using the term `transform`? We could make the interface cleaner if we followed TFP's naming and used the word `bijector` instead. Having two words for the same object makes the code harder to read and understand. This PR seems like the right place and time to update this?

gpflow/utilities/utilities.py
```diff
@@ -190,17 +190,22 @@ def leaf_components(input: tf.Module):
 def _merge_leaf_components(
     input: Dict[str, Union[tf.Variable, tf.Tensor, Parameter]]
 ) -> Dict[str, Union[tf.Variable, tf.Tensor, Parameter]]:
-    input_values = set([value.experimental_ref() for value in input.values()])
+
+    ref_fn = lambda x: (x if isinstance(x, Parameter) else x.ref())
```
vdutor, line 194: why not add a `ref` / `deref` method to `Parameter`? Seems like this is just missing from TFP's `TransformedVariable` API.
```
def ref(self):
    return self.unconstrained.ref()
```

gpflow/base.py
joelberkeley-pio: we still support 2.1

gpflow/utilities/utilities.py
awav: The parameter is not a tensor object and not a trait to the tensor class, it simply a wrapper to a tensor. You cannot make same reference for two different objects. That was one of the reasons why saving didn't work.

gpflow/base.py
awav: I don't like Bijector name, because not every bijector that TFP has is actually a bijection. Transform name has been used since the start of gpflow and I think keeping that name indeed reflects meaning better. And another reason for transform - keep the commit diff minimal.

tests/gpflow/config/test_config.py
```diff
@@ -58,7 +58,8 @@ def test_env_variables(attr_name, value, expected_value):
         assert getattr(config, attr_name) == expected_value
 
 
-@pytest.mark.parametrize("attr_name", set(list(zip(*_env_values))[0]))
+# @pytest.mark.parametrize("attr_name", set(list(zip(*_env_values))[0]))
```
sam-willis, line 61: remove?

gpflow/base.py
sam-willis: I think you're comment is right, the suggestion is wrong (I guess you know that though) - If you're going to do that, you can simply method the function, it'll inherit it from the parent.

I suppose what you're suggesting is
```
unconstrained_value = super().assign(value, use_locking, name, read_value)
return tf.debugging.assert_all_finite(unconstrained_value, message=message)
```
Which makes sense to me.

awav: In that case, the transformation will be performed 2 times. We redefine the method so that we could do extra steps before assigning value without sacrificing performance.

vdutor: Not sure I understand what you mean. The snippet @sam-willis posted seems correct to me?

doc/source/notebooks/understanding/models.pct.py
```diff
@@ -119,10 +119,11 @@
 p.transform.inverse(p)
 
 # %% [markdown]
-# You can also change the `transform` attribute in place:
+# To replace the `transform` of a parameter you need to recreate the parameter with updated transform:
 
 # %%
-m.kernel.kernels[0].variance.transform = tfp.bijectors.Exp()
```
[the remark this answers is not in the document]
awav: You cannot do this, transformation is read-only in TFP.

gpflow/base.py
sam-willis: it will only be performed twice if you call `_validate_unconstrained_value`, which is why I suggested just using `tf.debugging.assert_all_finite`. You could make the message a constant, if you don't want code duplication.

gpflow/utilities/utilities.py
vdutor: Would it not be cleaner to have
```python
def ref(self):
    return self
```
 in order to get rid of `ref_fn` and `deref_fn`? It seems to me that this referencing functionality could be handy elsewhere as well at some point.

gpflow/base.py
sam-willis: I think you can remove `_validate_unconstrained_value`, `_unconstrained_value` and `_constrained_value`, and just check that after either assignment or initialisation the unconstrained value is finite.

awav: I can't apply the Sam's suggestion because it changes the value first and then checks correctness, whereas it should be vice versa.

gpflow/utilities/utilities.py
awav: The `ref()`  belongs to tensor API, and `ref()` returns a different type. There must be a distinction between these two for clarity. I would prefer not to introduce this method to the parameter class because of a printing issue.

tests/gpflow/config/test_config.py
awav:
```suggestion
```

gpflow/base.py
sam-willis: ah I do see your point, maybe it's better to not change the value if it's invalid, although an exception is thrown, so I'm not sure how important it is. I guess if you catch the exception it makes recovering easier

sam-willis: `_constrained_value` is unused, and technically you should always have the identity for the now prior, so you don't need the if in `_unconstrained_value`, but I think those are very minor.

awav:
@vdutor, @sam-willis I added a test.

tests/gpflow/test_base.py
```diff
@@ -55,3 +56,24 @@ def test_cast_to_dtype_precision_issue():
     assert actual_value.dtype == np.float64
     expected_value = np.float64(0.2)
     assert actual_value == expected_value
+
+
+def test_parameter_saved():
```
vdutor, line 61: can you also check that the number of trainable parameters is the same in the original and reloaded model. I've had issues with that before.

awav: The exec returns the same result. If it were a different parameter the result would be different? Or do you mean something else?

vdutor: If you run `model.variables` on both models do you get the same number of parameters back? There used to be an issue that certain variables / parameters were dropped even though the execution of `exec` gave the same result.

awav: As far as I understand, `tf.save_model` saves models for deployment purposes only (https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/saved_model/README.md#builder). `m1` can be a different object that wraps graph computation.

awav: Keras uses `tf.save_model` format in a different way. Standard `tf.save_model` return not the same object that you pass in at saving time.

```python
> m1
<tensorflow.python.saved_model.load.Loader._recreate_base_user_object.<locals>._UserObject object at 0x13f553fd0>

> m1.variables
AttributeError: '_UserObject' object has no attribute 'variables'
```

Issue #1523, opened 2020-07-03
separate_independent_conditional fails if q_sqrt=None


aidanscannell:
Hello, I am using the conditionals method in a model that I am developing and I think that there is a bug, or at least some functionality missing which can easily be included.

In my model I need to calculate a sparse GP conditional by sampling the inducing output distribution q_samples~N(q_mu, q_sqrt q_sqrt^T) instead of analytically integrating them. In order to achieve this (given `conditional(Xnew, inducing_variable, kernel, f, full_cov, full_output_cov, q_sqrt, white)`) I set `f=q_sample` (a single sample as I map over the samples) and `q_sqrt=None`, which I think is the correct way to achieve this?

Using a separate independent multi-output kernel (and shared independent inducing variables) produces an error in `separate_independent_conditional(Xnew, inducing_variable, kernel, f, full_cov, full_output_cov, q_sqrt, white)` because `q_sqrt=None` is not handled.

## To reproduce

Here is some code to reproduce the error. I do not sample the inducing outputs here but instead simply pass an input and set `q_sqrt=None`.
```python
import numpy as np
import gpflow as gpf

def func(X):
    y1 = np.sin(X.flatten())
    y2 = np.cos(X.flatten())
    return np.stack([y1, y2]).T

num_inducing = 30
num_data = 100
num_test = 200
input_dim = 1

X = np.linspace(0, 3, num_data).reshape(num_data, input_dim)
Y = func(X)
output_dim = Y.shape[1]

# initialise SVGP with separate independent MOK and shared independent inducing variables
idx = np.random.choice(range(num_data), size=num_inducing, replace=False)
inducing_inputs = X[idx, ...].reshape(-1, input_dim)
inducing_variable = gpf.inducing_variables.SharedIndependentInducingVariables(
    gpf.inducing_variables.InducingPoints(inducing_inputs))
kern_list = [gpf.kernels.RBF() for _ in range(output_dim)]
kernel = gpf.kernels.SeparateIndependent(kern_list)
model = gpf.models.SVGP(kernel,
                        gpf.likelihoods.Gaussian(),
                        inducing_variable=inducing_variable,
                        num_latent_gps=output_dim)
gpf.utilities.print_summary(model)

# create test inputs
Xnew = np.linspace(-1, 4, num_test).reshape(num_test, input_dim)

# evaluate sparse gp conditional with q_sqrt=None
mu, var = gpf.conditionals.conditional(Xnew,
                                       model.inducing_variable,
[...]

PR #1533, opened 2020-07-20
Fixed separate_independent_conditional to correctly handle q_sqrt=None.

aidanscannell:
This PR adds a minor change to `separate_independent_conditional` to prevent it from raising the AttributeError,
`AttributeError: 'NoneType' object has no attribute 'shape'.` when `q_sqrt=None`. Previously if q_sqrt=None the method would fail as it attempt to access the shape of q_sqrt and transpose it for mapping through `base_conditional`.

This PR checks if `q_sqrt=None` and does not map it through `base_conditional` if so.

**PR type:** bugfix

**Related issue(s)/PRs:** #1523 
Please see [this issue](https://github.com/GPflow/GPflow/issues/1523) for more details on why a user may want to set `q_sqrt=None` and for the stack trace etc.

## Summary

**Proposed changes**

* Added an if statement to check if q_sqrt is None.
* If q_sqrt=None, the method no longer attempts to transpose q_sqrt and map it through single_gp_conditional.
* Instead, it removes q_sqrt from the tuple of arguments to be mapped through single_gp_conditinal.

Creating a correctly sized Tensor for q_sqrt containing Nones that can be mapped through base_conditional. I do not think this approach is good as  it would lead to unnecessary mapping of q_sqrt.

### Minimal working example

```python
import numpy as np
import gpflow as gpf


def func(X):
    y1 = np.sin(X.flatten())
    y2 = np.cos(X.flatten())
    return np.stack([y1, y2]).T


num_inducing = 30
num_data = 100
num_test = 200
input_dim = 1

X = np.linspace(0, 3, num_data).reshape(num_data, input_dim)
Y = func(X)
output_dim = Y.shape[1]

# initialise SVGP with separate independent MOK and shared independent inducing variables
idx = np.random.choice(range(num_data), size=num_inducing, replace=False)
inducing_inputs = X[idx, ...].reshape(-1, input_dim)
inducing_variable = gpf.inducing_variables.SharedIndependentInducingVariables(
    gpf.inducing_variables.InducingPoints(inducing_inputs))
kern_list = [gpf.kernels.RBF() for _ in range(output_dim)]
kernel = gpf.kernels.SeparateIndependent(kern_list)
model = gpf.models.SVGP(kernel,
                        gpf.likelihoods.Gaussian(),
                        inducing_variable=inducing_variable,
                        num_latent_gps=output_dim)
gpf.utilities.print_summary(model)

# create test inputs
Xnew = np.linspace(-1, 4, num_test).reshape(num_test, input_dim)

# evaluate sparse gp conditional with q_sqrt=None
mu, var = gpf.conditionals.conditional(Xnew,
                                       model.inducing_variable,
                                       model.kernel,
                                       model.q_mu,
                                       full_cov=False,
                                       full_output_cov=False,
                                       q_sqrt=None,
                                       white=model.whiten)

```
Please see corresponding Github issue for more details.

## PR checklist

- [ ] New features: code is well-documented - N/A
  - [ ] detailed docstrings (API documentation) - N/A
  - [ ] notebook examples (usage demonstration) - N/A
- [ ] Code has type annotations - I have not added type annotations to the method as they were not previously there.

### Release notes

Fixed separate_independent_conditional to correctly handle q_sqrt=None.
Previously if q_sqrt=None the method would fail with
AttributeError: 'NoneType' object has no attribute 'shape'.

Added if statement to check is q_sqrt is None.
If q_sqrt=None, method does not attempt to transpose q_sqrt and
map it through single_gp_conditional.


st--:
Hi @aidanscannell, thanks for contributing this fix! It looks good - ideally we'd also include a regression test (i.e., one that fails with the bug you discovered *before* your fix, and passes *with* the fix) to ensure we can't accidentally run into the same bug again in the future: would you be up for writing that as well? Happy to discuss with you on gpflow.slack.com (invite link in README.md) if you've got any questions. If not I'll try and add the test later this week, but it'd be a great help if you could help with that too :)

aidanscannell:
Hi @st--, yes I am happy to write a test. Please could you link an example that I can follow? Cheers, Aidan.

st--:
@aidanscannell- you should be able to simply copy the failing example from https://github.com/GPflow/GPflow/issues/1523 into a new test, you can add it to tests/gpflow/conditionals/test_multioutput.py. You can find an example in https://github.com/GPflow/GPflow/pull/1338: compare the minimal reproducible example for demonstrating the bug in the PR description to the regression test added to the code. Let me know if that's sufficient or if there's anything else I can help with!

aidanscannell:
Hi @st--, thanks for the help. I have added a test in conditionals/test_multioutput.py which I have tested with make check-all. It fails in develop and passes in aidanscannell/fix-separate_independent_conditional. Please let me know if there is anything else that needs to be done.

PR #1536, opened 2020-07-27
check dependency versions are valid on CI

### LINKED ISSUE #1532: Check deps on CI
`pip install gpflow` currently installs dependencies (setuptools, scipy) with versions that are incompatible with the tensorflow version installed.

This ticket isn't to fix the dependencies, per se, but suggests adding a `pip check -vvv` stage to CI, so that such problems are caught at PR stage.

joelberkeley-pio:
fixes #1532

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -27,7 +27,13 @@ if sys.version_info < (3, 7):
 # use autodoc_mock_imports instead. Hence we use this flag to decide whether or
 # not to append tensorflow and tensorflow_probability to the requirements:
 if os.environ.get("READTHEDOCS") != "True":
-    requirements.extend(["tensorflow>=2.1.0,<2.3", "tensorflow-probability>=0.9,<0.11"])
+    requirements.extend(
+        [
+            "tensorflow>=2.1.0,<2.3",
+            "tensorflow-probability>=0.9,<0.11",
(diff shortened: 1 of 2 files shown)


joelberkeley-pio:
> Is the dependency check executed with `READTHEDOCS` set?

pretty sure it won't. see https://docs.readthedocs.io/en/stable/faq.html#how-do-i-change-behavior-when-building-with-read-the-docs

setup.py
```diff
@@ -12,11 +12,10 @@
 # Dependencies of GPflow
 requirements = [
     "numpy>=1.10.0",
-    "scipy>=0.18.0",
+    "scipy>=0.18.0,==1.4.1",  # pinned to ==1.4.1 to satisfy tensorflow requirements
```
insysion, line 15: Johnny come lately here but we should remove the `>=0.18.0` constraint if we're going to pin it anyway. (Just to double check, is that pin strictly necessary?)

joelberkeley-pio: i left it in cos they're pinned for different reasons. i don't know the reason for `>=0.18.0`

joelberkeley-pio: if I we remove `>=0.18.0`, then down the line remove `==1.4.1` for some reason, we now have no constraint, while what we really want is `>=1.18.0`

PR #1542, opened 2020-08-09
Change shape of quadrature tensors for better broadcasting

gustavocmv:
**PR type:** Enhancement

**Related issue(s)/PRs:**  #1505

## Summary

**Proposed changes**

PR #1505 introduced a new approach for quadrature in the form of classes. In the original code, the penultimate dimension was used to broadcast with the quadrature points, resulting in applying the "quadratured" function f over tensors with shape `[N1, ..., Nb, N_gh_points, Dim]`. This resulted in awkward broadcasting  of the quadrature-tensors with other tensors. To change it, the quadrature code was modified to use  shape `[ N_gh_points, N1, ..., Nb, Dim]`

**What alternatives have you considered?**

I wanted to have it working that way when I worked with #1505, but couldn't make it work.

## PR checklist

- [ ] New features: code is well-documented
  - [ ] detailed docstrings (API documentation)
  - [ ] notebook examples (usage demonstration)
- [ ] The bug case / new feature is covered by unit tests
- [X] Code has type annotations
- [X] I ran the black formatter (`make format`)
- [X] I locally tested that the tests pass (`make check-all`)

PR #1549, opened 2020-08-26
Quadrature Likelihood + ScalarLikelihood Refactoring

gustavocmv:
**PR type:** Enhancement

**Related issue(s)/PRs:** #1505 , #1542 

## Summary

**Proposed changes**
This PR refactors the `ScalarLikelihood` class to make it work with the new quadrature class `NDiagGHQuadature`. This is done by inheriting from the `QuadratureLikelihood` class, which implements the general quadrature routines needed by the `Likelihood` classes by componentizing an `NDiagGHQuadature` object.

The methos of `ScalarLikelihood` where adapted following the idea of @markvdw from a discussion in the slack channel to make it possible to defer the initialization of the quadrature class until a method which passes data like `Fmu`, `Fvar` is called.

Quoting the discussion:

> Shouldn’t the observation_dim only need to be rebuilt once when the data is set? This doesn’t seem like much of a problem, right?
> It’s not nice for the user to have to specify redundant information. The observation_dim is redundant since it’s implied by the data. This used to be the case in kernels (only in the input_dim), and I’m happy we removed it.
> I think it’s ok that if you train a model on a particular dataset, that it fails when you run/test it on a dataset with a different dimensionality

The actual code uses the `ndiagquad` wrapper around  `NDiagGHQuadature` to implement the quadratures of `ScalarLikelihood`. This forces the method to re-initialize `NDiagGHQuadature` whenever it is called, resulting in a performance penalty (theoretically, I did not check explicitly).

## PR checklist

- [x] New features: code is well-documented

### Release notes


markvdw:
Hi @gustavocmv. This looks good but to properly review it I need to understand some context. What problem is this refactor trying to solve? I'm also happy to discuss this interactively on the Slack channel, if that's easier.

gustavocmv:
> Hi @gustavocmv. This looks good but to properly review it I need to understand some context. What problem is this refactor trying to solve? I'm also happy to discuss this interactively on the Slack channel, if that's easier.

Hi @markvdw , thank you for your reply! I will copy some parts of our Slack discussion so that @vdutor, @st-- and others can better understand the PR proposal:

----------------------------------

This PR is part of a series of works which began with some discussions between me and @stj regarding more complex likelihoods. The main target was heterocedastic likelihoods, but we concluded that the old `ndiadquad`  function was a bit clunky to work with. We discussed a re-design using classes, which was implemented in PR #1505 and enhanced in PR #1542. To keep backwards compatibility, `ndiagquad` was refactored as a wrapper around the new class NDiagGHQuadature.

The new PR #1549 tries to eliminate the dependency on the wrapper `ndiagquad`, which is mainly used in the ScalarLikelihood class. I refactored that class to use a QuadratureLikelihood class, which directly interacts with the `NDiagGHQuadature` through componentization. One of the key features of the QuadratureLikelihood class is to possibly infer the quadrature dimensions for `NDiagGHQuadature`  from the data and initialize it only once. This componentization approach saves the computational cost of always allocating the Points and Weights needed for quadrature methods, which was done by the old `ndiagquad` and also is the case with the wrapper running now.

Hope it clarifies the intention of the PR, which can be summarized as:
- Exposing the quadrature class NDiagGHQuadature through QuadratureLikelihood
- Eliminating the ndiaquad wrapper from the codebase by refactoring ScalarLikelihood
- Hopefully (I did not benchmark) enjoy some performance benefits

gpflow/quadrature/gauss_hermite.py
```diff
@@ -78,12 +78,15 @@ def __init__(self, dim: int, n_gh: int):
...
         """
+        print("PRINT - Initializing NDiagGHQuadrature")
+        tf.print("TF PRINT - Initializing NDiagGHQuadrature")
         self.dim = dim
         self.n_gh = n_gh
         self.n_gh_total = n_gh ** dim
-        Z, dZ = ndgh_points_and_weights(self.dim, self.n_gh)
-        self.Z = tf.ensure_shape(Z, (self.n_gh_total, self.dim))
-        self.dZ = tf.ensure_shape(dZ, (self.n_gh_total, 1))
+        with tf.init_scope():
```
st--, line 86: I've looked at the [docs for `tf.init_scope`](https://www.tensorflow.org/api_docs/python/tf/init_scope) but am still not quite sure why this is needed here - could you briefly explain (perhaps in a code comment) what motivated this introduction?

gpflow/quadrature/gauss_hermite.py
```diff
@@ -78,12 +78,15 @@ def __init__(self, dim: int, n_gh: int):
         :param n_gh: number of Gauss-Hermite points, integer
         :param dim: dimension of the multivariate normal, integer
         """
+        print("PRINT - Initializing NDiagGHQuadrature")
+        tf.print("TF PRINT - Initializing NDiagGHQuadrature")
```
st--, line 82: Good for testing & developing, but shouldn't be part of the release
```suggestion
```

gpflow/likelihoods/base.py
```diff
@@ -286,7 +286,71 @@ def _variational_expectations(self, Fmu, Fvar, Y):
...
+        if self._quadrature is None:
+            if self.latent_dim is None:
+                raise Exception(
+                    "latent_dim not specified. "
+                    "Either set likelihood.latent_dim directly or "
+                    "call a method which passes data to have it inferred."
+                )
+            with tf.init_scope():
+                print("PRINT - Initializing NDiagGHQuadrature")
+                tf.print("TF PRINT - Initializing NDiagGHQuadrature")
```
st--, line 348:
```suggestion
```
Good for testing & developing, but shouldn't be part of the release

gpflow/likelihoods/base.py
```diff
@@ -286,7 +286,71 @@ def _variational_expectations(self, Fmu, Fvar, Y):
...
+    def __init__(self, n_gh: int = 20, **kwargs):
+        super().__init__(**kwargs)
+        self.n_gh = n_gh
+        self._quadrature = None
+
+    @property
+    def quadrature(self):
+        if self._quadrature is None:
+            if self.latent_dim is None:
+                raise Exception(
```
st--, line 341: Should/could this be a more specific exception? Maybe a custom one?

gpflow/likelihoods/base.py
```diff
@@ -286,7 +286,71 @@ def _variational_expectations(self, Fmu, Fvar, Y):
...
+        :param Y: observation Tensor, with shape [..., observation_dim]:
+        :returns: log predictive density, with shape [...]
+        """
+        return tf.squeeze(self.quadrature(self._quadrature_log_prob, Fmu, Fvar, Y), -1)
+
+
+class NDiagGHQuadratureLikelihood(QuadratureLikelihood):
+    def __init__(self, n_gh: int = 20, **kwargs):
+        super().__init__(**kwargs)
+        self.n_gh = n_gh
```
st--, line 334: it was previously called `num_gauss_hermite_points`, how about keeping that name to retain backwards compatibility? (Could then potentially have a property-setter that re-creates the NDiagGH object)

gpflow/likelihoods/base.py
```diff
@@ -286,7 +286,71 @@ def _variational_expectations(self, Fmu, Fvar, Y):
         raise NotImplementedError
 
 
-class ScalarLikelihood(Likelihood):
+class QuadratureLikelihood(Likelihood):
```
st--, line 289: Is the code in here specific to how the ScalarLikelihood (now-sub)class works? would it work with another type of likelihood as well (e.g. Heteroskedastic)?

st-- (commented):
Some initial comments/questions (still need to finish reviewing the base.py changes)

gpflow/quadrature/gauss_hermite.py
gustavocmv: Removed.

gpflow/likelihoods/base.py
gustavocmv: Removed.

gpflow/likelihoods/base.py
gustavocmv: Do you mean implementing a custom gpflow error? I think it would be a better choice, indeed. To be honest, I've never done something like this, some guidance would be appreciated :).

gpflow/likelihoods/base.py
gustavocmv: I changed the name for backwards compatibility.

The idea of using a property-setter is very interesting, but it should be built involving both the `num_gauss_hermite_points` and the `num_latent` properties. Maybe we can discuss it in detail in Slack?

gpflow/likelihoods/base.py
gustavocmv: The `QuadratureLikelihood` class is an abstract class intended to componentize (any) quadrature object through the implementation of the property `quadrature`, as is done in `NDiagGHQuadratureLikelihood`. It serves the purpose of implementing the 3 main expectation-related methods for likelihoods using the componentized quadrature object, i.e.:
- `_predict_mean_and_var`
- `_predict_log_density`
- `_variational_expectations`
This is, in a way, what `ScalarLikelihood` implemented through `ndiagquad`.

My intention with the QuadratureLikelihood, however, is to have it working with "any" other likelihoods, but I did not try it yet.

gpflow/likelihoods/base.py
```diff
@@ -306,17 +368,12 @@ class ScalarLikelihood(Likelihood):
...
-
-    def _check_last_dims_valid(self, F, Y):
-        """
-        Assert that the dimensions of the latent functions and the data are compatible
-        :param F: function evaluation Tensor, with shape [..., latent_dim]
-        :param Y: observation Tensor, with shape [..., latent_dim]
-        """
-        tf.debugging.assert_shapes([(F, (..., "num_latent")), (Y, (..., "num_latent"))])
+    def __init__(
+        self, latent_dim: int = None, observation_dim: int = None, n_gh: int = 20, **kwargs
```
vdutor, line 372:
```suggestion
        self, latent_dim: Optional[int] = None, observation_dim: Optional[int] = None, n_gh: int = 20, **kwargs
```
```suggestion
        self, latent_dim: int = None, observation_dim: int = None, n_gh: int = 20, **kwargs
```

gpflow/likelihoods/base.py
```diff
@@ -340,10 +397,10 @@ def _variational_expectations(self, Fmu, Fvar, Y):
         :param Y: observation Tensor, with shape [..., latent_dim]:
         :returns: variational expectations, with shape [...]
         """
-        return tf.reduce_sum(
-            ndiagquad(self._scalar_log_prob, self.num_gauss_hermite_points, Fmu, Fvar, Y=Y),
-            axis=-1,
-        )
+        if self.latent_dim is None:
+            self.latent_dim = int(Fmu.shape[-1])
+            self.observation_dim = self.latent_dim
```
vdutor, line 402: Following the DRY-principle it would make sense to add a `_set_latent_and_observation_dimension_eagerly(self, Fmu)` function to this class.

gpflow/likelihoods/base.py
```diff
@@ -340,10 +397,10 @@ def _variational_expectations(self, Fmu, Fvar, Y):
         :param Y: observation Tensor, with shape [..., latent_dim]:
         :returns: variational expectations, with shape [...]
         """
-        return tf.reduce_sum(
-            ndiagquad(self._scalar_log_prob, self.num_gauss_hermite_points, Fmu, Fvar, Y=Y),
-            axis=-1,
-        )
+        if self.latent_dim is None:
+            self.latent_dim = int(Fmu.shape[-1])
```
vdutor, line 401: As you know, this will only work in eager mode. Would the `NDiagGHQuadrature(latent_dim,...)` also work when `latent_dim` is obtained using `tf.shape(Fmu)[-1]`?

gpflow/likelihoods/base.py
vdutor: I think we can simplify the design a little bit (and introduce less classes) by simply having quadrature default to the implementation in `NDiagGHQuadratureLikelihood`. If people want a special/another type of quadrature, they can always subclass and overwrite the `quadrature` property.

gpflow/likelihoods/base.py
```diff
@@ -286,7 +286,69 @@ def _variational_expectations(self, Fmu, Fvar, Y):
         raise NotImplementedError
 
 
-class ScalarLikelihood(Likelihood):
+class QuadratureLikelihood(Likelihood):
+    @property
+    def quadrature(self):
```
vdutor, line 291: I think it would be useful to add type annotations to this function.

gpflow/likelihoods/base.py
vdutor: In this PR I wouldn't introduce custom exceptions as AFAIK we don't have them anywhere else in the gpflow code base. If we want custom exceptions we should probably discuss this more generally and create a separate PR for it.

gpflow/likelihoods/base.py
gustavocmv: Done!

gpflow/likelihoods/base.py
gustavocmv: Not sure. Need to test it.

gpflow/likelihoods/base.py
gustavocmv: Done!

gpflow/likelihoods/base.py
gustavocmv: Changed `n_gh` to `num_gauss_hermite_points`

gpflow/likelihoods/base.py
gustavocmv: Resolving this one and saving the discussion of custom exceptions for another PR!

vdutor (approved):

st-- (changes requested):
I just tried out removing all analytic implementations `variational_expectations` & co. from `Gaussian` and the other scalar likelihoods, to have more test cases cover the quadrature implementation, and unfortunately this raises a bunch of failed tests - I think we should investigate why these failures are occurring before merging this PR.

gustavocmv:
> I just tried out removing all analytic implementations `variational_expectations` & co. from `Gaussian` and the other scalar likelihoods, to have more test cases cover the quadrature implementation, and unfortunately this raises a bunch of failed tests - I think we should investigate why these failures are occurring before merging this PR.

Solved in #1566

gustavocmv:
Closing this PR as it has been covered by #1566 #1559

PR #1551, opened 2020-08-27
Update min supported version of tfp

vdutor:
Develop (since PR #1518) does not support tfp 0.9.0 anymore. Updated the setup.py file.


awav (approved):
LGTM

PR #1553, opened 2020-09-07
Fixes for tf 2.3 / tfp 0.11.0 compatiblity

st--:
**PR type:** bugfix

## Summary

**Proposed changes**

* tensorflow_probability has been changing internals around their caching; here we update the cache-clearing for our deepcopy so that it works both for tensorflow_probability < 0.11.0, == 0.11.0 (with a workaround for the bug contained therein), and > 0.11.0 (which fixes the bug and changes internals again).

* removed pinning of numpy/scipy & changed order with tf/tfp dependencies so that pip can resolve versions correctly; also added the `--use-feature=2020-resolver` to CircleCI (as well as Makefile and README.md)

* for tf >= 2.3, disable a test case that was basically just checking that the test itself was written correctly (I don't think it's worth the effort to make something work that we wrote a workaround for as it's not doing what we need anyways)

* pin `black==19.10b0` as the 20.8b0 release changed formatting and I did not want this PR to get cluttered by that (we can bump the version in a separate PR if you like)

* We could remove deepcopy functionality (hurting current users needing this feature) or get tensorflow_probability to fix it properly upstream (preferred course of action, but would stop us from upgrading until we've convinced them and they've released the fix).

## PR checklist

### Release notes

* Compatibility with TensorFlow 2.3 and tensorflow_probability 0.11


gpflow/utilities/utilities.py
```diff
@@ -230,6 +231,33 @@ def update_state(parameter_or_variable, path, state):
     return state
 
 
+if Version(tfp.__version__) > Version("0.11.0"):
```
vdutor, line 234: Not sure, but do you mean this?
```suggestion
if Version(tfp.__version__) < Version("0.11.0"):
```

gpflow/utilities/utilities.py
```diff
@@ -230,6 +231,33 @@ def update_state(parameter_or_variable, path, state):
...
+
+    def _clear_bijector_cache(bijector: tfp.bijectors.Bijector):
+        bijector._cache.clear()
+
+
+elif Version(tfp.__version__) >= Version("0.11.0"):
```
vdutor, line 240: It's a bit weird that you have overlapping conditions. Was that intentional?
```suggestion
elif Version(tfp.__version__) == Version("0.11.0"):
```

st--: Good catch, it should be `==` only

vdutor (approved):

st--:
Had to change a few more things to get the build to pass with tf 2.3 ...

setup.py
```diff
@@ -9,35 +9,27 @@
...
+        "tensorflow-probability>0.10.0",  # tensorflow-probability==0.10.0 doesn't install correctly, https://github.com/tensorflow/probability/issues/991
+        "setuptools>=41.0.0",  # to satisfy dependency constraints
+    ]
+
+else:
+    requirements = []
+
+requirements.extend(
+    ["numpy", "scipy", "multipledispatch>=0.6", "tabulate", "typing_extensions",]
+)
```
joelberkeley-pio, line 29: i'd be surprised if this isn't enough to get it to build, so you could try it without the 2020-resolver

joelberkeley-pio: i think the problem is that pip currently doesn't update a dependency if it's further constrained later, so
```
numpy>1.10.0
numpy<1.11.0
```
might happily install 1.12

st--: ok, turns out we do need the new resolver :(

joelberkeley-pio: that's odd. I don't see this problem locally

tests/gpflow/utilities/test_multipledispatch.py
```diff
@@ -67,9 +68,19 @@ def test_a2_b2(x, y):
         assert len(w) == 0
 
 
+dispatcher_expectautographwarning_combinations_to_test = [
+    (gpflow.utilities.Dispatcher, False),  # no warnings with our custom Dispatcher
+]
+
+if Version(tf.__version__) < Version("2.3"):
```
vdutor, line 75: Not sure what our policy is, but do we really want to have backwards compatible tests? I'd be happy with staying up to date with TF 2.3.

st--: not in general necessarily, but in this case as it's very easy to write a test that passes because the test is written wrong, I wanted to keep the original test in there ...

st--: but on second thought it might be cleaner to simply skip the test when tf.version >= 2.3 ...

Issue #1449, opened 2020-05-01
Tracking model parameters during optimisation using Scipy


vr308:
# Documentation/tutorial notebooks

*Is there anything missing in the docs?* Yes 

*Are there any mistakes in the docs?* - No 

*Is there a feature that needs some example code in a notebook?* - Yes

There should be an example of how to track the evolution of variational parameters by utilizing the callback argument in  scipy minimize. It is not clear how to extract the state of the variables from the gplvm object. 

Using the snippets below in the GPLVM notebook just prints the gplvm object which is an numpy.ndarray which has the state of the trainable variables concatenated together ...ideally I want to be able to extract the trainable variable by name ...as I dont know the order in which they are stored.

```
def tracker(gplvm):
    print (gplvm)
    
opt = gpflow.optimizers.Scipy()
maxiter = ci_niter(10)
_ = opt.minimize(
    gplvm.training_loss,
    method="BFGS",
    variables=gplvm.trainable_variables,
    options=dict(maxiter=maxiter, disp=True),
    callback= tracker
)

```

*Do you know how to fix the docs?* If so, it'd be amazing if you'd be willing to directly contribute a pull request :)

Once I work out how to do, will be happy to do so.

Links:
* [latest GPflow documentation](https://gpflow.readthedocs.io/en/develop/)
* [latest GPflow tutorial notebooks](https://gpflow.readthedocs.io/en/develop/notebooks_file.html)

awav:
Hello @vr308! Thanks for the reported issue with documentation.

We have a docstring for the scipy's minimize method [here](https://github.com/GPflow/GPflow/blob/develop/gpflow/optimizers/scipy.py#L36) and we have a usage example in notebook [here](https://github.com/GPflow/GPflow/blob/develop/doc/source/notebooks/theory/FITCvsVFE.py#L158). I do agree that the notebook is hard to find and the scipy docstring could be better.

The simplest example of using `step_callback`:

```python
def step_callback(step, variables, values):
    print(f"Step {step}: variables{[v.numpy() for v in variables]}")

model = ...  # Create a GPR model
Scipy().minimize(model.training_loss, model.training_parameters, step_callback=step_callback)
```

Do you think the example above in the docstring will help in the future to understand how progress tracking works in scipy? The alternative would be a dedicated notebook. That will require bigger effort and you can help with it by initiating a PR, collecting use cases for scipy optimisation in a notebook file.

vr308:
Thanks! This is helpful..

[...]

PR #1558, opened 2020-09-10
Add code and documentation to monitor scipy optimization more easily.

vatsalaggarwal:
**PR type:** enhancement + doc improvement

**Related issue(s)/PRs:** 

## Summary

**Proposed changes**

* ... Adds a ScipyMonitor class. An instance of this class with the step_callback argument functionality of optimizers.Scipy().minimize to monitor trainings of GP with this optimizer.
* ... The monitoring notebook is updated with a comment informing the reader what to do if they want to monitor a training with the optimizers.Scipy optimizer.

- Adding another function to the Monitor class that returns the callable function that can be passed to step_callback. Instead, we went for adding another class as it seemed less dirty.
- For the notebook, an alternative is to add the scipy optimization + monitoring code to a runnable cell, but this breaks the flow of the notebook, so instead we chose to leave it as a comment.

### Minimal working example

```python
```

## PR checklist

### Release notes

Add code and documentation to monitor scipy optimization more easily.


vdutor (approved):
Thanks for the PR. LGTM.

gpflow/monitor/base.py
```diff
@@ -18,7 +18,6 @@
 
 import tensorflow as tf
 
-
 __all__ = ["MonitorTask", "ExecuteCallback", "MonitorTaskGroup", "Monitor"]
```
vdutor, line 21:
```suggestion
__all__ = ["MonitorTask", "ExecuteCallback", "MonitorTaskGroup", "Monitor", "ScipyMonitor"]
```

st--:
How about, instead of a new ScipyMonitor class, just adding a `scipy_callback` method or something like that that has the right function signature ?

doc/source/notebooks/basics/monitoring.pct.py
```diff
@@ -128,6 +128,21 @@ def plot_prediction(fig, ax):
...
+# something slightly different.
+#
+# ```
+# # define fast_tasks, slow_tasks, training_loss as in done in the previous 2 cells. Then,
+# from gpflow.monitor import ScipyMonitor
+# monitor = ScipyMonitor(fast_tasks, slow_tasks)
+#
+# opt = gpflow.optimizers.Scipy()
+# opt.minimize(training_loss, model.trainable_variables, step_callback=monitor)
+# ```
```
st--, line 144: Can you turn this into an actual code cell ? That would ensure the code is actually run in the continuous integration and give us more confidence it is (and will stay) correct.

st--:
This PR should resolve #1449

PR #1559, opened 2020-09-10
Multi Latent Likelihoods using new quadrature Likelihoods

vdutor:
This PR adds `MultiLatent` likelihoods, of which `HeteroskedasticTFPConditional` is an example, on top of the newly refactored `QuadratureLikelihood`s (PR #1549).

A notebook (courtesy of @st-- in  PR #1455 ) and tests are added for documentation purposes.


gpflow/likelihoods/base.py
```diff
@@ -286,6 +288,62 @@ def _variational_expectations(self, Fmu, Fvar, Y):
...
 
+class QuadratureLikelihood(Likelihood):
+    def __init__(self, num_gauss_hermite_points: int = 20, **kwargs):
+        super().__init__(**kwargs)
+        self.num_gauss_hermite_points = num_gauss_hermite_points
+        self._quadrature = None
+
+    @property
+    def quadrature(self):
+        if self._quadrature is None:
```
ChrisMorter, line 299: Just a stylistic point, others may disagree, but I find this sort of precondition checking easier to read if you invert the if statements so that the condition and what happens if it fails are grouped together, e.g. 

`

if self._quadrature is None:
        return self._quadrature

if self.latent_dim is None:
    raise Exception(
        "latent_dim not specified. "
        "Either set likelihood.latent_dim directly or "
        "call a method which passes data to have it inferred."
    )

with tf.init_scope():
    self._quadrature = NDiagGHQuadrature(self.latent_dim, self.num_gauss_hermite_points)

`

That way you end up with fewer levels of nesting, and it's easier to separate the precondition checking from what the method _really_ does.

gustavocmv (approved):
Looks good, ready to merge.

ChrisMorter (approved):

PR #1560, opened 2020-09-10
Fix SGPR upper bound for non zero mean functions

awav:


st-- (approved):
Woops, good catch! Thanks for this fix.

vdutor (approved):
It's pretty scary to think this type of bugs are still in the code... Was this code untested?

st--:
> It's pretty scary to think this type of bugs are still in the code... Was this code untested?

Good point. There already was a test for the `upper_bound` implementation (tests/integration/test_method_equivalence) - it even has a mean fucntion in it, but it's initialized to zero and the test data is such that it remained close enough to zero that it didn't catch the numeric discrepancy...

st--:
Please incorporate #1561 into this before merging so that we tighten up the test:)

PR #1566, opened 2020-09-14
Scalar Likelihood Refactoring

gustavocmv:
**PR type:** Enhancement / new feature / doc improvement

**Related issue(s)/PRs:** #1549
to 
## Summary

**Proposed changes**
This PR is a development/bugfix of half of #1549. It refactors the `ScalarLikelihood` class to move away from `ndiagquad` wrapper and use the `NDiagGHQuadrature` class directly as a component. The implementation componentizes an`NDiagGHQuadrature`  object with `dim=1` to be able to perform multiple 1-d quadratures through broadcasting. There's also a property+setter called `num_gauss_hermite_points` (named for backwards compatibility) which rebuilds the `NDiagGHQuadrature` object with the new value.

Side-Effect: Improvement in method `NDiagGHQuadrature._build_X_W` , which now includes a `tf.cast` of `Z, dZ` to match the `dtype` of `Fmu`.

*What alternatives have you considered?**
In the previous PR #1549, I used a naive implementation for the `ScalarLikelihood` which did not take advantage of the broadcasting to compute multiple 1-d quadratures. As such, when testing a `ScalarLlikelihood` with 10 dimensions and 20 gauss-hermite points, instead of 10 1-d quadratures, which require a 10*20 size grid (or only a 20 size grid with broadcasting, which is the case), a 10-d quadrature was being performed, using a 20**10 size grid, which caused (unnecessary) memory errors.
The idea of splitting this from #1549 is to avoid the multiple changes going through a big PR.

## PR checklist

- [X] New features: code is well-documented
  - [X] detailed docstrings (API documentation)
  - [ ] notebook examples (usage demonstration)
- [ ] The bug case / new feature is covered by unit tests
- [X] Code has type annotations
- [X] I ran the black formatter (`make format`)
- [X] I locally tested that the tests pass (`make check-all`)

### Release notes

**Fully backwards compatible:** yes

PR #1571, opened 2020-09-22
Likelihood quadrature cleanup

st--:
**PR type:** cleanup

**Related issue(s)/PRs:** #1505, #1559, #1566

## Summary

**Proposed changes**

* Global constant for num_gauss_hermite_points default
* QuadratureLikelihood with dependency injection (in preparation for unifying with MonteCarloLikelihood)
* ScalarLikelihood as QuadratureLikelihood subclass (with _quadrature_dim/_quadrature_log_prob/_quadrature_reduction)
* HeteroskedasticTFPConditional: rename argument to `scale_transform` and remove undocumented scaling
* DeprecationWarning for deprecated code

## PR checklist

### Release notes

* ...

### DIFF
--- a/gpflow/likelihoods/base.py
+++ b/gpflow/likelihoods/base.py
@@ -292,59 +292,67 @@ class Likelihood(Module, metaclass=abc.ABCMeta):
 
 
 class QuadratureLikelihood(Likelihood):
-    def __init__(self, num_gauss_hermite_points: int = DEFAULT_NUM_GAUSS_HERMITE_POINTS, **kwargs):
-        super().__init__(**kwargs)
+    def __init__(self, latent_dim: int, observation_dim: int, *, num_gauss_hermite_points: int = DEFAULT_NUM_GAUSS_HERMITE_POINTS, **kwargs):
+        super().__init__(latent_dim=latent_dim, observation_dim=observation_dim, **kwargs)
         self.num_gauss_hermite_points = num_gauss_hermite_points
         self._quadrature = None
 
     @property
     def quadrature(self):
-        if self._quadrature is None:
-            if self.latent_dim is None:
-                raise Exception(
-                    "latent_dim not specified. "
-                    "Either set likelihood.latent_dim directly or "
-                    "call a method which passes data to have it inferred."
-                )
-            with tf.init_scope():
-                self._quadrature = NDiagGHQuadrature(self.latent_dim, self.num_gauss_hermite_points)
+        if self._quadrature is not None:
+            return self._quadrature
+
+        if self.latent_dim is None:
+            raise Exception(
+                "latent_dim not specified. Either set likelihood.latent_dim directly "
+                "or call a method which passes data to have it inferred."
+            )
+
+        with tf.init_scope():
+            self._quadrature = NDiagGHQuadrature(self.latent_dim, self.num_gauss_hermite_points)
         return self._quadrature
 
-    def _predict_mean_and_var(self, Fmu, Fvar):
-        r"""
-        :param Fmu: mean function evaluation Tensor, with shape [..., latent_dim]
-        :param Fvar: variance of function evaluation Tensor, with shape [..., latent_dim]
-        :returns: mean and variance, both with shape [..., observation_dim]
-        """
-
-        def conditional_y_squared(F):
-            return self.conditional_variance(F) + self.conditional_mean(F) ** 2
-
-        integrands = [self.conditional_mean, conditional_y_squared]
-        E_y, E_y2 = self.quadrature(integrands, Fmu, Fvar)
-        V_y = E_y2 - E_y ** 2
-        return E_y, V_y
-
     def _quadrature_log_prob(self, F, Y):
-        return tf.expand_dims(self.log_prob(F, Y), -1)
+        return tf.expand_dims(self.log_prob(F, Y), axis=-1)
+
+    def _quadrature_reduction(self, v):
+        return tf.squeeze(v, axis=-1)
 
     def _predict_log_density(self, Fmu, Fvar, Y):
         r"""
         :param Fmu: mean function evaluation Tensor, with shape [..., latent_dim]
         :param Fvar: variance of function evaluation Tensor, with shape [..., latent_dim]
         :param Y: observation Tensor, with shape [..., observation_dim]:
-        :returns: variational expectations, with shape [...]
+        :returns: log predictive density, with shape [...]
         """
-        return tf.squeeze(self.quadrature.logspace(self._quadrature_log_prob, Fmu, Fvar, Y), -1)
+        return self._quadrature_reduction(
+            self.quadrature.logspace(self._quadrature_log_prob, Fmu, Fvar, Y=Y)
+        )
 
     def _variational_expectations(self, Fmu, Fvar, Y):
         r"""
         :param Fmu: mean function evaluation Tensor, with shape [..., latent_dim]
         :param Fvar: variance of function evaluation Tensor, with shape [..., latent_dim]
         :param Y: observation Tensor, with shape [..., observation_dim]:
-        :returns: log predictive density, with shape [...]
+        :returns: variational expectations, with shape [...]
         """
-        return tf.squeeze(self.quadrature(self._quadrature_log_prob, Fmu, Fvar, Y), -1)
+        return self._quadrature_reduction(
+            self.quadrature(self._quadrature_log_prob, Fmu, Fvar, Y=Y)
+        )
+
+    def _predict_mean_and_var(self, Fmu, Fvar):
+        r"""
+        :param Fmu: mean function evaluation Tensor, with shape [..., latent_dim]
+        :param Fvar: variance of function evaluation Tensor, with shape [..., latent_dim]
+        :returns: mean and variance of Y, both with shape [..., observation_dim]
+        """
+
+        def conditional_y_squared(*F):
+            return self.conditional_variance(*F) + tf.square(self.conditional_mean(*F))
+
+        E_y, E_y2 = self.quadrature([self.conditional_mean, conditional_y_squared], Fmu, Fvar)
+        V_y = E_y2 - E_y ** 2
+        return E_y, V_y
 
 
 class ScalarLikelihood(Likelihood):
--- a/gpflow/likelihoods/base.py
+++ b/gpflow/likelihoods/base.py
@@ -317,8 +317,9 @@ class QuadratureLikelihood(Likelihood):
 
     @num_gauss_hermite_points.setter
     def num_gauss_hermite_points(self, n_gh: int):
-        self._num_gauss_hermite_points = n_gh
-        self._set_quadrature()
+        if n_gh != self._num_gauss_hermite_points:
+            self._num_gauss_hermite_points = n_gh
+            self._set_quadrature()
 
     def _quadrature_log_prob(self, F, Y):
         return tf.expand_dims(self.log_prob(F, Y), axis=-1)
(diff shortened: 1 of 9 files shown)


gpflow/likelihoods/multilatent.py
```diff
@@ -112,11 +110,9 @@ def __init__(
         def conditional_distribution(Fs) -> tfp.distributions.Distribution:
             tf.debugging.assert_equal(tf.shape(Fs)[-1], 2)
             loc = Fs[..., :1]
-            scale = transform(Fs[..., 1:] / 2)
+            scale = transform(Fs[..., 1:])
```
st--, line 113: @vdutor flagging this up as it's changing behaviour - I'm assuming you had the division by 2 so that the latent GP models the log-variance, but that's only true if the transform is actually Exp, so that seemed a bit weird.

gpflow/likelihoods/multilatent.py
```diff
@@ -99,24 +97,23 @@ class HeteroskedasticTFPConditional(MultiLatentTFPConditional):
...
+        :param scale_transform: callable/bijector applied to the latent
+            function modelling the scale to ensure its positivity.
             Typically, `tf.exp` or `tf.softplus`, but can be any function f: R -> R^+.
         """
 
         def conditional_distribution(Fs) -> tfp.distributions.Distribution:
             tf.debugging.assert_equal(tf.shape(Fs)[-1], 2)
             loc = Fs[..., :1]
-            scale = transform(Fs[..., 1:] / 2)
+            scale = scale_transform(Fs[..., 1:])
```
st--, line 114: @vdutor flagging this up as it's mildly changing behaviour - I'm assuming you had the division by 2 so that the latent GP models the log-variance, but that's only true if the transform is actually Exp, so that seemed a bit weird for the general case here. (You can always include a scale in the transform itself!)

gpflow/likelihoods/multilatent.py
```diff
@@ -99,24 +97,23 @@ class HeteroskedasticTFPConditional(MultiLatentTFPConditional):
     def __init__(
         self,
         distribution_class: Type[tfp.distributions.Distribution] = tfp.distributions.Normal,
-        transform: tfp.bijectors.Bijector = positive(base="exp"),
-        num_gauss_hermite_points: int = 21,
+        scale_transform: tfp.bijectors.Bijector = positive(base="exp"),
```
st--, line 100: Updated name to make clear that this should be a transform for the *scale*

gpflow/likelihoods/base.py
```diff
@@ -366,15 +382,31 @@ class ScalarLikelihood(Likelihood):
...
         super().__init__(latent_dim=None, observation_dim=None, **kwargs)
-        self.num_gauss_hermite_points = 20
 
     @property
-    def num_gauss_hermite_points(self):
+    def num_gauss_hermite_points(self) -> int:
+        if not isinstance(self.quadrature, NDiagGHQuadrature):
+            raise TypeError(
+                "Can only query num_gauss_hermite_points if quadrature is a NDiagGHQuadrature instance"
+            )
```
gustavocmv, line 391: maybe add some note like "quadrature-specific attributes can be accessed with likelihood.quadrature"?

gpflow/likelihoods/multilatent.py
gustavocmv: tbh, I personally prefer to have the latent GP to log the log-scale, even though it is a little different from the HGP paper approach. Ofc, just a matter of taste.

vdutor: Happy for it to be the log-scale.

gpflow/likelihoods/base.py
```diff
@@ -366,15 +382,36 @@ class ScalarLikelihood(Likelihood):
...
         return self.quadrature.n_gh
 
     @num_gauss_hermite_points.setter
-    def num_gauss_hermite_points(self, n_gh):
-        self.quadrature = NDiagGHQuadrature(1, n_gh)
+    def num_gauss_hermite_points(self, n_gh: int):
```
vdutor, line 400: Can we not simply delete this instead of introducing deprecated code. I don't think many people changed the number of gh points on the fly any way.

gpflow/likelihoods/base.py
```diff
@@ -289,62 +292,75 @@ def _variational_expectations(self, Fmu, Fvar, Y):
...
-                    "call a method which passes data to have it inferred."
-                )
-            with tf.init_scope():
-                self._quadrature = NDiagGHQuadrature(self.latent_dim, self.num_gauss_hermite_points)
-        return self._quadrature
+    def _quadrature_dim(self) -> int:
+        return self.latent_dim
 
-    def _predict_mean_and_var(self, Fmu, Fvar):
+    def _quadrature_log_prob(self, F, Y):
```
vdutor, line 313: As `ScalarLikelihood` overwrites this method, does it make sense to have it private? As it seems to be important for inheriting classes, I think a docstring would be nice.

gpflow/likelihoods/base.py
```diff
@@ -289,62 +292,75 @@ def _variational_expectations(self, Fmu, Fvar, Y):
...
+        return tf.expand_dims(self.log_prob(F, Y), axis=-1)
+
+    def _quadrature_reduction(self, v):
```
vdutor, line 316: As `ScalarLikelihood` overwrites this method, does it make sense to have it private? As it seems to be important for inheriting classes, I think a docstring would be nice.

gpflow/likelihoods/multilatent.py
```diff
@@ -99,24 +97,23 @@ class HeteroskedasticTFPConditional(MultiLatentTFPConditional):
...
+        **kwargs,
```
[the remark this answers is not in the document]
st--: I agree in principle though this is deprecating the `num_gauss_hermite_points` argument

tests/gpflow/likelihoods/test_heteroskedastic.py
```diff
@@ -31,21 +31,18 @@ class Data:
...
-def likelihood() -> HeteroskedasticTFPConditional:
-    return HeteroskedasticTFPConditional(num_gauss_hermite_points=20)
-
-
-def test_analytic_mean_and_var(likelihood):
+def test_analytic_mean_and_var():
     """
     Test that quadrature computation used in HeteroskedasticTFPConditional
     of the predictive mean and variance is close to the analytical version, 
     which can be computed for the special case of N(y | mean=f1, variance=exp(f2)),
```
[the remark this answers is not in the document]
st--: thanks for the catch!

gpflow/likelihoods/base.py
```diff
@@ -289,62 +292,75 @@ def _variational_expectations(self, Fmu, Fvar, Y):
...
+        return tf.squeeze(v, axis=-1)
```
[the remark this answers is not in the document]
st--: Hm, i don't know. @gustavocmv ? I didn't want to mess about too much, just figure out how to get rid of the code duplication :thinking:

gpflow/likelihoods/base.py
st--: Then it'd be a breaking change though, and worst of all one users wouldn't even notice (they'd assign a different number, but it simply wouldn't have an effect anymore). If it's reasonably easy to add a backwards-compatible shim like in this case, that gives a deprecation warning, I think we should.

gpflow/likelihoods/base.py
st--: I don't think *other* classes would have to overwrite it, it's just because ScalarLikelihood behaves quite differently from the others ("broadcasting" over output dimensions and coupling latent_dim=observation_dim, without explicitly specifying either).

gpflow/likelihoods/base.py
st--: see below

gpflow/likelihoods/multilatent.py
```diff
@@ -22,20 +22,18 @@
...
     r"""
     A Likelihood which assumes that a single dimensional observation is driven
     by multiple latent GPs.
+
+    Note that this implementation does not allow for taking into account
+    covariance between outputs.
     """
 
-    def __init__(self, latent_dim: int, *, num_gauss_hermite_points: int = 21):
+    def __init__(self, latent_dim: int, **kwargs):
```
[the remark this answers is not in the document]
st--: see above

gpflow/likelihoods/base.py
vdutor: Can you write this down in the code somewhere - maybe where you overwrite them.

gpflow/likelihoods/multilatent.py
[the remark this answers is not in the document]
vdutor: yes - that be nice

gpflow/likelihoods/base.py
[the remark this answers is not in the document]
vdutor: If this is an easy change, I would be in favour of doing it. But it can be in a separate PR as well.

gpflow/likelihoods/multilatent.py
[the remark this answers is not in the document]
st--: On second thought, I'm not so sure about this. Apart from whether we want to do this for *all* likelihood classes (most just pass `**kwargs` up at the moment), it also means a _lot_ of changes if we decide to refactor it (though Pycharm might make that easier- do you know?)...

gpflow/likelihoods/base.py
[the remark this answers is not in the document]
gustavocmv: I believe the main affected part is this one:

`tf.reduce_sum(fun(X, *args, **kwargs) * W, axis=0)`

Some functions used in quadrature e.g. `conditional_mean` will have output shape `[n_gh_total,] + batch_shape + [d,]`. The `W` variable is set to be broadcastable with it, so it has shape `[n_gh_total,] + [1,]*len(batch_shape) + [1,]`. I tried to have this (indirectly) covered in the docstrings as quoted below:
```
:param fun: Callable or Iterable of Callables that operates elementwise, with
            signature f(X, *args, **kwargs). Moreover, it must satisfy the shape-mapping:
                X shape: [N_quad_points, b1, b2, ..., bX, d],
                    usually [N_quad_points, N, d]
                f(X) shape: [N_quad_points, b1, b2, ...., bX, d'],
                    usually [N_quad_points, N, 1] or [N_quad_points, N, d]
            In most cases, f should only operate over the last dimension of X
```
If we want to support functions returning `[n_gh_total,] + batch_shape`, which is the case of `log_prob`, we'd need to check for it and squeeze `W`(or expand `fun(...)`) to match the shapes. I believe it was the behaviour of the old `ndiagquad` function, but I think it caused more confusion than helped (ofc my opinion, open to discuss this).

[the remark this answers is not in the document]
st--: Thanks @gustavocmv ! I'm in favour of just leaving it as it is then; I've added these helpful explanations into the corresponding docstrings.

gpflow/likelihoods/base.py
st--: @vdutor @gustavocmv I've added docstrings, please check if a) they make sense and b) they satisfy your concerns!

gpflow/likelihoods/base.py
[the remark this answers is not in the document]
gustavocmv: Glad to help @st-- 
@vdutor what do you think about it?

gpflow/likelihoods/base.py
gustavocmv: Checked them, lgtm.

vdutor (approved):

PR #1574, opened 2020-09-29
fix BijectorCache workaround

st--:
Quickfix ... tensorflow_probability pushed 0.11.1 to PyPI (though there's no release beyond 0.11.0 on their GitHub yet), and against my expectations they didn't actually fix their broken `reset()` implementation in BijectorCache. I've now rewritten the work-around check in gpflow.utilities.utilities to use hasattr() on "clear" instead of solely basing it on package version, in case tfp does the same again with 0.11.2...

### DIFF
--- a/gpflow/utilities/utilities.py
+++ b/gpflow/utilities/utilities.py
@@ -245,22 +245,22 @@ def _get_leaf_components(input_module: tf.Module):
     return state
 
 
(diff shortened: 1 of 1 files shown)


vdutor (commented):
LGTM

vdutor (approved):

PR #1582, opened 2020-10-02
Explicit NotImplementedError on predict_y/predict_log_density with full_(output_)cov=True

### LINKED ISSUE #1569: Bug: Predictions with full_cov=True
The functions `model.predict_y()` and `model.predict_density()` do not work nicely with `full_cov=True`. For example, take `GPR`, where the `Gaussian` likelihood is used. In `predict_y()` the likelihood function `predict_mean_and_var()` is called. Here a scalar value of the likelihood variance is added to the covariance matrix. For `full_cov=True`, this should be added only to the diagonal.

I'm not sure what the right resolution for this is. In the short term, maybe throwing an error for `full_cov=True` would be helpful to show that this is unsupported.

st--:
Closes #1569

### DIFF
--- a/gpflow/models/model.py
+++ b/gpflow/models/model.py
@@ -212,7 +212,9 @@ class GPModel(BayesianModel):
         Compute the mean and variance of the held-out data at the input points.
         """
         if full_cov or full_output_cov:
-            raise NotImplementedError("predict_y currently supports only full_cov=full_output_cov=False")
+            raise NotImplementedError(
+                "predict_y currently supports only full_cov=full_output_cov=False"
+            )
 
         f_mean, f_var = self.predict_f(Xnew, full_cov=full_cov, full_output_cov=full_output_cov)
         return self.likelihood.predict_mean_and_var(f_mean, f_var)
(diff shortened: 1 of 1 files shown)


gpflow/models/model.py
```diff
@@ -211,6 +211,11 @@ def predict_y(
         """
         Compute the mean and variance of the held-out data at the input points.
         """
+        if full_cov or full_output_cov:
+            raise NotImplementedError(
+                "predict_y currently supports only full_cov=full_output_cov=False"
```
awav, line 216:
```suggestion
                "The  `predict_y` method currently supports argument values `full_cov=False` and `full_output_cov=False`"
```

gpflow/models/model.py
```diff
@@ -220,6 +225,11 @@ def predict_log_density(
         """
         Compute the log density of the data at the new data points.
         """
+        if full_cov or full_output_cov:
+            raise NotImplementedError(
+                "predict_log_density currently supports only full_cov=full_output_cov=False"
```
awav, line 230:
```suggestion
                "The  `predict_log_density ` method currently supports argument values `full_cov=False` and `full_output_cov=False`"
```

awav (approved):
LGTM. Please, address comments before merging.

Issue #818, opened 2018-07-26
Shape issues with multioutput kernels


kekeblom:
I have some questions regarding the `independent_interdomain_conditional` function in the multioutput module. 

I'm dealing with a case where I have a multiple output kernel and several independent GPs with separate inducing points. In this case the variables will have the following shapes:

`Kmn: M x L x N x P`
`Kmm: L x M x M`
`Knn: N x P (or P x N x N for full_cov case)`
`f: M x L`
`q_sqrt: L x M x M`

If `full_cov = False` and `full_ouput_cov = False` and `white = False`, `Ar` will have shape `(L, M, N, P)` as explicitly reshaped on line 311. Now on line 327 (link below) the code will crash as `Lm` will have rank 3 and `Ar` rank 4. The comment seems to be wrong as well. 

https://github.com/GPflow/GPflow/blob/54a276ab92ead568e969c7f073694b12bd8124e1/gpflow/multioutput/conditionals.py#L327

The problem might be fixed by replacing `Ar` with `A` on that line. However, then the output will have shape `(N * L, P)`. Shouldn't the mean and variance have shape `(N, L * P)` in this case?

Am I thinking about this the wrong way?

awav:
@kekeblom, could you post MWE?

kekeblom:
```
import gpflow
import tensorflow as tf
import numpy as np
import observations
from gpflow.multioutput.conditionals import independent_interdomain_conditional

(Xtrain, Ytrain), _ = observations.mnist('/tmp/mnist')
Xtrain -= Xtrain.mean(axis=0)

Zs = [np.random.randn(32, 392) for _ in range(2)]
rbf = gpflow.kernels.RBF(392)

def compute_Kmn(Z, X):
    X1 = X[:, 0:392]
    X2 = X[:, 392:]
    return tf.stack([
        rbf.K(Z, X1),
        rbf.K(Z, X2)
    ])

def compute_Knn(X):
    X1 = X[:, 0:392]
    X2 = X[:, 392:]
    return tf.stack([
        rbf.Kdiag(X1),
        rbf.Kdiag(X2)
    ])

batch = Xtrain[0:10]
# L = 2, P = 2, N = 10, M = 32
Kmm = tf.stack([rbf.K(Z) for Z in Zs]) # L x M x M
Kmn = tf.stack([compute_Kmn(Z, batch) for Z in Zs]) # L x P x M x N
[...]

Issue #1295, opened 2020-03-03
gpflow.conditionals.util.independent_interdomain_conditional not covered by tests


st--:
Noticed as part of #1219

st--:
We missed a bug (#818) because of this lack of tests. #1583 adds a minimal regression test, but more thorough tests for this case would still be valuable.

PR #1583, opened 2020-10-02
Bugfix for independent_interdomain_conditional

st--:
Fixes the issue brought up by #818


tests/gpflow/conditionals/test_multioutput.py
```diff
@@ -728,3 +729,48 @@ def test_separate_independent_conditional_with_q_sqrt_none():
...
+        X2 = X[:, D // 2 :]
+        return tf.stack([k(Z, X1), k(Z, X2)])
+
+    def compute_Knn(X):
+        X1 = X[:, 0 : D // 2]
+        X2 = X[:, D // 2 :]
+        return tf.stack([k(X1, full_cov=False), k(X2, full_cov=False)])
+
+    Xbatch = np.random.randn(N, D)
+    # L = 2, P = 2, N = 10, M = 32
```
vdutor, line 757: can we set L different to P? If so, would probably be a slightly better test to check this case.

st--: the independent model requires the number of latent GPs L to be equivalent to the number of outputs P, no?

st--: I've fixed the test accordingly! (We could still do with better tests - see #1295)

vdutor (approved):

PR #1585, opened 2020-10-05
Remove tiling workaround for lack of broadcasting in previous TensorFlow versions

### LINKED ISSUE #1321: Remove tiling in gauss_kl when tensorflow supports proper broadcasting
Currently, kullback_leiblers.gauss_kl() explicitly tiles as tf.linalg.triangular_solve does not broadcast (worse, it segfaults). Once TensorFlow fixes this issue (https://github.com/tensorflow/tensorflow/issues/37584), we should remove this workaround from our code, as it is very inefficient (lots of unnecessary memory allocation).

st--:
We used to tile in `gauss_kl()` to work around TensorFlow's lack of broadcasting. Since TensorFlow 2.2, this is finally working, and this PR removes the tiling for TensorFlow>=2.2. Resolves #1321.

### DIFF
--- a/gpflow/kullback_leiblers.py
+++ b/gpflow/kullback_leiblers.py
@@ -134,7 +134,7 @@ def gauss_kl(q_mu, q_sqrt, K=None, *, K_cholesky=None):
         else:
             if is_batched or Version(tf.__version__) >= Version("2.2"):
                 Lp_full = Lp
-            else:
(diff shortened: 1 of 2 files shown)


insysion (commented):
Looks good but let's add a test for the tiled case.

insysion (approved):
Beautiful test :)

PR #1598, opened 2020-10-13
HeteroskedasticTFPConditional should construct tensors at class-construction, not at module-import time

polaschwoebel:
**PR type:** enhancement

**Related issue(s)/PRs:** [discussion in gpflow slack, request-for-help, oct. 10]

## Summary

**Proposed changes**

As discussed in the slack channel, `likelihoods.multilatent.HeteroskedasticTFPConditional` currently creates tensors at import time. This is due to the call of `positive(base="exp")` as a keyword argument default. This causes issues when gpflow is imported before the CUDA environment variable is set. Suggested improvements: 

* don't call positive at import time, i.e. new input is `scale_transform: Optional[tfp.bijectors.Bijector] = None`
* move this call into the `__init__` method instead 
* make` self.scale_transform = scale_transform` an attribute of `HeteroskedasticTFPConditional`

Setting the CUDA environment variable before importing gpflow also works. 

### Minimal working example

```python
import os
import tensorflow as tf
import gpflow
os.environ["CUDA_VISIBLE_DEVICES"] = "0"
print('device is:', os.environ["CUDA_VISIBLE_DEVICES"])
print(tf.test.is_gpu_available())
```
cuda_error

## PR checklist

### Release notes

**Fully backwards compatible:** 

 change HeteroskedasticTFPConditional so it only constructs the tensor at class-construction, not at module-import time

### DIFF
--- a/gpflow/likelihoods/multilatent.py
+++ b/gpflow/likelihoods/multilatent.py
@@ -97,7 +97,7 @@ class HeteroskedasticTFPConditional(MultiLatentTFPConditional):
     def __init__(
         self,
         distribution_class: Type[tfp.distributions.Distribution] = tfp.distributions.Normal,
-        scale_transform: tfp.bijectors.Bijector = positive(base="exp"),
+        scale_transform: Optional[tfp.bijectors.Bijector] = None,
         **kwargs,
     ):
         """
@@ -107,6 +107,8 @@ class HeteroskedasticTFPConditional(MultiLatentTFPConditional):
             function modelling the scale to ensure its positivity.
             Typically, `tf.exp` or `tf.softplus`, but can be any function f: R -> R^+.
         """
+        if scale_transform is None:
+            scale_transform = positive(base="exp")
 
         def conditional_distribution(Fs) -> tfp.distributions.Distribution:
             tf.debugging.assert_equal(tf.shape(Fs)[-1], 2)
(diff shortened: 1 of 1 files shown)


st-- (approved):
looks good to me. thanks for contributing!

gpflow/likelihoods/multilatent.py
```diff
@@ -107,11 +107,14 @@ def __init__(
             function modelling the scale to ensure its positivity.
             Typically, `tf.exp` or `tf.softplus`, but can be any function f: R -> R^+.
```
st--, line 108: Could you add a line to the docstring explaining that the default is Exp if not explicitly specified?

vdutor (approved):
LGTM. Thanks for the PR!

gpflow/likelihoods/multilatent.py
polaschwoebel: Done! Is it correct to say it defaults to `tf.exp` if not specified? The code says  `positive(base="exp")`, is that the same?

st--: It's de-facto the same; `positive()` constructs a tfp.bijectors.Bijector instance instead of returning a function, but both exponentiate the argument. Could just call it "exp" instead of referring to `tf.`...

polaschwoebel: Ok, thanks for clarifying, did that.

PR #1619, opened 2020-12-17
fix ModelToTensorBoard max_size = -1

sam-willis:
**PR type:** bugfix

## Summary

Currently `ModelToTensorBoard` says in it's initialiser that passing -1 to max size will track all the parameters of a model, but currently this is not respected.

### Minimal working example

```python
```

## PR checklist

### Release notes

* ...

### DIFF
--- a/gpflow/monitor/tensorboard.py
+++ b/gpflow/monitor/tensorboard.py
@@ -117,11 +117,9 @@ class ModelToTensorBoard(ToTensorBoard):
                 "run the monitor outside the `tf.function`."
             )
 
-        if size == 1:
-            tf.summary.scalar(name, param[0], step=self.current_step)
-        else:
-            for i in range(min(size, self.max_size)):
-                tf.summary.scalar(f"{name}[{i}]", param[i], step=self.current_step)
+        it = range(size) if self.max_size == -1 else range(min(size, self.max_size))
+        for i in it:
+            tf.summary.scalar(f"{name}[{i}]", param[i], step=self.current_step)
 
 
 class ScalarToTensorBoard(ToTensorBoard):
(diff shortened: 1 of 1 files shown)


sam-willis:
The failing test is `test_scipy_jit`, this seem unrelated to the change I've made in this PR.

gpflow/monitor/tensorboard.py
```diff
@@ -117,11 +117,9 @@ def _summarize_parameter(self, name: str, param: Union[Parameter, tf.Variable]):
...
             )
 
-        if size == 1:
-            tf.summary.scalar(name, param[0], step=self.current_step)
-        else:
-            for i in range(min(size, self.max_size)):
-                tf.summary.scalar(f"{name}[{i}]", param[i], step=self.current_step)
+        it = range(size) if self.max_size == -1 else range(min(size, self.max_size))
+        for i in it:
+            tf.summary.scalar(f"{name}[{i}]", param[i], step=self.current_step)
```
st--, line 122: This changes behaviour (namely, it'll always add the `[0]` for scalar parameters). Might be nice to only have your change in the else: branch of the original code?

st--:
Hi @sam-willis , thanks for spotting & fixing this! The unit tests will be fixed by #1625. Would you be happy to address the one comment I left? We can merge it once the other PR is in.

gpflow/monitor/tensorboard.py
sam-willis: Ah, makes sense! Sure I'll make the change

st-- (approved):
looks good to me, thanks for the fix. (just wondering if it's worth bothering to add a regression test for this?)

sam-willis:
It probably would, but I don't have time to do it at the moment, it's up to you whether you'd rather merge now, or leave it with me in which case it'll get the test when I'm next at a loose end.

PR #1620, opened 2020-12-17
New SVGP with posterior object and faster predictions

stefanosele:
Posterior class with precomupations speeding up test

### DIFF
--- a/gpflow/models/svgp.py
+++ b/gpflow/models/svgp.py
@@ -331,4 +331,5 @@ class NewSVGP(OldSVGP):
         )
 
 
-SVGP = NewSVGP
+class SVGP(NewSVGP):
+    pass
--- a/gpflow/models/svgp.py
+++ b/gpflow/models/svgp.py
@@ -226,8 +226,8 @@ class NewSVGP(OldSVGP):
         self.posterior = Posterior(
             self.kernel,
             self.inducing_variable,
-            q_mu,
-            q_sqrt,
+            self.q_mu,
+            self.q_sqrt,
             whiten=self.whiten,
             mean_function=self.mean_function,
         )
--- a/gpflow/models/svgp.py
+++ b/gpflow/models/svgp.py
@@ -208,9 +208,7 @@ class NewSVGP(OldSVGP):
         then call
             posterior.predict_f(Xnew, ...)
         """
-        return self.posterior().predict_f(
-            Xnew, full_cov=full_cov, full_output_cov=full_output_cov
-        )
+        return self.posterior().predict_f(Xnew, full_cov=full_cov, full_output_cov=full_output_cov)
 
 
 class SVGP(NewSVGP):
--- a/gpflow/models/svgp.py
+++ b/gpflow/models/svgp.py
@@ -18,7 +18,7 @@ import numpy as np
 import tensorflow as tf
 
 from .. import kullback_leiblers
-from ..base import Parameter
+from ..base import Module, Parameter
 from ..conditionals import conditional
 from ..config import default_float
 from ..utilities import positive, triangular
@@ -27,7 +27,7 @@ from .training_mixins import ExternalDataTrainingLossMixin
 from .util import inducingpoint_wrapper
 
 
-class SVGP(GPModel, ExternalDataTrainingLossMixin):
+class OldSVGP(GPModel, ExternalDataTrainingLossMixin):
     """
     This is the Sparse Variational GP (SVGP). The key reference is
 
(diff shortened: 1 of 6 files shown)


gpflow/posterior.py
```diff
@@ -0,0 +1,194 @@
...
+    Kuf = covariances.Kuf(iv, kernel, Xnew)  # [(R), M, N]
+
+    fully_correlated = Kuf.shape.ndims == 4
+
+    if fully_correlated:
+        Knn = kernel(Xnew, full_cov=full_cov, full_output_cov=full_output_cov)
+        M, L, N, K = tf.unstack(tf.shape(Kuf), num=Kuf.shape.ndims, axis=0)
+        Kuf = tf.reshape(Kuf, (M * L, N * K))
+        if full_cov == full_output_cov:
+            Knn = tf.reshape(Knn, (N * K, N * K)) if full_cov else tf.reshape(Knn, (N * K,))
```
frgsimpson, line 186: Just for readability I'd prefer something alone the lines of 
```suggestion
            new_shape = (N * K, N * K)  if full_cov else (N * K,)
            Knn = tf.reshape(Knn, new_shape)
```

gpflow/posterior.py
```diff
@@ -0,0 +1,194 @@
...
+        mean = tf.matmul(Kuf, self.alpha, transpose_a=True)
+        if Kuf.shape.ndims == 3:
+            mean = tf.linalg.adjoint(tf.squeeze(mean, axis=-1))
+
+        if full_cov:
+            Kfu_Qinv_Kuf = tf.matmul(Kuf, self.Qinv @ Kuf, transpose_a=True)
+            if fully_correlated and not full_output_cov:
+                Kfu_Qinv_Kuf = tf.reshape(
+                    Kfu_Qinv_Kuf, tf.concat([tf.shape(Kfu_Qinv_Kuf)[:-2], (N, K, N, K)], axis=0)
+                )
```
frgsimpson, line 139:
```suggestion
                new_shape = tf.concat([tf.shape(Kfu_Qinv_Kuf)[:-2], (N, K, N, K)], axis=0)
                Kfu_Qinv_Kuf = tf.reshape(Kfu_Qinv_Kuf, new_shape)
```

frgsimpson (commented):
Looks good! Just left some small cosmetic suggestions.

gpflow/models/svgp.py
```diff
@@ -170,3 +175,75 @@ def predict_f(self, Xnew: InputData, full_cov=False, full_output_cov=False) -> M
...
         # tf.debugging.assert_positive(var)  # We really should make the tests pass with this here
         return mu + self.mean_function(Xnew), var
+
+
+class NewSVGP(OldSVGP):
+    """
+    Adds posterior() method and uses different math ordering for predict_f
+    """
+
+    def __init__(
```
vdutor, line 185: It doesn't look like you are changing anything in the init compared to the super class, so you could just drop the init here.

gpflow/models/svgp.py
```diff
@@ -170,3 +175,75 @@ def predict_f(self, Xnew: InputData, full_cov=False, full_output_cov=False) -> M
...
+        """
+        For backwards compatibility. We have to update posterior's cache here so we have acces to
+        the latest values during training.
+        For faster (cached) prediction, predict directly from the posterior object, i.e.,:
+            model.posterior.predict_f(Xnew, ...)
+        """
+        return self.posterior().predict_f(Xnew, full_cov=full_cov, full_output_cov=full_output_cov)
+
+
+class SVGP(NewSVGP):
```
vdutor, line 248: Why not keep SVGP as is, and let NewSVGP inherit from SVGP immediately?

gpflow/models/svgp.py
```diff
@@ -170,3 +175,75 @@ def predict_f(self, Xnew: InputData, full_cov=False, full_output_cov=False) -> M
         )
         # tf.debugging.assert_positive(var)  # We really should make the tests pass with this here
         return mu + self.mean_function(Xnew), var
+
+
+class NewSVGP(OldSVGP):
```
vdutor, line 180: Why not give this a more descriptive name: `FastPredictionSVGP`, for example?

vdutor: Also, see comment about immediately inheriting from `SVGP` instead of creating a phantom class in [L248](https://github.com/GPflow/GPflow/pull/1620/files#diff-e5e515abadf88d454f99405f4af8fb41b1d6d7aae79f11430f237e1a40395ddcR248).

gpflow/posterior.py
```diff
@@ -0,0 +1,194 @@
...
+
+
+class MvnNormal(Module):
+    def __init__(self, q_mu, q_sqrt):
+        self.q_mu = q_mu  # [M, L]
+        self.q_sqrt = q_sqrt  # [L, M, M], lower-triangular
+
+
+class Posterior(Module):
+    def __init__(self, kernel, iv, q_mu, q_sqrt, whiten=True, mean_function=None):
```
vdutor, line 40: Nit: everywhere else in the codebase we use `inducing_variable` and not `iv`, it would be nice to be consistent.

gpflow/posterior.py
```diff
@@ -0,0 +1,194 @@
...
+        self.iv = iv
+        self.kernel = kernel
+        self.mean_function = mean_function
```
vdutor, line 43: I guess you want to check if `mean_function` is `None` and set a `Zero` mean function in that case because now [L171](https://github.com/GPflow/GPflow/pull/1620/files#diff-b91e3b59b118c92df6de43aa80f1c546a3f183d81260e5ac8dd287bd03e7da17R171) will break when `None` is passed in.

gpflow/posterior.py
```diff
@@ -0,0 +1,194 @@
...
+            alpha = tf.linalg.triangular_solve(L, q_mu, adjoint=True)
+        # predictive mean = Kfu alpha
+        # predictive variance = Kff - Kfu Qinv Kuf
+        # S = q_sqrt q_sqrtT
+        if not self.whiten:
+            # Qinv = Kuu⁻¹ - Kuu⁻¹ S Kuu⁻¹
+            #      = Kuu⁻¹ - L⁻T L⁻¹ S L⁻T L⁻¹
+            #      = L⁻T (I - L⁻¹ S L⁻T) L⁻¹
+            #      = L⁻T B L⁻¹
+            if isinstance(self.q_dist, DiagNormal):
```
vdutor, line 77: you could make a method in `DiagNormal` and `MvnNormal` that returns you the correct q_sqrt parametrisation?

gpflow/posterior.py
```diff
@@ -0,0 +1,194 @@
...
+        self.q_sqrt = q_sqrt  # [M, L]
+
+
+class MvnNormal(Module):
+    def __init__(self, q_mu, q_sqrt):
+        self.q_mu = q_mu  # [M, L]
+        self.q_sqrt = q_sqrt  # [L, M, M], lower-triangular
+
+
+class Posterior(Module):
```
vdutor, line 39: A docstring would be welcome here.

gpflow/posterior.py
```diff
@@ -0,0 +1,194 @@
...
+
+    def update_cache(self):
+        self.alpha, self.Qinv = self._precompute()
+
+    def freeze(self):
+        alpha, Qinv = self._precompute()
+        self.alpha = Parameter(alpha, trainable=False)
+        self.Qinv = Parameter(Qinv, trainable=False)
+
+    def update_cache_with_variables(self):
```
vdutor, line 108: I don't immediately see the use of this. Could you add a docstring?

gpflow/posterior.py
```diff
@@ -0,0 +1,194 @@
...
+        Qinv = tf.linalg.triangular_solve(L, B_Linv, adjoint=True)
+
+        return alpha, Qinv
+
+    def update_cache(self):
+        self.alpha, self.Qinv = self._precompute()
+
+    def freeze(self):
+        alpha, Qinv = self._precompute()
+        self.alpha = Parameter(alpha, trainable=False)
```
vdutor, line 105: Question: why can we not convert then to (constant) tensors?

gpflow/posterior.py
```diff
@@ -0,0 +1,194 @@
...
+            self.alpha = Parameter(alpha, trainable=False)
+            self.Qinv = Parameter(Qinv, trainable=False)
+
+    def predict_f(
+        self, Xnew, full_cov: bool = False, full_output_cov: bool = False
+    ) -> MeanAndVariance:
+        # Qinv: [L, M, M]
+        # alpha: [M, L]
+
+        Kuf, Knn, fully_correlated = _get_kernels(
```
vdutor, line 123: Nit: is `_get_kernels` the most precise name? It looks to me like we are computing covariance matrices?

stefanosele (commented):
I'll address comments about docstrings once we finalise the interface if that's ok. My main concern is still usability and if the Posterior class can effectively replace the dispatcher and all the conditionals

gpflow/posterior.py
```diff
@@ -0,0 +1,194 @@
...
+
+        if not self.whiten:
+            # alpha = Kuu⁻¹ q_mu
+            alpha = tf.linalg.cholesky_solve(L, q_mu)
+        else:
+            # alpha = L⁻T q_mu
+            alpha = tf.linalg.triangular_solve(L, q_mu, adjoint=True)
+        # predictive mean = Kfu alpha
+        # predictive variance = Kff - Kfu Qinv Kuf
+        # S = q_sqrt q_sqrtT
```
awav, line 71: here and the rest:
```suggestion
        # S = q_sqrt q_sqrtᵀ
```

gpflow/models/svgp.py
st--: What @stefanosele said - this runs all pre-existing tests (and importantly the SVGP vs VGP method equivalence comparison) against our new code. Subclassing instead of aliasing `SVGP = NewSVGP` (which I actually started out with) means that `SVGP(...).__class__.__name__` is "SVGP" and we don't have to touch all the printing tests.

gpflow/posterior.py
st--: done

st--:
Superseded by #1636. @stefanosele do you want to have a look at the new PR? It's changed quite a bit since you last saw it. Would be great to get your thoughts on it.

stefanosele:
@st-- I can have a look closer to the end of the week. I'm a bit overwhelmed with starting on the new job. Sorry for the massive delay

PR #1626, opened 2021-01-12
Quickfix dynamic shapes in quadrature

vdutor:
Simple fix for when passing batches of unknown batch size to `ndiag_mc`.


st-- (approved):
lgtm

Issue #1599, opened 2020-10-22
Proposal: separate Predictor objects


st--:
# Feature request

`model.predict_f()` currently recomputes everything (kernel matrix, Cholesky, triangular_solve) from scratch at each call. Instead of recomputing these quantities, we should provide a predictor object that only does those computations that touch the new input points.

## Motivation

Recomputing lots of cacheable quantities from scratch is inefficient and unnecessarily slow. The slowness of predictions (particularly in a loop) is a well-known, recurring issue, e.g. https://github.com/GPflow/GPflow/issues/1554, https://github.com/GPflow/GPflow/issues/1189, https://github.com/GPflow/GPflow/issues/1030, https://github.com/GPflow/GPflow/issues/1003. We recently merged https://github.com/GPflow/GPflow/pull/1528 which was added so that a special-cased version of this proposal could be implemented down-stream. (Note that there are further efficiency gains possible that even #1528 does not allow for.) It would be much better to implement this properly, and make it available for all users out of the box.

## Proposal

**Describe the solution you would like**
Ideally, `model.predictor()` (or similar) returns an (optionally `tf.function`-decorated) closure that takes new input points `Xnew` and computes the prediction efficiently (having precomputing and cached all that can be computed without access to the test inputs).

Questions:
* Should this be extended to the other predict functions (predict mean and var, predict log density)?
* How will we be able to cache all the quantities needed within the various `conditional` branches?

pio-neil:
> Ideally, model.predictor() (or similar) returns an (optionally tf.function-decorated) closure that takes new input points Xnew and computes the prediction efficiently (having precomputing and cached all that can be computed without access to the test inputs).

Would it make sense for `model.predictor()` to return a `Predictor` class, which could then implement multiple methods (`predict_f`, `predict_y` etc).

st--:
> Would it make sense for `model.predictor()` to return a `Predictor` class, which could then implement multiple methods (`predict_f`, `predict_y` etc)?

Yes, that would certainly be an option as well! This could be a more extensive effort, related to thoughts/discussions we've had before around a `Posterior` object (that fully decouples from the model and contains all the prediction methods normally associated with the `Model` instance).

joelberkeley-secondmind:
this ticket is introduced as: here's a problem, implement this solution. How about: here's a problem, solve the problem, here's one possible way of doing it. So as to avoid biasing to particular solutions?

st--:
@joelberkeley-secondmind 

[...]

PR #1636, opened 2021-02-03
Posterior object for fast predictions

st--:
## Summary

GPflow had already been optimised for speed in the training loop. This PR addresses the speed of *predictions* and resolves (partially) #1599. We achieve this by introducing new Posterior objects that can cache the results of a "precompute" stage (everything that's independent of test points) and thereby significantly speeds up repeated predictions.

## Description of changes

* Posterior classes contain a precompute() and predict_f() that is significantly faster for repeated predictions.
* The "old" fused computations from conditionals() are incorporated as _fused_conditional() methods. This means the related code paths are always next to each other.
* All the conditionals() implementations now solely construct the corresponding Posterior object and call its fused_predict_f() method, thereby remaining backwards-compatible.
johnamcleod
This PR closes #1620, which is a previous iteration (left open for now for the review comments left on there already).

## Outstanding tasks:

- [ ] code clean-up
  - [x] ~Identify & remove all avoidable transposes!~ made an issue for this: #1665
  - [x] Posterior.freeze() removed
  - [x] Add warnings to conditionals() to use Posterior objects instead (Can we do this even on import? Ideally we’d move the various conditionals() into a single file now that they just construct posteriors…)
  - [ ] remove the explicit multi-output conditional() implementations (which now do exactly the same as this fallback)?
- [ ] design review of new posterior classes etc.
  - [x] ~Are we happy with the DeltaDist / DiagNormal / MvnNormal classes?~ renamed to _DeltaDist / _DiagNormal / _MvNormal, can revisit this in the future
- [ ] code review of implementation @awav @johnamcleod 
- [ ] add missing tests
  - [x] Numeric consistency check between the two code paths: for all “cached-posterior” predictions (all AbstractPosterior subclasses), we want to compare predict_f against the corresponding (“status-quo”) fused_predict_f (@johnamcleod resolved this in #1659)
  - [x] ~Test to confirm it all works within tf.function() and allows swapping out of models without having to retrace/recompile the graph. @sam-willis @nfergu/@pio-neil~ not done, but not actually required
  - [x] ~Test to confirm~ manually verified it all works within AOT compilation & gives expected speedup - 
  - [x] Test of update_cache_with_variables - #1672
  - [ ] Test of FullyCorrelatedPosterior with Kuf.shape.ndims == 3
@marcoadurno 
- [ ] add documentation
  - [ ] API docs?
  - [x] We need to explain this to users. Ideally, the VFF notebook could demonstrate the difference between caching and fused predict!
  - [ ] Notebook examples of how to use model.posterior()
- [x] reverted the backwards-incompatible a9d889b.
- [x] update RELEASE.md

Co-authored-by: stefanosele@users.no-reply.github.com
Co-authored-by: johnamcleod@users.no-reply.github.com


gpflow/posterior.py
```diff
@@ -183,7 +412,7 @@ def predict_f(
         # Qinv: [L, M, M]
         # alpha: [M, L]
 
-        Kuf = covariances.Kuf(inducing_variable, kernel, Xnew)
```
st--: Hasn't been covered by tests yet

st--:
Should revert the backwards-incompatible a9d889b before merging.

st--:
Note- this change will also allow us to speed up predictions even further by carrying out the post-precompute predictions in float32 rather than float64; to make that easy to use will require further modifications ("posterior.as_dtype(np.float32)") which should go in a different PR.

doc/source/notebooks/advanced/variational_fourier_features.pct.py
```diff
@@ -230,97 +230,105 @@ def gauss_kl_vff(q_mu, q_sqrt, K):
...
+
+
+class VFFPosterior(gpflow.posterior.AbstractPosterior):
+    def _precompute(self):
+        """
+        Not implemented yet
+        """
+        return np.nan, np.nan
+
+    def _conditional_with_precompute(self, Xnew, full_cov, full_output_cov):
```
johnamcleod, line 243: Can this method be removed by renaming `_conditional_fused` to `_conditional_with_precompute`?

gpflow/posterior.py
```diff
@@ -189,14 +466,14 @@ def predict_f(
         M, L, N, K = tf.unstack(tf.shape(Kuf), num=Kuf.shape.ndims, axis=0)
         Kuf = tf.reshape(Kuf, (M * L, N * K))
 
-        Knn = self.kernel(Xnew, full_cov=full_cov, full_output_cov=full_output_cov)
+        Kff = self.kernel(Xnew, full_cov=full_cov, full_output_cov=full_output_cov)
```
johnamcleod, line 469: Should the `full_output_cov` argument be removed?

gpflow/posterior.py
```diff
@@ -235,31 +512,67 @@ def predict_f(
...
-def _get_kernels(Xnew, inducing_variable, kernel, full_cov, full_output_cov):
+    def _conditional_fused(
+        self, Xnew, full_cov: bool = False, full_output_cov: bool = False
+    ) -> MeanAndVariance:
+        Kmm = covariances.Kuu(
+            self.inducing_variable, self.kernel, jitter=default_jitter()
+        )  # [M, L, M, L]
+        Kmn = covariances.Kuf(self.inducing_variable, self.kernel, Xnew)  # [M, L, N, P]
+        Knn = self.kernel(
+            Xnew, full_cov=full_cov, full_output_cov=full_output_cov
```
johnamcleod, line 525: Should the `full_output_cov` argument be removed?

gpflow/posterior.py
```diff
@@ -235,31 +512,67 @@ def predict_f(
...
+class FallbackIndependentLatentPosterior(FullyCorrelatedPosterior):  # XXX
+    def _conditional_fused(
+        self, Xnew, full_cov: bool = False, full_output_cov: bool = False
+    ) -> MeanAndVariance:
+        Kmm = covariances.Kuu(
+            self.inducing_variable, self.kernel, jitter=default_jitter()
+        )  # [L, M, M]
+        Kmn = covariances.Kuf(self.inducing_variable, self.kernel, Xnew)  # [M, L, N, P]
+        Knn = self.kernel(
+            Xnew, full_cov=full_cov, full_output_cov=full_output_cov
```
johnamcleod, line 563: Should the `full_output_cov` argument be removed?

st--: I don't think so; it changes the shape of Knn....

gpflow/posterior.py
st--: I don't think so; it changes the shape of Knn....

gpflow/posterior.py
st--: I don't think so; it changes the shape of Knn....

doc/source/notebooks/advanced/variational_fourier_features.pct.py
awav: Same question here. I don't think the name is important. I would like to know the reasoning for having two methods that do the same job.

doc/source/notebooks/advanced/variational_fourier_features.pct.py
```diff
@@ -77,14 +77,14 @@ def Kuu_matern12_fourierfeatures1d(inducing_variable, kernel, jitter=None):
         (b - a) * (tf.square(lamb) + tf.square(omegas)) / lamb / kernel.variance / two_or_four
     )  # eq. (111)
     v_cos = tf.ones_like(d_cos) / tf.sqrt(kernel.variance)  # eq. (110)
-    cosine_block = LowRank(Diag(d_cos), v_cos[:, None])
+    cosine_block = LowRank(Diag(d_cos, is_positive_definite=True), v_cos[:, None])
```
awav, line 80: Can you make `is_positive_definite=True` by default? If not, what is the reason for not doing it?

gpflow/conditionals/conditionals.py
```diff
@@ -66,13 +66,10 @@ def _conditional(
...
         about the shape of the variance, depending on `full_cov` and `full_output_cov`.
     """
-    Kmm = Kuu(inducing_variable, kernel, jitter=default_jitter())  # [M, M]
-    Kmn = Kuf(inducing_variable, kernel, Xnew)  # [M, N]
-    Knn = kernel(Xnew, full_cov=full_cov)
-    fmean, fvar = base_conditional(
-        Kmn, Kmm, Knn, f, full_cov=full_cov, q_sqrt=q_sqrt, white=white
-    )  # [N, R],  [R, N, N] or [N, R]
-    return fmean, expand_independent_outputs(fvar, full_cov, full_output_cov)
+    gp_posterior = IndependentPosteriorSingleOutput(
```
awav, line 69: Minor comment, but a `posterior` name w/o `gp` would be okay.

gpflow/posterior.py
```diff
@@ -30,10 +38,19 @@
     SeparateIndependentInducingVariables,
     SharedIndependentInducingVariables,
 )
-from .models.model import MeanAndVariance
+from .types import MeanAndVariance
 from .utilities import Dispatcher
 
 
+class DeltaDist(Module):
```
awav, line 45: This is an interesting and obscure object :) Could you add a bit more details about what it is for and how to use it?

gpflow/posterior.py
```diff
@@ -46,19 +63,117 @@ def __init__(self, q_mu, q_sqrt):
         self.q_sqrt = q_sqrt  # [L, M, M], lower-triangular
 
 
-class BasePosterior(Module, ABC):
-    def __init__(self, kernel, inducing_variable, q_mu, q_sqrt, whiten=True, mean_function=None):
+class AbstractPosterior(Module, ABC):
```
awav, line 66: `AbstractPosterior` doesn't look like an abstract object, it does pretty much lots of things that are not abstract.

gpflow/posterior.py
```diff
@@ -46,19 +63,117 @@ def __init__(self, q_mu, q_sqrt):
...
+    def __init__(
+        self,
+        kernel,
+        inducing_variable,
+        q_mu,
+        q_sqrt,
```
awav, line 72: Taking into account `DeltaDist` signature, this argument should be options with default `None`.

awav (commented):
Hi @st-- !

I do like that the old path is still in the code. If there is a plan to deprecate it, can we leave a deprecate message somewhere in the code or via warning?

It is hard to track technical changes/movements in the code. It would be nice to have guidance for a reviewer via GitHub or in-code comments. Files of the particular interest are `variational_fourier_features.pct.py`, `multioutput/conditionals.py`, `utils.py`.

The `types.py` file is not a must-have addition to the PR, considering the amount of other work and that you don't change the signature of most of the functions (they simply don't have type annotations).

doc/source/notebooks/advanced/variational_fourier_features.pct.py
st--: They do the same job from a mathematical point of view, but they don't do the same job from a finite-precision computational numerics point of view. When training, the fused conditional requires less ops than with the precompute. When testing, the one with precompute has significant advantages. Having both addresses @markvdw's concerns about completely changing the numerics at the core of GPflow.

doc/source/notebooks/advanced/variational_fourier_features.pct.py
st--: `Diag` is a TensorFlow class, so we have to explicitly set this at construction time.

gpflow/posterior.py
st--: Abstract in the sense that it's an abstract base class and you can't instantiate it without subclassing & actually implementing the _conditionals methods?

st--:
> It is hard to track technical changes/movements in the code. It would be nice to have guidance for a reviewer

To the best of my ability, I simply moved the code one-to-one from a gpflow.conditionals.conditional() implementation into the corresponding Posterior subclass's _fused_conditional() method.

> the code moving bit requires careful handling. I do want to trust tests in this situation. However, I don't have full confidence that we actually have tests in GPflow with good coverage for many different failing/successful scenarios.

Yes, that's a concern I share. In the course of working on this PR @johnamcleod discovered #1651 and #1653.

gpflow/posterior.py
st--: We could remove it again altogether. Happy to hear what you think on this, or whether we could improve on it further - I just wanted to take the opportunity of working on this to see if we can do better than having if:else branches based on whether q_sqrt is None, rank 2, or rank 3..

gpflow/posterior.py
st--: I was thinking being explicit and requiring `None` to be passed might be better. Adding type-hints, this would be
```suggestion
        q_sqrt: Union[None, TensorType],
```

st--: or we can move to explicitly passing in one of the DeltaDist/DiagDist/... classes

st--:
If we're all happy with this PR in-principle, should we change the merge target to `develop` and close #1620?

awav:
> Identify & remove all avoidable transposes!

As far as I know we don't have actual benchmarks that transpose make things worse in those (all) cases.

> Are we happy with the DeltaDist / DiagNormal / MvnNormal classes?

No :)

> Test to confirm it all works within tf.function() and allows swapping out of models without having to retrace/recompile the graph.

Either I don't understand what you mean or the answer is that it is not possible (you can't replace parts of a computational graph with another parts, a user have to recompile all the time)

> Test to confirm it all works within AOT compilation & gives expected speedup

AOT is simply a compilation step from a graph representation to a native code, why would you test it?

st--:
> > Are we happy with the DeltaDist / DiagNormal / MvnNormal classes?
>
> No :)

To clarify (following discussion) - it's not ideal as it is, but good enough to get the PR merged in. Let's rename them to _DeltaDist etc to ensure they're not seen as part of the public API.

st--:
> > Identify & remove all avoidable transposes!
>
> As far as I know we don't have actual benchmarks that transpose make things worse in those (all) cases.

This does have a measurable effect, but it's not crucial - this already speeds things up much more. Let's keep track of it in an issue for potential future work. Though ideally this issue will be solved by smarter linear algebra compilers @awav is working on separately.

gpflow/conditionals/conditionals.py
```diff
@@ -66,13 +66,12 @@ def _conditional(
...
-    fmean, fvar = base_conditional(
-        Kmn, Kmm, Knn, f, full_cov=full_cov, q_sqrt=q_sqrt, white=white
-    )  # [N, R],  [R, N, N] or [N, R]
-    return fmean, expand_independent_outputs(fvar, full_cov, full_output_cov)
+    posterior_class = get_posterior_class(kernel, inducing_variable)
+
+    posterior = posterior_class(
+        kernel, inducing_variable, f, q_sqrt, whiten=white, mean_function=None, precompute=False
+    )
+    return posterior.fused_predict_f(Xnew, full_cov=full_cov, full_output_cov=full_output_cov)
```
st--, line 74: Fallback to enable conditionals() to still work without users having to register both get_posterior_class() *and* conditional().

Questions:
1. should we add a deprecation warning when users try to register on conditional() directly?
2. should we just remove the explicit multi-output conditional() implementations (which now do exactly the same as this fallback)?

gpflow/posteriors.py
```diff
@@ -0,0 +1,536 @@
...
+        self.alpha, self.Qinv = self._precompute()
+
+    def update_cache_with_variables(self):
+        alpha, Qinv = self._precompute()
+        if isinstance(self.alpha, Parameter) and isinstance(self.Qinv, Parameter):
+            self.alpha.assign(alpha)
+            self.Qinv.assign(Qinv)
+        else:
+            self.alpha = Parameter(alpha, trainable=False)
+            self.Qinv = Parameter(Qinv, trainable=False)
```
st--, line 121: Not tested yet.

gpflow/posteriors.py
```diff
@@ -0,0 +1,536 @@
...
+        if full_cov == full_output_cov:
+            new_shape = (N * K, N * K) if full_cov else (N * K,)
+            Kff = tf.reshape(Kff, new_shape)
+
+        N = tf.shape(Xnew)[0]
+        K = tf.shape(Kuf)[-1] // N
+
+        mean = tf.matmul(Kuf, self.alpha, transpose_a=True)
+        if Kuf.shape.ndims == 3:
+            mean = tf.linalg.adjoint(tf.squeeze(mean, axis=-1))
```
st--, line 399: This path seems to not be tested yet, can we think of an easy way of adding it?

tests/gpflow/models/test_svgp_posterior.py
```diff
@@ -0,0 +1,77 @@
...
+from typing import Tuple
+
+import numpy as np
+import pytest
+import tensorflow as tf
+
+import gpflow
+from gpflow.models.svgp import NewSVGP, OldSVGP
+
+input_dim = 7
```
awav, line 11: Hide the global variable?

```suggestion
__input_dim = 7
```

awav, line 11: I'd make it a private module (global) variable.

```suggestion
input_dim = 7
```

```suggestion
__input_dim = 7
```

tests/gpflow/optimizers/test_scipy.py
```diff
@@ -63,4 +63,7 @@ def get_values(model):
     # The tolerance of the following test had to be loosened slightly from atol=1e-15
     # due to the changes introduced by PR #1213, which removed some implicit casts
     # to float32. With TF 2.4 / TFP 0.12, we had to further increase atol from 1e-14.
+    np.testing.assert_allclose(
+        get_values(m1), get_values(m2), rtol=4e-14, atol=2e-14
+    )  # for Predictor
```
awav, line 68: What do you mean by "for Predictor"?

tests/gpflow/posteriors/test_posteriors.py
```diff
@@ -0,0 +1,503 @@
...
+
+        def fn_2(n_inducing_points, num_latent_gps):
+            # qsqrt: [L, M, M]
+            return tf.linalg.band_part(
+                tf.random.normal(
+                    (num_latent_gps, n_inducing_points, n_inducing_points), dtype=tf.float64
+                ),
+                -1,
+                0,
+            )
```
awav, line 67:
```suggestion
            shape = (num_latent_gps, n_inducing_points, n_inducing_points)
            return tf.linalg.band_part(tf.random.normal(shape, dtype=tf.float64), -1, 0)
```

RELEASE.md
```diff
@@ -35,35 +35,48 @@ This release contains contributions from:
...
+PR #1636, which allows for a significant speed-up of post-training predictions
+with the `SVGP` model (partially resolving #1599).
+
+* For end-users, by default nothing changes; see Breaking Changes below if you
+  have written your own _implementations_ of `gpflow.conditionals.conditional`.
+* After training an `SVGP` model, you can call `model.posterior()` to obtain a
+  Posterior object that precomputes all quantities not depending on the test
+  inputs (e.g. Choleskty of Kuu), and provides a `posterior.predict_f()` method
+  that reuses these cached quantities. `model.predict_f` computes exactly the
+  same quantities as before and does NOT give any speed-up.
```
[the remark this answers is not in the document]
st--:
```suggestion
  same quantities as before and does **not** give any speed-up.
```

doc/source/notebooks/advanced/variational_fourier_features.pct.py
```diff
@@ -230,98 +236,167 @@ def gauss_kl_vff(q_mu, q_sqrt, K):
...
+                # won't work  # make ticket for this?
+                raise NotImplementedError
+            elif q_sqrt.get_shape().ndims == 3:
+                # L = tf.matrix_band_part(tf.transpose(q_sqrt, (2, 0, 1)), -1, 0)  # K x M x M
+
+                # K x M x N
+                # A_tiled = tf.expand_dims(A.get(), 0) * tf.ones((num_func, 1, 1), dtype=float_type)
+
+                # LTA = tf.matmul(L, A_tiled, transpose_a=True)  # K x M x N
+                # TODO the following won't work for K > 1
```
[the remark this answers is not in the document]
st--: they were there before, I'd be happy for someone else to clean up the notebook further but it's beyond the scope of this PR

gpflow/models/svgp.py
```diff
@@ -170,3 +170,93 @@ def predict_f(self, Xnew: InputData, full_cov=False, full_output_cov=False) -> M
...
+
+        For faster (cached) prediction, predict directly from the posterior object, i.e.,:
+            model.posterior().predict_f(Xnew, ...)
+        """
+        return self.posterior(posteriors.PrecomputeCacheType.NOCACHE).fused_predict_f(
+            Xnew, full_cov=full_cov, full_output_cov=full_output_cov
+        )
+
+
+class SVGP(NewSVGP):
```
st--, line 261:
```suggestion
class SVGP(SVGP_with_posterior):
```

tests/gpflow/models/test_svgp_posterior.py
st--: it's just a test though - what would you hide it from, and why?

tests/gpflow/models/test_svgp_posterior.py
```diff
@@ -0,0 +1,77 @@
...
+@pytest.mark.parametrize("white", [True, False])
+@pytest.mark.parametrize(
+    "multioutput",
+    [None]
+    + list(
+        product(
+            ("LinearCoregionalization", "SeparateIndependent", "SharedIndependent"),
+            ("SeparateIndependent", "SharedIndependent"),
+        )
+    ),
```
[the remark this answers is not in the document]
st--: Ouh that's a great suggestion!

tests/gpflow/models/test_svgp_posterior.py
```diff
@@ -0,0 +1,77 @@
+from itertools import product
+from typing import Tuple
+
+import numpy as np
+import pytest
+import tensorflow as tf
+
+import gpflow
+from gpflow.models.svgp import NewSVGP, OldSVGP
```
st--, line 9:
```suggestion
from gpflow.models.svgp import SVGP_with_posterior, SVGP_deprecated
```

tests/gpflow/models/test_svgp_posterior.py
```diff
@@ -0,0 +1,77 @@
...
+            raise NotImplementedError
+
+    lik = gpflow.likelihoods.Gaussian(0.1)
+    q_mu = np.random.randn(M, L)
+    if q_diag:
+        q_sqrt = np.random.randn(M, L) ** 2
+    else:
+        q_sqrt = np.tril(np.random.randn(L, M, M))
+    mold = OldSVGP(k, lik, Z, q_diag=q_diag, q_mu=q_mu, q_sqrt=q_sqrt, whiten=whiten)
+    mnew = NewSVGP(k, lik, Z, q_diag=q_diag, q_mu=q_mu, q_sqrt=q_sqrt, whiten=whiten)
```
st--, line 50:
```suggestion
    mold = SVGP_deprecated(k, lik, Z, q_diag=q_diag, q_mu=q_mu, q_sqrt=q_sqrt, whiten=whiten)
    mnew = SVGP_with_posterior(k, lik, Z, q_diag=q_diag, q_mu=q_mu, q_sqrt=q_sqrt, whiten=whiten)
```

tests/gpflow/models/test_svgp_posterior.py
```diff
@@ -0,0 +1,77 @@
...
+    ),
+)
+def test_old_vs_new_svgp(q_diag, white, multioutput):
+    mold, mnew = make_models(q_diag=q_diag, whiten=white, mo=multioutput)
+
+    X = np.random.randn(100, input_dim)
+    Xt = tf.convert_to_tensor(X)
+
+    for full_cov in (True, False):
+        for full_output_cov in (True, False):
```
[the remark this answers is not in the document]
st--: Could do, but that would slow down the tests as there'd be four times as many, even though we don't need to re-create the models, so I'd prefer to keep it as is

tests/gpflow/optimizers/test_scipy.py
st--, line 68:
```suggestion
```
I don't think it does anything actually

st--: removed line - not actually needed

st--: (this was left over from tests where SVGP.predict_f wasn't using the fused_predict_f code path but the precomputed one, which slightly deteriorated performance; probably a merge remnant)

gpflow/posterior.py
st--: addressed

gpflow/models/svgp.py
```diff
@@ -170,3 +170,93 @@ def predict_f(self, Xnew: InputData, full_cov=False, full_output_cov=False) -> M
...
+          quantities and stores them as tensors (which allows differentiating
+          through the prediction). This is the default.
+        - PrecomputeCacheType.VARIABLE (or `"variable"`): Precomputes the cached
+          quantities and stores them as variables, which allows for updating
+          their values without changing the compute graph (relevant for AOT
+          compilation).
+        - PrecomputeCacheType.NOCACHE (or `"nocache"` or `None`): Avoids
+          immediate cache computation.  This is useful for avoiding extraneous
+          computations when you only want to call the posterior's
+          fused_predict_f method.
```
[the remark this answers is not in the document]
st--: I think with the state the GPflow documentation is in at the moment it's more helpful to users if this documentation is at the point of contact where you actually need it.

awav (approved):
LGTM

st--:
Most likely not perfect yet, but let's merge it and see what happens. :-)

PR #1637, opened 2021-02-03
Mypy build fix

st--:
Mypy now recursively checks all directories. The notebooks do not pass
the type check (for one, they tend to redefine objects!), so we simply
restrict the type-checking to the gpflow source and tests explicitly.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,7 +38,7 @@ format-check:
 	isort --check-only $(ISORT_CONFIG) $(ISORT_TARGETS)
 
 type-check:
-	mypy .
+	mypy gpflow tests
 
 test:
 	pytest -n auto --dist loadfile -v --durations=10 tests/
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -38,7 +38,7 @@ format-check:
 	isort --check-only $(ISORT_CONFIG) $(ISORT_TARGETS)
 
 type-check:
-	mypy .
+	mypy gpflow tests
```
joelberkeley, line 41: there's also a mention of mypy in the contrib guidelines.
> Use `pylint` and `mypy` for formatting and _type checking_
I don't know if you want to clarify there that the notebooks don't need to pass. Whatever you think makes more sense

joelberkeley (approved):

joelberkeley-secondmind (approved):

PR #1640, opened 2021-02-12
Remove positive transform for anisotropic lengthscales

### LINKED ISSUE #1639: Anisotropic lengthscales
Currently, by default, all stationary kernels impose a positive transform on the lengthscales Parameter. For the case of AnisotropicStationary kernels, this reduces the viable parameter space by a factor of 2^(D-1), where D is the number of input dimensions. So it is acceptable in the case of a 1D problem, but in higher dimensions we would miss out. For example, for the Cosine kernel there is a significant difference between lengthscales of [+1, +1], and [+1, -1], they correspond to waves propagating in perpendicular directions.

frgsimpson:
Resolves #1639


gpflow/kernels/stationaries.py
```diff
@@ -115,6 +115,12 @@ class AnisotropicStationary(Stationary):
     input dimension.
     """
 
+    def __init__(self, *args, **kwargs):
```
vdutor, line 118: Best to copy the base class `__init__` parameters for better docstrings and autocomplete.

gpflow/kernels/stationaries.py
```diff
@@ -115,6 +115,12 @@ class AnisotropicStationary(Stationary):
...
+        super().__init__(*args, **kwargs)
+
+        if tf.size(self.lengthscales) > 1:
```
vdutor, line 121: I would prefer to do it like we do it in `ard`:
```suggestion
  if self.lengthscales.shape.ndims > 1
```

vdutor (approved):
LGTM

gpflow/kernels/stationaries.py
frgsimpson: Nice, I've updated to just use self.ard itself

gpflow/kernels/stationaries.py
frgsimpson: Done!

st--:
Superseded by #1843

PR #1642, opened 2021-02-15
Fix API for using SciPy optimizer with monitoring.

vatsalaggarwal:
**PR type:** enhancement + doc improvement

**Related issue(s)/PRs:** https://github.com/GPflow/GPflow/issues/1449, https://github.com/GPflow/GPflow/pull/1558

## Summary

**Proposed changes**

* The SciPy optimizer currently misbehaves when used with Monitoring objects. The changes remove the 'variables' and 'step' parameters from being used in SciPy with this object. I tried to remove these parameters completely, but it seems they are being used for custom callback objects being used in other parts of the code.

Having a separate monitoring class for SciPy as in #1558. After discussion with STJohn, it was decided that this is likely to cause confusion.

### Minimal working example

See `doc/source/notebooks/basics/monitoring.pct.py`

## PR checklist

### DIFF
--- a/doc/source/notebooks/basics/monitoring.pct.py
+++ b/doc/source/notebooks/basics/monitoring.pct.py
@@ -134,14 +134,16 @@ for step in range(optimisation_steps):
 
 
 # %% [markdown]
-# For optimal performance, we can also wrap the monitor call inside `tf.function`:
+# ### For optimal performance, we can also wrap the monitor call inside `tf.function`:
 
 # %%
 opt = tf.optimizers.Adam()
 
-log_dir = f"{log_dir}/compiled"
-model_task = ModelToTensorBoard(log_dir, model)
-lml_task = ScalarToTensorBoard(log_dir, lambda: model.training_loss(), "training_objective")
+log_dir_compiled = f"{log_dir}/compiled"
+model_task = ModelToTensorBoard(log_dir_compiled, model)
+lml_task = ScalarToTensorBoard(
+    log_dir_compiled, lambda: model.training_loss(), "training_objective"
+)
 # Note that the `ImageToTensorBoard` task cannot be compiled, and is omitted from the monitoring
 monitor = Monitor(MonitorTaskGroup([model_task, lml_task]))
 
@@ -163,3 +165,27 @@ for i in tf.range(optimisation_steps):
 
 # %% [markdown]
 # When opening TensorBoard, you may need to use the command `tensorboard --logdir . --reload_multifile=true`, as multiple `FileWriter` objects are used.
+
+# %% [markdown]
+# ### Scipy Optimization monitoring
+#
+# Note that if you want to use the `Scipy` optimizer provided by GPflow, and want to monitor the training progress, then you need to simply replace
+# the optimization loop with a single call to its `minimize` method and pass in the monitor as a `step_callback` keyword argument:
+#
+
+# %%
+opt = gpflow.optimizers.Scipy()
+
+log_dir_scipy = f"{log_dir}/scipy"
+model_task = ModelToTensorBoard(log_dir_scipy, model)
+lml_task = ScalarToTensorBoard(log_dir_scipy, lambda: model.training_loss(), "training_objective")
+image_task = ImageToTensorBoard(log_dir_scipy, plot_prediction, "image_samples")
+
+monitor = Monitor(
+    MonitorTaskGroup([model_task, lml_task], period=1), MonitorTaskGroup(image_task, period=5)
+)
+
+# %%
+opt.minimize(training_loss, model.trainable_variables, step_callback=monitor)
+
+# %%
--- a/gpflow/optimizers/scipy.py
+++ b/gpflow/optimizers/scipy.py
@@ -19,10 +19,12 @@ import scipy.optimize
 import tensorflow as tf
 from scipy.optimize import OptimizeResult
 
+from ..monitor.base import Monitor
+
 __all__ = ["Scipy"]
 
 Variables = Iterable[tf.Variable]  # deprecated
-StepCallback = Callable[[int, Sequence[tf.Variable], Sequence[tf.Tensor]], None]
+StepCallback = Union[Callable[[int, Sequence[tf.Variable], Sequence[tf.Tensor]], None], Monitor]
 LossClosure = Callable[[], tf.Tensor]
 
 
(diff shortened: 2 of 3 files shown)


doc/source/notebooks/basics/monitoring.pct.py
```diff
@@ -163,3 +165,27 @@ def step(i):
 
 # %% [markdown]
 # When opening TensorBoard, you may need to use the command `tensorboard --logdir . --reload_multifile=true`, as multiple `FileWriter` objects are used.
+
+# %% [markdown]
+# ### Scipy Optimization monitoring
+#
+# Note that if you want to use the scipy optimizer, and want to monitor the training progress, then you need to simply replace
```
st--, line 172:
```suggestion
# Note that if you want to use the `Scipy` optimizer provided by GPflow, and want to monitor the training progress, then you need to simply replace
```

doc/source/notebooks/basics/monitoring.pct.py
```diff
@@ -163,3 +165,27 @@ def step(i):
...
+# the optimization loop with a single call to SciPy's `minimize` and pass in the monitor as a `step_callback` keyword-argument:
```
st--, line 173:
```suggestion
# the optimization loop with a single call to its `minimize` method and pass in the monitor as a `step_callback` keyword argument:
```

tests/gpflow/test_monitor.py
```diff
@@ -289,3 +289,9 @@ def tf_func(step):
 
     for step in tf.range(100):
         tf_func(step)
+
+
+def test_scipy_monitor(monitor, model):
+    opt = gpflow.optimizers.Scipy()
+
+    opt.minimize(model.training_loss, model.trainable_variables, step_callback=monitor)
```
st--, line 297: This test does a basic check whether the code _runs_ (doesn't blow up) but it doesn't, for example, check whether monitor actually got called. You can use mocks for this - we've not been using them across GPflow yet, though I think that'd make a lot of tests faster and easier to understand! There's an overview here: https://changhsinlee.com/pytest-mock/

would you be interested in having a go at figuring that out yourself, or would you prefer if I take over that bit?

st-- (approved):
Looks good to me, thanks for contributing:)

@vatsalaggarwal let me know whether you'd like to figure out mocking (see comment on the test you added) or whether you prefer me to do that separately.

We'll still have to figure out the typing. @joelberkeley do you happen to know the answer to this mypy question?

tests/gpflow/test_monitor.py
vatsalaggarwal: Oooh, good point! So one solution is that I can write a simple custom step_callback function that progresses a step counter. Then, I can query the step and check that the counter has increased... this will ensure that a monitor is actually being called... But, it won't check if everything works properly in this particular scenario...

I've had a quick look at this link, thanks! That's cool. For these tests, what would we mock?

gpflow/optimizers/scipy.py
```diff
@@ -121,8 +125,13 @@ def callback_func(
...
         def _callback(x: np.ndarray) -> None:
             nonlocal step
-            values = cls.unpack_tensors(variables, x)
-            step_callback(step, variables, values)
+
+            if isinstance(step_callback, Monitor):
+                step_callback(step)
+            else:
+                values = cls.unpack_tensors(variables, x)
+                step_callback(step, variables, values)
```
joelberkeley-secondmind, line 133: `step_callback` could be a function of one argument without being a `Monitor`, for example, for
```
def foo(x):
    return None
```
`isinstance(f, Monitor)` is False, but `f(step, variables, values)` will still fail. Does this answer the question? I don't think there is any (sensible) way to check at runtime for the number of parameters to a function, but if you only expect a `Monitor` or a 3-parameter function, you could just
```
StepCallback = Union[
    Callable[[int, Sequence[tf.Variable], Sequence[tf.Tensor]], None], Monitor
]
```
Have I understood correctly?

vatsalaggarwal: Yes, you have! That makes sense... thanks!

joelberkeley-secondmind: that said, would it not just be possible to wrap a monitor in a function that ignores some of it arguments, like
```
monitor = Monitor()

def callback(step, variables, values):
    return monitor(step)

optimizer.callback_func(..., callback)
```
or do you particularly care about avoiding the call to `unpack_tensors`

vatsalaggarwal: Do you mean something similar to https://github.com/GPflow/GPflow/pull/1558? I discussed with STJohn offline - this would cause users to change their code (or use a new API) when they used, for example, Adam as opposed to BFGS/L-BFGS-B...

With the given implementation, these details are hidden from the user?

Or, is there something we can do in the middle which doesn't change the API but also wraps the monitor?

joelberkeley-secondmind: ok, if there's a larger reason for it, sure

vatsalaggarwal: Cool, will keep it as-is then. Thanks for your comments!

tests/gpflow/test_monitor.py
vatsalaggarwal: Have made this change to check that the monitor does indeed get called...

tests/gpflow/test_monitor.py
```diff
@@ -21,6 +22,14 @@ class Data:
     num_steps = 2
 
 
+class DummyMonitor(MonitorTask):
+    def run(self, **kwargs):
+        pass
+
+    def get_internal_step(self):
+        return self.current_step
```
st--, line 30:
```suggestion
class DummyTask(MonitorTask):
    def run(self, **kwargs):
        pass
```
This name makes a bit clearer what we're doing. Also, in the test we can access `current_step` directly, no need for the method indirection.

tests/gpflow/test_monitor.py
```diff
@@ -31,6 +40,12 @@ def model():
     return gpflow.models.GPR(data, kernel, noise_variance=0.01)
 
 
+@pytest.fixture
+def dummy_monitor():
+    dm = DummyMonitor()
+    return Monitor(MonitorTaskGroup(dm, period=1),), dm
+
+
```
st--, line 48:
```suggestion
```
It's only used once, so I think the code will be easier to read without

tests/gpflow/test_monitor.py
```diff
@@ -289,3 +304,19 @@ def tf_func(step):
...
+def test_scipy_monitor_called(dummy_monitor, model):
+    opt = gpflow.optimizers.Scipy()
+
+    monitor, dummy_monitor_task = dummy_monitor
+
+    opt.minimize(model.training_loss, model.trainable_variables, step_callback=monitor)
+
+    assert dummy_monitor_task.get_internal_step() >= 1
```
st--, line 322:
```suggestion
def test_scipy_monitor_called(model):
    task = DummyTask()
    monitor = Monitor(MonitorTaskGroup(task, period=1))
    opt = gpflow.optimizers.Scipy()
    opt.minimize(model.training_loss, model.trainable_variables, step_callback=monitor)
    assert task.current_step > 1
```

PR #1643, opened 2021-02-22
One backprop step for Natgrad & Adam

vatsalaggarwal:
**PR type:**  new feature

**Related issue(s)/PRs:** 

## Summary

**Proposed changes**

Stochastic Variational Inference based GP models typically combine NatGrad to train the variational parameters along with Adam to train the non-variational parameters. However, currently, this occurs as two distinct forward-backward passes - one for each optimiser. This PR, instead, aims to reduce this to a single forward-backward pass to improve computational efficiency.

Thus, we will take a step in the variational parameters space and the hyperparameters space at the same time. This has implications for the behaviour of NatGrad - for example, in the Gaussian likelihood fully Variational case, the NatGrad step (with gamma=1) will no longer be "exact". The variational parameters computed would be exactly optimal with respect to the previous iterations hyperparameters, but no longer with the new hyperparameters.

Initial experiment on a very small toy problem, attached as a notebook, suggests that this nonetheless could be useful. The intuition here is that if we begin with a good setting of the variational parameters (perhaps using some warm-up) and then apply the joint optimiser, then we are likely to end up in the correct place. This is because, as we get closer to the optima, the steps in hyperparameter space are likely to get small, so our variational parameters are likely not to be too far apart from the optima. If this turns out be problematic, it is always possible to finish off the optimisation process with a few steps focusing only on the variational parameter space.

This intuition remains to be robustly and empirically tested. Nonetheless, the hope is that by making this code available via an easy API, the wider audience could already start to benefit from speed-ups in training times.

### Minimal working example

c.f. attached notebook example.

## PR checklist

### DIFF
--- a/gpflow/optimizers/natgrad.py
+++ b/gpflow/optimizers/natgrad.py
@@ -120,6 +120,107 @@ class XiSqrtMeanVar(XiTransform):
         return natural_to_meanvarsqrt(nat1, nat2)
 
 
+def _assert_shapes(q_mu, q_sqrt):
+    tf.debugging.assert_shapes(
+        [(q_mu, ["M", "L"]), (q_sqrt, ["L", "M", "M"]),]
+    )
+
+
+def natgrad_apply_gradients(
+    gamma: Scalar,
+    q_mu_grad: tf.Tensor,
+    q_sqrt_grad: tf.Tensor,
+    q_mu: Parameter,
+    q_sqrt: Parameter,
+    xi_transform: XiTransform,
+):
+    """
+    This function does the backward step on the q_mu and q_sqrt parameters,
+    given the gradients of the loss function with respect to their unconstrained
+    variables. I.e., it expects the arguments to come from
+
+        with tf.GradientTape() as tape:
+            loss = loss_function()
+        q_mu_grad, q_mu_sqrt = tape.gradient(loss, [q_mu, q_sqrt])
+
+    (Note that tape.gradient() returns the gradients in *unconstrained* space!)
+
+    Implements equation [10] from
+
+    @inproceedings{salimbeni18,
+        title={Natural Gradients in Practice: Non-Conjugate Variational Inference in Gaussian Process Models},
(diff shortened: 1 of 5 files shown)


gpflow/optimizers/natgrad.py
```diff
@@ -120,6 +120,107 @@ def naturals_to_xi(nat1, nat2):
...
+        # we need these to calculate the relevant gradients
+        meanvarsqrt = expectation_to_meanvarsqrt(eta1, eta2)
+
+        if not isinstance(xi_transform, XiNat):
+            nat1, nat2 = meanvarsqrt_to_natural(q_mu, q_sqrt)
+            xi1_nat, xi2_nat = xi_transform.naturals_to_xi(nat1, nat2)
+            dummy_tensors = tf.ones_like(xi1_nat), tf.ones_like(xi2_nat)
+            with tf.GradientTape(watch_accessed_variables=False) as forward_tape:
+                forward_tape.watch(dummy_tensors)
+                dummy_gradients = tape.gradient(
```
vatsalaggarwal, line 195: I'm not sure why it's showing that L190-L195 aren't covered by tests... I have checked and these lines do get reached by the current suite of tests? @st--

gpflow/optimizers/natgrad.py
```diff
@@ -120,6 +120,107 @@ def naturals_to_xi(nat1, nat2):
...
+                    [xi1_nat, xi2_nat], [nat1, nat2], output_gradients=dummy_tensors
+                )
+
+    # 2) the chain rule to get ∂L/∂η, where η (eta) are the expectation parameters
+    dL_deta1, dL_deta2 = tape.gradient(
+        meanvarsqrt, [eta1, eta2], output_gradients=[dL_dmean, dL_dvarsqrt]
+    )
+
+    if not isinstance(xi_transform, XiNat):
+        nat_dL_xi1, nat_dL_xi2 = forward_tape.gradient(
```
vatsalaggarwal, line 205: I'm not sure why it's showing that L205 aren't covered by tests... I have checked and these lines do get reached by the current suite of tests? @st-- could you help?

PR #1649, opened 2021-03-15
fix bug (that masked test failure!) in test_method_equivalence

st--:
In working on #1620 / #1636 I noticed that
* test_method_equivalence does model optimisation *as part of test collection*, not just in the tests - if the optimisation fails, this breaks the test suite as a whole! and slows down all tests, even if you don't actually run this specific one
* and there was a bug in `_create_approximate_models()`: it doesn't actually construct an SVGP model with `whiten=False` - this test case does not actually pass though; this is due to *significantly* worse convergence than any of the other models - we can get the test to pass by optimizing more on the non-whitened SVGP model.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,8 +5,8 @@ runtest: &runtest
         name: Install GPflow
         command: |
           pip install -U pip
-          pip install --use-feature=2020-resolver -e .
-          pip install --use-feature=2020-resolver -r tests_requirements.txt
+          pip install -e .
+          pip install -r tests_requirements.txt
     - run:
         name: Run tests
         command: |
@@ -28,7 +28,7 @@ jobs:
       - run:
           name: Check installed dependencies are compatible
           command: |
-            pip install --use-feature=2020-resolver --progress-bar=off -e .
+            pip install --progress-bar=off -e .
             pip check -vvv
             python -c "import gpflow"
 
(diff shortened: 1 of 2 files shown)


st--:
Hm, it was passing locally... maybe have to up the number of optimisation steps for non-whitened SVGP some more...

.circleci/config.yml
```diff
@@ -5,8 +5,8 @@ runtest: &runtest
         name: Install GPflow
         command: |
           pip install -U pip
-          pip install --use-feature=2020-resolver -e .
-          pip install --use-feature=2020-resolver -r tests_requirements.txt
```
awav, line 9: Is that a necessary part of the PR? What is the reasoning for removing it?

awav (approved):
```
test_method_equivalence does model optimisation as part of test collection, not just in the tests - if the optimisation fails, this breaks the test suite as a whole!
```

Correct me, if I understand it wrong: optimisation fails for all tests in the file scope and the slow down of tests is also for a give file scope. Right?

LGTM, I left another comment.

.circleci/config.yml
st--: I could've made a separate PR for that but seemed a bit ridiculous. When looking at the CircleCI output I saw that pip's most recent version is now warning that as of pip 21.0 this will become an error (which would fail all our builds then on).

PR #1650, opened 2021-03-19
Clean up utilities subpackage

st--:
For now, just separates module-traversion utilities from other miscellaneous. Including proper deprecation forwarding. Thinking if we should rename "utilities" to something more meaningful now that it's _just_ the module-traversion tools.... :thinking:


awav (commented):
Hi!

1. What is the reason for splitting the utilities?
2. Git doesn't show if methods changed or not, can you confirm that you didn't change methods (it would be easier w/o renaming/splitting)
3. You added tests, could you explain what they do?

st--:
>     1. What is the reason for splitting the utilities?

I have found the utilities submodule very confusing for a long time, with too many functions doing very different things mixed all in one place. I wanted to separate at least the "module traversal" utilities from the others. I think this makes the code easier to understand.

>     2. Git doesn't show if methods changed or not, can you confirm that you didn't change methods (it would be easier w/o renaming/splitting)

Yes, I've just moved code around. You can use `git diff develop:gpflow/utilities/utilities.py gpflow/utilities/misc.py` to compare different paths across branches. If you have a GUI tool like meld installed you can also replace `git diff` with `git difftool` which makes it even easier to spot that compared to develop the two edited files only have removals of the functions respectively left in the other module.

>     3. You added tests, could you explain what they do?

Yeah, turns out not all the utilities functions were properly covered by tests, so once I split them into separate files the code coverage was no longer hitting the required level for the build to pass. So I added some more tests.

st--:
I'd prefer to rename `gpflow.utilities.utilities` to `gpflow.utilities.traversal` or `gpflow.utilities.module_traversal` or something like that as well, that gives it a more meaningful name.

awav (approved):
`gpflow.utilities.traversal` is okay. LGTM

Issue #1600, opened 2020-10-22
Handling unconnected gradients in Scipy optimizer


st--:
# Feature request

Our Scipy optimizer wrapper should explicitly handle unconnected gradients.

## Motivation

Currently, the GPflow Scipy optimizer wrapper (`gpflow.optimizers.Scipy`) simply errors (with one of the typically unhelpful TensorFlow error messages) when one of the passed-in `variables` is not connected to the loss to be minimized. This can easily happen, for example when using an SVGP model with a `Constant` kernel (which might be used as a very simple baseline) but having inducing locations not explicitly set to non-trainable. This could also happen erroneously, e.g. when calling `Scipy().minimize(model1.training_loss, model2.trainable_variables)`.

## Proposal

**Describe the solution you would like**
An explicit error message when the variables are not connected to the loss (and hence the `grads` list contains `None` entries).

**What alternatives have you considered?**
We could also simply change the call for the gradient computation to use `unconnected_gradients=tf.UnconnectedGradients.ZERO`. This might silently hide bug cases like `minimize(model1.training_loss, model2.trainable_variables)`, so perhaps this option could be an extra flag to the minimize() call?

antonykamp:
Hey,
If this issue is still active, I would be happy to work on it. 
One question: is checking whether `gradient` returns a zero tensor with `tf.math.count_nonzero` appropriate?

st--:
Hi @antonykamp ! thanks for your interest. Yes, nothing has changed about this issue - if you try to optimize a loss that's not connected to the variables you pass in, you still get an unhelpful error message.
Regarding your question, can you expand a bit more on how you'd want to use count_nonzero?

antonykamp:
> Can you expand a bit more on how you'd want to use count_nonzero?

Of course! When using `unconnected_gradients=tf.UnconnectedGradients.ZERO` and the loss is not connected to the variables, the associated item in the gradient equals a zero tensor. GPflow could test for this by counting the nonzeros of each item. Should the situation arise that every value equals zero, GPflow warns the user, eg.

https://github.com/antonykamp/GPflow/blob/599bc49ed902c2296a550c8f0409dcc9bdcef900/gpflow/optimizers/scipy.py#L181

st--:
Ah! You mean to give a warning, yet continue running?
[...]

PR #1655, opened 2021-03-27
Added warning for unconnected gradients

antonykamp:
**PR type:** enhancement

**Related issue(s)/PRs:** #1600 

## Summary

**Proposed changes**

* Throw a warning if the computation of loss and gradient in the scipy wrapper returns an object consisting exclusively of zeros (-> gradients are possibly unconnected).
* Let the user choose what the computation of loss and gradient in the scipy wrapper returns if the gradients are unconnected (default: zero tensor).

I considered a `[None]` as default case of the second change. I rejected this idea, because this throws an error in most cases. Otherwise, some users could take advantage of this "feature" at the moment because it was the default return value before this PR (that's why I've added the possibility to choose).

### Minimal working example
(Example from `test/gpflow/optimizers/test_scipy.py`)

```python

import numpy as np
import gpflow

class Datum:
    X = rng.rand(20, 1) * 10
    Y = np.sin(X) + 0.9 * np.cos(X * 1.6) + rng.randn(*X.shape) * 0.8
    Y = np.tile(Y, 2)  # two identical columns

def _create_full_gp_model():
    return gpflow.models.GPR(
        (Datum.X, Datum.Y),
        kernel=gpflow.kernels.SquaredExponential(),
        mean_function=gpflow.mean_functions.Constant(),
    )

# create two different models
m1 = _create_full_gp_model()
m2 = _create_full_gp_model()

# try to optimize two separate models -> gradients are unconnected!
opt.minimize(
    m1.training_loss,
    variables=m2.trainable_variables,
    options=dict(maxiter=50),
    compile=False,
)

# >>> RuntimeWarning: At least one of the passed-in variables is not connected to the loss to be minimized. Therefore, this entry of the gradient is a zero tensor.
```

## PR checklist

### Release notes

**Fully backwards compatible:** not sure (see _What alternatives have you considered?_)

* Added warning for unconnected gradients and associated tests

### DIFF
--- a/gpflow/optimizers/scipy.py
+++ b/gpflow/optimizers/scipy.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import warnings
 from typing import Callable, Iterable, List, Optional, Sequence, Tuple, TypeVar, Union
 
 import numpy as np
@@ -29,6 +30,9 @@ LossClosure = Callable[[], tf.Tensor]
 
 
 class Scipy:
+    def __init__(self):
+        self.gradient_already_checked = False
+
     def minimize(
         self,
         closure: LossClosure,
@@ -36,6 +40,7 @@ class Scipy:
         method: Optional[str] = "L-BFGS-B",
         step_callback: Optional[StepCallback] = None,
         compile: bool = True,
+        unconnected_gradients: str = tf.UnconnectedGradients.ZERO,
         **scipy_kwargs,
     ) -> OptimizeResult:
         """
@@ -59,6 +64,10 @@ class Scipy:
             compile: If True, wraps the evaluation function (the passed `closure`
                 as well as its gradient computation) inside a `tf.function()`,
                 which will improve optimization speed in most cases.
+            unconnected_gradients: A value that can either hold 'none' or 'zero',
+                which will influence the handling when one of the passed-in variables
+                is not connected to the loss to be minimized. As 'zero'  (default)
+                the entry of the gradient equals a zero tensor, [None] otherwise.
 
             scipy_kwargs: Arguments passed through to `scipy.optimize.minimize`
                 Note that Scipy's minimize() takes a `callback` argument, but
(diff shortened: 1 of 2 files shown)


antonykamp:
This is the follow-up-PR of the discussion in #1600.

gpflow/optimizers/scipy.py
```diff
@@ -97,20 +114,34 @@ def initial_parameters(cls, variables: Sequence[tf.Variable]) -> tf.Tensor:
 
     @classmethod
     def eval_func(
-        cls, closure: LossClosure, variables: Sequence[tf.Variable], compile: bool = True
+        cls,
+        closure: LossClosure,
+        variables: Sequence[tf.Variable],
+        compile: bool = True,
+        unconnected_gradients: str = tf.UnconnectedGradients.ZERO,
```
jesnie, line 121: Are we sure this is the behaviour we want? Maybe it would be safer to default to `None`, so that there is an obvious failure if the unconnected gradient is indeed a bug? And users can pick `ZERO` if they're ok with a bit more risk?

gpflow/optimizers/scipy.py
```diff
@@ -97,20 +114,34 @@ def initial_parameters(cls, variables: Sequence[tf.Variable]) -> tf.Tensor:
...
             return loss, cls.pack_tensors(grads)
 
         if compile:
             _tf_eval = tf.function(_tf_eval)
 
         def _eval(x: np.ndarray) -> Tuple[np.ndarray, np.ndarray]:
             loss, grad = _tf_eval(tf.convert_to_tensor(x))
+
+            if not gradient_already_checked and any(
+                [tf.math.count_nonzero(gr) == 0 for gr in grad]
```
jesnie, line 138: I don't think this technically is correct? You could have a zero-gradient for perfectly legitimate reasons? Also, we probably need to handle the case where the user sets `unconnected_gradients = tf.UnconnectedGradients.NONE`?

gpflow/optimizers/scipy.py
```diff
@@ -97,20 +114,34 @@ def initial_parameters(cls, variables: Sequence[tf.Variable]) -> tf.Tensor:
...
+            ):
+                warnings.warn(
+                    "At least one of the passed-in variables is not connected to the loss to be minimized. Therefore, this entry of the gradient is a zero tensor.",
```
jesnie, line 141: Can we do something to tell the user exactly which gradients it is that is unconnected?

jesnie (commented):
Hi.
I'm sorry my colleagues have forgotten to look at this for a very long time.
Are you still up for working on it? Otherwise I can take this over.

antonykamp:
hey,
I'm sorry too: Currently, I don't have the time to take a look into it again :( Maybe later, but feel free to take it over :D

jesnie:
I'm closing this as it has been rendered redundant by #1745.

Issue #1567, opened 2020-09-16
Named mean and variance


joelberkeley-pio:
# Feature request

It might be really nice to have the mean and variance returned by `GPModel.predict_f` and `GPModel.predict_y` to be `typing.NamedTuple`s.

## Motivation

This has two advantages. The main one is being able to calculate and use the mean _or_ variance straightforwardly in one line, and in a clear, safe manner. Rather than
```
mean, _ = model.predict_f(...)
foo(mean)
```
which is a little verbose, or
```
foo(model.predict_f(...)[0])
```
which is prone to error and hard to read, you can do
```
foo(model.predict_f(...).mean)
```
The other advantage is that it's a little safer. Using `res.mean` is more explicit and less prone to error than `res[0]` or `mean, _ = res`.

## Proposal

`MeanAndVariance` becomes a `typing.NamedTuple` like
```
class MeanAndVariance(NamedTuple):
    mean: tf.Tensor
    variance: tf.Tensor
```
Since `NamedTuple` is a typed version of `namedtuple`, which returns a subclass of `tuple`, I'm pretty sure this change is backwards compatible

**What alternatives have you considered?**

None

**Are you willing to open a pull request?** (We really appreciate contributions!)

Maybe, not just now though

antonykamp:
Hey,
If this is still a requested feature, I would be really happy to work on it too :)

st--:
Yes, that'd be great!

PR #1657, opened 2021-03-28
predict_f and predict_y as NamedTuple

antonykamp:
**PR type:** enhancement

**Related issue:** #1567 

## Summary

**Proposed changes**

* By requesting `predict_f` or `predict_y`, GPflow returns an instance of `MeanAndVariance`, a subclass of NamedTuple. By that, the user can access mean and variance individually.  
* You can find a hint to this enhancement in the example "Basic (Gaussian likelihood) GP regression model".

I've considered shorter names for`.mean` and `.variance` but rejected this idea because the current resulting code is a) more readable b) already shorter than before.
### Minimal working example

```python
import gpflow
import numpy as np

data = rng.randn(100, 2), rng.randn(100, 1)
Xtest, _ = rng.randn(Ntest, D), rng.randn(Ntest, 1)
kernel = Matern32() + gpflow.kernels.White()
model_gp = gpflow.models.GPR(data, kernel=kernel)

mu_f = model_gp.predict_f(Xtest).mean # instead of 'model.predict_f(Xtest)[0]' or 'mu_f, _ = model.predict_f(Xtest)'
var_y = model_gp.predict_y(Xtest).variance # instead of 'model.predict_y(Xtest)[1]' or '_, var_y = model.predict_y(Xtest)'
```

## PR checklist

### Release notes

* MeanAndVariance as NamedTuple
* Added test about backwards compatibility
* Added hint in docs, format

(Is likely to be rebased)

### DIFF
--- a/doc/source/notebooks/basics/regression.pct.py
+++ b/doc/source/notebooks/basics/regression.pct.py
@@ -180,7 +180,7 @@ _ = plt.xlim(-0.1, 1.1)
 
 
 # %% [markdown]
-# Moreover, you can get the mean and variance data individually for example as `m.predict_f(xx).mean` instead of `m.predict_f(xx)[0]`.
+# Moreover, you can get the mean and variance predictions individually, using for `m.predict_f(xx).mean` instead of `m.predict_f(xx)[0]` and `m.predict_f(xx).variance` instead of `m.predict_f(xx)[1]`.
 
 # %% [markdown]
 # ## GP regression in higher dimensions
--- a/doc/source/notebooks/basics/regression.pct.py
+++ b/doc/source/notebooks/basics/regression.pct.py
@@ -180,7 +180,7 @@ _ = plt.xlim(-0.1, 1.1)
 
 
 # %% [markdown]
-# Moreover, you can get the mean and variance predictions individually, using for `m.predict_f(xx).mean` instead of `m.predict_f(xx)[0]` and `m.predict_f(xx).variance` instead of `m.predict_f(xx)[1]`.
+# Moreover, you can get the mean and variance predictions individually, using for `m.predict_f(xx).mean` instead of `m.predict_f(xx)[0]` and `m.predict_f(xx).variance` instead of `m.predict_f(xx)[1]`. If you pass `full_cov=True` or `full_output_cov=True` to `m.predict_f`, you will get information about the covariance instead of variance prediction. You can access it with `m.predict_f(xx).covariance`.
 
 # %% [markdown]
 # ## GP regression in higher dimensions
--- a/doc/source/notebooks/basics/regression.pct.py
+++ b/doc/source/notebooks/basics/regression.pct.py
@@ -180,6 +180,9 @@ _ = plt.xlim(-0.1, 1.1)
 
 
 # %% [markdown]
+# Moreover, you can get the mean and variance data individually for example as `m.predict_f(xx).mean` instead of `m.predict_f(xx)[0]`.
+
+# %% [markdown]
 # ## GP regression in higher dimensions
 #
 # Very little changes when the input space has more than one dimension. By default, the `lengthscales` is an isotropic (scalar) parameter. It is generally recommended that you allow to tune a different lengthscale for each dimension (Automatic Relevance Determination, ARD): simply initialize `lengthscales` with an array of length $D$ corresponding to the input dimension of `X`.  See [Manipulating kernels](../advanced/kernels.ipynb) for further information.
--- a/gpflow/models/model.py
+++ b/gpflow/models/model.py
@@ -261,8 +261,6 @@ class GPModel(BayesianModel):
             )
 
         f_mean, f_var = self.predict_f(Xnew, full_cov=full_cov, full_output_cov=full_output_cov)
-        if full_cov or full_output_cov:
-            return MeanAndCovariance(*self.likelihood.predict_mean_and_var(f_mean, f_var))
         return MeanAndVariance(*self.likelihood.predict_mean_and_var(f_mean, f_var))
 
     def predict_log_density(
--- a/gpflow/models/model.py
+++ b/gpflow/models/model.py
@@ -27,7 +27,7 @@ from .training_mixins import InputData, RegressionData
 
 
 class MeanAndVariance(NamedTuple):
-    """ NamedTuple to access mean- and variance-function separately """
+    """ NamedTuple that holds mean and variance as named fields """
 
     mean: tf.Tensor
     variance: tf.Tensor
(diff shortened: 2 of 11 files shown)


doc/source/notebooks/basics/regression.pct.py
```diff
@@ -179,6 +179,9 @@
 _ = plt.xlim(-0.1, 1.1)
 
 
+# %% [markdown]
+# Moreover, you can get the mean and variance data individually for example as `m.predict_f(xx).mean` instead of `m.predict_f(xx)[0]`.
```
st--, line 183:
```suggestion
# Moreover, you can get the mean and variance predictions individually, using for `m.predict_f(xx).mean` instead of `m.predict_f(xx)[0]` and `m.predict_f(xx).variance` instead of `m.predict_f(xx)[1]`.
```
(note- if we rename the variance field we'll have to update this)

gpflow/models/model.py
```diff
@@ -25,7 +25,12 @@
 from ..utilities import to_default_float
 from .training_mixins import InputData, RegressionData
 
-MeanAndVariance = Tuple[tf.Tensor, tf.Tensor]
+
+class MeanAndVariance(NamedTuple):
+    """ NamedTuple to access mean- and variance-function separately """
```
st--, line 30:
```suggestion
    """ NamedTuple that holds mean and variance as named fields """
```

gpflow/models/model.py
```diff
@@ -25,7 +25,12 @@
...
+
+    mean: tf.Tensor
+    variance: tf.Tensor
```
st--, line 33: More of a general question at other maintainers (@vdutor @markvdw @awav) & anyone else ...: what should the name of this field be? In code we often abbreviate the variance to "var", but also it sometimes represents the covariance... or maybe that should be a different type? MeanAndVariance and MeanAndCov (and return a different one depending on full_cov etc)?

st-- (commented):
Thanks for the contribution - I think this will be really helpful! I'd just like to hear from some of the other people involved with the GPflow project what their thoughts on the naming of the variance field are. :)

gpflow/models/model.py
joelberkeley: i like the idea of the two types. Haven't thought about it in detail

joelberkeley: do you always know for sure if you've got the cov or var?

st--: If you pass in full_cov=False and full_output_cov=False, you get the marginals back. If one of full_cov or full_output_cov is True, you get the covariance over inputs or outputs, respectively. If both are True, you should get the N P x N P covariance matrix (though this combination isn't actually implemented in several cases, I believe). So the output type is solely determined by the full_cov and full_output_cov arguments.

joelberkeley: btw you can use `typing.overload` and `typing.Literal` to give more information than just `MeanAndVariance | MeanAndCov`, if you wanted to do that. Probably not sth for this PR though

antonykamp: > btw you can use typing.overload and typing.Literal to give more information than just MeanAndVariance | MeanAndCov, if you wanted to do that

Alright! I'll look into it the next week :D

st--:
Hi @antonykamp could you make sure to update RELEASE.md and CONTRIBUTORS.md as part of this PR? (See #1660 and #1661).

gpflow/models/model.py
antonykamp: @joelberkeley , do you suggest adding Literals to the class-attributes-type and overloading the "variance" attribute by "covariance" (and the class-name too)? :)

antonykamp:
> Hi @antonykamp could you make sure to update RELEASE.md and CONTRIBUTORS.md as part of this PR? (See #1660 and #1661).

I'll take care of it after the "literal-overload-naming"-question is resolved. Otherwise, I would have to change it multiple times probably :)

gpflow/models/model.py
joelberkeley: i'd suggest leaving it for this PR. It's not vital, it's fiddly and tbh I've not used `Literal` yet so I'm kind of guessing how it works. Plain overloads would let you do sth like this (I've omitted some arguments)
```python
@overload
def predict_f(self, ...) -> MeanAndVariance: ...

def predict_f(self, ..., full_cov: bool = False) -> Union[MeanAndCovariance, MeanAndVariance]:
    ...
```
for literals you could add
```python
@overload
def predict_f(self, ..., full_cov: Literal[True]) -> MeanAndCovariance: ...

@overload
def predict_f(self, ..., full_cov: Literal[False]) -> MeanAndVariance: ...
```

antonykamp: Ah. I thought of using it in the definition of `MeanAndVariance | MeanAndCov`. I didn't consider using `Literal` and `overload` in `predict_f`.  That means that the user would see different return types of `predict_f` depending on full_cov, etc. Sounds pretty nice!

Regardless of this PR, I would play a bit with this idea :D

antonykamp: I've created a small example, which should work as wanted:
on_hover
on_space

But I' not comfortable with the if-else statement :/

```python
from typing import Literal, NamedTuple, overload

class MeanAndVariance(NamedTuple):
    mean: int
    variance: int

class MeanAndCovariance(NamedTuple):
    mean: int
    covariance: int

@overload
def predict_f(auto_cov: Literal[False]) -> MeanAndVariance:
    ...

@overload
def predict_f(auto_cov: Literal[True]) -> MeanAndCovariance:
    ...

def predict_f(auto_cov: bool = False) -> (MeanAndVariance | MeanAndCovariance):
    # calculations
    return  MeanAndCovariance(1, 2) if auto_cov else MeanAndVariance(1, 2)
```

st--: I don't think there's a good way around the if-else, and I think it's better to be explicit than the ambiguity of having to remember whether it's a [N, Q] or [Q, N, N] tensor ...:) I'd be happy with this.

antonykamp: As it pointed out, [typing.Literal](https://docs.python.org/3/library/typing.html#typing.Literal) is only available from Python 3.8 and up :D

st--: @antonykamp the [typing_extensions module](https://github.com/python/typing/blob/master/typing_extensions/README.rst) provides backports for older versions of Python, it does seem to include Literal. :)

antonykamp: I added this pattern with `overload` and `Tuple` to the abstract method  `predict_f` in `gpflow/models/models.py`. Should the overloads of `predict_f` be marked as _abstract_ too? In any case I see no chance to test the Ellipsis operator of each overloaded funciton :/

Also, I wanted to ask if the parameters of the model constructor should be listed in the parametrization?

antonykamp:
Closed because it's old :)

PR #1661, opened 2021-04-06
Add RELEASE.md

st--:
Adapted from TensorFlow's running release notes.

Also includes release notes since the last release (2.1.4).


RELEASE.md
```diff
@@ -1 +1,64 @@
...
+* <NOTES SHOULD BE GROUPED PER AREA>
+
+## Thanks to our Contributors
+
+This release contains contributions from:
+
+<INSERT>, <NAME>, <HERE>, <USING>, <GITHUB>, <HANDLE>
+
+
+# Release 2.1.5
```
vdutor, line 36: can you make it more clear this is the current release people have to add notes to?

RELEASE.md
```diff
@@ -1 +1,64 @@
-The release notes have been moved to the ['Releases' section](https://github.com/GPflow/GPflow/releases) of the GPflow GitHub Repo
+Release notes for past releases are available in the ['Releases' section](https://github.com/GPflow/GPflow/releases) of the GPflow GitHub Repo.
+
+# Release x.y.z
```
vdutor, line 3: can we not have the template at the bottom of the page?

st--: We could, but that makes it more hassle when creating a new release to copy things from the bottom back to the top - to me it makes more sense to follow TensorFlow in this and leave as is.

RELEASE.md
st--: done

vdutor (approved):
LGTM. Can you add an extra step to our release procedure that the person in charge of making the release also copies the template and starts the release notes for the next release?

st--:
Great idea - I added it as a HOWTO_RELEASE.md document to the repo so everyone can access it & propose edits.

PR #1669, opened 2021-04-15
Change URL for GPflow logo in README.md to absolute

st--:
so that it will render correctly on PyPI


johnamcleod (approved):
Looks good to me!

PR #1673, opened 2021-04-21
Posterior object: improve handling of cache precomputation

st--:


gpflow/posteriors.py
```diff
@@ -531,6 +562,23 @@ def _get_posterior_linearcoregionalization_mo_efficient(kernel, inducing_variabl
...
 
-def create_posterior(kernel, inducing_variable, q_mu, q_sqrt, whiten=True, mean_function=None):
+def create_posterior(
+    kernel,
+    inducing_variable,
+    q_mu,
+    q_sqrt,
+    whiten,
+    mean_function,
+    precompute_cache: Union[str, bool, PrecomputeCacheType],
```
st--: I'm not sure whether these _should_ or should _not_ have a default value...

gpflow/posteriors.py
```diff
@@ -531,6 +562,23 @@ def _get_posterior_linearcoregionalization_mo_efficient(kernel, inducing_variabl
...
 
-def create_posterior(kernel, inducing_variable, q_mu, q_sqrt, whiten=True, mean_function=None):
+def create_posterior(
+    kernel,
+    inducing_variable,
+    q_mu,
+    q_sqrt,
+    whiten,
+    mean_function=None,
+    precompute_cache: Union[str, bool, PrecomputeCacheType] = PrecomputeCacheType.TENSOR,
```
st--, line 572: I'm not sure if these _should_ or should _not_ have a default value...

(Currently reverted the default so I don't have to change all the test_posteriors calls to create_posterior().)

gpflow/posteriors.py
```diff
@@ -70,27 +71,48 @@ def __init__(self, q_mu, q_sqrt):
...
+        q_sqrt: tf.Tensor,
+        whiten: bool = True,
+        mean_function: Optional[mean_functions.MeanFunction] = None,
+        *,
+        precompute_cache: PrecomputeCacheType,
     ):
+        """
+        Concrete implementations should not overwrite this constructor.
+
+        :param kernel:
```
johnamcleod, line 108: Is the intention to fill out this docstring?

gpflow/posteriors.py
johnamcleod: It looks like `bool` is not fully supported for this type. `False` with be interpreted as "no cache", but it looks like `True` will throw the "not a valid type..." exception. Supporting a string is nice.

Having a default value of `TENSOR` makes sense to me.

johnamcleod (approved):
Looks good to me!

gpflow/posteriors.py
st--: yeah ...

gpflow/posteriors.py
st--: yeah, the type hint isn't quite right. basically I think it makes a lot of sense to say `precompute_cache=False` or `precompute_cache=None`, but as there's two ways of storing the cache, `True` would be ambiguous. any better idea for howto handle this?

st--: Ok having written that I just thought of one, we can have it `Union[None, str, PrecomputeCacheType]`, then we avoid the half-bool support

johnamcleod: Sounds good!

PR #1684, opened 2021-05-17
Fix unit test which fails when running with TensorFlow 2.5.0

johnamcleod:


setup.py
```diff
@@ -15,7 +15,7 @@
 # not to append tensorflow and tensorflow_probability to the requirements:
 if os.environ.get("READTHEDOCS") != "True":
     requirements = [
-        "tensorflow>=2.2.0",
+        "tensorflow>=2.5.0",
```
st--, line 18: This would mean GPflow will stop supporting TF 2.2 - 2.4 -- can we not find a workaround to support those as well?

tests/gpflow/utilities/test_traversal.py
```diff
@@ -209,14 +209,15 @@ def create_model():
...
-+-------------------------+------------------+-------------+---------+-------------+-----------+---------+----------+"""
++-------------------------------+------------------+-------------+---------+-------------+-----------+---------+----------+\n\
+| name                          | class            | transform   | prior   | trainable   | shape     | dtype   | value    |\n\
++===============================+==================+=============+=========+=============+===========+=========+==========+\n\
+| C._trainable_weights[0]       | ResourceVariable |             |         | True        | (2, 2, 1) | float32 | [[[0.... |\n\
+| C.variable                    |                  |             |         |             |           |         |          |\n\
++-------------------------------+------------------+-------------+---------+-------------+-----------+---------+----------+\n\
+| C._self_tracked_trackables[0] | Parameter        | Identity    |         | True        | ()        | float64 | 0.0      |\n\
+| C.param                       |                  |             |         |             |           |         |          |\n\
```
st--, line 219: How about simply changing the expected string based on the TF version? We've got similar test branching in the tests of clearing the bijector caches.

setup.py
johnamcleod: Yep, your other comment achieved that.

tests/gpflow/utilities/test_traversal.py
johnamcleod: Done.

st-- (approved):
looks good, thanks for fixing so quickly!

PR #1694, opened 2021-06-16
upgrade to black==20.8b1

st--:
Only non-whitespace change is in replacing `.resize(-1,)` with `.flatten()`


johnamcleod (approved):
LGTM!

PR #1695, opened 2021-06-17
Small refactor of the posteriors to support other models

johnamcleod:
**PR type:** enhancement / doc improvement

**Related issue(s)/PRs:** 

## Summary

**Proposed changes**

This PR refactors the posterior objects. The current posterior objects are tightly coupled to the SVGP model, by requiring both inducing points and variational parameters. This PR changes the constructor of the base class (`AbstractPosterior`) to accept the `alpha` and `Qinv` values (alongside the inducing points or training data points). This makes the base class sufficiently flexible to support the GPR, SGPR and VGP models, which can be added later.

The posterior object is not necessarily responsible for calculating `alpha` and `Qinv`. Subclasses of `AbstractPosterior` for different posteriors could have a `precompute` method like the SVGP `BasePosterior`, or the models could calculate the tensor values when creating a posterior object.

The base class for the SVGP posteriors is still `BasePosterior`. Updating the cached values has been left on the `BasePosterior`. 

This PR also includes a draft of a notebook about the caching functionality, with examples of the maths for the GPR, SGPR and (S)VGP models.

### Minimal working example

See the timing example at the end of the new notebook.

### Release notes

Refactor the posterior base classes to support other models than the SVGP model.

## PR checklist

  - [x] notebook examples (usage demonstration)
- [x] Build checks
- [ ] Release management

### DIFF
--- a/doc/source/notebooks/advanced/fast_predictions.pct.py
+++ b/doc/source/notebooks/advanced/fast_predictions.pct.py
@@ -16,7 +16,7 @@
 # # Faster predictions by caching
 
 # + [markdown] id="PLuPjfS7KLQ-"
-# The default behaviour of `predict_f` in GPflow models is to compute the predictions from scratch on each call. This is convenient when predicting and training are interleaved, and simplifies the use of these models. There are some use cases, such as Bayesian optimisation, where prediction (at different test points) happens much more frequently than training. In these cases it is convenient to cache parts of the calculation which do not depend upon the test points, and reuse those parts between predictions. 
+# The default behaviour of `predict_f` in GPflow models is to compute the predictions from scratch on each call. This is convenient when predicting and training are interleaved, and simplifies the use of these models. There are some use cases, such as Bayesian optimisation, where prediction (at different test points) happens much more frequently than training. In these cases it is convenient to cache parts of the calculation which do not depend upon the test points, and reuse those parts between predictions.
 #
 # There are three models to which we want to add this caching capability: GPR, (S)VGP and SGPR. The VGP and SVGP can be considered together; the difference between the models is whether to condition on the full training data set (VGP) or on the inducing variables (SVGP).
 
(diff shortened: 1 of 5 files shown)


gpflow/posteriors.py
```diff
@@ -123,13 +120,89 @@ def __init__(
...
+        if self.alpha is None or self.Qinv is None:
+            raise ValueError(
+                "Cache has not been precomputed yet. Call update_cache first or use fused_predict_f"
+            )
+        mean, cov = self._conditional_with_precompute(
+            Xnew, full_cov=full_cov, full_output_cov=full_output_cov
+        )
+        return self._add_mean_function(Xnew, mean), cov
+
+    @abstractmethod
```
st--, line 174: If it's abstract, why does it have an implementation?

gpflow/posteriors.py
```diff
@@ -123,13 +120,89 @@ def __init__(
...
+        q_mu: tf.Tensor,
+        q_sqrt: tf.Tensor,
+        whiten: bool = True,
+        mean_function: Optional[mean_functions.MeanFunction] = None,
+        *,
+        precompute_cache: Optional[PrecomputeCacheType],
+    ):
+
+        super().__init__(kernel, inducing_variable, mean_function=mean_function)
         self.mean_function = mean_function
```
st--, line 202:
```suggestion
```

doc/source/notebooks/advanced/fast_predictions.pct.py
```diff
@@ -0,0 +1,108 @@
+# ---
+# jupyter:
```
st--, line 2: Still need to link this document from the notebook table of contents

st-- (commented):
This generally seems fine; my main question is regarding follow-up PRs to enable support for GPR and SGPR: will we have to maintain duplicate code for all combinations of multi-output kernels and models? Is that something we could avoid, or inherent in the maths?

doc/source/notebooks/advanced/fast_predictions.pct.py
johnamcleod: Good point - I have added this to the `intro.md` file.

gpflow/posteriors.py
johnamcleod: An oversight. I have removed the decorator.

johnamcleod (commented):
> will we have to maintain duplicate code for all combinations of multi-output kernels and models? Is that something we could avoid, or inherent in the maths?

There should be no need to have duplicate code to support the combinations of kernels and models.

st-- (approved):
lgtm

PR #1699, opened 2021-06-30
John/parameters

johnamcleod:
**PR type:** enhancement

**Related issue(s)/PRs:** 

## Summary

**Proposed changes**

This PR modifies the constructor of a `Parameter` object so it can be constructed from another `Parameter` and retain the properties which were set on the original, while allowing the user to override properties as desired.

At the moment many components of GPflow models will internally construct `Parameter`s with the values which are passed in. They may set additional properties on these objects, such as the kernel variance `Parameter` which is constructed with the `positive` transform.

The `Parameter` objects have a set of useful properties which a user might specify, but when the `Parameter` is constructed inside another GPflow class these have to be set after construction. For example, if the user wants to add a prior to the kernel variance it is necessary to set the prior on the variance `Parameter` after constructing the kernel. 

In this example, the effect of this PR would be that the user could construct a `Parameter` with the chosen prior, and then pass that into the kernel constructor, which would retain the prior and add the `positive` transform as usual.

### Minimal working example

To construct a non-trainable constant mean function it is necessary to do this:
```python
import gpflow

cm = gpflow.mean_functions.Constant(1.0)
cm.c.trainable=False
```

After this PR it is possible to completely specify the mean function at construction time:
```python
import gpflow

cm = gpflow.mean_functions.Constant(gpflow.Parameter(1.0, trainable=False))
```

### Release notes

**Fully backwards compatible:** no

At present it is possible to construct a `Parameter` from another `Parameter`, but all of the properties are lost. This can lead to subtle bugs where properties are removed.

## PR checklist

- [x] Build checks
- [ ] Release management


gpflow/base.py
```diff
@@ -12,14 +12,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import functools
 from enum import Enum
+from inspect import currentframe, getargvalues
```
st--, line 16:
```suggestion
```
doesn't seem to be required

tests/gpflow/optimizers/test_mcmc.py
```diff
@@ -34,12 +34,14 @@ def build_model(data):
...
 
-def build_model_with_uniform_prior(data, prior_on, prior_width):
+def build_model_with_uniform_prior_no_transforms(data, prior_on, prior_width):
     def parameter(value):
         low_value = -100
         high_value = low_value + prior_width
         prior = Uniform(low=np.float64(low_value), high=np.float64(high_value))
-        return gpflow.Parameter(value, prior=prior, prior_on=prior_on)
+        return gpflow.Parameter(
+            value, transform=tfp.bijectors.Identity(), prior=prior, prior_on=prior_on
```
st--, line 43:
```suggestion
            value, prior=prior, prior_on=prior_on
```
shouldn't this test the setting of the default (identity) transform?

st-- (approved):
LGTM, subject to a few minor suggestions:)

st--:
(And you need to update RELEASE.md before merging!)

gpflow/base.py
```diff
@@ -117,14 +116,27 @@ def __init__(
...
         A prior can be imposed either on the constrained version (default) or on the unconstrained version of the parameter.
         """
-        if transform is None:
-            transform = tfp.bijectors.Identity()
+        if isinstance(value, Parameter):
+            transform = transform if transform else value.transform
+            prior = prior if prior else value.prior
+            prior_on = prior_on if prior_on else value.prior_on
+            trainable = trainable if trainable is not None else value.trainable
+            name = name if name else value.bijector.name
```
frgsimpson, line 124:
```suggestion
            transform = transform or value.transform
            prior = prior or value.prior
            prior_on = prior_on or value.prior_on
            name = name or value.bijector.name
            trainable = value.trainable if trainable is None else trainable
```
Perhaps just a matter of personal preference, but I tend to find the or operator a little more readable in these cases.

gpflow/base.py
```diff
@@ -117,14 +116,27 @@ def __init__(
...
+            if transform is None:
+                transform = tfp.bijectors.Identity()
+
+            prior_on = prior_on if prior_on else PriorOn.CONSTRAINED
+            trainable = trainable if trainable is not None else True
 
-        value = _cast_to_dtype(value, dtype)
-        _validate_unconstrained_value(value, transform, dtype)
+            value = _cast_to_dtype(value, dtype)
+            _validate_unconstrained_value(value, transform, dtype)
```
frgsimpson, line 136: Is there a risk that we're now not always validating the parameter values?  (though there are times when I don't care about invalid values.)

frgsimpson (approved):
LGTM!

gpflow/base.py
johnamcleod: Good point. If the transformation changed then the value (which was valid under the previous transformation) might no longer be valid. I will move this check so it is always called.

tests/gpflow/optimizers/test_mcmc.py
johnamcleod: When `value` is a `Parameter` and `transform=None`, the `value.transform` is used to set the transform of the new parameter. You are correct to say that this test is checking the identity transform, so to ensure that it is being used we set it directly in the test. Otherwise we have to assume that the incoming `Parameter` is using the identity transform which is not always true.

PR #1710, opened 2021-09-09
Start to simplify hetgpr

frgsimpson:


gpflow/likelihoods/heteroskedastic.py
```diff
@@ -25,7 +25,7 @@
 from ..utilities import positive
 
 
-class HeteroskedasticGaussianLikelihood(HeteroskedasticTFPConditional):
+class ParameterisedGaussianLikelihood(HeteroskedasticTFPConditional):
```
johnamcleod, line 28: Can this class be deleted? It looks like all it adds to the `HeteroskedasticTFPConditional` is the Gaussian distribution.

johnamcleod (approved):

gpflow/likelihoods/heteroskedastic.py
frgsimpson: Yes as it stands, though perhaps this can become the class which inherits from MultiLatentTFPConditional and permits dependence on both X and f?

johnamcleod: I think the `MultiLatentTFPConditional` already provides that behaviour. It is the `HeteroskedasticTFPConditional` which restricts it to the Xs.

PR #1734, opened 2021-11-10
Fix #1733

### LINKED ISSUE #1733: GPR Posterior Calculations incomplete
# Bug

When you call the `posterior()` method, `posterior.alpha` should be calculated, but currently[ it just sets `posterior.alpha` equal to y](https://github.com/GPflow/GPflow/blob/d088c2c4583df606a6f0ba79eea78d27c6210f7c/gpflow/posteriors.py#L264).

## To reproduce

```python
gpr_model = gpflow.models.GPR( data=(train_x, train_y), kernel=kernel)
posterior = gpr_model.posterior()
assert  all( posterior.alpha.numpy() == train_y )
```

## Expected behavior

Obviously there are calculation steps missing in the `posterior` method.

## System information

* GPflow version: 2.3.0

johnamcleod:
Fix #1733

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -55,11 +55,11 @@ This release contains contributions from:
 
 ## Bug Fixes and Other Changes
 
-* Fixed broken CircleCi build.
-* Update CircleCi build to use next-gen Docker images.
-* Fixed broken link in `README.md`.
-* Make `make dev-install` also install the test requirements.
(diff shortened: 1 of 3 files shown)


Andrew878 (approved):
thanks for spotting/fixing my bug

jesnie (approved):
Approving for @Andrew878 , since it appears he doesn't have the right access.

PR #1738, opened 2021-11-25
Fix broken CircleCi build.

jesnie:
**PR type:** bugfix

**Related issue(s)/PRs:** N/A

## Summary

**Proposed changes**

* Install dependencies into virtual environments, to avoid conflicts with pre-installed packages.
* Fix the version of Tensorflow and Tensorflow Probability used on CircleCi, to fix issue where the current `latest` versions aren't compatible.
* Consistently use the newest version of `pip` on CircleCi.
* Consistently turn off progress bars in `pip` on CircleCi, as progress bars are not useful in logs.

* Constrain the version of Tensorflow / Tensorflow Probability in our `setup.py`.

### Minimal working example

N/A

### Release notes

N / A

## PR checklist

- [x] Build checks
- [x] Release management

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,29 +1,51 @@
-runtest: &runtest
-  steps:
-    - checkout
-    - run:
-        name: Install GPflow
-        command: |
-          # Run in a fresh virtual environment, to avoid conflicts with preinstalled packages.
-          virtualenv -p python3.6 .venv
-          source .venv/bin/activate
-          pip install --progress-bar=off -U pip
-          # Ensure consistency between tensorflow and tensorflow-probability
-          pip install --progress-bar=off tensorflow==2.6.2 tensorflow-probability==0.14.0
-          pip install --progress-bar=off -e .
-          pip install --progress-bar=off -r tests_requirements.txt
-    - run:
-        name: Run tests
-        command: |
-          source .venv/bin/activate
-          pytest -v -W ignore::UserWarning --durations=10 -m "${PYTEST_FILTER}" --cov=./gpflow ./tests
-    - run:
-        name: Upload coverage report
-        command: |
-          source .venv/bin/activate
-          bash <(curl -s https://codecov.io/bash) -t "${CODECOV_TOKEN}"
-
 version: 2.1
+
+commands:
+  setup_venv:
+    steps:
+      - checkout
+      - run:
+          name: Setup virtual environment
+          command: |
+            # Run in a fresh virtual environment, to avoid conflicts with preinstalled packages.
+            virtualenv -p python3.6 .venv
+            source .venv/bin/activate
+            pip install --progress-bar=off -U pip
+
+  install_gpflow:
+    steps:
+      - setup_venv
+      - run:
+          name: Install GPflow
+          command: |
+            source .venv/bin/activate
+            # Ensure consistency between tensorflow and tensorflow-probability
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -4,16 +4,23 @@ runtest: &runtest
...
         command: |
-          pip install -U pip
-          pip install -e .
-          pip install -r tests_requirements.txt
+          # Run in a fresh virtual environment, to avoid conflicts with preinstalled packages.
+          virtualenv -p python3.6 .venv
+          source .venv/bin/activate
+          pip install --progress-bar=off -U pip
+          # Ensure consistency between tensorflow and tensorflow-probability
+          pip install --progress-bar=off tensorflow==2.6.2 tensorflow-probability==0.14.0
```
st--, line 12: Should it really be pinning patch version, or just the minor versions, 2.6 and 0.14 so that we will automatically pull in e.g. tf 2.6.3 or tfp 0.14.1?

st--:
In principle I'm happy with anything that gets the build running again. Is there some way that allows for "consistent tf/tfp version" whilst making it harder to forget to check whether it works with an updated tensorflow version at all?

.circleci/config.yml
```diff
@@ -4,16 +4,23 @@ runtest: &runtest
     - run:
         name: Install GPflow
         command: |
-          pip install -U pip
-          pip install -e .
-          pip install -r tests_requirements.txt
+          # Run in a fresh virtual environment, to avoid conflicts with preinstalled packages.
+          virtualenv -p python3.6 .venv
```
st--, line 8: Does CircleCI always provide a fresh directory? (I remember Jenkins having leftover stuff from previous runs)
(and if it doesn't, what happens if .venv already exists?)

.circleci/config.yml
```diff
@@ -28,6 +35,12 @@ jobs:
       - run:
           name: Check installed dependencies are compatible
           command: |
+            # Run in a fresh virtual environment, to avoid conflicts with preinstalled packages.
+            virtualenv -p python3.6 .venv
+            source .venv/bin/activate
+            pip install --progress-bar=off -U pip
+            # Ensure consistency between tensorflow and tensorflow-probability
+            pip install --progress-bar=off tensorflow==2.6.2 tensorflow-probability==0.14.0
```
st--, line 43: same point about 2.6 / 0.14 so we would automatically get 2.6.3 / 0.14.1 ...

st-- (approved):
I'm happy for you to merge this as a work-around to get the build going again, assuming that

- you can address the minor vs patch pinning points before merging, and
- you can confirm CircleCI always provides a fresh environment so there won't be any issue with pre-existing .venvs (I don't think that would be the case, just good to check).

If you think it might be possible to come up with a better solution (than manually bumping tf/tfp versions), that'd be great, but doesn't need to be part of this PR!

.circleci/config.yml
st--: Actually, to keep it consistent across the file, how about specifying tensorflow versions as config constants or even turning this into a [reusable command](https://circleci.com/docs/2.0/reusing-config/) ?

jesnie: > Actually, to keep it consistent across the file, how about specifying tensorflow versions as config constants or even turning this into a reusable command ?

I have refactored this config to use reusable commands.

PR #1745, opened 2021-11-30
Handle unconnected gradients in gpflow.optimizers.Scipy.

### LINKED ISSUE #1600: Handling unconnected gradients in Scipy optimizer

jesnie:

**Related issue(s)/PRs:** Fixes #1600

## Summary

**Proposed changes**

* Update `gpflow.optimizers.Scipy.minimize` to take a new parameter `allow_unused_variables`.
* If a variable is unused/unconnected and `allow_unused_variables=True`, then a warning is printed, but optimization continues.
* If a variable is unused/unconnected and `allow_unused_variables=False`, then an exception with an appropriate message is raised.

This PR competes with #1655 .
I think it's an open question whether `allow_unused_variables` should default to `True` or `False`. The current suggestion is to default to `True`.

### Minimal working example

```python
import tensorflow as tf
import gpflow
from gpflow.config import default_float

target1 = [0.2, 0.8]
v1 = tf.Variable([0.5, 0.5], dtype=default_float(), name="v1")
v2 = tf.Variable([0.5], dtype=default_float(), name="v2")

def f() -> tf.Tensor:
    # v2 not used.
    return tf.reduce_sum((target1 - v1) ** 2)

print(opt.minimize(f, [v1, v2], allow_unused_variables=True))
```

### Release notes

**Fully` backwards compatible:** yes

N/A

## PR checklist

- [x] New features: code is well-documented
  - [x] detailed docstrings (API documentation)
- [x] Build checks
- [x] Release management

### DIFF
--- a/gpflow/optimizers/scipy.py
+++ b/gpflow/optimizers/scipy.py
@@ -108,11 +108,26 @@ class Scipy:
         compile: bool = True,
         allow_unused_variables: bool = False,
     ) -> Callable[[np.ndarray], Tuple[np.ndarray, np.ndarray]]:
+        first_call = True
+
         def _tf_eval(x: tf.Tensor) -> Tuple[tf.Tensor, tf.Tensor]:
+            nonlocal first_call
+
             values = cls.unpack_tensors(variables, x)
             cls.assign_tensors(variables, values)
-            loss, grads = _compute_loss_and_gradients(closure, variables)
-            grads = cls._filter_unused_variables(variables, grads, allow_unused_variables)
+
+            if first_call:
+                # Only check for unconnected gradients on the first function evaluation.
+                loss, grads = _compute_loss_and_gradients(
+                    closure, variables, tf.UnconnectedGradients.NONE
+                )
+                grads = cls._filter_unused_variables(variables, grads, allow_unused_variables)
+                first_call = False
+            else:
+                loss, grads = _compute_loss_and_gradients(
+                    closure, variables, tf.UnconnectedGradients.ZERO
+                )
+
             return loss, cls.pack_tensors(grads)
 
         if compile:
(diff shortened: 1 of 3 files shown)


frgsimpson (commented):
Looks like a really nice improvement!

My only concern is whether it's necessary to perform these gradient checks at each call to eval. 
I suspect gradients can't become connected/disconnected during gradient descent, so we perhaps need only perform this test once for each call to 'minimise'.

gpflow/optimizers/scipy.py
```diff
@@ -97,13 +102,32 @@ def initial_parameters(cls, variables: Sequence[tf.Variable]) -> tf.Tensor:
...
     @classmethod
     def eval_func(
-        cls, closure: LossClosure, variables: Sequence[tf.Variable], compile: bool = True
+        cls,
+        closure: LossClosure,
+        variables: Sequence[tf.Variable],
+        compile: bool = True,
+        allow_unused_variables: bool = False,
     ) -> Callable[[np.ndarray], Tuple[np.ndarray, np.ndarray]]:
+        first_call = True
```
frgsimpson, line 111: Does this need to be tf.constant(True), to ensure the if statement is revisited?

frgsimpson (approved):
LGTM! Just a small question about whether the if statement will be re-executed at each iteration.

gpflow/optimizers/scipy.py
jesnie: > Does this need to be tf.constant(True), to ensure the if statement is revisited?

Depends... I don't really think it's necessary. If this is compiled, then we will always visit the `if first_call` branch, however all the interesting/expensive stuff inside that branch is pure Python, which won't be compiled into the TF graph, so it doesn't hurt that we're always in that branch...

frgsimpson: If you're sure autograph doesn't compile the enumerate call then I agree that should be fine.

jesnie: I wrote a small benchmark and tried using a `tf.Variable` for this if-statement - that seems to make optimisation 15-100% slower, so I'm not going to do that.

PR #1747, opened 2021-12-07
Andrewl/fix gpr posterior leading dim

Andrew878:
**PR type:** bugfix 

**Related issue(s)/PRs:** 

## Summary

**Proposed changes**

* The GPR posterior implementation was not working for `predict_f` calls with inputs containing leading dimensions.
* This PR addresses the problem and also amends the test so that it is covered in the future
* ...

### Minimal working example

 

This would work:
```python
# refer to test_gpr_posterior.py
m = GPR_with_posterior(data=some_regression_data, kernel=k)
X_new = np.random.randn(5, 2)
m.posterior(cache_type).predict_f(X_new, full_cov=full_cov, full_output_cov=full_output_cov)

```
But this wouldn't
```python
# refer to test_gpr_posterior.py
m = GPR_with_posterior(data=some_regression_data, kernel=k)
X_new = np.random.randn(10, 5, 2)
m.posterior(cache_type).predict_f(X_new, full_cov=full_cov, full_output_cov=full_output_cov)

```
and would result in `InvalidArgumentError`

The PR addresses this. 
### Release notes

## PR checklist

- [ X] Build checks
  - [X ] I locally tested that the tests pass (`make check-all`)
- [ X] Release management
  - [X ] RELEASE.md updated with entry for this change
  - [ X] New contributors: I've added myself to CONTRIBUTORS.md

### DIFF
--- a/gpflow/posteriors.py
+++ b/gpflow/posteriors.py
@@ -242,20 +242,52 @@ class GPRPosterior(AbstractPosterior):
         Computes predictive mean and (co)variance at Xnew, *excluding* mean_function.
         Relies on cached alpha and Qinv.
         """
-        Kmn = self.kernel(self.X_data, Xnew)
-        Knn = self.kernel(Xnew, full_cov=full_cov)
 
+        Kmn = self.kernel(self.X_data, Xnew)
+        # compute kernel stuff
+        num_func = tf.shape(self.Y_data)[-1]  # R
+        N = tf.shape(Kmn)[-1]
+
+        # get the leading dims in Kmn to the front of the tensor Kmn
+        K = tf.rank(Kmn)
+        perm = tf.concat(
+            [
+                tf.reshape(tf.range(1, K - 1), [K - 2]),  # leading dims (...)
+                tf.reshape(0, [1]),  # [M]
+                tf.reshape(K - 1, [1]),
+            ],
+            0,
+        )  # [N]
+        Kmn = tf.transpose(Kmn, perm)  # [..., M, N]
+        leading_dims = tf.shape(Kmn)[:-2]
+
+        # get the leading dims in Knm to the front of the tensor Knm
+        perm_T = tf.concat(
+            [
+                tf.reshape(0, [1]),  # [M]
+                tf.reshape(K - 1, [1]),
+                tf.reshape(tf.range(1, K - 1), [K - 2]),  # leading dims (...)
+            ],
+            0,
+        )  # [N]
+        Knm = tf.transpose(Kmn, perm=perm_T)
+
+        Knn = self.kernel(Xnew, full_cov=full_cov)
         err = self.Y_data - self.mean_function(self.X_data)  # type: ignore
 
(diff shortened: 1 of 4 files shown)


gpflow/posteriors.py
```diff
@@ -242,20 +242,52 @@ def _conditional_with_precompute(
...
+                tf.reshape(0, [1]),  # [M]
+                tf.reshape(K - 1, [1]),
+                tf.reshape(tf.range(1, K - 1), [K - 2]),  # leading dims (...)
+            ],
+            0,
+        )  # [N]
+        Knm = tf.transpose(Kmn, perm=perm_T)
 
+        Knn = self.kernel(Xnew, full_cov=full_cov)
         err = self.Y_data - self.mean_function(self.X_data)  # type: ignore
```
henrymoss, line 276: Thanks for doing this @Andrew878 . I think some of this functionality could be replaced with the ` leading_transpose` method in `gpflow.utilities.ops`

tests/gpflow/models/test_gpr_posterior.py
```diff
@@ -21,9 +20,9 @@ def make_models(regression_data):
 
 def _get_data_for_tests():
     """Helper function to create testing data"""
-    X = np.random.randn(100, input_dim)
-    Y = np.random.randn(100, output_dim)
-    X_new = np.random.randn(100, input_dim)
+    X = np.random.randn(5, 2)
+    Y = np.random.randn(output_dim)
+    X_new = np.random.randn(10, 5, 2)
```
henrymoss, line 25: Perhaps worth testing for even more leading dims? i.e. `X_new = np.random.rand(3,10,5,2)`

gpflow/posteriors.py
Andrew878: I was able to replace the `perm_T` with `leading_transpose` but I could not for the `perm` term. Happy to be shown a nice solution here (if possible)

tests/gpflow/models/test_gpr_posterior.py
Andrew878: yes... I had to rework my solution to handle this....so good call. It now works and I've updated the tests to reflect this use case

Andrew878:
It's not clear to me why the type checker if failing. I don't think I have used `leading_dims` any differently to other parts of the code (still looking into it)

tests/gpflow/models/test_gpr_posterior.py
henrymoss: Great!

Andrew878:
> It's not clear to me why the type checker if failing. I don't think I have used `leading_dims` any differently to other parts of the code (still looking into it)

I spoke to Jesper and we weren't able to understand why the mypy errors were occurring, so I have replaced `List[Union[int, type(...)]` with , `List[Any]`. While not ideal, any other input type will likely break the function (and hence flag a misspecified input). I have also made a note in the docstring about the expected type too.

frgsimpson (approved):
LGTM!

PR #1759, opened 2022-01-21
Make dataclasses dependency conditional at install time

tmct:
Ensures dataclasses is not treated as a dependency of the "any" wheel on Python 3.7+

**PR type:** bugfix

**Related issue(s)/PRs:** -

## Summary

**Proposed changes**
* Make dataclasses dependency conditional at install time rather than conditional at wheel build time

* Asking you to deprecate Python 3.6
* Asking you to create separate wheels for py36, py38 etc.

### Minimal working example

```bash
python --version  # 3.8
pip install gpflow
# dataclasses has been installed into site-packages! :(
```

### Release notes

## PR checklist

- [ ] Build checks
- [ ] Release management

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -33,7 +33,7 @@ requirements.extend(
         "typing_extensions",
         "packaging",
         "deprecated",
-        "dataclasses; python_version < "3.7",
+        "dataclasses;python_version<'3.7'",
     ]
 )
 
--- a/setup.py
+++ b/setup.py
@@ -33,11 +33,9 @@ requirements.extend(
         "typing_extensions",
         "packaging",
         "deprecated",
+        "dataclasses; python_version < "3.7",
     ]
 )
-
-if sys.version_info < (3, 7):
-    requirements.append("dataclasses")  # became part of stdlib in python 3.7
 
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -33,12 +33,10 @@
         "typing_extensions",
         "packaging",
         "deprecated",
+        "dataclasses; python_version < "3.7",
```
jesnie, line 36: This doesn't look right. Maybe something like this:
```suggestion
        "dataclasses; python_version < '3.7'",
```
Our continuous integration tools are also complaining.
Maybe you can test this on your local machine?

jesnie (commented):
Thanks for contributing!

Feel free to add yourself to the contributors list, and update the release notes while you're at it, so everybody can know about the hard work you are doing! :-)

setup.py
tmct: Apologies! I will correct and run the test locally

jesnie: This looks much better. Thanks!

jesnie (approved):
Looks good. Thanks!

PR #1766, opened 2022-01-31
Make mypy a bit more strict, and fix revealed problems.

jesnie:
Make `mypy` a bit more strict, and fix all the new problems.

I've felt it necessary to make a small refactor / breaking change to the API of `InducingVariables`. Previously the `InducingVariables` was built around overriding `__len__`, which was annotated to return `int`. However, it actually returned a `tf.Tensor`, and the code depends on this actually being a `tf.Tensor`. However having `__len__` return a tensor feels dangerously weird. So I've refactored the api to be based on a `num_inducing -> tf.Tensor` instead.

Merged `types.py` into `base.py`. `types.py` only had `MeanAndVariance` and all the other types were already in `base.py`.
Moved `InputData`, `OutputData` and `RegressionData` to `base.py` so they can be imported without circular dependencies.


frgsimpson (approved):
LGTM!

PR #1767, opened 2022-02-02
Make posteriors work for Bayesian Optimisation.

### LINKED ISSUE #1765: The posterior objects work poorly with Bayesian Optimisation
# Bug

## To reproduce

1. Create a model where the data are `tf.Variable`s.
2. Create a `Posterior` from the model.
3. Wrap `posterior.predict_f` in `tf.function`.
4. Add rows to the data.
5. Call `posterior.update_cache()`.
6. Call the compiled `predict_f`.

Observe crashes and wrong predictions.

## Expected behavior

The `Posterior` object should reflect the state of the data, even when it changes shape.

jesnie:
1. Add posterior unit tests, with variable-sized input data.
2. Fix a small bug with `x.shape` vs `tf.shape(x)`.
3. Refactor `posterior._precompute` to track which tensor dimensions are dynamic, so we can create `tf.Variable`s that are dynamic in the right ways.

Fixes #1765.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -127,7 +127,9 @@ disable=missing-docstring,
         unsubscriptable-object,
         no-else-return,
         import-error,
-        misplaced-comparison-constant
+        misplaced-comparison-constant,
+        too-many-lines,
+        fixme
 
 # Enable the message, report, category or checker with the given id(s). You can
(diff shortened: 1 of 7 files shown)


Andrew878 (approved):
As previously discussed, looks good.

PR #1768, opened 2022-02-02
Refactor check_shapes into multiple files.

jesnie:
Refactor `check_shapes` into multiple files, and write some more tests.

### DIFF
--- a/gpflow/experimental/check_shapes/base_types.py
+++ b/gpflow/experimental/check_shapes/base_types.py
@@ -11,15 +11,9 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
-# flake8: noqa
 """
-Library for annotating and checking the shapes of tensors.
+Definitions of commonly used types.
 """
+from typing import Any, Callable, TypeVar
 
-from .check_shapes import check_shapes
-from .errors import ArgumentReferenceError, ShapeMismatchError
-from .inheritance import inherit_check_shapes
-from .specs import ArgumentSpec, DimensionSpec, ShapeSpec
-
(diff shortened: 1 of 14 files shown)


gpflow/experimental/check_shapes/base_types.py
```diff
@@ -0,0 +1,24 @@
...
+Definitions of commonly used types.
+"""
+from typing import Any, Callable, TypeVar, Union
+
+import numpy as np
+import tensorflow as tf
+
+C = TypeVar("C", bound=Callable[..., Any])
+
+Shaped = Union[np.ndarray, tf.Tensor]
```
vdutor, line 24: Seems like a duplicate of `TensorType` or `TensorLike`: https://github.com/GPflow/GPflow/blob/c9f88c1f10cc869fcd3556948d3a9a188ddb3246/gpflow/base.py#L34?

jesnie: > Seems like a duplicate of `TensorType` or `TensorLike`:
> 
> https://github.com/GPflow/GPflow/blob/c9f88c1f10cc869fcd3556948d3a9a188ddb3246/gpflow/base.py#L34
> 
> ?

Two things:

1. I'd kind of like for the check_shapes stuff not to depend on the rest of GPflow, so that it's easy to spin into a separate repository later.
2. I don't think the list `[np.ndarray, tf.Tensor]` is the same as the list `[tf.Tensor, tf.Variable, "Parameter"]`. Though maybe both lists should be extended. If we weren't using an ancient version of TensorFlow there is also: https://www.tensorflow.org/api_docs/python/tf/types/experimental/TensorLike

jesnie: I've replaced `Shaped` with `TensorType`.

vdutor (approved):

PR #1773, opened 2022-02-08
Add mypy annotations to some of the tests.

jesnie:
Mostly adding type hints to some tests.
More completely:

1. Add type hints to some tests.
2. Replace some use of `os.` with `pathlib.Path.`, where we were already using `pathlib`.
3. Removed some unused imports.
4. Reformatted `mypy.ini`.

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -4,6 +4,7 @@ strict = True
 ignore_missing_imports = True
 disallow_subclassing_any = False
 disallow_untyped_calls = False
+disallow_untyped_decorators = False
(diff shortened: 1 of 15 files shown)


mypy.ini
```diff
@@ -4,6 +4,7 @@ strict = True
 ignore_missing_imports = True
 disallow_subclassing_any = False
 disallow_untyped_calls = False
+disallow_untyped_decorators = False
```
uri-granta, line 7: (isn't this the default?)

uri-granta (approved):
LGTM!

PR #1774, opened 2022-02-09
Make the VGP support variable-shaped data.

jesnie:
Make the VGP support variable-shaped data.

This includes:
1. Adding `shape` parameters to the `Parameter` class, to allow it to have dynamic shapes.
2. Adding an `after_data_changed` method to `InternalDataTrainingLossMixin` to allow the model to update internal values after data has changed.
3. Updating the `VGP` class itself.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -61,6 +61,8 @@ This release contains contributions from:
 ## Major Features and Improvements
 
 * Add new posterior class to enable faster predictions from the VGP model. (#1761)
+* VGP class bug-fixed to work with variable-sized data. Note you can use
+  `gpflow.models.vgp.update_vgp_data` to ensure variational parameters are updated sanely. (#1774).
 
 * Added `experimental` sub-package for features that are still under developmet.
   * Added `gpflow.experimental.check_shapes` for checking tensor shapes. (#1760)
--- a/gpflow/base.py
+++ b/gpflow/base.py
@@ -114,6 +114,9 @@ class Parameter(tfp.util.TransformedVariable):
         trainable: Optional[bool] = None,
         dtype: Optional[DType] = None,
         name: Optional[str] = None,
+        pretransformed_shape: Optional[Sequence[Optional[int]]] = None,
+        transformed_shape: Optional[Sequence[Optional[int]]] = None,
+        shape: Optional[Sequence[Optional[int]]] = None,
     ):
         """
         A parameter retains both constrained and unconstrained
@@ -121,6 +124,10 @@ class Parameter(tfp.util.TransformedVariable):
         It is often challenging to operate with unconstrained parameters. For example, a variance cannot be negative,
         therefore we need a positive constraint and it is natural to use constrained values.
         A prior can be imposed either on the constrained version (default) or on the unconstrained version of the parameter.
+
+        :param pretransformed_shape: Declare the shape of the pretransformed values. Useful for setting dynamic shapes.
+        :param transformed_shape: Declare the shape of the transformed values. Useful for setting dynamic shapes.
+        :param shape: Convenience shortcut for setting both `pretransformed_shape` and `transformed_shape`.
         """
         if isinstance(value, Parameter):
             transform = transform or value.transform
@@ -141,9 +148,30 @@ class Parameter(tfp.util.TransformedVariable):
             value = _cast_to_dtype(value, dtype)
 
         _validate_unconstrained_value(value, transform, dtype)
-        super().__init__(value, transform, dtype=value.dtype, trainable=trainable, name=name)
 
-        self.prior = prior  # type: Optional[Prior]
+        if shape is not None:
+            assert (
+                pretransformed_shape is None
+            ), "Cannot set both `shape` and `pretransformed_shape`."
+            assert transformed_shape is None, "Cannot set both `shape` and `transformed_shape`."
+            pretransformed_shape = shape
+            transformed_shape = shape
+
+        super().__init__(
+            value,
+            transform,
+            dtype=value.dtype,
+            trainable=trainable,
+            name=name,
+            shape=pretransformed_shape,
+        )
+
+        # TransformedVariable.__init__ doesn't allow us to pass a pre-transformed shape, so we
+        # manually override it.
+        if transformed_shape is not None:
+            self._shape = tf.TensorShape(transformed_shape)
+
+        self.prior: Optional[Prior] = prior
         self.prior_on = prior_on  # type: ignore  # see https://github.com/python/mypy/issues/3004
 
     def log_prior_density(self) -> tf.Tensor:
@@ -248,6 +276,8 @@ def _validate_unconstrained_value(
 ) -> tf.Tensor:
     value = _cast_to_dtype(value, dtype)
     unconstrained_value = _to_unconstrained(value, transform)
+    if unconstrained_value.dtype.is_integer:
+        return unconstrained_value
     message = (
         "gpflow.Parameter: the value to be assigned is incompatible with this parameter's "
         "transform (the corresponding unconstrained value has NaN or Inf) and hence cannot be "
--- a/gpflow/base.py
+++ b/gpflow/base.py
@@ -114,20 +114,23 @@ class Parameter(tfp.util.TransformedVariable):
         trainable: Optional[bool] = None,
         dtype: Optional[DType] = None,
         name: Optional[str] = None,
-        pretransformed_shape: Optional[Sequence[Optional[int]]] = None,
-        transformed_shape: Optional[Sequence[Optional[int]]] = None,
+        unconstrained_shape: Optional[Sequence[Optional[int]]] = None,
+        constrained_shape: Optional[Sequence[Optional[int]]] = None,
         shape: Optional[Sequence[Optional[int]]] = None,
     ):
-        """
-        A parameter retains both constrained and unconstrained
-        representations. If no transform is provided, these two values will be the same.
-        It is often challenging to operate with unconstrained parameters. For example, a variance cannot be negative,
-        therefore we need a positive constraint and it is natural to use constrained values.
-        A prior can be imposed either on the constrained version (default) or on the unconstrained version of the parameter.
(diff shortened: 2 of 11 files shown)


gpflow/base.py
```diff
@@ -140,9 +147,30 @@ def __init__(
...
+            pretransformed_shape = shape
+            transformed_shape = shape
+
+        super().__init__(
+            value,
+            transform,
+            dtype=value.dtype,
+            trainable=trainable,
+            name=name,
+            shape=transformed_shape,
```
uri-granta, line 165: Is this the right way round? Looking at `TransformedVariable.__init__` suggests that `shape` gets passed to the _pretransformed_ variable as part of kwargs, whie `self._shape` is the calculated shape of the _transformed_ value. Or am I misreading it?

gpflow/models/vgp.py
```diff
@@ -70,12 +70,67 @@ def __init__(
...
+
+    def after_data_changed(self) -> None:
+        new_X_data, new_Y_data = self.data
+        new_num_data = tf.shape(new_X_data)[0]
+        f_mu, f_cov = self.predict_f(new_X_data, full_cov=True)  # [N, L], [L, N, N]
+
+        # This model is hard-coded to use the whitened representation, i.e.  q_mu and q_sqrt
+        # parametrize q(v), and u = f(X) = L v, where L = cholesky(K(X, X)) Hence we need to
+        # back-transform from f_mu and f_cov to obtain the updated new_q_mu and new_q_sqrt:
+        Knn = self.kernel(new_X_data, full_cov=True)  # [N, N]
```
uri-granta, line 118: This is going to significantly simplify the trieste `VariationalGaussianProcess` wrapper class :-)

tests/gpflow/test_base.py
```diff
@@ -213,3 +214,38 @@ def test_construct_parameter_from_existing_parameter_value_becomes_invalid() ->
...
+    for value in values:
+        parameter.assign(value)
+        np.testing.assert_equal(value, parameter.numpy())
+
+
+def test_construct_parameter_with_variable_shape__different_transformed_shape() -> None:
+    parameter = gpflow.Parameter(
+        [[1, 0], [2, 3]],
+        transform=triangular(),
+        pretransformed_shape=[None, None],
```
uri-granta, line 238: Again, unless I'm misunderstanding, `[None, None]` is the _transformed_ shape (after applying `triangular()`) and `[None]` is the _pretransformed_ shape (which we get by applying `triangular().inverse`)?

gpflow/models/training_mixins.py
```diff
@@ -72,6 +72,12 @@ def training_loss_closure(self, *, compile=True) -> Callable[[], tf.Tensor]:
             return tf.function(self.training_loss)
         return self.training_loss
 
+    def after_data_changed(self) -> None:
```
uri-granta, line 75: (there's something a bit ugly about the method name, though I'm not sure what. maybe the lack of a verb in the name of a method that returns `None`? how about: )

```suggestion
    def update_after_data_change(self) -> None:
```

uri-granta (approved):
Looks great! Just one important question regarding transformed vs. pretransformed.

gpflow/base.py
jesnie: > Is this the right way round? Looking at `TransformedVariable.__init__` suggests that `shape` gets passed to the _pretransformed_ variable as part of kwargs, whie `self._shape` is the calculated shape of the _transformed_ value. Or am I misreading it?

You're right. Those have been swapped. I'll swap them (back).

tests/gpflow/test_base.py
jesnie: > Again, unless I'm misunderstanding, `[None, None]` is the _transformed_ shape (after applying `triangular()`) and `[None]` is the _pretransformed_ shape (which we get by applying `triangular().inverse`)?

Fixed.

gpflow/models/training_mixins.py
jesnie: > (there's something a bit ugly about the method name, though I'm not sure what. maybe the lack of a verb in the name of a method that returns `None`? how about: )

Dunno. That's much too verbose for my taste. I was going for something that looks like a listener pattern. Maybe `on_data_change` or `data_changed`?

gpflow/models/training_mixins.py
```diff
@@ -72,6 +72,12 @@ def training_loss_closure(self, *, compile=True) -> Callable[[], tf.Tensor]:
...
+        """
+        If the initial data passed to the model were `tf.Variable`s, call this after the values
+        have changed so that the model can refresh intermediate values it may have cached.
+        """
+
```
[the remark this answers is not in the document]
jesnie: > Maybe clearer to have an explicit `pass` here (e.g. to emphasise that this isn't an abstract method)?
> 
> Also maybe mention this method in the class docstring alongside the other two methods?

Yeah, I had that first, but had to remove it. We require 97% test coverage, and adding a `pass` is an untested line and I dropped below the test coverage threshold. :shrug:

gpflow/models/vgp.py
jesnie: > This is going to significantly simplify the trieste `VariationalGaussianProcess` wrapper class :-)

I know. Did you seem my proof-of-concept PR? :-)

gpflow/models/training_mixins.py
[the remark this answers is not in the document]
uri-granta: Lol! FYI you can mark code blocks as not requiring coverage (see https://coverage.readthedocs.io/en/coverage-4.3.3/excluding.html#advanced-exclusion), though arguably the default `after_data_changed` noop _should_ have a simple test for it.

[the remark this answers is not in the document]
jesnie: > Lol! FYI you can mark code blocks as not requiring coverage (see https://coverage.readthedocs.io/en/coverage-4.3.3/excluding.html#advanced-exclusion), though arguably the default `after_data_changed` noop _should_ have a simple test for it.

I mean yes, but also no. I'm not going to write a unit test for `pass`.

gpflow/models/training_mixins.py
jesnie: > I mildly prefer `on_data_change` to `after_data_changed`, but also happy for you to leave it as it is.

I've renamed the method.

awav:
Thanks, @jesnie!

I'm worried about a couple of things:
- This solution covers only VGP case. Is there a plan / request to add this feature to other models? The code looks a bit like an unnecessary change at the moment and targeting a specific issue for a specific user of GPflow.
- The solution changes internal state of the model. Out of philosophical reasons, I would prefer to keep the immutability feature of models (of course on python level, not tensorflow). Indeed, in some scenarios like BO it is necessary to update models with new data, however, the same effect can be achieved without intervening into integrity of the base model class. In fact, before we switched to the tensorflow probability variables, the feature of dynamic shapes was supported, and it was straightforward to cope with it.

jesnie:
> * This solution covers only VGP case. Is there a plan / request to add this feature to other models? The code looks a bit like an unnecessary change at the moment and targeting a specific issue for a specific user of GPflow.

This is solving a problem that only affects the VGP, at least among the "main" models. The problem is that the VGP has variational parameters that depend on the data, so when the data changes shape the variational parameters also need to change shape.

- The GPR and SGPR does not have this problem because they don't have variational parameters.
- The SVGP does not have this problem because its variational parameters are defined with respect to the inducing points, and they don't change shape.

> * The solution changes internal state of the model. Out of philosophical reasons, I would prefer to keep the immutability feature of models (of course on python level, not tensorflow). Indeed, in some scenarios like BO it is necessary to update models with new data, however, the same effect can be achieved without intervening into integrity of the base model class. In fact, before we switched to the tensorflow probability variables, the feature of dynamic shapes was supported by Parameter class, and updating data in BO case was straightforward. I would suggest we take a look at this direction?!

I'm not sure what you mean: On a Python level I believe the model still is immutable? `on_data_change` only `assign`s to tensorflow variables. Or do you mean something else?

uri-granta:
This is in fact one of a couple of issues that we've encountered with using gpflow in a BO environment. The others are:

- VPG and SVGP store num_data as a number not a Variables, preventing us from updating it every loop. Furthermore, we can't just replace it with a Variable after initialisation either, as the elbo method assumes it's a number, so we've had to turn it into a property.
- GPR and VGP don't support initialising data from Variables with dynamic shape, though we can overwrite the .data attribute after initialisation.

So far we've been fixing these in trieste because there was nobody to do the work in gpflow. However, it certainly seems far more natural (and simpler) to support the BO use case out of the box in gpflow?

gpflow/models/vgp.py
```diff
@@ -181,6 +199,44 @@ class VGP(VGP_with_posterior):
     pass
 
 
+def update_vgp_data(vgp: VGP_deprecated, new_data: RegressionData) -> None:
+    """
+    Set the data on the given VGP model, and update it's variational parameters.
```
uri-granta, line 204:
```suggestion
    Set the data on the given VGP model, and update its variational parameters.
```

RELEASE.md
```diff
@@ -61,6 +61,8 @@ This release contains contributions from:
 ## Major Features and Improvements
 
 * Add new posterior class to enable faster predictions from the VGP model. (#1761)
+* VGP class bug-fixed to work with variable-sized data. Note you will have to call
+  `model.after_data_changed` whenever your data has changed. (#1774).
```
uri-granta, line 65:
```suggestion
  `update_vgp_data` whenever your data has changed. (#1774).
```

uri-granta (approved):
Looks good, thank you!

RELEASE.md
jesnie: I've made a different, but similar change. Done.

awav (commented):
I would rename the `pretransformed_shape` and `transformed_shape` to reflect their destinations: unconstrained and constrained shapes respectively. Otherwise, we will end up with two different names for the same entities.

PS1: are you sure about creating `triangular_size` utility function? :)

The rest LGTM.

PR #1783, opened 2022-02-22
Add type hints to kernels.

jesnie:
Add type hints to kernels.

### DIFF
--- a/gpflow/kernels/misc.py
+++ b/gpflow/kernels/misc.py
@@ -17,7 +17,7 @@ from typing import Optional
 import numpy as np
 import tensorflow as tf
 
-from ..base import Parameter
+from ..base import Parameter, TensorType
 from ..utilities import positive, to_default_float
 from .base import ActiveDims, Kernel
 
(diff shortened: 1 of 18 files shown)


gpflow/kernels/misc.py
```diff
@@ -101,8 +101,9 @@ def _J(self, theta):
             return 3.0 * tf.sin(theta) * tf.cos(theta) + (np.pi - theta) * (
                 1.0 + 2.0 * tf.cos(theta) ** 2
             )
+        assert False, f"Don't know how to handle order {self.order}."
```
uri-granta, line 104: (if you want codecov to leave you alone, you could instead replace the `elif self.order == 2:` above with an `else:` followed by an `assert self.order == 2`)

uri-granta (approved):
LGTM!

gpflow/kernels/misc.py
jesnie: > (if you want codecov to leave you alone, you could instead replace the `elif self.order == 2:` above with an `else:` followed by an `assert self.order == 2`)

Good idea. Done.

PR #1787, opened 2022-02-23
Add type hints to conditionals.

jesnie:
Add type hints to conditionals.

### DIFF
--- a/tests/gpflow/conditionals/test_multioutput.py
+++ b/tests/gpflow/conditionals/test_multioutput.py
@@ -1,12 +1,16 @@
+from typing import Callable, List, Sequence, Tuple
+
 import numpy as np
 import pytest
 import scipy
 import tensorflow as tf
+from _pytest.fixtures import SubRequest
 
 import gpflow
 import gpflow.inducing_variables.multioutput as mf
 import gpflow.kernels.multioutput as mk
 from gpflow import set_trainable
+from gpflow.base import MeanAndVariance, RegressionData
 from gpflow.conditionals import sample_conditional
 from gpflow.conditionals.util import (
     fully_correlated_conditional,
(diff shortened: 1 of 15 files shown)


tests/gpflow/conditionals/test_multioutput.py
```diff
@@ -23,38 +27,36 @@
 float_type = default_float()
 rng = np.random.RandomState(99201)
 
-
 # ------------------------------------------
 # Helpers
 # ------------------------------------------
 
 
-def predict(model, Xnew, full_cov, full_output_cov):
```
uri-granta, line 32: (what an odd function!)

tests/gpflow/conditionals/test_multioutput.py
```diff
@@ -23,38 +27,36 @@
...
     for model in models:
-        m, v = predict(model, Xnew, full_cov, full_output_cov)
+        m, v = model.predict_f(Xnew, full_cov=full_cov, full_output_cov=full_output_cov)
         ms.append(m)
         vs.append(v)
     return ms, vs
 
 
-def assert_all_array_elements_almost_equal(arr, decimal):
+def assert_all_array_elements_almost_equal(arr: Sequence[tf.Tensor]) -> None:
```
uri-granta, line 49: Do you think the original intent might have been to set `atol` to `0.1**decimal`?

uri-granta (approved):
Looks good!

tests/gpflow/conditionals/test_multioutput.py
jesnie: > Do you think the original intent might have been to set `atol` to `0.1**decimal`?

I think the original intent was to do `np.testing.assert_almost_equal(..., decimal=decimal)`.
Apparently that turned into `np.testing.asert_allclose(..., atol=1e-5)`.
:shrug:

PR #1789, opened 2022-02-24
Add type hints to conditionals, monitoring, and optimizers.

jesnie:


gpflow/monitor/base.py
```diff
@@ -77,30 +77,34 @@ class MonitorTaskGroup:
...
         """
         :param task_or_tasks: a single instance or a list of `MonitorTask` instances.
             Each `MonitorTask` in the list will be run with the given `period`.
         :param period: defines how often to run the tasks; they will execute every `period`th step.
             For large values of `period` the tasks will be less frequently run. Defaults to
             running at every step (`period = 1`).
         """
-        self.tasks = task_or_tasks
+        self._tasks: Collection[MonitorTask] = []
+        self.tasks = task_or_tasks  # type: ignore
```
uri-granta, line 91: Does mypy not handle the property setter?

gpflow/monitor/base.py
```diff
@@ -77,30 +77,34 @@ class MonitorTaskGroup:
...
         self._period = period
 
     @property
-    def tasks(self) -> List[MonitorTask]:
+    def tasks(self) -> Collection[MonitorTask]:
```
uri-granta, line 95: Since you're converting it to a list anyway, wouldn't it be more polite to return it as one too (or at least as a `Sequence`)?

uri-granta (approved):
LGTM

PR #1792, opened 2022-03-01
Add back a len(inducing_points) to break our users less.

jesnie:
Add back a `len(inducing_points)` to break our users less.

`len(inducing_points)` was removed because its typing was weird/confusing/misleading. However this turns out to cause breakages downstream. Add it back, but marked as `deprecated`.


uri-granta (approved):
Looks good, thanks

PR #1796, opened 2022-03-01
Support anonymous dimensions in check_shapes.

jesnie:


gpflow/experimental/check_shapes/check_shapes.py
```diff
@@ -54,14 +54,14 @@ def _check_shapes(func: C) -> C:
         def wrapped(*args: Any, **kwargs: Any) -> Any:
             try:
                 bound_arguments = signature.bind(*args, **kwargs)
-            except TypeError:
+            except TypeError as e:
                 # TypeError is raised if *args and **kwargs don't actually match the arguments of
                 # `func`. In that case we just call `func` normally, which will also result in an
                 # error, but an error with the error message the user is used to.
```
vdutor, line 60: I don't understand what is going on here, can you elaborate please?

jesnie: > I don't understand what is going on here, can you elaborate please?

Let's say the user writes something like:
```python
@check_shapes(
   ...
)
def f(x: tf.Tensor) -> tf.Tensor:
    ...

f(x, y)  # <---Notice how the parameters are wrong
```

Now, because of the `@check_shapes` the user is not actually calling `f` directly, but instead calling the `wrapped` in the code you are commenting on. The `signature.bind` above will raise a `TypeError` because the user is calling `f` with the wrong parameters. Of course we want to give the user an excellent and easy-to-understand error message. How do we do that? We make `wrapped` call `f(x, y)` (which is wrong), but that will generate exactly the same error message as if the user had called `f(x, y)` directly, which hopefully is an error that is easy to understand for the user.

Does this make sense?

vdutor: thanks!

vdutor (approved):

PR #1797, opened 2022-03-01
Support Optional values for check_shapes.

jesnie:


vdutor (commented):
Can you point me to the docstring where the behaviour of passing an argument as optional is explained? Seems to be missing from this PR.

jesnie:
> Can you point me to the docstring where the behaviour of passing an argument as optional is explained? Seems to be missing from this PR.

It's here: https://github.com/GPflow/GPflow/pull/1797/files#diff-5aba21148f5f3e45da7c29df2864f03b773e3669944637a8788554373a7c822bR65

vdutor (approved):

PR #1799, opened 2022-03-02
Add type hints to models.

jesnie:


gpflow/models/gplvm.py
```diff
@@ -145,7 +149,8 @@ def __init__(
         assert self.X_prior_var.shape[0] == self.num_data
         assert self.X_prior_var.shape[1] == self.num_latent_gps
 
-    def maximum_log_likelihood_objective(self) -> tf.Tensor:
+    # type-ignore is because of changed method signature:
+    def maximum_log_likelihood_objective(self) -> tf.Tensor:  # type: ignore
```
uri-granta, line 153: Why does this have an incompatible method? Doesn't that cause issues? Is it basically the case that you can only call `maximum_log_likelihood_objective` if you know the precise type of GPModel?

Presumably the abstract `maximum_log_likelihood_objective` and `_training_loss` methods include args and kwargs parameters to suggest that submethods might expect additional arguments?

gpflow/models/gplvm.py
```diff
@@ -262,5 +267,6 @@ def predict_f(
             var = tf.tile(tf.expand_dims(var, 1), shape)
         return mean + self.mean_function(Xnew), var
 
-    def predict_log_density(self, data: OutputData) -> tf.Tensor:
+    # type-ignore is because of changed method signature:
+    def predict_log_density(self, data: OutputData) -> tf.Tensor:  # type: ignore
```
uri-granta, line 271: Why does this even change the abstract signature, given that it's not implemented?!

gpflow/models/sgpr.py
```diff
@@ -151,15 +150,17 @@ class SGPR_deprecated(SGPRBase_deprecated):
...
         return self.elbo()
 
-    def _common_calculation(self):
+    def _common_calculation(self) -> "SGPR.CommonTensors":
         """
         Matrices used in log-det calculation
 
-        :return: A , B, LB, AAT with :math:`LLᵀ = Kᵤᵤ , A = L⁻¹K_{uf}/σ, AAT = AAᵀ, B = AAT+I, LBLBᵀ = B`
-        A is M x N, B is M x M, LB is M x M, AAT is M x M
+        :return: A , B, LB, AAT with :math:`LLᵀ = Kᵤᵤ , A = L⁻¹K_{uf}/σ, AAT = AAᵀ,
```
uri-granta, line 161: Those superscript and subscript unicode characters inside the :math: block are giving me a headache, but that's not part of this PR!

tests/gpflow/models/test_svgp_posterior.py
```diff
@@ -10,7 +11,14 @@
...
 
-def make_models(M=64, D=input_dim, L=3, q_diag=False, whiten=True, mo=None):
+def make_models(
+    M: int = 64,
+    D: int = input_dim,
+    L: int = 3,
+    q_diag: bool = False,
+    whiten: bool = True,
+    mo: Optional[Tuple[str, str]] = None,
+) -> Tuple[SVGP_deprecated, SVGP_with_posterior]:
```
uri-granta, line 21: Out of curiosity, why is the `SVGP_deprecated` base class called that?

uri-granta (approved):
Looks good!

gpflow/models/gplvm.py
jesnie: > Why does this have an incompatible method? Doesn't that cause issues? Is it basically the case that you can only call `maximum_log_likelihood_objective` if you know the precise type of GPModel?
> 
> Presumably the abstract `maximum_log_likelihood_objective` and `_training_loss` methods include args and kwargs parameters to suggest that submethods might expect additional arguments?

Yes, it's because you can pass arbitrary `*args` and `**kwargs` in the super class. Yes, that violates the "Liskov Substitution Principle", which isn't great. But this isn't the time and place to work on that.

gpflow/models/gplvm.py
jesnie: > Why does this even change the abstract signature, given that it's not implemented?!

You're right. I'll update it to have the correct signature instead.

PR #1802, opened 2022-03-03
Fix problems discovered by numpy with type hints.

jesnie:
Basically various fixes to allow us to run with both older and newer versions of `{Python, TensorFlow, numpy, mypy}`.

### DIFF
--- a/gpflow/base.py
+++ b/gpflow/base.py
@@ -26,12 +26,19 @@ if TYPE_CHECKING:
     from IPython.lib import pretty
 
 DType = Union[np.dtype, tf.DType]
-VariableData = Union[List[Any], Tuple[Any], np.ndarray, int, float]  # deprecated
+
+# It would be nice to use something more interesting than `Any` here, but it looks like the
+# infrastructure in the rest of the ecosystem isn't really set up for this yet. Maybe when we
+# get Python 3.11?
+# Also, this type hint is as str, because With numpy < 1.20.0 the other one doesn't work at runtime.
+AnyNDArray = Union["np.ndarray[Any, Any]"]
+
+VariableData = Union[List[Any], Tuple[Any], AnyNDArray, int, float]  # deprecated
 Transform = Union[tfp.bijectors.Bijector]
 Prior = Union[tfp.distributions.Distribution]
 
 
-TensorType = Union[np.ndarray, tf.Tensor, tf.Variable, "Parameter"]
+TensorType = Union[AnyNDArray, tf.Tensor, tf.Variable, "Parameter"]
 """
 Type alias for tensor-like types that are supported by most TensorFlow and GPflow operations.
 
--- a/gpflow/expectations/expectations.py
+++ b/gpflow/expectations/expectations.py
@@ -121,7 +121,8 @@ def _init_expectation(
     obj1, feat1 = obj1 if isinstance(obj1, tuple) else (obj1, None)
     obj2, feat2 = obj2 if isinstance(obj2, tuple) else (obj2, None)
     return (
-        cast(ProbabilityDistribution, p),
+        # type-ignore instead of cast, because it dependes on versions whether a cast is necessary.
+        p,  # type: ignore
         cast(ExpectationObject, obj1),
         cast(Optional[InducingVariables], feat1),
         cast(ExpectationObject, obj2),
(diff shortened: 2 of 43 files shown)


gpflow/base.py
```diff
@@ -26,12 +26,22 @@
     from IPython.lib import pretty
 
 DType = Union[np.dtype, tf.DType]
-VariableData = Union[List[Any], Tuple[Any], np.ndarray, int, float]  # deprecated
+
+if TYPE_CHECKING:
+    # It would be nice to use something more interesting than `Any` here, but it looks like the
+    # infrastructure in the rest of the ecosystem isn't really set up for this yet. Maybe when we
+    # get Python 3.11?
+    AnyNdArray = Union[np.ndarray[Any, Any]]
```
uri-granta, line 34: I ended up using just writing `"np.ndarray[Any, Any]"` when I needed it (with quotes), which works both at type checking and runtime, but this is probably nicer. Fortunately we have very few ndarrays in trieste, as it's nearly all tensors.

gpflow/expectations/expectations.py
```diff
@@ -121,7 +121,8 @@ def _init_expectation(
     obj1, feat1 = obj1 if isinstance(obj1, tuple) else (obj1, None)
     obj2, feat2 = obj2 if isinstance(obj2, tuple) else (obj2, None)
     return (
-        cast(ProbabilityDistribution, p),
+        # type-ignore instead of cast, because it dependes on versions whether a cast is necessary.
+        p,  # type: ignore
```
uri-granta, line 125: Do you not get an unnecessary type ignore error then in one of the versions? Update: just spotted the config change.

gpflow/base.py
uri-granta: Also, to be consistent with `numpy.typing.NDArray`, you may want to name this `AnyNDArray`

gpflow/models/util.py
```diff
@@ -12,20 +12,26 @@
...
 from .training_mixins import Data, ExternalDataTrainingLossMixin
 
+_I = TypeVar("_I", bound=InducingVariables)
+_InducingILike = Union[_I, tf.Tensor, AnyNdArray]
 
-def inducingpoint_wrapper(
-    inducing_variable: Union[InducingVariables, tf.Tensor, np.ndarray]
-) -> InducingVariables:
+
+InducingVariablesLike = _InducingILike[InducingVariables]
```
uri-granta, line 30: Do you use `_InducingILike` anywhere else? If not, isn't it less confusing to just define:

```python
InducingVariablesLike = Union[InducingVariables, tf.Tensor, AnyNdArray]
InducingPointsLike = Union[InducingPoints, tf.Tensor, AnyNdArray]
```

tests/gpflow/models/test_variational.py
```diff
@@ -30,38 +31,44 @@
...
+    meanA: AnyNdArray, meanB: AnyNdArray, varA: AnyNdArray, varB: AnyNdArray
+) -> AnyNdArray:
     # KL[ qA | qB ] = E_{qA} \log [qA / qB] where qA and qB are univariate normal distributions.
     return 0.5 * (
-        np.log(varB) - np.log(varA) - 1.0 + varA / varB + (meanB - meanA) * (meanB - meanA) / varB
+        np.log(varB)
+        - np.log(varA)
+        - 1.0
+        + varA / varB
+        + cast(AnyNdArray, meanB - meanA) * cast(AnyNdArray, meanB - meanA) / varB
```
uri-granta, line 57: (why is this necessary? what does mypy think otherwise?)

uri-granta (approved):
Not too painful!

gpflow/models/util.py
jesnie: > Do you use `_InducingILike` anywhere else? If not, isn't it less confusing to just define:
> 
> ```python
> InducingVariablesLike = Union[InducingVariables, tf.Tensor, AnyNdArray]
> InducingPointsLike = Union[InducingPoints, tf.Tensor, AnyNdArray]
> ```

Sure. I used to use it, but then stuff didn't work, and I guess I don't need it anymore.

gpflow/expectations/expectations.py
jesnie: > Do you not get an unnecessary type ignore error then in one of the versions? Update: just spotted the config change.

:+1:  Yes, there are a couple of places where I've given up on writing code that is type-correct with all versions of numpy.

PR #1807, opened 2022-03-07
Test builds with both minimum and maximum supported versions.

jesnie:
1. Test builds with both minimum and maximum supported versions of Python, Tensorflow and Tensorflow-probability.
2. Fix a couple of tests/bugs that fail with the newer versions.
3. Update package classifiers.


.circleci/config.yml
```diff
@@ -25,14 +56,24 @@ commands:
             pip install --progress-bar=off
               -e .
               -r tests_requirements.txt
-              tensorflow==2.6.* tensorflow-probability==0.14.* \n
+              tensorflow==<<parameters.tf_ver>>.*
```
uri-granta, line 59: Can you omit the trailing `.*`? (in case you ever want a min/max version that includes the point version)

setup.py
```diff
@@ -70,12 +70,21 @@ def read_file(filename: str) -> str:
...
         "Natural Language :: English",
         "Operating System :: MacOS :: MacOS X",
         "Operating System :: Microsoft :: Windows",
         "Operating System :: POSIX :: Linux",
         "Programming Language :: Python :: 3.7",
+        "Programming Language :: Python :: 3.8",
+        "Programming Language :: Python :: 3.9",
+        "Programming Language :: Python :: 3.10",
         "Topic :: Scientific/Engineering :: Artificial Intelligence",
+        "Typing :: Typed",
```
uri-granta, line 88: cute

uri-granta (approved):
Looks good!

.circleci/config.yml
jesnie: > Can you omit the trailing `.*`? (in case you ever want a min/max version that includes the point version)

Hmm, it appears that I can.

setup.py
jesnie: :smile:

PR #1809, opened 2022-03-08
Make posterior predict more memory efficient

uri-granta:
Calling `predict_f(Xnew, full_cov=False)` on a GPR posterior has an unnecessary quadratic space usage. Specifically, calculating the diagonal covariance matrix in `GPRPosterior._conditional_with_precompute` calculates an intermediate NxN matrix, which is unnecessary.

Note that this code does not seem to be tested in `test_posteriors.py` but it is covered by `test_bo_integration`.


jesnie (approved):
Very nice. Thank you.

uri-granta:
> Very nice. Thank you.

Don't thank me yet! CI revealed some tests that I didn't spot. Will take a look to see why they're failing once I'm in the office.

PR #1811, opened 2022-03-10
Speed up posterior computation

uri-granta:
Turns out the recent change to posteriors to make computations more memory efficient also slowed them down slightly for small sample sizes. This version appears to be as fast as previously for small sample sizes (and significantly faster for large sample sizes).


jesnie (approved):
LGTM

PR #1812, opened 2022-03-10
Enable warnings when returning 'Any'.

jesnie:
Enable warnings when returning 'Any'.
Most of the fixes are "type-ignores" or `cast`s, but in several places incorrect type annotations were discovered.


uri-granta (approved):
Looks good!

PR #1816, opened 2022-03-15
Be uptight about __all__, and fix revealed mypy errors.

jesnie:
Disallow implicit reexports in `mpyp`, which requires us to be more correct about our use of `__all__`.
Added some unit tests to make sure of systematic use of `__all__`.
For some reason this PR makes `mypy` angry about some uses of `np.ndarray`, so I've fixed those problems.


tests/gpflow/test_all.py
```diff
@@ -0,0 +1,110 @@
...
+        if is_all_assignment(t):
+            all_assignment = t  # type: ignore
+    assert all_assignment is not None, f"No `__all__` found in module {module}."
+
+    error_message = "__all__ must be a static list of constant strings. Some tools expect this."
+
+    all_value = all_assignment.value
+    assert isinstance(all_value, ast.List), error_message
+
+    assert all(
```
uri-granta, line 106: (nice tests!)

mypy.ini
```diff
@@ -8,6 +8,3 @@ disallow_untyped_decorators = False
 ; This is needed because version 1.20.0 of NumPy introduces typing, so some ignores are necessary /
 ; unnecessary depending on the version of numpy:
 warn_unused_ignores = False
-; It would actually be good to disallow implicit reexport, but that will take some effort to be
-; strict about our use of __all__.
-implicit_reexport = True
```
uri-granta, line 13: Don't you need to explicitly set this to False? (the docs suggest it defaults to True)

tests/gpflow/conditionals/test_broadcasted_conditionals.py
```diff
@@ -65,7 +65,7 @@ def test_conditional_broadcasting(full_cov: bool, white: bool, conditional_type:
     and conditional should broadcast over the [...].
     """
     q_mu = np.random.randn(Data.M, Data.Dy)
-    q_sqrt = np.tril(np.random.randn(Data.Dy, Data.M, Data.M), -1)
+    q_sqrt: AnyNDArray = np.tril(np.random.randn(Data.Dy, Data.M, Data.M), -1)
```
uri-granta, line 68: I'm really curious why these have suddenly started being necessary. What error do you get without the annotations?

uri-granta (approved):
Just one question re the config

mypy.ini
jesnie: > Don't you need to explicitly set this to False? (the docs suggest it defaults to True)

The `strict = True` above sets this to `False`.

tests/gpflow/test_all.py
jesnie: > (nice tests!)

Thanks!

tests/gpflow/conditionals/test_broadcasted_conditionals.py
jesnie: > I'm really curious why these have suddenly started being necessary. What error do you get without the annotations?

```
tests/gpflow/conditionals/test_broadcasted_conditionals.py:68: error: Need type annotation for "q_sqrt"
```
:shrug:

PR #1822, opened 2022-03-23
Rework check_shapes errors.

jesnie:
1. Build a framework of `ErrorContext`s for describing where/why an error handled.
2. Use those to make sure all user-facing exceptions are written in a consistent format.
3. This particularly improves error messages in case of errors in the definition of the expected shape; and when we're unable to extract any shape from an object.


gpflow/experimental/check_shapes/error_contexts.py
```diff
@@ -0,0 +1,436 @@
...
+        flat: List[ErrorContext] = []
+        # Merge any nested parallel contexts.
+        self._flatten(self, flat)
+
+        # If any `StackContexts`s have the same (grand-) parent, merge them, so that we don't repeat
+        # stuff in the output.
+        by_head: Dict[ErrorContext, List[ErrorContext]] = {}
+        for child in flat:
+            if isinstance(child, StackContext):
+                head, body = self._split_head(child)
```
uri-granta, line 213: (can you not just `._flatten` child to get the head, since you'll be doing that later anyway?)

gpflow/experimental/check_shapes/parser.py
```diff
@@ -246,41 +257,113 @@ def info_field_args(self, tree: Tree[Token]) -> str:
...
+_ARGUMENT_SPEC_CACHE: Dict[str, ParsedArgumentSpec] = {}
+_SPHINX_DOCSTRING_CACHE: Dict[Tuple[str, Tuple[ParsedArgumentSpec, ...]], str] = {}
+
+
+def parse_argument_spec(argument_spec: str, context: ErrorContext) -> ParsedArgumentSpec:
     """
     Parse a `check_shapes` argument specification.
     """
-    tree = _ARGUMENT_SPEC_PARSER.parse(argument_spec)
+    cached_value = _ARGUMENT_SPEC_CACHE.get(argument_spec)
```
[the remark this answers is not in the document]
jesnie: > I might be being stupid, but are you actually updating the caches anywhere?

Doh! Well spotted. Thanks!

gpflow/experimental/check_shapes/parser.py
```diff
@@ -246,41 +257,113 @@ def info_field_args(self, tree: Tree[Token]) -> str:
...
+            StackContext(
+                context,
+                LarkUnexpectedInputContext(argument_spec, e, _ARGUMENT_SPEC_TERMINAL_DESCRIPTIONS),
+            )
+        ) from e
     specs: ParsedArgumentSpec = _ParseArgumentSpec().visit(tree)
     return specs
 
 
-@lru_cache(maxsize=None)
```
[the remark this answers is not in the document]
jesnie: > As an aside, the cachetools library supports specifying the cache key. Something like:
> 
> ```python
> @cached(cache={}, key=lambda docstring, argument_specs, context: hashkey((dostring, argument_specs)))
> def _parse_and_rewrite_sphinx_docstring(
> ```

Yeah, no, dunno. It's another dependency, it is seems like it's actually simpler to just manually update the cache (though apparently I'm not very good at that...)

tests/gpflow/experimental/check_shapes/test_check_shapes.py
```diff
@@ -382,33 +394,65 @@ def h(a: TensorType, b: TensorType) -> TensorType:
...
         "return: [d...]",
     )
-    def g(a: TensorType, b: TensorType) -> TensorType:
+    def g(a: TestShaped, b: TestShaped) -> TestShaped:
         return a
 
     with pytest.raises(ShapeMismatchError):
         g(t(2, 3), t(2, 4))
-        get_check_shapes(g)(h)  # Don't crash.
+        get_check_shapes(g)(h)  # pylint: disable=not-callable    # Don't crash.
```
uri-granta, line 415: (out of curiosity: any idea why pylint complains about this?)

tests/gpflow/experimental/check_shapes/test_check_shapes.py
```diff
@@ -382,33 +394,65 @@ def h(a: TensorType, b: TensorType) -> TensorType:
...
+
+    with pytest.raises(ShapeMismatchError) as e:
+        f(t(2, 3), t(3, 4))
+
+    (message,) = e.value.args
+    assert (
+        f"""
+Tensor shape mismatch in call to function.
+  Function: test_check_shapes__error_message.<locals>.f
+    Declared: {__file__}:430
```
uri-granta, line 446: Would be nice to avoid having to hardcode the line number so the test doesn't break whenever anyone updated the tests above it. Can you get the check shape line number using something like https://stackoverflow.com/questions/3056048/filename-and-line-number-of-python-script ?

tests/gpflow/experimental/check_shapes/test_parser.py
```diff
@@ -745,19 +747,155 @@ def __str__(self) -> str:
...
+    [
+        (
+            "a [batch..., x]",
+            f"""
+Unable to parse shape specification.
+  check_shapes called at: {__file__}:{_CHECK_SHAPES_LINE}
+    Argument number (0-indexed): 0
+      Line:     "a [batch..., x]"
+                    ^
+      Expected: integer (re=(?:[0-9])+)
```
uri-granta, line 782: (a pity it can't infer the missing : but I guess it's true that it expects an integer now)

tests/gpflow/experimental/check_shapes/test_parser.py
```diff
@@ -745,19 +747,155 @@ def __str__(self) -> str:
...
+        check_shapes(spec)
+    (message,) = e.value.args
+    assert expected_message == message
+
+
+def test_parse_and_rewrite_docstring__error() -> None:
+    # I don't actually know how to provoke an error from parsing the docstring.
+    # The problem is that *any* string is allow as free-form documentation.
+    # Update this test if you find a good example."
+    assert True
```
uri-granta, line 901: So if the parser encounters a badly formatted info field does it assume everything until then must have been part of the free-form documentation? (I guess it won't matter too much as a badly formatted info string would fail to compile with sphinx anyway.)

tests/gpflow/experimental/check_shapes/test_error_contexts.py
```diff
@@ -0,0 +1,575 @@
...
+        b1.add_columned_line(2, 22)
+        with b1.indent() as b2:
+            b2.add_columned_line(3, 333)
+        with b1.indent() as b2:
+            b2.add_columned_line(444, 4)
+        b1.add_columned_line(2222, 2)
+    builder.add_columned_line(1, 11)
+    assert (
+        """
+11 11111
```
uri-granta, line 127: Was just going to suggest this test :-)

tests/gpflow/experimental/check_shapes/test_error_contexts.py
```diff
@@ -0,0 +1,575 @@
...
+    assert expected == to_str(context)
+
+
+def test_function_call_context() -> None:
+    def f() -> str:
+        return to_str(FunctionCallContext(f))
+
+    assert (
+        f"""
+f called at: {__file__}:305
```
uri-granta, line 307: (again you may be able to extract 305 using `getframeinfo(currentframe()).lineno`)

uri-granta (approved):
Looks good! I really like the error messages.

tests/gpflow/experimental/check_shapes/test_parser.py
jesnie: > So if the parser encounters a badly formatted info field does it assume everything until then must have been part of the free-form documentation? (I guess it won't matter too much as a badly formatted info string would fail to compile with sphinx anyway.)

Yes.

tests/gpflow/experimental/check_shapes/test_parser.py
jesnie: > (a pity it can't infer the missing : but I guess it's true that it expects an integer now)

Yeah, it's limited what the poor LALR parsers can do.

tests/gpflow/experimental/check_shapes/test_check_shapes.py
jesnie: > Would be nice to avoid having to hardcode the line number so the test doesn't break whenever anyone updated the tests above it. Can you get the check shape line number using something like https://stackoverflow.com/questions/3056048/filename-and-line-number-of-python-script ?

It would be nice, but I don't like the idea of using `inspect` and looking a the stack, because that's exactly what we're doing in the production code. I have seen a couple of instances of people writing the same algorithm in the production code and the test and then checking that they give the same result. That doesn't work if the algorithm is fundamentally wrong - you just check that the wrong result is the same as the same wrong result.

tests/gpflow/experimental/check_shapes/test_error_contexts.py
jesnie: > (again you may be able to extract 305 using `getframeinfo(currentframe()).lineno`)

Again, I don't really like the idea of `currentframe`, because that is (mostly) the same as what we're doing in the production code.

tests/gpflow/experimental/check_shapes/test_check_shapes.py
jesnie: > (out of curiosity: any idea why pylint complains about this?)

Hmm, it doesn't seem to do it any more. Maybe I had a buggy version of `pylint`? It does seem like a weird error.

PR #1828, opened 2022-03-24
Don't repeat documentation for functions used multiple times in multiple-dispatch

jesnie:
Notice how the documentation for `gpflow.conditionals.multioutput.conditionals.separate_independent_conditional` is repeated multiple times in https://gpflow.github.io/GPflow/2.4.0/api/gpflow/conditionals/index.html#gpflow-conditionals-conditional
because it is registered for multiple different types fo multiple dispatch.
This PR merges those sections into one.


frgsimpson (approved):
LGTM

PR #1829, opened 2022-03-24
Remove numpydoc and fix unexpected sections.

jesnie:
`numpydoc`, and its associated subsections seems to be poorly used, and causing the majority of our documentation warnings.
This PR reduces the number of warnings from 462 to 142.


frgsimpson (approved):
LGTM!

PR #1833, opened 2022-03-28
Rewrite generate_module_rst to fix various bugs.

jesnie:
The previous implementation:

1. Attempted to do everything in a single pass, which is of course a noble goal, but
2. Would not take into account that functions that are part of a multiple-dispatch would be printed as part of that dispatch, so they were printed twice.
3. Would first check whether a module is empty, and then filter it's contents, which can cause a module to become empty, so it would create empty modules.

I've concluded that it's infeasible to fix those bugs while retaining the single pass, and in the process I ended up completely rewriting the entire file. New process is:

1. Collect everything.
2. Filter out duplicate functions.
3. Filter out empty modules.
4. Write out result.

I've also slightly changed the layout of the generated page, so the output for a module now is:

1. Docstring for the module itself.
2. List of sub-modules.
3. List of classes.
4. List of functions.

Instead of the sub-modules, classes, and functions all being mixed together.


frgsimpson (approved):
LGTM!

PR #1834, opened 2022-03-29
Fix new version of click breaking old version of black.

jesnie:
Newer versions of `black` are incompatible with TensorFlow 2.4, which we must support, so we've pinned `black` to an old version.
There's just been a new release of `click`, 8.1.0, which now breaks our old version of `black`, so now we have to pin `click` as well...

### DIFF
--- a/tests_requirements.txt
+++ b/tests_requirements.txt
@@ -1,13 +1,12 @@
 mypy
-black==20.8b1
+black==20.8b1  # Use old version of black, because newer version is incompatible with our TF 2.4.
 isort>=5.1
 pytest>=3.5.0
 pytest-cov
 pytest-random-order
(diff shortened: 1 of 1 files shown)


tests_requirements.txt
```diff
@@ -1,13 +1,12 @@
...
-black==20.8b1
+black==20.8b1  # Use old version of black, because newer version is incompatible with our TF 2.4.
 isort>=5.1
 pytest>=3.5.0
 pytest-cov
 pytest-random-order
-pytest-xdist  # for local tests only
+pytest-xdist
 codecov
-types-dataclasses
```
jesnie, line 9: This is unnecessary since we've upgraded to Python 3.7.

PR #1838, opened 2022-03-30
Attempt to make `test_check_shapes__disable__speed[tf.function]` less flaky

jesnie:
I'm not happy about doing this, but the TensorFlow compilation time seems to have much higher variance than the difference between with- / without `check_shapes`.

### DIFF
--- a/tests/gpflow/experimental/check_shapes/test_integration.py
+++ b/tests/gpflow/experimental/check_shapes/test_integration.py
@@ -112,6 +112,14 @@ def test_check_shapes__tensorflow_compilation(
 
 @pytest.mark.parametrize("func_wrapper", [lambda x: x, tf.function], ids=["none", "tf.function"])
 def test_check_shapes__disable__speed(func_wrapper: Callable[[Any], Any]) -> None:
+    if func_wrapper is tf.function:
+        pytest.skip(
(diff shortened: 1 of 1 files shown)


tests/gpflow/experimental/check_shapes/test_integration.py
```diff
@@ -110,8 +110,17 @@ def loss() -> tf.Tensor:
     np.testing.assert_allclose(target, v.numpy(), atol=0.01)
 
 
-@pytest.mark.parametrize("func_wrapper", [lambda x: x, tf.function], ids=["none", "tf.function"])
-def test_check_shapes__disable__speed(func_wrapper: Callable[[Any], Any]) -> None:
+@pytest.mark.parametrize(
+    "func_wrapper, scale",
+    [
+        (lambda x: x, 1.0),
+        # Large scale becasue this is super flaky. Presumable because the TensorFlow compilation time is
```
uri-granta, line 117: Why not restrict the time measurements to the execution of f then?

alansaul:
LGTM since I don't think this existing test is currently testing the speed anyway, but will still sporadically fail and break the build. We should make some valid speed tests soon to test that there isn't a performance regression from using shape checking, maybe make an issue to remind us?

alansaul (approved):

PR #1840, opened 2022-03-31
Support checking all elements of collections.

jesnie:
What we're trying to achieve is to enable the checking of all members of collections. For example we might have a function (`concat_rows`) that takes a collection of array, and require:

1. All inputs have rank 2.
2. The inputs may have any number of rows.
3. The inputs must have the same number of columns.
4. The output has rank 2.
5. The output may have any number of rows.
6. The output has the same number of columns as the input.

```python
check_shapes(
    "xs[all]: [., n_cols]",
    "return: [., n_cols]",
)
def concat_rows(xs: Sequence[np.ndarray]) -> np.ndarray:
    return np.concatenate(xs, axis=0)
```

To do that we:

1. Update `ArgumentRef` and friends to handle collections. Previously `get` would return a single value and `error_context` would return a single `ErrorContext`. In this PR I remove `error_context` and update `get` to return a list of `(value, error_context)` tuples. Then I update all the `ArgumentRef` implementations to handle sequences. Note that previously missing values were handled by returning `None` - now I handle them by returning an empty list.
7. Update `check_shapes` to handle that the `ArgumentRef` now returns a list, and check all elements of that list.
8. Create a common abstract base class (`DelegatingArgumentRef`) for `ArgumentRef`s that are a wrapper around another `ArgumentRef` - which is most of the implementations. This is just for code reuse.
9. Add new syntax for selecting all elements of collections to `check_shapes.lark`. I use `foo[all]` to select all elements of a sequence, `foo.keys()` to select the keys of a mapping, and `foo.values()` to select the values from a mapping. I'm not super happy about the inconsistent syntax - any ideas?
10. Add new `ArgumentRef` implementations (`AllElementsRef`, `KeysRef`, and `ValuesRef`) corresponding to the three new syntaxes.
11. Add new `ErrorContext`s (`MappingKeyContext` and `MappingValueContext`) to be used by errors generated by (`KeysRef` and `ValuesRef`) respectively. (`AllElementsRef` uses the pre-existing `IndexContext`.)
12. Update `parser.py` to translate the new syntax into the new arguments refs.
13. Update our documentation to mention the new syntax.
14. Update our unit tests.

### DIFF
--- a/gpflow/experimental/check_shapes/__init__.py
+++ b/gpflow/experimental/check_shapes/__init__.py
@@ -53,6 +53,9 @@ The ``<argument specifier>`` can then be modified to refer to elements of the ob
 * Use ``.<name>`` to refer to attributes of the object.
 * Use ``[<index>]`` to refer to elements of a sequence. This is particularly useful if your
   function returns a tuple of values.
+* Use ``[all]`` to select all elements in a collection.
+* Use ``.keys()`` to select all keys in a mapping.
+* Use ``.values()`` to select all values in a mapping.
 
 We do not support looking up values in a  ``dict``.
 
@@ -65,7 +68,10 @@ For example::
         "data.training_data: ...",
         "return: ...",
         "return[0]: ...",
-        "something[0].foo.bar[23]: ...",
+        "data_list[all]: ...",
+        "data_dict.keys(): ...",
+        "data_dict.values(): ...",
+        "something[all].foo[0].bar.values(): ...",
     )
     def f(...):
         ...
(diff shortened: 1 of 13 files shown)


gpflow/experimental/check_shapes/__init__.py
```diff
@@ -55,6 +55,9 @@ def linear_model(
     * Use ``.<name>`` to refer to attributes of the object.
     * Use ``[<index>]`` to refer to elements of a sequence. This is particularly useful if your
       function returns a tuple of values.
+    * Use ``[all]`` to select all elements in a collection.
+    * Use ``.keys()`` to select all keys in a mapping.
+    * Use ``.values()`` to select all values in a mapping.
```
jesnie, line 60: I'm not sure how happy I am about the syntax here. Any better ideas?

awav:
Hi @jesnie, can you please add a description to this PR? It is hard to parse what is going on in the diff :)

jesnie:
> Hi @jesnie, can you please add a description to this PR? It is hard to parse what is going on in the diff :)

Done. Hope it helps.

gpflow/experimental/check_shapes/__init__.py
awav: I haven't used `check_shapes` yet, so it is hard to say what is best in that case. I will leave it to you to decide.

awav (approved):
LGTM

gpflow/experimental/check_shapes/__init__.py
```diff
@@ -53,6 +53,9 @@ def linear_model(
 * Use ``.<name>`` to refer to attributes of the object.
 * Use ``[<index>]`` to refer to elements of a sequence. This is particularly useful if your
   function returns a tuple of values.
+* Use ``[all]`` to select all elements in a collection.
```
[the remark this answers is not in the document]
jesnie: > I assume this works with input arguments and with the special "return" keyword?

Yes. Those are orthogonal.

PR #1843, opened 2022-04-01
Remove positive transform for anisotropic lengthscales in higher dims

### LINKED ISSUE #1639: Anisotropic lengthscales

frgsimpson:
Resolves #1639


jesnie (approved):
LGTM

PR #1845, opened 2022-04-01
Factor shape checking out into a separate class.

jesnie:
Split `check_shapes.py` into two:

1. `checker.py`, which only has the shape checking logic.
2. `decorator.py`, which only has a function decorator, that applies the logic from `checker.py`.

This allows the shape checking logic to be used without decorating a function.


gpflow/experimental/check_shapes/checker.py
```diff
@@ -0,0 +1,220 @@
...
+        return shaped
+
+    def check_shapes(
+        self,
+        checks: Iterable[
+            Union[Tuple[Any, TensorSpecLike], Tuple[Any, TensorSpecLike, ErrorContext]]
+        ],
+    ) -> None:
+        """
+        Raise an eror if any tensor has the wrong shape.
```
frgsimpson, line 117:
```suggestion
        Raise an error if any tensor has the wrong shape.
```

gpflow/experimental/check_shapes/checker.py
```diff
@@ -0,0 +1,220 @@
...
+
+        This remembers observed shapes and specifications, so that tensors can be checked for
+        companibility across multiple calls, and so that we can provide good error messages.
```
frgsimpson, line 120:
```suggestion
        compatibility across multiple calls, and so that we can provide good error messages.
```

gpflow/experimental/check_shapes/checker.py
```diff
@@ -0,0 +1,220 @@
...
+
+        :param checks: Checks to perform. The elements can either be `(shaped, tensor_spec)` or
+            `(shaped, tensor_spec, context)` tuples. Where: `shaped` is the tensor whose shape to
+            check; `tensor_spec` is the specification to check it against (see `Check
+            specification`_); and `context` is contains (optional) information about where `shaped`
```
frgsimpson, line 125:
```suggestion
            specification`_); and `context` contains (optional) information about where `shaped`
```

frgsimpson (approved):
LGTM!

PR #1847, opened 2022-04-04
Fix breakage with old versions of Python and new versions of Numpy.

jesnie:
Apparently the `np.ndarray[Any, Any]` syntax doesn't work for `python < 3.9`.

(I stumbled over this while trying to fix another bug that also relates to Python versions.)

### DIFF
--- a/gpflow/base.py
+++ b/gpflow/base.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import sys
 from enum import Enum
 from typing import TYPE_CHECKING, Any, List, Optional, Sequence, Tuple, Union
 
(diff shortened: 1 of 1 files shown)


st--:
What do you think of having a version matrix of checking at least a subset of the tests for a combination of all supported python / tensorflow versions?

jesnie:
> What do you think of having a version matrix of checking at least a subset of the tests for a combination of all supported python / tensorflow versions?

Dunno. Maybe. I'd be a bit worried about our budget on CircleCI, but maybe if we're sufficiently picky about which tests we run.

jesnie:
This doesn't really work, and we'll make sure this gets fixed in a more systematic way, see #1848.

PR #1849, opened 2022-04-05
Add support for checking shapes that allow broadcasting.

jesnie:


gpflow/experimental/check_shapes/__init__.py
```diff
@@ -85,6 +85,8 @@ def f(...):
     * ``*<name>`` or ``<name>...``, to bind *any* number of dimensions to a variable. Again,
       multiple uses of the same variable name must match the same dimension sizes.
     * ``*`` or ``...``, to allow *any* number of dimensions without constraints.
+    * Any of the above can be prefixed with the keyword ``broadcast``, to allow any value that
+      broadcasts to the specification.
```
vdutor, line 89: I think it would be useful to add an example of this feature to the docs.

gpflow/experimental/check_shapes/checker.py
```diff
@@ -49,17 +50,93 @@ def _as_parsed_tensor_spec(tensor_spec: TensorSpecLike, context: ErrorContext) -
...
+        if self.size is None:
+            self.size = actual
+
+        return self.size == actual
+
+
+@dataclass
+class _ObservedDims:
+    """
+    Storage of observed sizes of a var-rank / batch varible.
```
vdutor, line 91:
```suggestion
    Storage of observed sizes of a var-rank / batch variable.
```

gpflow/experimental/check_shapes/checker.py
```diff
@@ -49,17 +50,93 @@ def _as_parsed_tensor_spec(tensor_spec: TensorSpecLike, context: ErrorContext) -
...
 
+@dataclass
+class _ObservedDim:
+    """
+    Storage of observed size of a single dimension variable.
+    """
+
+    size: Optional[int] = None
+
+    def update(self, actual: Optional[int], broadcast: bool) -> bool:
```
vdutor, line 72: would `check_then_update` not be a better name if we use this method as part of an assert as [in this line](https://github.com/GPflow/GPflow/pull/1849/files#diff-b8b7cd3ef17cb6b9a4464a873ba9d26050f3cc6403361319b6da1b3b01ca67a8R271)?

gpflow/experimental/check_shapes/checker.py
```diff
@@ -49,17 +50,93 @@ def _as_parsed_tensor_spec(tensor_spec: TensorSpecLike, context: ErrorContext) -
...
+        """
+        Update this observation with new data.
```
vdutor, line 74: Only "updates" if the `self.size` is `None` - not sure I'm following what this method is supposed to do?

vdutor (approved):
LGTM

gpflow/experimental/check_shapes/__init__.py
jesnie: > I think it would be useful to add an example of this feature to the docs.

Fair enough. Done.

gpflow/experimental/check_shapes/checker.py
jesnie: > would `check_then_update` not be a better name if we use this method as part of an assert as [in this line](https://github.com/GPflow/GPflow/pull/1849/files#diff-b8b7cd3ef17cb6b9a4464a873ba9d26050f3cc6403361319b6da1b3b01ca67a8R271)?

Fair call. I'll call it `check_and_update`, though.

gpflow/experimental/check_shapes/checker.py
jesnie: > Only "updates" if the `self.size` is `None` - not sure I'm following what this method is supposed to do?

I've rewritten the documentation.

gpflow/experimental/check_shapes/checker.py
jesnie: Done.

PR #1853, opened 2022-04-06
Add support for checking the shapes of intermediate computations.

jesnie:


gpflow/experimental/check_shapes/__init__.py
```diff
@@ -139,11 +138,36 @@ def f(...):
...
+    @check_shapes(
+        "weights: [n_features, n_labels]",
+        "test_features: [n_rows, n_features]",
+        "test_labels: [n_rows, n_labels]",
+        "return: []",
+    )
+    def loss(
+        weights: np.ndarray, test_features: np.ndarray, test_labels: np.ndarray
+    ) -> np.ndarray:
+        prediction = check_shape(test_features @ weights, "[n_rows, n_labels]")
```
uri-granta, line 157: Unrelated suggestion: make `config._enabled ` default to `False` if `__debug__` is `False` (so we emulate the behaviour of `assert`)?

gpflow/experimental/check_shapes/__init__.py
```diff
@@ -343,14 +370,15 @@ def get_linear_model_shape(model: LinearModel) -> Shape:
     )
     def loss(
         model: LinearModel, test_features: np.ndarray, test_labels: np.ndarray
-    ) -> None:
+    ) -> np.ndarray:
```
uri-granta, line 373: (good catch)

gpflow/experimental/check_shapes/checker_context.py
```diff
@@ -0,0 +1,72 @@
...
+"""
+from contextlib import contextmanager
+from contextvars import ContextVar
+from typing import Iterator, Optional
+
+from .checker import S, ShapeChecker, TensorSpecLike
+from .config import get_enable_check_shapes
+from .error_contexts import ErrorContext, FunctionCallContext
+
+_shape_checker: ContextVar[ShapeChecker] = ContextVar("_shape_checker")
```
uri-granta, line 25: (TIL about ContextVar!)

gpflow/experimental/check_shapes/checker_context.py
```diff
@@ -0,0 +1,72 @@
...
+        yield
+    finally:
+        _shape_checker.reset(token)
+
+
+def get_shape_checker() -> ShapeChecker:
+    """
+    Get the :class:`ShapeChecker` from the wrapping :func:`check_shapes` decorator.
+
+    Behaviour is undefined if you call this from a function that is not directly wrapped in
```
uri-granta, line 44: Why "undefined" rather than "returns a `LookupError`"?

uri-granta (approved):
Lovely!

gpflow/experimental/check_shapes/__init__.py
jesnie: > Unrelated suggestion: make `config._enabled ` default to `False` if `__debug__` is `False` (so we emulate the behaviour of `assert`)?

Mmmm, dunno. Maybe. Definitely another PR, though.

gpflow/experimental/check_shapes/checker_context.py
jesnie: > Why "undefined" rather than "returns a `LookupError`"?

Well, what happens depends...
I can do something like:

```python
def g():
    get_shape_checker() 

@check_shapes()
def f():
     g()
```

With my current implementation `g` will see the `ShapeChecker` from `f` which is *weird*. I'd like to avoid that but I don't know a fast/easy way to do that. So, I declare the behaviour undefined, so I'm allowed to fix it in the future if I find a way to do it.

PR #1854, opened 2022-04-07
Simplify parsing of argument references.

jesnie:
Small simplification of the definition and parsing of argument references.


frgsimpson (approved):
LGTM!

PR #1855, opened 2022-04-11
Add conditional shapes.

jesnie:
Add support for conditional shapes.

Including:

1. Extend the specification language with an `if` statement and minimal support for boolean expressions.
2. Type of `ParallelContext.children` has changed from `List` to `Tuple`, which allows `ParallelContext` to be hashed, which is necessary to nest a `ParallelContext` within another `ParallelContext`.
3. Updating the `check_shapes` decorator to respect the new `if` statements.
4. Added a couple of new `ErrorContext`s to describe errors within conditionals.
5. Updated test `utils.py` with some functions, `barg`, `bor`, `band`, and `bnot` to make it easy to create boolean expressions to tests. Made the `note` argument optional on a couple of existing functions.

### DIFF
--- a/gpflow/experimental/check_shapes/bool_specs.py
+++ b/gpflow/experimental/check_shapes/bool_specs.py
@@ -1,0 +1,102 @@
+# Copyright 2022 The GPflow Contributors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+Code for specifying and evaluating boolean expressions.
+"""
+from abc import ABC, abstractmethod
+from dataclasses import dataclass
+from typing import Any, Mapping, Tuple
+
+from .argument_ref import ArgumentRef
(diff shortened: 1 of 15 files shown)


gpflow/experimental/check_shapes/parser.py
```diff
@@ -204,13 +239,44 @@ def _argument_spec_to_sphinx(self, argument_spec: ParsedArgumentSpec) -> str:
...
+
+        out.append(".")
+
         if tensor_spec.note is not None:
             note_spec = tensor_spec.note
             out.append(" ")
             out.append(note_spec.note)
         return "".join(out)
 
+    def _bool_spec_to_sphinx(self, bool_spec: ParsedBoolSpec, paren_wrap: bool) -> str:
```
uri-granta, line 256: (you could do the same trick here if you'd like to avoid unnecessary brackets in "a or b or c" or "a and b or c" etc)

uri-granta (approved):
Looks great!

gpflow/experimental/check_shapes/bool_specs.py
```diff
@@ -0,0 +1,102 @@
...
+
+class ParsedBoolSpec(ABC):
+    """ A boolean expression. """
+
+    @abstractmethod
+    def get(self, arg_map: Mapping[str, Any], context: ErrorContext) -> Tuple[bool, ErrorContext]:
+        """ Evaluate this boolean value. """
+
+
+def _paren_repr(spec: ParsedBoolSpec) -> str:
```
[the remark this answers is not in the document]
jesnie: I'm not sure I want to do that. If I understand correctly you want to do:

```
a or b and c
```

instead of

```
a or (b and c)
```

? They might technically be the same, but I find the second option to much more readable, even if it's a bit longer.

gpflow/experimental/check_shapes/parser.py
jesnie: Again, I think I like the extra brackets. I think they add readability.

PR #1856, opened 2022-04-11
Fix doc warnings.

jesnie:
Fix doc warnings. Fixes #1254

Fix all warnings generated by Sphinx. This doesn't mean our documentation is far perfect - far from it.
But at least some classes of problems have been fixed.

For maths it has been a bit of a mix whether I've formatted it as a code example:

```
Example::

    math goes here
```

or as LaTeX:

```
Example:

... math::

    math goes here
```

obviously LaTeX is perttier, but it's also much more work. I figure we'll need to go back and look over all of it again anyway.

PR #1857, opened 2022-04-12
Add test of documentation warnings.

jesnie:
Add a unit test that our documentation build doesn't produce any warnings.

(This doesn't build the notebooks, we have other tests for that.)

Basically this PR consists of:

1. Modify `doc/build_docs.py` with a flag that makes it fail if there are any warnings.
2. Modify `doc/build_docs.py` so that if `--max_notebooks` it generates an empty file, instead of no file, because the missing files would cause errors/warnings.
3. Add an integration test that runs `doc/build_docs.py` with `--max_notebooks=0` and `--fail_on_warning`.
4. Set up CircleCI to run the new test.

Currently this test fails, because our documentation is broken, but it should be fixed by #1856

I've set this up as a separate "fast" test on CircelCI - it seems like it takes about 3 minutes to run: 2 minutes to install GPflow and 1 minute to actually run the test. Alternatively we could run this as a "slow" test, or fold it into the existing "unit" or "notebook" tests. Folding it into an existing test would save the cost of installing GPflow.

### DIFF
--- a/tests/integration/test_docs.py
+++ b/tests/integration/test_docs.py
@@ -1,0 +1,45 @@
+# Copyright 2022 the GPflow authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import subprocess
+from pathlib import Path
+
+import pytest
+
+BUILD_DOCS_PATH = Path(__file__).parent.parent.parent / "doc" / "build_docs.py"
+
+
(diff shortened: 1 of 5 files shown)


tests/integration/test_docs.py
```diff
@@ -0,0 +1,36 @@
...
+        [
+            "python",
+            str(BUILD_DOCS_PATH),
+            "develop",
+            str(build_path),
+            "--max-notebooks",
+            "0",
+            "--fail-on-warning",
+        ]
+    ).check_returncode()
```
alansaul, line 36: Might be nice to make a slightly nicer error message here to help a user understand what's gone wrong, currently it just fails on a `subprocess.CalledProcessError`, which might be a bit confusing. Judging by how broken are docs have been in the past, I guess it might happen a fair amount.

alansaul (approved):
LGTM

tests/integration/test_docs.py
jesnie: > Might be nice to make a slightly nicer error message here to help a user understand what's gone wrong, currently it just fails on a `subprocess.CalledProcessError`, which might be a bit confusing. Judging by how broken are docs have been in the past, I guess it might happen a fair amount.

Sure, (though the point of this PR is to prevent us from getting into a badly broken state to begin with...)

PR #1859, opened 2022-04-21
Fix problems with different versions of Python and Numpy.

### LINKED ISSUE #1848: Make sure all reasonable and supported combinations of Python and numpy (and tensorflow?) works.
Right now Python3.9 + numpy 1.20 refuses to load.

The problem comes from:

1. `numpy < 1.20.0` has no types.
2. For `1.20.0 <= numpy < 1.22.0` arrays has type `np.ndarray`.
3. For `1.22.0 <= numpy` arrays has type `np.ndarray[Any, Any]`.

We need a setup that works both for `mypy` and when actually running the code.

jesnie:
This has been tested with:

* Python 3.7 + Numpy 1.19
* Python 3.7 + Numpy 1.20
* Python 3.7 + Numpy 1.21
* Python 3.8 + Numpy 1.19
* Python 3.8 + Numpy 1.20
* Python 3.8 + Numpy 1.21
* Python 3.8 + Numpy 1.22
* Python 3.9 + Numpy 1.20
* Python 3.9 + Numpy 1.21
* Python 3.9 + Numpy 1.22
* Python 3.10 + Numpy 1.21
* Python 3.10 + Numpy 1.22

Fixes #1848

### DIFF
--- a/gpflow/base.py
+++ b/gpflow/base.py
@@ -18,25 +18,27 @@ from typing import TYPE_CHECKING, Any, List, Optional, Sequence, Tuple, Union
 import numpy as np
 import tensorflow as tf
 import tensorflow_probability as tfp
-from packaging.version import Version
 from typing_extensions import Final
 
 from .config import default_float, default_summary_fmt
+from .type_flags import GENERIC_NP_ARRAYS, NP_TYPE_CHECKING
 
 if TYPE_CHECKING:  # pragma: no cover
     from IPython.lib import pretty
 
 DType = Union[np.dtype, tf.DType]
 
-# It would be nice to use something more interesting than `Any` here, but it looks like the
-# infrastructure in the rest of the ecosystem isn't really set up for this yet. Maybe when we
-# get Python 3.11?
-if TYPE_CHECKING:  # pragma: no cover
-    AnyNDArray = Union["np.ndarray[Any, Any]"]
-elif Version(np.__version__) < Version("1.20.0"):
-    AnyNDArray = np.ndarray
+
+if TYPE_CHECKING and (not NP_TYPE_CHECKING):  # pragma: no cover
+    AnyNDArray = Any
 else:
-    AnyNDArray = np.ndarray[Any, Any]
+    if GENERIC_NP_ARRAYS:
+        # It would be nice to use something more interesting than `Any` here, but it looks like
+        # the infrastructure in the rest of the ecosystem isn't really set up for this
+        # yet. Maybe when we get Python 3.11?
+        AnyNDArray = np.ndarray[Any, Any]  # type: ignore
+    else:
+        AnyNDArray = Union[np.ndarray]  # type: ignore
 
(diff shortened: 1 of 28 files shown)


gpflow/base.py
```diff
@@ -28,14 +29,20 @@
...
-# get Python 3.11?
 if TYPE_CHECKING:  # pragma: no cover
-    AnyNDArray = Union["np.ndarray[Any, Any]"]
-elif Version(np.__version__) < Version("1.20.0"):
+    # It would be nice to use something more interesting than `Any` here, but valid values depend on
+    # the version of `numpy`, and `mypy` doesn't support branching on that. Maybe, when we bump
+    # minimal versions of Python and TensorFlow to > 3.7 and 2.4, we can set a minimum version of
+    # numpy to >= 1.21, and we can get rid of this.
+    AnyNDArray = Any  # Valid for all versions of numpy.
+    # AnyNDArray = Union["np.ndarray"]  # Valid for numpy < 1.21.0
```
uri-granta, line 38: I think the following workaround works:

```python
if TYPE_CHECKING:
    GENERIC_NDARRAYS = True
    if GENERIC_NDARRAYS:
        AnyNDArray = Union["np.ndarray[Any, Any]"]
    else:
        AnyNDArray = Union["np.ndarray"] # type: ignore
```

then for numpy<1.21.0 you call mypy with `--always-false GENERIC_NDARRAYS` to get the second definition.

jesnie:
@uri-granta I've had to write a bit of infrastructure to set `--always-false` flags for `mypy`, as you suggested. Do you want to take another look?

gpflow/type_flags.py
```diff
@@ -0,0 +1,39 @@
...
+
+GENERIC_NP_ARRAYS = (sys.version_info >= (3, 9)) and (NP_VERSION >= Version("1.22.0"))
+"""
+Whether to use generic numpy arrays. This is not applied at all, if type checking and not
+`NP_TYPE_CHECKING`.
+"""
+
+MYPY_FLAGS = {
+    "NP_TYPE_CHECKING": True,
+    "GENERIC_NP_ARRAYS": NP_VERSION >= Version("1.21.0"),
```
uri-granta, line 38: why does MYPY_FLAGS's GENERIC_NP_ARRAYS say 1.21+ while the runtime GENERIC_NP_ARRAYS (which I'm not sure is ever used?) says 1.22+?

uri-granta (approved):
Looks good. Just one question.

gpflow/type_flags.py
jesnie: > why does MYPY_FLAGS's GENERIC_NP_ARRAYS say 1.21+ while the runtime GENERIC_NP_ARRAYS (which I'm not sure is ever used?) says 1.22+?

Because that's the value that works. :-) I can't be bothered to figure out why. I assume it has something to do with types in a `.pyi` file being different from those in a `.py` file.

The `GENERIC_NP_ARRAYS` default value is used at runtime, if `TYPE_CHECKING` is `False`.

PR #1865, opened 2022-04-26
Iterate on check_shapes documentation.

jesnie:
Improve `check_shapes` documentation - mostly focussed on the `check_shapes` module landing page.

### DIFF
--- a/gpflow/experimental/check_shapes/__init__.py
+++ b/gpflow/experimental/check_shapes/__init__.py
@@ -14,31 +14,25 @@
 """
 A library for annotating and checking the shapes of tensors.
 
-This library is compatible with both TensorFlow and NumPy.
+The main entry point is :func:`check_shapes`.
 
-The main entry point is :func:`check_shapes.check_shapes`.
+Example:
 
-For example::
-
-    @tf.function
-    @check_shapes(
-        "features: [batch_shape..., n_features]",
-        "weights: [n_features]",
-        "return: [batch_shape...]",
-    )
-    def linear_model(
-        features: tf.Tensor, weights: tf.Tensor
-    ) -> tf.Tensor:
-        ...
+.. literalinclude:: /examples/test_check_shapes_examples.py
+   :start-after: [basic]
+   :end-before: [basic]
+   :dedent:
 
 
 Check specification
 +++++++++++++++++++
 
 The shapes to check are specified by the arguments to :func:`check_shapes`. Each argument is a
-string of the format::
+string of the format:
 
-    <argument specifier>: <shape specifier> [if <condition>] [# <note>]
+.. code-block:: bnf
+
+   <argument specifier> ":" <shape specifier> ["if" <condition>] ["#" <note>]
 
 
 Argument specification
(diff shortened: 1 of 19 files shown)


gpflow/experimental/check_shapes/__init__.py
```diff
@@ -48,247 +42,282 @@ def linear_model(
...
-        "return: ...",
-        "return[0]: ...",
-        "data_list[all]: ...",
-        "data_dict.keys(): ...",
-        "data_dict.values(): ...",
-        "something[all].foo[0].bar.values(): ...",
-    )
-    def f(...):
-        ...
+We do not support looking up a specific key or value in a  ``dict``.
```
vdutor, line 84: Can you make this a `.. note::`-block so that it pops up in the docs?

tests/examples/test_check_shapes_examples.py
```diff
@@ -0,0 +1,739 @@
...
+        "a: [broadcast batch...] if check_a",
+        "b: [broadcast batch...] if check_b",
+        "return: [batch...]",
+    )
+    def add(a: AnyNDArray, b: AnyNDArray, check_a: bool = True, check_b: bool = True) -> AnyNDArray:
+        return a + b
+
+    # [bool_spec_argument_ref]
+
+    add(np.ones((3, 1)), np.ones((1, 4)), check_b=False)
```
vdutor, line 370: I would include the call in the code example to show that the user can set these arguments.
```suggestion

    add(np.ones((3, 1)), np.ones((1, 4)), check_b=False)
    # [bool_spec_argument_ref]
```

tests/examples/test_check_shapes_examples.py
```diff
@@ -0,0 +1,739 @@
...
+    def add(
+        a: AnyNDArray,
+        b: AnyNDArray,
+        check_all: bool = False,
+        check_a: bool = True,
+        check_b: bool = True,
+    ) -> AnyNDArray:
+        return a + b
+
+    # [bool_spec_or]
```
vdutor, line 392: Similar to the above, I would include the call in this example.

tests/examples/test_check_shapes_examples.py
```diff
@@ -0,0 +1,739 @@
...
+    def add(
+        a: AnyNDArray,
+        b: AnyNDArray,
+        enable_checks: bool = True,
+        check_a: bool = True,
+        check_b: bool = True,
+    ) -> AnyNDArray:
+        return a + b
+
+    # [bool_spec_and]
```
vdutor, line 416: Similar to the above, I would include the call in this example.

tests/examples/test_check_shapes_examples.py
```diff
@@ -0,0 +1,739 @@
...
+
+    add(np.ones((3, 1)), np.ones((1, 4)), check_b=False)
+
+
+def test_example__bool_spec_not() -> None:
+    # pylint: disable=unused-argument
+
+    # [bool_spec_not]
+
+    @check_shapes(
```
vdutor, line 426: From reading the docs it wasn't clear if this is allowed?
```python
@check_shapes(
    "a: [n1] if flag",
    "a: [n2] if not flag",
 )
 def foo(a):
    pass
```

gpflow/experimental/check_shapes/__init__.py
```diff
@@ -48,247 +42,282 @@ def linear_model(
...
+.. literalinclude:: /examples/test_check_shapes_examples.py
+   :start-after: [reuse__get_check_shapes]
+   :end-before: [reuse__get_check_shapes]
+   :dedent:
 
 
 Checking shapes without a decorator
 +++++++++++++++++++++++++++++++++++
 
 While the :func:`check_shapes` decorator is the recommend way to use this library, it is possible to
```
vdutor, line 313: why are we exposing this in the public API? Maybe best to keep hidden from the user in case you want to make changes to this later?

vdutor (approved):
Great work!

gpflow/experimental/check_shapes/__init__.py
jesnie: > why are we exposing this in the public API? Maybe best to keep hidden from the user in case you want to make changes to this later?

Hmmm, maybe. Not sure. I like giving people a fallback in case there is some reason they cannot use the decorator. As you've seen yourself, using decorators can sometimes give weird side-effects, so I don't like forcing people to use them.

tests/examples/test_check_shapes_examples.py
jesnie: > I would include the call in the code example to show that the user can set these arguments.

Sure.

tests/examples/test_check_shapes_examples.py
jesnie: > Similar to the above, I would include the call in this example.

Sure.

tests/examples/test_check_shapes_examples.py
jesnie: > Similar to the above, I would include the call in this example.

Sure.

gpflow/experimental/check_shapes/__init__.py
jesnie: > Can you make this a `.. note::`-block so that it pops up in the docs?

Sure.

PR #1870, opened 2022-04-28
Fix bug in release process.

jesnie:
The release of `2.5.0` stalled without triggering the `fast-tests` job.

I think this should fix it, and start the release of `2.5.1`.


alansaul (approved):
LGTM when you've fixed the coverage issue

PR #1879, opened 2022-05-05
Fix documentation import logic.

jesnie:
Fix logic of where functions and classes are placed in the documentation.
The previous logic was that objects are placed at the place closest to the root, where they are imported. This leads to weird behaviour, such as `gpflow.utilities.model_utils.add_noise_cov` being documented as `gpflow.posteriors.add_noise_cov`, just because it happens to be imported by `gpflow.posteriors`. The new logic only considers places where the object either is declared or added to `__all__`, which places `gpflow.utilities.model_utils.add_noise_cov` under `gpflow.utilities.add_noise_cov`.

### DIFF
--- a/doc/generate_module_rst.py
+++ b/doc/generate_module_rst.py
@@ -123,19 +123,9 @@ class DocumentableModule:
     @staticmethod
     def collect(
         root: ModuleType,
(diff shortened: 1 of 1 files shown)


frgsimpson (approved):
LGTM!

PR #1880, opened 2022-05-05
Add check_shapes to inducing_variables.

jesnie:
Add shape checking to the `inducing_variables` sub-package:

1. I don't know what's going on with that `kufs.py`. See: https://secondmind-ai.slack.com/archives/CUFMWJN4T/p1651748445091629
2. I use `get_shape.register` to define a shape of inducing variables.
3. This triggered some `mypy` warnings in `posteriors.py`, which I fix with a `cast`. `X_data` has type `Union[tf.Tensor, InducingVariables]`, though in practice they should never be `InducingVariables` at that point in the code.
4. Extended tests of `get_shapes`, as I though of some special cases I hadn't covered.
5. Extended tests of `InducingVariables`, particularly around shapes.

### DIFF
--- a/gpflow/inducing_variables/inducing_variables.py
+++ b/gpflow/inducing_variables/inducing_variables.py
@@ -20,10 +20,11 @@ import tensorflow_probability as tfp
 from deprecated import deprecated
 
 from ..base import Module, Parameter, TensorData, TensorType
+from ..experimental.check_shapes import ErrorContext, Shape, check_shapes, get_shape
 from ..utilities import positive
 
 
-class InducingVariables(Module):
+class InducingVariables(Module, abc.ABC):
     """
     Abstract base class for inducing variables.
     """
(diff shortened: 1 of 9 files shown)


gpflow/inducing_variables/inducing_variables.py
```diff
@@ -57,7 +80,15 @@ def __init__(self, Z: TensorData, name: Optional[str] = None):
 
     @property
     def num_inducing(self) -> Optional[tf.Tensor]:
-        return tf.shape(self.Z)[0]
+        cs = ShapeChecker().check_shape  # mypy doesn't support decorating properties.
```
uri-granta, line 83: Is this better than using a decorator and adding a type ignore?

uri-granta (approved):
Lovely! Just one small question.

gpflow/inducing_variables/inducing_variables.py
jesnie: > Is this better than using a decorator and adding a type ignore?

Huh, `# type: ignore` hadn't even occurred to me. I'll do that. Thanks!

PR #1882, opened 2022-05-09
Stop using uninitialised variables.

### LINKED ISSUE #1881: Variables without an initial value prevents TensorFlow compilation.
# Bug / performance issue

Our code has some variables without an initial value:

```
foo: str
foo = ...
```

Apparently TensorFlow doesn't know how to compile these, and it causes performance problems. See:
https://github.com/tensorflow/tensorflow/issues/56024

## To reproduce

```
pytest tests/gpflow/conditionals/test_multioutput.py
```

**Stack trace, or error message**
```
WARNING:tensorflow:AutoGraph could not transform <bound method Dispatcher.dispatch of <dispatched prior_kl>> and will run it as-is.
Please report this to the TensorFlow team. When filing the bug, set the verbosity to 10 (on Linux, `export AUTOGRAPH_VERBOSITY=10`) and attach the full output.
Cause: invalid value for "node": expected "ast.AST", got "<class 'NoneType'>"; to visit lists of nodes, use "visit_block" instead
To silence this warning, decorate the function with @tf.autograph.experimental.do_not_convert
```

## Expected behavior

Code is compiled and runs fast.

## System information

* GPflow version: 2.5.1
* GPflow installed from: "pip install gpflow"
* TensorFlow version: 2.8.0
* Python version: 3.10.4
* Operating system: Ubuntu 18.04.6 LTS

jesnie:
Fixes #1881 

Apparently TensorFlow cannot compile code that has variables without a value. See #1881 .
This PR:

- Adds a unit test that all assignments have a value.
- Removes any cases of assignments without a value that we had.
- (And drive-by fixes a couple of unused imports...)

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -55,15 +55,13 @@ This release contains contributions from:
 
 ## Bug Fixes and Other Changes
 
-* <SIMILAR TO ABOVE SECTION, BUT FOR OTHER IMPORTANT CHANGES / BUG FIXES>
-* <IF A CHANGE CLOSES A GITHUB ISSUE, IT SHOULD BE DOCUMENTED HERE>
-* <NOTES SHOULD BE GROUPED PER AREA>
+* Fixed some bugs that prevented TensorFlow compilation and had negative performance impact. (#1882)
 
 ## Thanks to our Contributors
 
 This release contains contributions from:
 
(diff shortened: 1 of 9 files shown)


awav (approved):
LGTM

PR #1886, opened 2022-05-10
Apply check_shapes to gpflow.utilities.

jesnie:


gpflow/utilities/ops.py
```diff
@@ -82,24 +95,32 @@ def leading_transpose(tensor: tf.Tensor, perm: List[Any], leading_dim: int = 0)
     return tf.transpose(tensor, perm)
 
 
+@check_shapes(
+    "a: [a_shape...]",
+    "b: [b_shape...]",
+    "return: [a_shape_b_shape...]",
```
uri-granta, line 101: This is just a naming convention right? (i.e. there's no check that a_shape_b_shape = a_shape + b_shape)

uri-granta (approved):
Nice!

gpflow/utilities/ops.py
jesnie: > This is just a naming convention right? (i.e. there's no check that a_shape_b_shape = a_shape + b_shape)

Yes. Maybe one day we'll support this, but not today.

PR #1893, opened 2022-05-18
Make sure all code compiles correctly.

jesnie:
Sometimes TensorFlow will (almost) silently not compile your code.
This PR adds a test that none of our other tests prints: "AutoGraph could not transform", and fixes the couple of cases there were discovered.

(Surprisingly TensorFlow *really* don't like type annotations...)

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,0 +1,26 @@
+# Copyright 2022 The GPflow Contributors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


tests/conftest.py
```diff
@@ -0,0 +1,26 @@
...
+from _pytest.logging import LogCaptureFixture
+
+
+@pytest.fixture(autouse=True)
+def test_auto_graph_compile(caplog: LogCaptureFixture) -> Iterable[None]:
+    yield
+
+    for when in ["setup", "call", "teardown"]:
+        for record in caplog.get_records(when):
+            assert not record.msg.startswith("AutoGraph could not transform"), record.getMessage()
```
awav, line 26: TF can change the message, but I don't know how we can protect ourselves against that.

awav (approved):
LGTM

tests/conftest.py
jesnie: > TF can change the message, but I don't know how we can protect ourselves against that.

We could write a test of our test, but that would require us to write some code that TensorFlow reliably can't compile - and the compile can compile more and more stuff - so that isn't reliable either.

PR #1897, opened 2022-05-26
Fix bug in ShapeChecker related to multiple varrank dimensions.

jesnie:


gpflow/experimental/check_shapes/checker.py
```diff
@@ -445,8 +445,9 @@ def _match_dims(self, shape_check: _ShapeCheck) -> Iterable[Tuple[ParsedDimensio
 
         # Take as much as we can from the left side:
 
+        allocated_i = 0
         while not shape_check.finished:
-            expected_rank = allocated_sizes[shape_check.expected_begin]
```
jesnie, line 449: Basically `allocated_sizes` will correspond to `expected[expected_begin:expected_end]` (for the initial values of `expected_begin` and `expected_end`. Often `expected_begin` starts at 0, and if that is the case this works. However, if `expected_begin` starts at a value larger than 0, this code crashes.

frgsimpson (approved):
LGTM!

PR #1898, opened 2022-05-26
Make sure all ErrorContext-s are hashable.

jesnie:
Make sure all `ErrorContext`s are hashable.

A couple of places we hash the error context, but actually they were not always hashable, leading to crashes. This PR ensure that all the `ErrorContext`s are hashable.


uri-granta (approved):
LGTM!

PR #1900, opened 2022-05-31
Improve check_shapes's handling of broadcasting.

### LINKED ISSUE #1888: Improve check_shapes handling of leading bradcast.
A pattern I see often in GPflow is something like:

```python
@check_shapes(
    "X: [N, D]",
    "lengthscales: [broadcast D]",
)
def foo(X, lengthscales):
    return X / lengthscales
```

`lengthscales` should have length `D` or `1` or be a scalar. We don't currently support this. You can kind of hack it with:

```python
@check_shapes(
    "X: [N, D]",
    "X: [N, D_tuple...]",
    "lengthscales: [broadcast D_tuple...]",
)
def foo(X, lengthscales):
    return X / lengthscales
```

but that is inelegant. Figure out a better solution.

jesnie:
Fixes #1888

Improve `check_shapes`'s handling of broadcasting:

1. Allows `broadcast` single dimensions to have length 0 or 1, instead of always 1.
2. Allows variable-rank dimensions to be broadcast, even if they're not the first dimension.


gpflow/experimental/check_shapes/checker.py
```diff
@@ -388,42 +389,60 @@ def _match_dims(self, shape_check: _ShapeCheck) -> Iterable[Tuple[ParsedDimensio
         allocated_sizes: List[Optional[int]] = []
 
         # Count of `None`s in `allocated_sizes`:
-        n_unallocated = 0
+        n_allocated_none = 0
```
jesnie, line 392: Renamed for readability.

uri-granta (approved):
Looks good!

Issue #1902, opened 2022-06-06
check_shapes should generate documentation if it doesn't exist


jesnie:
Currently `check_shapes` only updates existing documentation. It will not document parameters that does not have any other documentation, or functions that don't have a docstring at all. However, for some code there really isn't anything interesting to say other than the shapes of the input and output, so `check_shapes` should generate this documentation.

PR #1904, opened 2022-06-06
Make check_shapes add info-fields for arguments that don't have any.

jesnie:
Fixes: #1902

Make check_shapes add info-fields for arguments that don't have any.


frgsimpson (approved):
LGTM!

Issue #1241, opened 2020-02-03
Add shape asserts across code base


st--:
We have now built `gpflow.experimental.check_shapes`. It should be applied consistently to our code:

* it helps in uncovering bugs in our code and gives us more confidence our code is behaving as intended,
* it makes clear _in code_ what shapes our code expects / returns (without relying on shape comments that may easily get out of date with the code itself),
* it gives more helpful error messages to users who supply wrongly-shaped tensors as inputs.

The following modules have relevant tensorflow code that would probably benefit from a bunch of shape assertions:

- [x] gpflow.conditionals + tests.gpflow.conditionals
- [x] gpflow.config + tests.gpflow.config
- [x] gpflow.covariances + tests.gpflow.covariances
- [x] gpflow.expectations + tests.gpflow.expectations
- [x] gpflow.experimental + tests.gpflow.experimental
- [x] gpflow.inducing_variables
- [x] gpflow.kernels + tests.gpflow.kernels
- [x] gpflow.likelihoods + tests.gpflow.likelihoods
- [x] gpflow.models + tests.gpflow.models
- [x] gpflow.monitor
- [x] gpflow.optimizers + tests.gpflow.optimizers
- [x] gpflow.quadrature + tests.gpflow.quadrature
- [x] gpflow.utilities + tests.gpflow.utilities
- [x] gpflow.* + tests.gpflow.posteriors
- [x] tests.integration

If you would be willing to help with some of these, please go ahead and open PRs! :)

awav:
Good first issue? ;)

st--:
Well, I'm assuming people using GPflow and interested in contributing may have some familiarity with arrays and shapes - the shape _comments_ tend to be there already, and there's plenty of examples of how to do it in #1219, so it's just extrapolation from there...

joelberkeley:
i don't think we should be including asserts on return values. that kind of thing should be checked with tests not runtime asserts. I'm also dubious about expecting users to ever have to look at the code to find out what a function does

joelberkeley-pio:
should we be including asserts on return values? I'd have thought that kind of thing should be checked with tests not runtime asserts.

I'm also wary of expecting users to ever have to look at a function's body to find out what it does or requires. For example, I would expect the docs to include the type signature and docstring, but not the source code, and I would expect that to be sufficient to know how to use functionality

st--:
[...]

PR #1907, opened 2022-06-07
Add shape checks to covariances.

jesnie:
Add shape checks to covariances.

This involves significant changes to `tests/gpflow/covariances/test_multioutput.py`, because it was actually using inconsistent shapes.

Partial: #1241


jesnie:
Blocked by everything.

PR #1912, opened 2022-06-10
Micro optimisations of check_shapes.

jesnie:
A couple of optimisations to `check_shapes` after looking at it through a profiler:

1. It turns out we were spending a lot of time computing hashes. Instead of having many dicts, each containing some information for each variable, we create a `_VariableState` structure, and have a single dict. This allows us to reduce many dict look-ups to a single call. (And is arguably more readable anyway...) In the process of this `_check_variable_use` got merged into `_parse_checks`.
2. Have a config option for whether to pre-compute file and line numbers for error-messages, defaulting to `False`, since this is particularly slow.

Together these seem to reduce the overhead of `check_shapes` by about 40%, which isn't as much as I'd like, but is better than nothing.

### DIFF
--- a/gpflow/experimental/check_shapes/checker.py
+++ b/gpflow/experimental/check_shapes/checker.py
@@ -14,8 +14,8 @@
 """
 Class responsible for remembering and checking shapes.
 """
-from dataclasses import dataclass
-from typing import Any, Dict, Iterable, List, Optional, Set, Tuple, TypeVar, Union
+from dataclasses import dataclass, field
+from typing import Any, DefaultDict, Dict, Iterable, List, Optional, Set, Tuple, TypeVar, Union
 
 from ..utils import experimental
 from .base_types import Dimension, Shape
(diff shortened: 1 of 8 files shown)


gpflow/experimental/check_shapes/checker.py
```diff
@@ -185,6 +185,37 @@ def finished(self) -> bool:
...
+
+    Set this `None` if this variable is rank-1.
+    """
+
+    waiting_for_varrank: Set[_ShapeCheck]
+    """ Checks that are waiting for the rank of this variable to be determined. """
+
+    @staticmethod
+    def new() -> "_VariableState":
+        return _VariableState([], None, None, set())
```
uri-granta, line 216: Why not use field default values instead (using `field(default_factory=list)` etc for the mutable fields)? And then you can just write `DefaultDict(_VariableState)`.

uri-granta (approved):
Looks good!

gpflow/experimental/check_shapes/checker.py
jesnie: > Why not use field default values instead (using `field(default_factory=list)` etc for the mutable fields)? And then you can just write `DefaultDict(_VariableState)`.

That probably is more elegant. I'll do that.

PR #1917, opened 2022-06-27
Stop using singledispatch because TensorFlow cannot compile it.

jesnie:
Basically the interaction of TensorFlow and `@singledispatch` is super flaky, so I'm removing `@singledispatch` from `check_shapes`.

### DIFF
--- a/tests/gpflow/experimental/check_shapes/test_integration.py
+++ b/tests/gpflow/experimental/check_shapes/test_integration.py
@@ -16,7 +16,7 @@
 # pylint: disable=no-member  # PyLint struggles with TensorFlow.
 
 from time import perf_counter
-from typing import Any, Callable
+from typing import Any, Callable, Optional
 
 import numpy as np
 import pytest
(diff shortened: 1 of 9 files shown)


tests/gpflow/experimental/check_shapes/test_integration.py
```diff
@@ -115,58 +115,108 @@ def test_layer(join: SuperLayer) -> None:
...
-            tf.function(input_signature=[]),
-        ),
-        (
-            # See: https://github.com/tensorflow/tensorflow/issues/56414
-            tf.function(experimental_relax_shapes=Version(tf.__version__) < Version("2.9.0")),
-            tf.function(experimental_relax_shapes=True),
-        ),
+        (_IdWrapper, _IdWrapper),
+        (_TfFunction, _TfFunction),
+        (_ShapedTfFunctionErr, _ShapedTfFunctionLoss),
```
uri-granta, line 151: If you're always testing identical error and loss wrappers, would it be better to have just one parameter?

uri-granta (approved):
Looks good! (and an impressive test)

tests/gpflow/experimental/check_shapes/test_integration.py
jesnie: > If you're always testing identical error and loss wrappers, would it be better to have just one parameter?

But they're not always the same. `_ShapedTfFunctionErr` is not  `_ShapedTfFunctionLoss`.

PR #1918, opened 2022-06-28
Consistently raise errors where full_output_cov isn't supported.

jesnie:
We have some methods that do not support `full_cov=True` and many that do not support `full_output_cov=True`. Right now some of those methods raise various error messages, but many also silently ignore the invalid argument.

With this PR we systematically raise the same error message in these cases.

### DIFF
--- a/gpflow/models/cglb.py
+++ b/gpflow/models/cglb.py
@@ -19,7 +19,7 @@ import tensorflow as tf
 from ..base import InputData, MeanAndVariance, Parameter, RegressionData, TensorType
 from ..config import default_float, default_int
 from ..covariances import Kuf
-from ..utilities import add_noise_cov, to_default_float
+from ..utilities import add_noise_cov, assert_params_false, to_default_float
 from .sgpr import SGPR
 
 
@@ -181,6 +181,8 @@ class CGLB(SGPR):
         :param cg_tolerance: float or None: If None, the cached value of
             :math:`v` is used. If float, conjugate gradient is run until :math:`rᵀQ⁻¹r < ϵ`.
         """
+        assert_params_false(self.predict_f, full_output_cov=full_output_cov)
+
         x, y = self.data
         err = y - self.mean_function(x)
         kxx = self.kernel(x, x)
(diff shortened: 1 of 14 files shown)


gpflow/utilities/model_utils.py
```diff
@@ -1,9 +1,29 @@
+from typing import Any, Callable
+
 import tensorflow as tf
 
 from ..base import TensorType
 from ..experimental.check_shapes import check_shapes
 
 
+def assert_params_false(
```
uri-granta, line 9: Nice. (I'm automatically trying to generalise this, but there's definitely no need at the moment.)

gpflow/models/cglb.py
```diff
@@ -181,6 +181,8 @@ def predict_f(
         :param cg_tolerance: float or None: If None, the cached value of
             :math:`v` is used. If float, conjugate gradient is run until :math:`rᵀQ⁻¹r < ϵ`.
         """
+        assert_params_false(self.predict_f, full_output_cov=full_output_cov)
```
uri-granta, line 184: A pity we have to pass in the method name each time as I'm sure we'll end up with copy and paste errors.

You could probably replace this with a decorator of some sort, though I expect it would confuse mypy.

```python
@assert_param(full_output_cov=False)
def predict_f(...):
```

uri-granta (approved):
:+1:

gpflow/utilities/model_utils.py
jesnie: > Nice. (I'm automatically trying to generalise this, but there's definitely no need at the moment.)

Yeah, I did spend some trying to think of more clever ways to do this, but I'm worried about making it too complicated and creating problems with TensorFlow compilation.

gpflow/models/cglb.py
jesnie: > A pity we have to pass in the method name each time as I'm sure we'll end up with copy and paste errors.
> 
> You could probably replace this with a decorator of some sort, though I expect it would confuse mypy.
> 
> ```python
> @assert_param(full_output_cov=False)
> def predict_f(...):
> ```

Again, I'm more worried about confusing TensorFlow.

PR #1919, opened 2022-06-29
Add an X parameter to the Likelihood.

jesnie:
Add an `X` parameter to the `Likelihood` class methods.

This will make it easy/possibly to write heteroskedastic likelihoods.

I plan to do more with heteroskedastic likelihoods in downstream PRs, but I want to get the "controversial" part of it out of the way first.

Basically this is the controversial part, because this is a breaking change.
To guesstimate the scale of the breakage, consider our notebooks. We have 35 notebooks doing all kinds of advanced stuff. A single one of those breaks.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -39,6 +39,8 @@ This release contains contributions from:
 
 ## Breaking Changes
 
+* All likelihood methods now take an extra `X` argument. If you have written custom likelihoods or
+  you have custom code calling likelihoods directly you will need to add this extra argument.
 * Change to `InducingVariables` API. `InducingVariables` must now have a `shape` property.
 * `gpflow.experimental.check_shapes.get_shape.register` has been replaced with
   `gpflow.experimental.check_shapes.register_get_shape`.
@@ -50,6 +52,9 @@ This release contains contributions from:
 
 ## Major Features and Improvements
 
+* Improved handling of variable noise
+  - All likelihood methods now take an `X` argument, allowing you to easily implement
+    heteroscedastic likelihoods.
 * `gpflow.experimental.check_shapes`
   - Can now be in three different states - ENABLED, EAGER_MODE_ONLY, and DISABLE.
   - Now support multiple variable-rank dimensions at the same time, e.g. `cov: [n..., n...]`.
--- a/gpflow/likelihoods/base.py
+++ b/gpflow/likelihoods/base.py
@@ -27,8 +27,10 @@ explicitly passed to likelihood constructor.
 """
 
 
-class Likelihood(Module, metaclass=abc.ABCMeta):
-    def __init__(self, latent_dim: Optional[int], observation_dim: Optional[int]) -> None:
+class Likelihood(Module, abc.ABC):
+    def __init__(
+        self, input_dim: Optional[int], latent_dim: Optional[int], observation_dim: Optional[int]
+    ) -> None:
         """
         A base class for likelihoods, which specifies an observation model
         connecting the latent functions ('F') to the data ('Y').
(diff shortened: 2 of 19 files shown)


gpflow/likelihoods/multiclass.py
```diff
@@ -34,17 +34,17 @@ class Softmax(MonteCarloLikelihood):
     """
 
     def __init__(self, num_classes: int, **kwargs: Any) -> None:
-        super().__init__(latent_dim=num_classes, observation_dim=None, **kwargs)
+        super().__init__(input_dim=None, latent_dim=num_classes, observation_dim=None, **kwargs)
         self.num_classes = self.latent_dim
 
-    def _log_prob(self, F: TensorType, Y: TensorType) -> tf.Tensor:
+    def _log_prob(self, X: TensorType, F: TensorType, Y: TensorType) -> tf.Tensor:
```
vdutor, line 40: Could you explain why you choose this order of inputs?

gpflow/likelihoods/base.py
```diff
@@ -48,97 +50,119 @@ def __init__(self, latent_dim: Optional[int], observation_dim: Optional[int]) ->
         :param observation_dim: the dimension of the observation vector Y for a single data point
```
vdutor, line 50: Forgot to add `input_dim` to docstrings

gpflow/likelihoods/base.py
```diff
@@ -48,97 +50,119 @@ def __init__(self, latent_dim: Optional[int], observation_dim: Optional[int]) ->
...
-        :param Y: observation Tensor, with shape [..., observation_dim]
+        :param result: result tensor, with shape [...]
+        :param X: input tensor, with shape [..., input_dim]
+        :param F: function evaluation tensor, with shape [..., latent_dim]
+        :param Y: observation tensor, with shape [..., observation_dim]
         """
         expected_shape = tf.broadcast_dynamic_shape(tf.shape(F)[:-1], tf.shape(Y)[:-1])
         tf.debugging.assert_equal(tf.shape(result), expected_shape)
 
+    def _check_input_dims(self, X: TensorType) -> None:
```
vdutor, line 84: I assume you will be adding your shape-checking functionality in a later PR?

gpflow/likelihoods/base.py
```diff
@@ -274,7 +313,7 @@ def _quadrature_dim(self) -> int:
         assert self.latent_dim is not None
         return self.latent_dim
 
-    def _quadrature_log_prob(self, F: TensorType, Y: TensorType) -> tf.Tensor:
+    def _quadrature_log_prob(self, F: TensorType, X: TensorType, Y: TensorType) -> tf.Tensor:
```
vdutor, line 316: why this change in order? Is there a way of working around this?

gpflow/likelihoods/base.py
```diff
@@ -48,97 +50,119 @@ def __init__(self, latent_dim: Optional[int], observation_dim: Optional[int]) ->
         :param observation_dim: the dimension of the observation vector Y for a single data point
         """
         super().__init__()
+        self.input_dim = input_dim
         self.latent_dim = latent_dim
         self.observation_dim = observation_dim
 
-    def _check_last_dims_valid(self, F: TensorType, Y: TensorType) -> None:
+    def _check_last_dims_valid(self, X: TensorType, F: TensorType, Y: TensorType) -> None:
```
vdutor, line 57: Can you explain why you choose this order for the input args?

vdutor (commented):
In principle, I am happy with this proposal. The only thing is that sometimes the order of input args seems inconsistent. It would be good to know why that is the case.

gpflow/likelihoods/base.py
jesnie: > Can you explain why you choose this order for the input args?

It seems most intuitive to me. You start with an X, then you map that to F, then you map that to Y?

gpflow/likelihoods/base.py
jesnie: > Forgot to add `input_dim` to docstrings

Done.

gpflow/likelihoods/base.py
jesnie: > I assume you will be adding your shape-checking functionality in a later PR?

Yes.

gpflow/likelihoods/base.py
jesnie: > why this change in order? Is there a way of working around this?

This is because this method is used as `self.quadrature.logspace(self._quadrature_log_prop, ...)` and `self.quadrature` operates over the first parameter, so the first parameter has to be `F`.

To work around this we'd have to update all our quadrature stuff to have special-case handling of `X`, which seems like a bad idea.

jesnie:
> In principle, I am happy with this proposal. The only thing is that sometimes the order of input args seems inconsistent. It would be good to know why that is the case.

Basically, our quadrature code operates over the first argument, so a couple of places I have had to change parameter order to have `F` as the first argument.

RELEASE.md
```diff
@@ -39,6 +39,8 @@ This release contains contributions from:
 
 ## Breaking Changes
 
+* All likelihood methods now take an extra `X` parameter. If you have written custom likelihoods or
+  you have custom code calling likelihoods directly you will need to add this extra parameter.
```
frgsimpson, line 43:
```suggestion
* All likelihood methods now take an extra `X` argument. If you have written custom likelihoods or
  you have custom code calling likelihoods directly you will need to add this extra argument.
```
Just to avoid confusion with the gpflow Parameter object.

RELEASE.md
```diff
@@ -50,6 +52,9 @@ This release contains contributions from:
 
 ## Major Features and Improvements
 
+* Improved handling of variable noise
+  - All likelihood methods now take an `X` parameter, allowing you to easily implement
```
frgsimpson, line 56:
```suggestion
  - All likelihood methods now take an `X` argument, allowing you to easily implement
```

frgsimpson (approved):
Looks great!

RELEASE.md
jesnie: > Just to avoid confusion with the gpflow Parameter object.

Sure.

RELEASE.md
jesnie: Sure. done.

PR #1920, opened 2022-06-29
Add a Polynomial mean function.

jesnie:
Add a Polynomial mean function.

Mostly because I intend to use it for a heteroskedastic variance in a later PR.

### DIFF
--- a/gpflow/mean_functions.py
+++ b/gpflow/mean_functions.py
@@ -26,13 +26,15 @@ allows this to be done whilst additionally learning parameters of the
 parametric function.
 """
 
-from typing import Collection, Optional
+from typing import Collection, Iterator, Optional, Tuple
 
 import numpy as np
 import tensorflow as tf
 
 from .base import Module, Parameter, TensorType
 from .config import default_float, default_int
+from .experimental.check_shapes import check_shape as cs
+from .experimental.check_shapes import check_shapes
 
 
 class MeanFunction(Module):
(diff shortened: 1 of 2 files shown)


gpflow/mean_functions.py
```diff
@@ -152,6 +154,46 @@ def __call__(self, X: TensorType) -> tf.Tensor:
...
+            return
+        for i in range(degree + 1):
+            for inner in self._compute_powers(degree - i, input_dim - 1):
+                yield (i,) + inner
+
+    @check_shapes(
+        "X: [batch..., input_dim]",
+        "return: [batch..., output_dim]",
+    )
+    def __call__(self, X: TensorType) -> tf.Tensor:
```
vdutor, line 191: Maybe we can reply on `tf.math.polyval` to evaluate the polynomial: https://www.tensorflow.org/api_docs/python/tf/math/polyval. I believe uses Horner's method, a slightly more stable alternative to evaluate polynomials.

gpflow/mean_functions.py
```diff
@@ -152,6 +154,46 @@ def __call__(self, X: TensorType) -> tf.Tensor:
...
+
+    @check_shapes("w: [broadcast output_dim, broadcast n_terms]")
+    def __init__(
+        self, degree: int, input_dim: int = 1, output_dim: int = 1, w: Optional[TensorType] = None
+    ) -> None:
+        """
+        :param degree: The degree of the polynomial.
+        :param input_dim: Number of inputs / variables this polynomial is defined over.
+        :param output_dim: Number of outputs / polynomials.
+        :param w: Initial weights of the terms of the polynomial.
```
vdutor, line 170: If you pass in weights, how should they be ordered, and how many weights do you expect given the `input_dim`, `output_dim` and `degree`? I assume first the constant term, linear terms, etc. Probably best to add this to the docstring.

tests/gpflow/test_mean_functions.py
```diff
@@ -188,6 +190,42 @@ def test_linear_mean_functions_associative_property(mean_functions: Sequence[Mea
...
+            [1.0 + 2.0 * 2.0 + 3.0 * (2.0 ** 2)],
+        ],
+        Y,
+    )
+
+
+def test_polynomial__linear() -> None:
+    # Test on a 3D linear function, where we can easily work out the maths.
+    p = Polynomial(degree=1, input_dim=3, w=[1.0, 2.0, 3.0, 4.0])
+    X = np.array([1.0, 2.0, 3.0])
```
vdutor, line 224: I'm surprised we can pass in a rank-1 tensor, should this not be `[[1.,2.,3.]]` (i.e. `1 x 3` shape) to represent a single 3D point?

gpflow/mean_functions.py
jesnie: > If you pass in weights, how should they be ordered, and how many weights do you expect given the `input_dim`, `output_dim` and `degree`? I assume first the constant term, linear terms, etc. Probably best to add this to the docstring.

I've now tried explaining this, though it's annoyingly difficult for the multivariate case.

gpflow/mean_functions.py
jesnie: > Maybe we can reply on `tf.math.polyval` to evaluate the polynomial: https://www.tensorflow.org/api_docs/python/tf/math/polyval. I believe uses Horner's method, a slightly more stable alternative to evaluate polynomials.

Yeah, I was looking at whether there was a library function for this - I think `tf.math.polyval` only applies to the single-variate case.

tests/gpflow/test_mean_functions.py
jesnie: > I'm surprised we can pass in a rank-1 tensor, should this not be `[[1.,2.,3.]]` (i.e. `1 x 3` shape) to represent a single 3D point?

You can do either. If your input shape is `[1, 3]` then the output will have shape `[1, 1]`. If your input has shape `[3]`, then your output will have shape `[1]`. In general the input shape is `[batch..., input_dim]` where `batch...` can be whatever you want, including empty. :-)

vdutor (approved):

PR #1928, opened 2022-07-06
Rename mean functions, and allow function parameters in likelihoods.

jesnie:
This PR:

1. Renames `mean_functions` -> `functions`.
2. Create aliases for `mean_functions` for backwards compalitibility.
3. Allows many of the existing likelihood classes to be configured with a functions instead of a constant value.

Effectively this allows the models that currently accept an external `Likelihood` to easily model heteroscedastic likelihoods, for example the VGP and the SVGP.


awav:
I disagree with renaming of `mean_functions` to `functions`. It is a bit generic and confusing. If were were shortening things I would choose `means` instead.

gpflow/likelihoods/scalar_continuous.py
```diff
@@ -36,61 +45,92 @@ class Gaussian(ScalarLikelihood):
...
-    DEFAULT_VARIANCE_LOWER_BOUND = 1e-6
-
     def __init__(
         self,
-        variance: float = 1.0,
-        variance_lower_bound: float = DEFAULT_VARIANCE_LOWER_BOUND,
+        variance: Optional[ConstantOrFunction] = None,
+        *,
+        scale: Optional[ConstantOrFunction] = None,
+        variance_lower_bound: float = DEFAULT_LOWER_BOUND,
```
awav, line 53: Would it make sense to leave the name as before, i.e. DEFAULT_VARIANCE_LOWER_BOUND?

awav, line 53: Would it make sense to leave the name as before, i.e. DEFAULT_VARIANCE_LOWER_BOUND?

awav (commented):
Does this work prepare the ground for heteroscedastic case? The renaming part looks a bit off in this PR.

gpflow/likelihoods/scalar_continuous.py
jesnie: > Would it make sense to leave the name as before, i.e. DEFAULT_VARIANCE_LOWER_BOUND?

Well, below it's used for `scale_lower_bound` and `shape_lower_bound` as well, so `DEFAULT_VARIANCE_LOWER_BOUND` would be a slightly weird name. I suppose I can just have a lot of constants instead, if you like that better?

jesnie:
> Does this work prepare the ground for heteroscedastic case? The renaming part looks a bit off in this PR.

It more than "prepares the ground". This PR is enough for the models that already take a `likelihood` to support the heteroscedastic case. So with this PR the `VGP` and `SVGP` are "done".

awav:
> > I disagree with renaming of `mean_functions` to `functions`. It is a bit generic and confusing. If were were shortening things I would choose `means` instead.
> 
> I'm not doing it to make the name shorter. I'm doing it because I want to use the same functions / code, for things other than means, and:
> 
> ```python
> variance: MeanFunction = mean_functions.Linear()
> ```
> 
> seems weird. With the rename it looks like:
> 
> ```python
> variance: Function = functions.Linear()
> ```

Ok, it makes sense, but still `functions` is a bit vague. However, I don't have immediate proposal for a better alternative.

gpflow/likelihoods/scalar_continuous.py
jesnie: > Why do we introduce different parameterisations for Gaussian?

If you want to parameterise the `Gaussian` with a function, then it's a lot more intuitive to use the `scale` as that is "linear". If you use the `variance` then you'd need a `sqrt` function to model linear noise, which is a bit weird.

vdutor (approved):

vdutor:
I'm happy with the backwards compatibility, in which we can still import the functions through `gpflow.mean_functions`. I also did not come up with a better name.

PR #1933, opened 2022-07-08
Add shape checks to models.

jesnie:


gpflow/models/cglb.py
```diff
@@ -159,9 +169,10 @@ def quad_term(self, common: SGPR.CommonTensors) -> tf.Tensor:
 
         return -ub
 
+    @inherit_check_shapes
     def predict_f(
         self,
-        xnew: InputData,
+        Xnew: InputData,
```
Corwinpro, line 175: Hi, may I ask about this? This would be backwards incompatible change for users explicitly specifying `xnew` as a keyword argument. What is the motivation for this change? Are we sure nothing would break on the user side if we rename this?

Thank you!

jesnie: > Hi, may I ask about this? This would be backwards incompatible change for users explicitly specifying `xnew` as a keyword argument. What is the motivation for this change? Are we sure nothing would break on the user side if we rename this?
> 
> Thank you!

Yes, I'm aware this is a breaking change, and no I'm not sure I'm not breaking anything. BUT the current state is that it's incompatible with all the other models where this is called `Xnew` which prevents polymorphism so... today you cannot write:
```python
    model.predict_f(xnew=my_new_x)
```
without knowing the type of `model`, because it's called `Xnew` in most of the models, but `xnew` in the cglb...

GPflow is full of this kind of stuff, and while I try to keep breaking changes reasonably rare, we'll probably need to be able to clean this kind of stuff up if we want to be able to continue development while remaining reasonably sane.

Corwinpro: Thank you @jesnie , I understand.

Would you like to take this in two steps, and only add the shape checks in this PR whilst renaming the input variable in other one? I can take care of the second PR, given that you have already done all the work.

Thank you, and please let me know what you think.

jesnie: > Thank you @jesnie , I understand.
> 
> Would you like to take this in two steps, and only add the shape checks in this PR whilst renaming the input variable in other one? I can take care of the second PR, given that you have already done all the work.
> 
> Thank you, and please let me know what you think.

I'm a big fan of "one thing per PR", but the `@inherit_check_shapes` won't work without the renaming, because it's looking for an argument called `Xnew`, which should have been inherited from the parent. There are ways to deal with that, but it seems like overkill to me, for the sake of the naming of one argument...

Corwinpro: Makes sense, thank you!

I guess it would be useful to add a line to the release notes for that change.

frgsimpson (approved):
LGTM!

gpflow/models/gplvm.py
```diff
@@ -254,8 +261,8 @@ def predict_f(
                 + tf.linalg.matmul(tmp2, tmp2, transpose_a=True)
                 - tf.linalg.matmul(tmp1, tmp1, transpose_a=True)
             )
-            shape = tf.stack([1, 1, tf.shape(Y_data)[1]])
-            var = tf.tile(tf.expand_dims(var, 2), shape)
+            shape = tf.stack([tf.shape(Y_data)[1], 1, 1])
+            var = tf.tile(tf.expand_dims(var, 0), shape)
```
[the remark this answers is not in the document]
jesnie: > This makes sense for consistency with the standard GPR predict_f, though does make me wonder why the ordering is swapped for full_cov. Perhaps the GPR should be the one to change?!

(I'm kind of impressed you've been able to read what's going on here, I should totally have documented that better...)

Yes, it's annoying that those parameters are swapped and part of me would love to swap them back, BUT they're consistent everywhere else. If we change this we'll break all our users and have to rewrite a significant chunk of GPflow. It's probably not happening, and if it does then it a separate project.

PR #1941, opened 2022-07-14
Extend the GPR and SGPR to support heteroscedastic noise.

jesnie:


frgsimpson (approved):
Looks great,  I think we just ought to verify that SGPR performance on a larger dataset is unaffected by these changes (both in terms of time and precision), as these won't be picked up by the tests.

Issue #1949, opened 2022-07-20
Write a notebook about GPflow utilities


jesnie:
Mostly the stuff in `gpflow.utilities`. We'd have to think a bit about what kind of stuff we really want to advertise externally, and what is more of an internal helper.

Issue #1948, opened 2022-07-20
Write a notebook on GPflow architecture


jesnie:

mattyred:
What do you exactly want from this notebook? Showing the architecture would mean showing how GPFlow works for new users?

Issue #1946, opened 2022-07-20
Write a notebook about models with latent variables


jesnie:
This is possibly already covered by:
https://gpflow.github.io/GPflow/2.5.2/notebooks/basics/GPLVM.html
https://gpflow.github.io/GPflow/2.5.2/notebooks/advanced/heteroskedastic.html
At least make sure that between these documents there's a coherent story.

Issue #1943, opened 2022-07-20
Write a notebook on how to handle large datasets.


jesnie:
1. On sparse models.
2. Maybe cover mini batching?

PR #1950, opened 2022-07-20
Clean up empty documents.

jesnie:
Delete notebooks without content. They have been replaced with the following tickets: #1943, #1944, #1945, #1946, #1947, #1948, #1949.

Also, reduce logging in [`multioutput.pct.py`](https://gpflow.github.io/GPflow/2.5.2/notebooks/advanced/multioutput.html) a bit, to make it more readable.


frgsimpson (approved):
LGTM!

PR #1954, opened 2022-08-09
Add a benchmark suite to GPflow.

jesnie:


benchmark/dataset_api.py
```diff
@@ -0,0 +1,337 @@
...
+
+def _get_data_file_from_url(cache_dir: Path, url: str, archive_member: Optional[Path]) -> Path:
+    file_name = url.rsplit("/", 1)[-1]
+    file_path = cache_dir / file_name
+    if not file_path.is_file():
+        cache_dir.mkdir(parents=True, exist_ok=True)
+        with urlopen(url) as response:
+            data = response.read()
+        file_path.write_bytes(data)
+        if archive_member:
```
uri-granta, line 269: Not 100% sure what archive_member is doing, but why not just extract the file you're returning?

uri-granta (commented):
(some initial comments on the first few files. need to look at something else now but will get back to this asap.)

benchmark/grouping.py
```diff
@@ -0,0 +1,214 @@
...
+
+class GroupingKeySource(Enum):
+    """ Sources of data to group by. """
+
+    METADATA = "metadata"
+    METRICS_DF = "metrics_df"
+    PLOTTER = "plotter"
+
+
+class GroupingKey(Enum):
```
uri-granta, line 36: Maybe include the member value type in the class definition? (mypy won't check it for you but it won't complain either)

```suggestion
class GroupingKey(Tuple[GroupingKeySource, type, bool, float], Enum):
```

benchmark/grouping.py
```diff
@@ -0,0 +1,214 @@
...
+        whether values of this key need context.
+        """
+        return self.value[2]  # type: ignore[no-any-return]
+
+    @property
+    def key_cost(self) -> float:
+        """
+        How much we don't want to show this key.
+
+        Should be positive.
```
uri-granta, line 91: (do you want to police this?)

benchmark/grouping.py
```diff
@@ -0,0 +1,214 @@
...
+        if plotter_by_columns:
+            df = df.drop(columns=plotter_by_columns)
+        result.append((keys, df, key_plotters))
+    return result
+
+
+def _sanitise_keys(
+    by: Sequence[GroupingKey], keys: Any, plotters: Collection[Plotter]
+) -> Tuple[Tuple[str, ...], Collection[Plotter]]:
+    result_keys: List[str] = []
```
uri-granta, line 184: (docstring?)

tests/benchmark/test_plot.py
```diff
@@ -0,0 +1,64 @@
...
+    setup_source(source1, 1)
+    setup_source(source21, 21)
+    setup_source(source22, 22)
+
+    cycle = source22 / "cycle"
+    cycle.symlink_to("../..")
+
+    metrics_df, metadata = _crawl_sources([source1, source2])
+    assert len(metrics_df) == 6
+    assert {"test_id_1", "test_id_21", "test_id_22"} == set(m.run_id for m in metadata)
```
uri-granta, line 64: (out of curiosity, is there a good way to test actual plots with any benefit?)

uri-granta (approved):
Looks good! Various comments but nothing too important

Issue #1959, opened 2022-08-16
GPR.posterior is not numerically stable.


jesnie:
See https://github.com/secondmind-labs/trieste/issues/603 for context.

The `GPRPosterior` does:
```python
Kmm_plus_s_inv = tf.linalg.cholesky_solve(Lm, tf.eye(tf.shape(X_data)[0], dtype=Lm.dtype))
```
which is not numerically stable, and is causing covariance matrices that are not symmetric.

The `SGPRPosterior` has similar code, look into that as well.

PR #1960, opened 2022-08-17
Rewrite (S)GPRPosterior for numerical stability.

jesnie:
Fixes: #1959

Our previous `GPRPosterior` and `SGPRPosterior` objects computes matrix inverses by solving against an identity matrix. This is not numerically stable.

In this PR I rewrite those posterior objects. My strategy has simply been to take the old (proven) `predict_f` method, and copy everything that does not depend on `Xnew` to `_precompute` and the rest to `_conditional_with_precompute`. `_conditional_fused` then just becomes a composition of `_precompute` and `_conditional_with_precompute`, which is exactly the same code as the old `predict_f`.

I've benchmarked the new code using our new benchmarking framework:

- The timings are generally on the same order of magnitude.
- Often the old code is a bit faster.
- Except sometimes the new code is massively faster for GPR on a GPU.

metrics_posterior_build_time
metrics_posterior_prediction_time

### DIFF
--- a/gpflow/posteriors.py
+++ b/gpflow/posteriors.py
@@ -32,7 +32,14 @@ from .conditionals.util import (
 )
 from .config import default_float, default_jitter
 from .covariances import Kuf, Kuu
-from .experimental.check_shapes import check_shapes, inherit_check_shapes
+from .experimental.check_shapes import (
+    ErrorContext,
+    Shape,
+    check_shapes,
+    get_shape,
+    inherit_check_shapes,
+    register_get_shape,
+)
 from .inducing_variables import (
     FallbackSeparateIndependentInducingVariables,
     FallbackSharedIndependentInducingVariables,
@@ -160,6 +167,11 @@ class PrecomputedValue:
             PrecomputedValue(alpha, alpha_dynamic),
             PrecomputedValue(Qinv, Qinv_dynamic),
         )
+
+
+@register_get_shape(PrecomputedValue)
+def get_precomputed_value_shape(shaped: PrecomputedValue, context: ErrorContext) -> Shape:
+    return get_shape(shaped.value, context)
 
 
 def _validate_precompute_cache_type(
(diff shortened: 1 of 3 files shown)


gpflow/posteriors.py
```diff
@@ -549,44 +477,45 @@ def _conditional_with_precompute(
...
+        L, LB, c = cache
 
         Kus = Kuf(self.inducing_variable, self.kernel, Xnew)
-        Knn = self.kernel(Xnew, full_cov=full_cov)
-
-        Ksu = tf.transpose(Kus)
-        mean = Ksu @ alpha
-
+        tmp1 = tf.linalg.triangular_solve(L, Kus, lower=True)
+        tmp2 = tf.linalg.triangular_solve(LB, tmp1, lower=True)
```
vdutor, line 484: Would it be possible to give these more meaningful names?

gpflow/posteriors.py
```diff
@@ -381,66 +381,43 @@ def _conditional_with_precompute(
...
+            Knn=Knn,
+            f=err,
+            full_cov=full_cov,
+            q_sqrt=None,
+            white=False,
+        )
 
     @check_shapes(
-        "return[0].value: [M, M]",
+        "return[0].value: [M, D]",
```
vdutor, line 400: Suggestion: you could register the shape of a `PrecomputedValue` for `check_shapes`...

gpflow/posteriors.py
```diff
@@ -381,66 +381,43 @@ def _conditional_with_precompute(
...
-        Kmm_plus_s_inv = tf.linalg.cholesky_solve(Lm, tf.eye(tf.shape(X_data)[0], dtype=Lm.dtype))
 
+        D = err.shape[1]
         M = X_data.shape[0]
+        D_dynamic = D is None
         M_dynamic = M is None
 
-        return (PrecomputedValue(Kmm_plus_s_inv, (M_dynamic, M_dynamic)),)
+        return (
+            PrecomputedValue(err, (M_dynamic, D_dynamic)),
```
vdutor, line 418: I would expect `err` to have potentially leading dims. Is this covered?

vdutor (approved):

st--:
@jesnie just wanted to check, how does this affect the many-predictions use-case, for different numbers of inducing points (SVGP/SGPR) or data points (VGP/GPR)? Which was the initial motivation for creating the Posterior objects in the first place - e.g. if I want to call predict_f() with different Xs many many times inside a loop. There seems to be a definite regression in the benchmarks overview you posted, but I'm also wondering how the gap changes with M/N.

jesnie:
> @jesnie just wanted to check, how does this affect the many-predictions use-case, for different numbers of inducing points (SVGP/SGPR) or data points (VGP/GPR)? Which was the initial motivation for creating the Posterior objects in the first place - e.g. if I want to call predict_f() with different Xs many many times inside a loop. There seems to be a definite regression in the benchmarks overview you posted, but I'm also wondering how the gap changes with M/N.

Sorry about the late reply - I've been on vacation, but that is a very fair question and now I have some data for you:

boring_posterior_build_time
boring_posterior_prediction_time

(For the SGPR the number of inducing points is half the number of data points...)

1. The VGP and SVGP are not affected by this PR.
2. The new code is no slower than the old code for building the posterior objects.
3. Generally the new code tends to faster than the old code, when on a GPU and with larger data sizes.
4. For the GPR the new code is always faster than the old code, when on a GPU - it's slower on a CPU.
5. For the SGPR the new code is faster than the old code, when you run large datasets on a GPU - it's slower otherwise.

I think, if users care about speed they will be using a GPU, so we can disregard the CPU results.
This means that this is a clear improvement for the GPR, and a mixed bag for the SGPR.

gpflow/posteriors.py
jesnie: > Suggestion: you could register the shape of a `PrecomputedValue` for `check_shapes`...

Done.

gpflow/posteriors.py
jesnie: > I would expect `err` to have potentially leading dims. Is this covered?

I'm pretty sure GPflow doesn't support the initial (`X`, `Y`) training data (and hence `err`) having batch dimensions - though it is supported for `Xnew`.

gpflow/posteriors.py
jesnie: > Would it be possible to give these more meaningful names?

I have no clue what to call them, though - and this is just copy n' pasted code, it might be better to keep it like this to make it easier to see that it's copy n' pasted.

Issue #1952, opened 2022-07-22
Add a benchmark suite to GPflow.


jesnie:
Add a benchmark suite to monitor regressions.

1. Have some "minimal" set of benchmarks for continuously monitoring regression.
2. Should be extensible to make it easy to test/benchmark new features, (but without adding them to the set of minimal benchmarks).

It would be nice with integration into CI.

jesnie:
The current solution doesn't work, because the CircleCI worker cannot push to `develop` because that branch has special protections...

PR #1966, opened 2022-08-31
Regularly run benchmarks through CI.

jesnie:
Fixes: #1952 

Set up CircleCI to regularly run our benchmarks.

1. I've created [a trigger called `run_benchmarks`](https://app.circleci.com/settings/project/github/GPflow/GPflow/triggers?return-to=https%3A%2F%2Fapp.circleci.com%2Fprojects%2Fproject-dashboard%2Fgithub%2FGPflow%2F) that runs every Monday between 10am and 11am.
2. It runs benchmarks and stores the results in the `develop` branch, under `GPflow/ci_benchmark_results`.
3. I've created a symbolic link from `GPflow/doc/sphinx/benchmark_plots` to `GPflow/ci_benchmark_results/plots`.
4. I've created a small page in our documentation that displays the plot.
5. The plot I'm checking in here is temporary - to make sure plot generation doesn't crash while we have no data.
6. Some small tuning of the `ci` benchmark suite.


.circleci/config.yml
```diff
@@ -261,6 +261,64 @@ jobs:
...
+      - run:
+          name: Run benchmarks
+          command: |
+            source .venv/bin/activate
+            rm -rf <<parameters.tmp_results_dir>>
+            mkdir -p <<parameters.tmp_results_dir>>
+            python -m benchmark.run --no_plot ci <<parameters.tmp_results_dir>>
+      - add_ssh_keys:
+          fingerprints:  # Add key to give write-access below
+            - "b5:9b:f3:74:26:00:b3:c3:de:3a:b3:03:ee:0b:0b:64"
```
vdutor, line 294: Is this fine from a security point of view?

vdutor (approved):
LGTM. Just a few general questions:
- how does it get access to the previous results for the plots that show performance over time?
- It is unclear with which data the timeline.png image was made.
- Is it possible to add uncertainty intervals in the plots?
- How long does it take to run the benchmarks?

jesnie:
> LGTM. Just a few general questions:
> 
> * how does it get access to the previous results for the plots that show performance over time?

They're stored in `GPflow/ci_benchmark_results` in the `develop` branch.

> * It is unclear with which data the timeline.png image was made.

Yeah, it's based on some temporary, small-scale, experiments I did, just to have something to plot. I've deleted that data.

> * Is it possible to add uncertainty intervals in the plots?

Yes. It already plots uncertainty - it just only has a single data-point, because I didn't want to wait for real data.

> * How long does it take to run the benchmarks?

Well, it took 3 minutes for a single repetition of two models. The configuration is to run 5 repetitions of 4 models so I'd expect ~ 30 minutes?

.circleci/config.yml
jesnie: > Is this fine from a security point of view?

...I think so... it's not the key itself, only a fingerprint of one...

PR #1968, opened 2022-09-05
Remove generator from Polynomial function.

jesnie:
TensorFlow complains that it cannot compile our `Polynomial` function, because it doesn't like generators, so update the relevant code to create a list instead.

Also fix a couple of documentation formatting issues.


uri-granta (approved):
LGTM!

PR #1971, opened 2022-09-06
Push benchmark results to gh-pages instead of develop.

jesnie:
Fixes: #1952

With the current set-up we write benchmark results directly to the `develop` branch. This doesn't work because the `develop` branch is protected and the CircleCI worker doesn't have access privileges to overwrite that protection.

With this PR we move benchmark results to the `gh-pages` branch instead, which the CircleCI worker pushes to all the time anyway.


uri-granta (approved):
:+1:

Issue #1970, opened 2022-09-06
CircleCI jobs time out.


jesnie:
Without a paid account CircleCI has a task time limit of 1 hour. We often hit that with our documentation, and now also with our benchmarks.

Figure out a way to parallelise those jobs. I think we can do something with CircleCI workspaces.

PR #1972, opened 2022-09-06
Paralellise building of documentation.

jesnie:
Part of #1970 .

We're having problems with some builds timing out. This PR paralellises the slow documentation build.

This PR splits notebooks between shards naively by filename. We may want to consider a future PR to keep track of notebook build timings, and split them more intelligently.


.circleci/config.yml
```diff
@@ -212,6 +212,35 @@ jobs:
...
+          tfp_ver: <<pipeline.parameters.max_tfp_ver>>
+      - run:
+          name: Build notebooks
+          no_output_timeout: 60m
+          command: |
+            source .venv/bin/activate
+            mkdir -p /tmp/gpflow_build_docs
+            time python doc/build_docs.py --shard ${CIRCLE_NODE_INDEX}/${CIRCLE_NODE_TOTAL}
+            # Each shard must output unique files, so clean up output directory:
+            rm $(find /tmp/gpflow_build_docs -type f -and -not -name \*.ipynb)
```
uri-granta, line 238: I think you can do this (but don't quote me on it):

```suggestion
            find /tmp/gpflow_build_docs -type f -and -not -name \*.ipynb -delete
```

.circleci/config.yml
```diff
@@ -212,6 +212,35 @@ jobs:
...
+          py_ver: <<pipeline.parameters.max_py_ver>>
+          tf_ver: <<pipeline.parameters.max_tf_ver>>
+          tfp_ver: <<pipeline.parameters.max_tfp_ver>>
+      - run:
+          name: Build notebooks
+          no_output_timeout: 60m
+          command: |
+            source .venv/bin/activate
+            mkdir -p /tmp/gpflow_build_docs
+            time python doc/build_docs.py --shard ${CIRCLE_NODE_INDEX}/${CIRCLE_NODE_TOTAL}
```
uri-granta, line 236: (does the `time` result in a log output?)

doc/build_docs.py
```diff
@@ -38,6 +38,34 @@
 _DOCTREE_TMP = _TMP / "doctree"
 
 
+class ShardingStrategy:
+    def __init__(self, spec: str) -> None:
+        self.spec = spec
+        if spec == "no":
```
uri-granta, line 44: maybe docstring or comments summarising the three modes?

doc/build_docs.py
```diff
@@ -38,6 +38,34 @@
...
+            self.shard_i = 0
+            self.shard_n = 1
+        elif spec == "collect":
+            self.setup_tmp = False
+            self.build_notebooks = False
+            self.build_other = True
+            self.shard_i = 0
+            self.shard_n = 1
+        else:
+            i_str, n_str = spec.split("/")
```
uri-granta, line 57: Might be friendlier to explicitly validate the input rather than rely on split and int. (As an aside, at the moment "٢/٤" is valid input.)

uri-granta (approved):
Nice!

.circleci/config.yml
jesnie: Honestly that `time` is some debugging stuff. I'll remove it.

.circleci/config.yml
jesnie: Good call. Seems to work. Thanks!

doc/build_docs.py
jesnie: > Might be friendlier to explicitly validate the input rather than rely on split and int. (As an aside, at the moment "٢/٤" is valid input.)

What would you do? Write a regular expression? I don't think that's worth it. Remember this isn't directly customer facing code - it only runs in CI.

doc/build_docs.py
jesnie: > maybe docstring or comments summarising the three modes?

Done.

PR #1976, opened 2022-09-15
Upgrade for TensorFlow 2.10.

jesnie:
Start testing GPflow for TensorFlow 2.10.
Also slight update to format of `pip install`.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -9,16 +9,16 @@ parameters:
     default: "3.10"
   min_tf_ver:
     type: string
-    default: "2.4"
+    default: "2.4.0"
   max_tf_ver:
     type: string
-    default: "2.9"
+    default: "2.10.0"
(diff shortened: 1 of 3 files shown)


uri-granta (approved):
LGTM

Corwinpro:
I think this now fails of m1 macos. I will investigate and update.

jesnie:
> I think this now fails of m1 macos. I will investigate and update.
> 
> UPD: seems to be running fine on `python3.9` but fails on 3.10.

I assume it's because Google hasn't released the right packages yet, and that there isn't really anything we can do about it?

Corwinpro:
I think so: using `python3.9` for default `gpflow` installation works just fine.

It might actually be a bug in the google's packages - there is `tensorflow-macos==2.10` for `python3.10` (and even 2.8, 2.9) but they don't work.

PR #1980, opened 2022-09-20
Work around bug in MyPy.

jesnie:
Work around bad interaction between Python, MyPy and NumPy versions, by pinning a specific version of Python `3.10.x`.

We should be able to remove this fix after the next release of MyPy.

See: https://github.com/python/mypy/issues/13627 for more details.


uri-granta (approved):
Looks good! Any easy way to make sure we remember to revert this once mypy is updated?

jesnie:
> Looks good! Any easy way to make sure we remember to revert this once mypy is updated?

Not that I can think of. In the worst case I expect this will be fixed when we upgrade to Python 3.11, which shouldn't be too long. :shrug: .

PR #1982, opened 2022-09-22
Reformat notebooks to width 80.

jesnie:
Reformat notebooks to width 80, because that is the width that is available in the compiled documentation, and this avoids horizontal scrolling in the compiled documentation.

Also remove some unused imports.


uri-granta (approved):
Nice

PR #1985, opened 2022-09-25
Change likelihood defaults

awav:

## Summary

GPflow had a feature of setting the minimal positive value using `gpflow.config.set_default_positive_minimum`. Recently, the default minimum boundary for positive values in scalar likelihoods has been changed to in module global variable with fixed value. That introduced a couple of issues:
* New approach created confusion around `positive_minimum` config value, as it was supposed to control **all** positive values in the GPflow package.
* Default global value in likelihoods module creates another way of controlling behaviour of GPflow package. `config` was supposed to be a single controlling centre.

This PR introduces `likelihoods_positive_minimum` in config that controls positive values in scalar likelihoods.

## PR checklist

- [ ] Build checks
- [ ] Release management


gpflow/config/__config__.py
```diff
@@ -37,6 +37,7 @@
 * ``GPFLOW_FLOAT``: "float16", "float32", or "float64"
 * ``GPFLOW_POSITIVE_BIJECTOR``: "exp" or "softplus"
 * ``GPFLOW_POSITIVE_MINIMUM``: Any positive float number
+* ``LIKELIHOODS_POSITIVE_MINIMUM``: Any positive float number
```
jesnie, line 40: We should probably prepend `GPFLOW_` for consistency with the above...

```suggestion
* ``GPFLOW_LIKELIHOODS_POSITIVE_MINIMUM``: Any positive float number
```

jesnie (approved):
LGTM

PR #1988, opened 2022-09-27
Loosen Python version requirements again.

jesnie:
MyPy 0.981 seems to have been released, and we can get rid of our workaround.


jesnie:
Superseded by https://github.com/GPflow/GPflow/pull/1990

PR #2002, opened 2022-10-17
Remove `gpflow.experimental.check_shapes` and import `check_shapes` instead.

jesnie:


sc336 (approved):
Looks simple and fine, happy to approve this if it passes automated testing.

PR #2020, opened 2022-11-30
Avoid loss of precision in to_default_float when input is a python float

ChrisMorter:
**PR type:** bugfix 

## Summary

It seems that `to_default_float` can result in a loss of precision when the input value is a python float. This was previously not the case because `to_default_float` used to use gpflow's `cast` function internally, which used `tf.convert_to_tensor(..., dtype=tf.float64)` in the case that the input is a python float. 

However, this function was declared obsolete in gpflow 2.6.1, and was replaced by Tensorflow's `tf.cast` instead. The implementation of `tf.cast` is such that it converts the python float to a tf.float32 tensor (thereby losing precision), which it then casts to tf.float64.

You can see the essence of the problem as follows (note that 0.99999999 and 1.0 are distinct at 64 bit precision, so the fact that the lhs is rounded to 1.0 demonstrates a loss of precision):

```
lhs = tf.cast(0.99999999, dtype=tf.float64)
rhs = tf.convert_to_tensor(1.0, dtype=tf.float64)
assert lhs != rhs
```

### Release notes

## PR checklist

- [x] Build checks
- [x] Release management

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -83,5 +83,6 @@ Because GitHub's [graph of contributors](http://github.com/GPflow/GPflow/graphs/
 [@tmct](https://github.com/tmct)
 [@ltiao](https://github.com/ltiao)
 [@corwinpro](https://github.com/corwinpro)
+[@ChrisMorter](https://github.com/ChrisMorter)
 
 Add yourself when you first contribute to GPflow's code, tests, or documentation!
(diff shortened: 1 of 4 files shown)


ChrisMorter:
Closing to recreate with master as target to facilitate doing a quick release ([recreated PR](https://github.com/GPflow/GPflow/pull/2021)).

PR #2022, opened 2022-12-02
Some corrections to RELEASE.md

sc336:
Some typos in the RELEASE.md for the quick 2.6.4 bugfix patch.


uri-granta (approved):
Looks good. (At some point we should chat about the gpflow release process, as trieste's process is a bit ugly and could use updating.)

PR #2024, opened 2022-12-02
Sc336/fix conflicts

sc336:
This is an attempt to merge the v2.6.4 changes into develop.


tests/gpflow/utilities/test_misc.py
```diff
@@ -36,3 +36,44 @@
 )
 def test_is_variable(value: Any, expected: bool) -> None:
     assert expected == is_variable(value)
+
+
+@pytest.mark.parametrize(
+    "x",
+    [
+        0.999999999,
```
avullo, line 44: What is the purpose to parametrise over a single value?

sc336: I think it's to allow the test to be extended to other values should that be required in the future.  I also think it fits the style of the rest of the tests.  @ChrisMorter wrote the test, he might have further commentary.

ChrisMorter: No good reason to be honest, I think I anticipated having multiple test cases, but only ended up needing one. This can be changed to remove the parameterisation if you would like?

avullo (approved):
LGTM

PR #2032, opened 2023-01-06
Rotated deploy key due to CircleCI security incident 2023-01-04

sc336:
**PR type:** security issue

**Related issue(s)/PRs:** CircleCI had a [security issue](https://circleci.com/blog/january-4-2023-security-alert/) that requires some action

## Summary

**Proposed changes**
Change the CircleCI deploy key (that is what this PR is doing; there are other key rotations also required)

### Release notes

**Fully backwards compatible:** yes

## PR checklist

- [ ] New features: code is well-documented
  - [ ] detailed docstrings (API documentation)
  - [ ] notebook examples (usage demonstration)
- [ ] The bug case / new feature is covered by unit tests
- [ ] Code has type annotations
- [ ] Build checks
  - [ ] I ran the black+isort formatter (`make format`)
  - [ ] I locally tested that the tests pass (`make check-all`)
- [ ] Release management
  - [ ] RELEASE.md updated with entry for this change
  - [ ] New contributors: I've added myself to CONTRIBUTORS.md

PR #2033, opened 2023-01-11
Sc336/key rotation

sc336:
This is a continuation of #2032.  The old key remained in one place and it is replaced here with the new key.


uri-granta (approved):
:+1:

PR #2041, opened 2023-01-24
Sc336/2.7.0 master merge

sc336:
After some frustration trying to merge develop into master while both are protected, this merges in a new branch (sc336/2.7.0-master-merge-again) which is created from develop and has master merged into it.


st--:
I think the issue might have happened by _squash-merging_ develop into master in a previous release. Or something like that. I remember squashing/rebasing commits on one of the branches during a merge was the issue...

PR #2043, opened 2023-01-26
Use PyPi token for package uploading

awav:
Use PyPi [token](https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#id71) for uploading the GPflow's python package via CircleCI instead of the password.


sc336 (approved):
That is easier than I thought.

PR #2047, opened 2023-03-10
Always copy the final point found by scipy

khurram-ghani:
**PR type:** bugfix

## Summary
We have observed some instability during model training, where occasionally the SGPR model loss is unexpectadly very poor. We have tracked the problem down to how the _gpflow_ wrapper calls the _scipy_ optimiser (using BFGS in our case).

The wrapper uses `tf.Variable`s for the model hyperparameters and these are updated on every evaluation of the training loss. On those occasions where the BFGS line-search fails for an iteration, the _scipy_ optimiser returns the previous good point that it found and sets appropriate flags in the `OptimizeResult` class. However, the model hyperparameters will have the value of the last evaluation of the failed line-search, which can be quite poor. One solution is for the caller to look at the returned result, realise that there was a fail and copy over the final point from `OptimizeResult` back to the model hyperparameters. However it's perhaps better for the wrapper to always copy the final _scipy_ optimiser point to the model just before returning.

**Proposed changes**

* ...
* ...
* ...

### Minimal working example

```python
```

### Release notes

## PR checklist

- [ ] Build checks
- [ ] Release management

### DIFF
--- a/gpflow/optimizers/scipy.py
+++ b/gpflow/optimizers/scipy.py
@@ -39,7 +39,6 @@ class Scipy:
         step_callback: Optional[StepCallback] = None,
         compile: bool = True,
         allow_unused_variables: bool = False,
-        always_copy_last_good_state: bool = False,
         **scipy_kwargs: Any,
     ) -> OptimizeResult:
         """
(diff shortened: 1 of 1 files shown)


awav:
Why is that a draft? Suggestions: remove the option from the minimize. The only meaningful final values for variables is the last successful iteration of the optimizer!

gpflow/optimizers/scipy.py
```diff
@@ -39,6 +39,7 @@ def minimize(
         step_callback: Optional[StepCallback] = None,
         compile: bool = True,
         allow_unused_variables: bool = False,
+        always_copy_last_good_state: bool = False,
```
awav, line 42: Please, remove this. An explanation in the main thread.

awav (approved):

gpflow/optimizers/scipy.py
awav: @khurram-ghani, @sc336 I reverted the last commit

PR #2050, opened 2023-03-20
TF bugfix in `NaturalGradient._natgrad_steps`

ericpena:
**PR type:** bugfix

**Related issue(s)/PRs:** 

## Summary

**Proposed changes**
* This change is an update to the `_natgrad_steps` method within the `NaturalGradient` class.
* Specifically, the line (`tf.name_scope(f"{self._name}/natural_gradient_steps"):`) was throwing an `AttributeError` because `_name` should be `name` (i.e., without an underscore).
* I experienced this error when attempting to replicate the [Heteroskedastic Likelihood and Multi-Latent GP](https://gpflow.github.io/GPflow/2.7.1/notebooks/advanced/heteroskedastic.html) example.
* Change that I made:
**Before**: `with tf.name_scope(f"{self._name}/natural_gradient_steps"):`
**After**: `with tf.name_scope(f"{self.name}/natural_gradient_steps"):`

* I also added my name to the list of contributors and documented changed in `release.md`.

**What alternatives have you considered?**
* I tried changing different versions of Tensorflow before I took a look at the source code and found the issue.

#### Before the bug fix

The code snippet that produces the error within the Heteroskedastic Likelihood and Multi-Latent GP example is the following:

```python
epochs = 100
log_freq = 20

for epoch in range(1, epochs + 1):
    optimisation_step()

    # For every 'log_freq' epochs, print the epoch and plot the predictions against the data
    if epoch % log_freq == 0 and epoch > 0:
        print(f"Epoch {epoch} - Loss: {loss_fn().numpy() : .4f}")
        Ymean, Yvar = model.predict_y(X)
        Ymean = Ymean.numpy().squeeze()
        Ystd = tf.sqrt(Yvar).numpy().squeeze()
        plot_distribution(X, Y, Ymean, Ystd)

model
```

Then, I would get the following error
```python
AttributeError                            Traceback (most recent call last)
<ipython-input-11-aee9c220de71> in <module>
      3 
      4 for epoch in range(1, epochs + 1):
----> 5     optimisation_step()
```

and this continues with:

```python
/usr/local/lib/python3.9/dist-packages/gpflow/optimizers/natgrad.py in tf___natgrad_steps(self, loss_fn, parameters)
     21                     loss = ag__.converted_call(ag__.ld(loss_fn), (), None, fscope)
[...]

PR #2061, opened 2023-04-19
test CI build

st--:
The build of #2050 is failing, but this doesn't seem to be anything introduced by the PR. This PR is a null change to check whether the build system is working correctly.


st--:
seems to be working...

PR #2062, opened 2023-04-19
Copy of #2050

st--:
version of #2050 in branch inside repo to test out build system


st--:
Hm build seems to be working normally again now

PR #2063, opened 2023-04-19
Fix api doc generation for aliases

### LINKED ISSUE #2056: Missing documentation for SquaredExponential kernel
# Documentation

*Is there anything missing in the docs?*

Documentation of gpflow.kernels.stationaries.SquaredExponential went missing sometime after 2.4.0

It is there in [2.4.0](https://gpflow.github.io/GPflow/2.4.0/api/gpflow/kernels/index.html#gpflow-kernels-squaredexponential).

In [2.5.0](https://gpflow.github.io/GPflow/2.5.0/api/gpflow/kernels/index.html#gpflow-kernels-rbf) (and [currently](https://gpflow.github.io/GPflow/develop/api/gpflow/kernels/index.html#gpflow-kernels-rbf)), the entry was replaced by the RBF "as an alias of `gpflow.kernels.stationaries.SquaredExponential`", but there is no `SquaredExponential` entry.

khurram-ghani:
**PR type:** doc improvement

**Related issue(s)/PRs:** resolves #2056 

## Summary

**Proposed changes**

* Fix issues with API doc generation for `SquaredExponential` and `Constant` kernels.
* Due to use of aliases, previously the doc generation for these two kernels was missing. Only the minimal docs for `RBF` and `Bias` aliases (respectively) were present.

Doc example with missing classes: https://gpflow.github.io/GPflow/2.7.1/api/gpflow/kernels/index.html#gpflow-kernels-rbf

Generated fixed documentation locally and checked the classes are present.

## PR checklist

- [X] Build checks
- [ ] Release management

### DIFF
--- a/doc/generate_module_rst.py
+++ b/doc/generate_module_rst.py
@@ -170,9 +170,11 @@ class DocumentableModule:
             module.seen_in_dispatchers(seen)
         for function in self.functions:
             if isinstance(function, DocumentableDispatcher):
-                impls = function.obj.funcs.values()
-                for impl in impls:
-                    seen.add(id(impl))
+                # See comment below (for classes) about aliases.
+                if function.name.endswith("." + function.obj.__name__):
+                    impls = function.obj.funcs.values()
+                    for impl in impls:
+                        seen.add(id(impl))
 
     def prune_duplicates(self) -> None:
         seen: Set[int] = set()
--- a/doc/generate_module_rst.py
+++ b/doc/generate_module_rst.py
@@ -170,11 +170,11 @@ class DocumentableModule:
             module.seen_in_dispatchers(seen)
         for function in self.functions:
             if isinstance(function, DocumentableDispatcher):
-                # See comment below (for classes) about aliases.
-                if function.name.endswith("." + function.obj.__name__):
-                    impls = function.obj.funcs.values()
-                    for impl in impls:
-                        seen.add(id(impl))
+                impls = function.obj.funcs.values()
+                for impl in impls:
+                    # See comment below (for classes) about aliases.
+                    key = (impl.__name__, id(impl))
+                    seen.add(key)
 
     def prune_duplicates(self) -> None:
         seen: Set[int] = set()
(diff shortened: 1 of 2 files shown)


doc/generate_module_rst.py
```diff
@@ -186,14 +186,19 @@ def prune_duplicates(self) -> None:
             new_classes = []
             for c in module.classes:
                 if id(c.obj) not in seen:
-                    seen.add(id(c.obj))
+                    # Special treatment for aliases, i.e. when names don't match.
+                    # Don't mark them as seen to avoid the original object being skipped.
+                    if c.name.endswith("." + c.obj.__name__):
```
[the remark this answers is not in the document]
khurram-ghani: The `RBF` entry was already present in the generated `.rst` file. The original code however did not list `SquaredExponential` entry in the `.rst` file, as it detected it as a duplicate. Sphinx then takes the `.rst` file and does the appropriate thing from there for the present entries. This fix simply ensures that both `RBF` and `SquaredExponential` are listed in the `.rst` file.

doc/generate_module_rst.py
```diff
@@ -186,14 +186,19 @@ def prune_duplicates(self) -> None:
...
+                        seen.add(id(c.obj))
                     new_classes.append(c)
             module.classes = new_classes
 
             new_functions = []
             for f in module.functions:
                 if id(f.obj) not in seen:
-                    seen.add(id(f.obj))
+                    # See comment above (for classes) about aliases.
+                    if f.name.endswith("." + f.obj.__name__):
```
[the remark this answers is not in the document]
khurram-ghani: Good spot; I missed that. I'll add that as well for completeness.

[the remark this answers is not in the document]
khurram-ghani: I have added that now ([3bc0aac](https://github.com/GPflow/GPflow/pull/2063/commits/3bc0aac1920f28109b740bc4ace9df14e0865dcc)).

doc/generate_module_rst.py
```diff
@@ -170,9 +170,11 @@ def seen_in_dispatchers(self, seen: Set[int]) -> None:
             module.seen_in_dispatchers(seen)
         for function in self.functions:
             if isinstance(function, DocumentableDispatcher):
-                impls = function.obj.funcs.values()
-                for impl in impls:
-                    seen.add(id(impl))
+                # See comment below (for classes) about aliases.
+                if function.name.endswith("." + function.obj.__name__):
```
uri-granta, line 174: Shouldn't this be checking for aliases of the implementations `impl` rather than of the dispatcher `function`?

doc/generate_module_rst.py
```diff
@@ -186,14 +188,19 @@ def prune_duplicates(self) -> None:
...
+                    if c.name.endswith("." + c.obj.__name__):
+                        seen.add(id(c.obj))
                     new_classes.append(c)
             module.classes = new_classes
 
             new_functions = []
             for f in module.functions:
                 if id(f.obj) not in seen:
-                    seen.add(id(f.obj))
+                    # See comment above (for classes) about aliases.
```
uri-granta, line 201: Thinking about this a bit more, wouldn't it make more sense to just use a different key? That way we would also remove duplicate aliases.

```suggestion
                    key = (id(f.obj), f.name[f.name.rfind(".")+1:])
                    if key not in seen:
                        seen.add(key)
                        new_functions.append(f)
```

doc/generate_module_rst.py
khurram-ghani: I guess you could have aliases at both levels, dispatcher and implementations. Adding a check at the implementation level is more tricky since there is no existing Documentable* wrapper class for the implementations -> so there is no pathname to reference, only the function objects themselves. It seems to me that it is unlikely we will have aliases for implementations, since there is not a need for users to call them directly. What do you think?

doc/generate_module_rst.py
khurram-ghani: That makes sense. I have updated the keys.
This required changes to the dispatcher part as well to use the function implementation `__name__` in the key, so that it then later matches correctly here.

khurram-ghani: This addresses your previous comment as well I think. The check is now against the implementation instead of the dispatcher itself.

uri-granta (approved):
LGTM!

PR #2064, opened 2023-04-21
Add ability to set tf.function args for scipy optimizer

khurram-ghani:
**PR type:** enhancement

**Related issue(s)/PRs:** None 

## Summary

**Proposed changes**

* The main aim of this PR is to add XLA compilation support for the `Scipy` optimizer. This can be achieved by passing `jit_compile=True` argument to `tf.function`.
* However the implemented mechanism is more general, allowing any argument to be passed to `tf.function` call in `Scipy.minimize`.
* Note the new argument added to `Scipy.minimize` is a normal dictionary and not keyword-args. There is already a kwargs argument for scipy optimizer itself.

Considered changing the existing `compile: bool` argument to `Scipy.minimize()` to an enum; with options for "no-compilation", "default-compilation" and "XLA-compilation" -- or something similar. However, decided that it is better to add a more general purpose mechanism allowing setting of any `tf.function` argument. There are other existing arguments (and there might be others in the future) that some users may want to set.

### Minimal working example

```python
# Enable XLA compilation
result = opt.minimize(..., tffun_args={'jit_compile': True})
```

### Release notes

## PR checklist

- [X] Build checks
- [ ] Release management


gpflow/optimizers/scipy.py
```diff
@@ -39,6 +39,7 @@ def minimize(
         step_callback: Optional[StepCallback] = None,
         compile: bool = True,
         allow_unused_variables: bool = False,
+        tffun_args: Dict[str, Any] = {},
```
uri-granta, line 42: 1. More general to allow any mapping, not just dicts.
2. Generally best to avoid mutable default arguments, since if you mutate them in (or outside) the function they end up being changed for all subsequent calls. For dicts it's conventional to use None (since there isn't a built in immutable dict structure):

```suggestion
        tffun_args: Optional[Mapping[str, Any]] = None,
    ....
        if ttfun_args is None:
            ttfun_args = {}
```
4.

gpflow/optimizers/scipy.py
```diff
@@ -109,6 +117,7 @@ def eval_func(
         variables: Sequence[tf.Variable],
         compile: bool = True,
         allow_unused_variables: bool = False,
+        tffun_args: Dict[str, Any] = {},
```
uri-granta, line 120: (ditto)

tests/gpflow/optimizers/test_scipy.py
```diff
@@ -64,15 +75,57 @@ def test_scipy_jit() -> None:
...
 
 
+@unittest.mock.patch("tensorflow.function")
 @pytest.mark.parametrize("compile", [True, False])
-def test_scipy__optimal(compile: bool) -> None:
+@pytest.mark.parametrize(
+    "tffun_args",
+    [{}, dict(jit_compile=True), dict(jit_compile=False, other_arg="dummy")],
+)
+def test_scipy__tffun_args(
```
uri-granta, line 99: Nice test

uri-granta (approved):
LGTM! Just one comment.

awav:
* Are we sure that `tffun_args` is a good name?
* Which use case do you want to cover?

> Note the new argument added to Scipy.minimize is a normal dictionary and not keyword-args. There is already a kwargs argument for scipy optimizer itself.

Yes. Wouldn't adding a dictionary with argument to tensorflow `tf.function` be even more confusing?

gpflow/optimizers/scipy.py
khurram-ghani: I reasoned that `tffun_args` wasn't ever going to be mutated; but I agree it is better to use None. I have made both those suggested improvements.

uri-granta:
> * Are we sure that `tffun_args` is a good name?
> 
> Wouldn't adding a dictionary with arguments to tensorflow `tf.function` be even more confusing?

Personally, I think the current approach is the best one (though the docstring for `ttfun_args` should be expanded to include an example of how to use it for XLA).

* A generic thin wrapper is easier to maintain and more future proof than an adhoc XLA option.
* We already pass through arbitrary arguments to the scipy optimizer.
* The name `ttfun_args` is very much in the style of `scipy_kwargs`.
* The `options` parameter often used in `scipy_kwargs` is already an explicit dictionary.

I therefore don't think `tffun_args` being a dictionary is an issue, especially if we give an example in the docstring. Though we should perhaps use the dict syntax to make it even clearer, as in:

```python
minimize(
        m3.training_loss,
        variables=m3.trainable_variables,
        options=dict(maxiter=50),
        compile=True,
        tffun_args=dict(jit_compile=True),
    )
```

khurram-ghani:
I have added an example to the docstring.

gpflow/optimizers/scipy.py
```diff
@@ -109,7 +123,10 @@ def eval_func(
         variables: Sequence[tf.Variable],
         compile: bool = True,
         allow_unused_variables: bool = False,
+        tffun_args: Optional[Mapping[str, Any]] = None,
```
khurram-ghani, line 126: I am thinking of making this a non-optional arg. This will remove the 2 lines below and improve the code coverage. This method is not directly tested and since I believe this is an internal method, it is okay to change its interface. Is that okay?

khurram-ghani: I have gone ahead and made this change, as otherwise the codecov check was failing.

uri-granta: :+1:

PR #2066, opened 2023-04-27
Don't use enums for default values

uri-granta:
**PR type:** bugfix

**Related issue(s)/PRs:** related to [1985](https://github.com/GPflow/GPflow/pull/1985) 

## Summary

**Proposed changes**

The [wrapper](https://github.com/GPflow/GPflow/blob/awav/likelihoods-positive-minimum/gpflow/config/__config__.py#L97-L99) for config constants is implemented as an Enum, even though it's not really being used as an enumeration. More problematically, Enums don't work correctly when two members have the same value and result in an alias for the first entry. The PR [1985](https://github.com/GPflow/GPflow/pull/1985) has this issue since both `LIKELIHOODS_POSITIVE_MINIMUM` and `JITTER` have the value `1e-6`. 

As proposed by this PR, we simply stop making the wrapper an enum. This forces us to explicitly state the environment variable names alongside the default values, but this duplication is not necessarily a bad thing.

We could use an external enum package such as `aenum` that does support duplicate entries, but this introduces an unnecessary package dependency to GPflow.
### Minimal working example

```python
# Buggy behaviour
class _Values(enum.Enum):
     POSITIVE_MINIMUM = 0.0
     LIKELIHOODS_POSITIVE_MINIMUM = 1e-6
     JITTER = 1e-6

print(_Values.JITTER)
<_Values.LIKELIHOODS_POSITIVE_MINIMUM: 1e-06>
```

### Release notes

## PR checklist

- [X] Build checks
- [ ] Release management

### DIFF
--- a/gpflow/config/__config__.py
+++ b/gpflow/config/__config__.py
@@ -52,7 +52,7 @@ import contextlib
 import enum
 import os
 from dataclasses import dataclass, field, replace
-from typing import Any, Dict, Generator, List, Mapping, Optional, Union
+from typing import Any, Dict, Generator, List, Mapping, Optional, Tuple, Union
 
 import numpy as np
 import tabulate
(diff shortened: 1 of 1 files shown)


khurram-ghani (approved):
Looks good!

PR #2071, opened 2023-06-12
Allow setting lengthscale transform in kernel init

ChrisMorter:
**PR type:** enhancement 

**Related issue(s)/PRs:** 

## Summary

This PR allow the lengthscale parameter transform used by `Stationary` kernels to be configured when constructing the kernel. AFAICS, currently it is only possible to control which transformation is used by setting the default using `set_default_positive_bijector`.

The motivation for this is that I've encountered a repeatable cholesky decomposition error when training a SGPR model (using a  custom kernel) on a particular dataset. Unfortunately I can't share the reproduction here because it uses customer data, but the root of the problem is that the lengthscale hits zero for one of the `CosineKernels` from which the custom kernel is comprised, resulting in the cholesky error. This doesn't happen if I use an `exp` transform rather than `softplus`, presumably because it's better suited to handling small values. For this reason, I'd like to be able to use the `exp` transformation for those cosine terms.

However, this codebase is used for various other model and kernel types, all of which would need to be re-validated if I were to use `set_default_positive_bijector`, because that would change the transform for all of them. Also, this validation could quite possible uncover cases when softplus is better than exp. Even just setting it temporarily when constructing this kernel wouldn't be a great solution because the kernel contains several other terms, some of which wouldn't we wouldn't necessarily want to use the `exp` transform for. Instead it would be better to be able to specify the correct transform to use on a more granular level by specifying it through `__init__`.

- Using `set_default_positive_bijector` to set the default everywhere, and re-validating the behaviour or all other model/kernels types used.
- Just setting the default when constructing this particular kernel type, and then resetting it to softplus afterwards.
 

### Release notes

## PR checklist

- [ ] Build checks
- [ ] Release management

### DIFF
--- a/gpflow/kernels/stationaries.py
+++ b/gpflow/kernels/stationaries.py
@@ -53,13 +53,19 @@ class Stationary(Kernel):
             slice of indices which controls which columns of X are used (by
             default, all columns are used).
         """
+
+        self._lengthscale_transform = kwargs.pop("lengthscale_transform", None)
+
         for kwarg in kwargs:
             if kwarg not in {"name", "active_dims"}:
                 raise TypeError(f"Unknown keyword argument: {kwarg}")
 
         super().__init__(**kwargs)
         self.variance = Parameter(variance, transform=positive())
-        self.lengthscales = Parameter(lengthscales, transform=positive())
+
+        self.lengthscales = Parameter(
+            lengthscales, transform=positive(base=self._lengthscale_transform)
+        )
         self._validate_ard_active_dims(self.lengthscales)
 
     @property
@@ -165,7 +171,9 @@ class AnisotropicStationary(Stationary):
         super().__init__(variance, lengthscales, **kwargs)
 
         if self.ard:
-            self.lengthscales = Parameter(self.lengthscales.numpy())
+            self.lengthscales = Parameter(
+                self.lengthscales.numpy(), transform=positive(base=self._lengthscale_transform)
+            )
 
     @inherit_check_shapes
     def K(self, X: TensorType, X2: Optional[TensorType] = None) -> tf.Tensor:
@@ -229,8 +237,11 @@ class RationalQuadratic(IsotropicStationary):
         lengthscales: TensorType = 1.0,
         alpha: TensorType = 1.0,
         active_dims: Optional[ActiveDims] = None,
+        **kwargs: Any,
     ) -> None:
-        super().__init__(variance=variance, lengthscales=lengthscales, active_dims=active_dims)
+        super().__init__(
+            variance=variance, lengthscales=lengthscales, active_dims=active_dims, **kwargs
+        )
         self.alpha = Parameter(alpha, transform=positive())
 
     @inherit_check_shapes
--- a/tests/gpflow/kernels/test_kernels.py
+++ b/tests/gpflow/kernels/test_kernels.py
@@ -21,7 +21,7 @@ from check_shapes import check_shapes
 from numpy.testing import assert_allclose
 
 import gpflow.ci_utils
-from gpflow.base import AnyNDArray, TensorType
+from gpflow.base import AnyNDArray, TensorType, Transform
 from gpflow.config import default_float, default_positive_bijector
 from gpflow.kernels import (
     RBF,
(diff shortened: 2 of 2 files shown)


khurram-ghani:
Under the alternative solution of setting the `positive_bijector` temporarily, have you considered using a context manager with `as_content`? For example, one option could be to change the default inside the custom kernel only when creating the cosine kernel in question:
```
with as_context(Config(positive_bijector="exp")):
    cosine_kernel = kernels.Cosine()
```

awav:
Hi @chris, you can assign the transform of the lengthscale by replacing the parameter with the transform you need. Is that really necessary to change the whole class of kernels for that? This is change goes against the initial design choices of composability: when you need to change something in the model you replace that block of the model.

ChrisMorter:
> Hi @chris, you can assign the transform of the lengthscale by replacing the parameter with the transform you need. Is that really necessary to change the whole class of kernels for that? This change goes against the initial design choices of composability: when you need to change something in the model you replace that block of the model.

@awav I've just tried this, and it seems to work. I was initially hesitant to just replace the parameter because I wasn't sure that that was safe to do, e.g. because there may be initialisation logic that happens if the parameter is set in `__init__`, which would otherwise be skipped. Looking at the code more closely, the only example I can find of this is in `Stationary`, the call to `self._validate_ard_active_dims` would be missed if the caller just sets `lengthscales` directly after `__init__`. That's a fairly minor issue though, and is easily mitigated by the caller just taking care to alter the shape of the `lengthscales`. 

Perhaps it would be nice if `Stationary` had an `update_lengthscales` method? This would make it clearer to users of GPflow, who may not be entirely familiar with its design decisions, that changing the lengthscales parameter post `__init__` is fine to do, and it could call `_validate_ard_active_dims`, so that logic isn't missed. Anyway, I'll close this PR now, because I agree it's easier to just change the lengthscale parameter post `__init__.`

ChrisMorter:
> Under the alternative solution of setting the `positive_bijector` temporarily, have you considered using a context manager with `as_context`? For example, one option could be to change the default inside the custom kernel only when creating the cosine kernel in question:
> 
> ```
> with as_context(Config(positive_bijector="exp")):
>     cosine_kernel = kernels.Cosine()
> ```

I did consider this, but wasn't particularly keen because it would be dangerous if this was every called in a multi-threaded context because it's effectively modifying global state.

I think I'll go with @awav's suggestion now that I know that it's ok to simply replace the parameter, that seems like a simpler solution and doesn't have the potential to cause threading issues.

gpflow/kernels/stationaries.py
```diff
@@ -53,13 +53,19 @@ def __init__(
             slice of indices which controls which columns of X are used (by
             default, all columns are used).
         """
+
+        self._lengthscale_transform = kwargs.pop("lengthscale_transform", None)
```
uri-granta, line 57: Probably clearer to add this as an explicit, typed optional argument, rather than incorporate it into kwargs and then immediately pop it?

Either way, can you also update the docstring, as well as RELEASE.md with a brief description?

gpflow/kernels/stationaries.py
```diff
@@ -165,7 +171,9 @@ def __init__(
         super().__init__(variance, lengthscales, **kwargs)
 
         if self.ard:
-            self.lengthscales = Parameter(self.lengthscales.numpy())
+            self.lengthscales = Parameter(
+                self.lengthscales.numpy(), transform=positive(base=self._lengthscale_transform)
```
uri-granta, line 175: Was this intentionally not previously being transformed with positive?

gpflow/kernels/stationaries.py
```diff
@@ -229,8 +237,11 @@ def __init__(
         lengthscales: TensorType = 1.0,
         alpha: TensorType = 1.0,
         active_dims: Optional[ActiveDims] = None,
```
uri-granta, line 239: Could probably remove `active_dims` now, and let it merge into `kwargs`.

uri-granta (commented):
LGTM! Just a couple of questions/comments.

uri-granta:
Oops. Page hadn't loaded properly so didn't spot the subsequent discussion. Feel free to ignore.

gpflow/kernels/stationaries.py
uri-granta: (Still kinda curious about this.)

PR #2074, opened 2023-06-19
Cache scipy compile graphs and re-use if args are same

khurram-ghani:
**PR type:** enhancement

**Related issue(s)/PRs:** None 

## Summary

**Proposed changes**

Added a cache to the scipy optimizer for compiled graphs. Previously, a new `tf.function` was created for each call to `minimize`, even if called on exactly the same loss and variables. This wasted compute time and potentialy memory. With this PR, the compiled functions are cached and re-used if all the call arguments (affecting the compiled function) are exactly the same. Current cache size is 2, with the oldest entry replaced when the cache is full.

### Minimal working example

See unit-test for example: https://github.com/GPflow/GPflow/blob/fa8ee98634801971d4dd9e3d20a3f83c5b6d4a23/tests/gpflow/optimizers/test_scipy.py#L424-L447

### Release notes

## PR checklist

- [X] Build checks
- [ ] Release management


gpflow/optimizers/scipy.py
```diff
@@ -31,6 +43,13 @@
 
 
 class Scipy:
+    def __init__(self) -> None:
```
uri-granta, line 46: How about making `compile_cache_limit: int = 2` an initialistion argument? That would allow both changing the cache size and (more importantly) disabling it (by setting it to 0) if it turns out there is an issue that we haven't foreseen.

uri-granta (commented):
(quick initial comment; will look at the tests after in a bit)

gpflow/optimizers/scipy.py
khurram-ghani: I have added this now.

tests/gpflow/optimizers/test_scipy.py
```diff
@@ -221,3 +228,276 @@ def f() -> tf.Tensor:
...
+#   - miss: the same closure but different variables, same reaming arguments
+#   - miss: Different closure but the same variables, same reaming arguments
+#   - miss: the same closure and variables but different tf_fun_args
+#   - miss: the same closure and variables but different allow_unused_variables
+#   - miss: the same closure, variables and args, but no-compile
+@pytest.mark.parametrize(
+    "expect_cache_hit, same_closure2, same_variables2, same_tf_fun_args2, "
+    "allow_unused_variables2, compile",
+    [
+        (True, True, True, True, False, True),
```
uri-granta, line 376: Maybe use `pytest.param` to give the test cases human-readable names so it's clearer what each one does in case one fails? (you could probably just move the comment descriptions above into the id)

RELEASE.md
```diff
@@ -58,12 +58,15 @@ This release contains contributions from:
 * <SIMILAR TO ABOVE SECTION, BUT FOR OTHER IMPORTANT CHANGES / BUG FIXES>
 * <IF A CHANGE CLOSES A GITHUB ISSUE, IT SHOULD BE DOCUMENTED HERE>
 * <NOTES SHOULD BE GROUPED PER AREA>
+* Scipy minimize wrapper caches compiled graphs and re-uses them if called with the same arguments.
```
uri-granta, line 61: Maybe mention that this can be disabled by setting the new compile_cache_size parameter?

uri-granta (approved):
Looks good! Happy to discuss your testing concerns.

RELEASE.md
khurram-ghani: Done

Issue #2078, opened 2023-07-11
dataclass(frozen=True) doesn't like ndarray field defaults


brianwa84:
In python 3.11, dataclass complains when processing Datum: https://github.com/GPflow/GPflow/blob/develop/tests/gpflow/models/test_methods.py#L30

```
/<embedded stdlib>/dataclasses.py:815: in _get_field
    raise ValueError(f'mutable default {type(f.default)} for field '
E   ValueError: mutable default <class 'numpy.ndarray'> for field X is not allowed: use default_factory
```

Issue #2007, opened 2022-10-25
Update CI to test against Python 3.11


jesnie:
There's a new release of Python. We should make sure GPflow works with it.

sc336:
[TensorFlow doesn't support 3.11 yet](https://github.com/tensorflow/tensorflow/issues/58032) - they're aiming to add support in the next release, TF2.12 - we can revisit this issue then.

micmicpx:
TensorFlow now supports python 3.11, but check_shapes>=1.0.0 does not. Any chance we can up this requirement?

NicolasDurrande:
What is the current status of this? I'm happy to have a look into it but I'm not sure what the blockers are...

jesnie:
I'm looking into this. Most of the work is happening on `check_shapes`.

PR #2081, opened 2023-08-04
Test GPflow with Python 3.11.

jesnie:
Fixes: #2007 
Fixes: #2078 

Test GPflow with Python 3.11.

I had to remove some `@dataclass` from some tests, because a dataclass cannot have default values that are numpy arrays in Python 3.11 (because it has more stricts tests that the default values must be immutable). Actually the values were only used statically anyways, so `@dataclass` was not needed.

Also removed some unused imports.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,7 +6,7 @@ parameters:
     default: "3.7"
   max_py_ver:
     type: string
-    default: "3.10"
+    default: "3.11"
   min_tf_ver:
     type: string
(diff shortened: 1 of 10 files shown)


uri-granta (approved):
Looks good. My one concern is that we would like to continue testing with Python 3.10 as we're using that upstream. Really though that should be addressed with an additional `production` test target, in addition to `min` and `max`. Will probably look at adding that in a separate PR.

PR #2096, opened 2023-12-08
fix bug of issue #2091 by type checking A

### LINKED ISSUE #2091: Set a prior over a linear mean function
Bug: I set a Gaussian prior over a linear mean function and when a I see the monitor, gpflow only assigns a prior over the b parameter, and no for the A parameter.

## To reproduce

I just put the code for that part of the model, for simplicity:
```python
import gpflow
import numpy as np
import tensorflow as tf

#trivial example, from [here](https://gpflow.github.io/GPflow/2.5.2/notebooks/advanced/mcmc.html) 
rng = np.random.RandomState(14)
X = np.linspace(-3, 3, 20).reshape(1,20)
y = rng.exponential(np.sin(X) ** 2)

A = gpflow.Parameter(np.zeros(X.shape[1]),dtype=np.float64)
A.prior=tfp.distributions.Normal((np.zeros(X.shape[1])),np.ones(X.shape[1]))
b = gpflow.Parameter(np.zeros((1,)),dtype=np.float64)
b.prior=tfp.distributions.Normal(np.zeros((1,)),np.ones((1,)))
mf = gpflow.functions.Linear(A,b)

k = gpflow.kernels.SquaredExponential()
k.lengthscales.prior = tfp.distributions.Normal(np.zeros(X.shape[1]),np.ones(X.shape[1]))
k.variance.prior = tfp.distributions.Normal(np.zeros(1),np.ones(1))

lik = gpflow.likelihoods.Gaussian()
lik.variance.prior = tfp.distributions.Normal(np.zeros(1),np.ones(1))

model = gpflow.models.GPMC((X, y), kernel=k, mean_function=mf, likelihood=lik)
model
```
So, there is no warning, but, as can be seen from the monitor, gpflow doesn't put the prior over A:

imagen

System information

* GPflow version: 2.9.0
* GPflow installed from: "pip install gpflow"
* TensorFlow version: 2.10.1
* Python version: 3.10.9 | packaged by conda-forge | (main, Jan 11 2023, 15:15:40) [MSC v.1916 64 bit (AMD64)]
* Operating system: win11/Ubuntu 18.046 LTS (I probed it in both os)

I guess the setting is correct, but I don't understand why gpflow set the prior over b and not over A.

JSchmiegel:
**PR type:** bugfix

**Related issue(s)/PRs:** resolves #2091

## Summary

**Proposed changes**

* type check A
* throws error if the Parameter does not have the correct shape

### Minimal working example

* refer to https://github.com/GPflow/GPflow/issues/2091

### Release notes

## PR checklist

- [x] Build checks
- [x] Release management

### DIFF
--- a/gpflow/functions.py
+++ b/gpflow/functions.py
@@ -114,7 +114,7 @@ class Linear(MeanFunction, Function):
             if len(A._shape) >= 2:
                 self.A = A
             else:
-                raise TypeError(
+                raise ValueError(
                     "Error 'gpflow.funcitons.Linear()' mean function. A has not the correct shape (at least 2d)."
                 )
         else:
--- a/gpflow/functions.py
+++ b/gpflow/functions.py
@@ -110,7 +110,15 @@ class Linear(MeanFunction, Function):
         MeanFunction.__init__(self)
         A = np.ones((1, 1), dtype=default_float()) if A is None else A
         b = np.zeros(1, dtype=default_float()) if b is None else b
-        self.A = Parameter(np.atleast_2d(A))
+        if isinstance(A, Parameter):
+            if len(A._shape) >= 2:
+                self.A = A
+            else:
+                raise TypeError(
+                    "Error 'gpflow.funcitons.Linear()' mean function. A has not the correct shape (at least 2d)."
+                )
+        else:
+            self.A = Parameter(np.atleast_2d(A))
         self.b = Parameter(b)
 
     @inherit_check_shapes
--- a/tests/gpflow/test_functions.py
+++ b/tests/gpflow/test_functions.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import unittest
 from typing import Any, Sequence, Tuple, Type
 
 import numpy as np
@@ -234,8 +235,8 @@ def test_polynomial__1d() -> None:
     Y = p(X)
     assert_allclose(
         [
-            [1.0 + 2.0 * 1.0 + 3.0 * (1.0 ** 2)],
-            [1.0 + 2.0 * 2.0 + 3.0 * (2.0 ** 2)],
+            [1.0 + 2.0 * 1.0 + 3.0 * (1.0**2)],
+            [1.0 + 2.0 * 2.0 + 3.0 * (2.0**2)],
         ],
         Y,
     )
@@ -371,3 +372,20 @@ def test_models_with_mean_functions_changes(model_class: Type[Any]) -> None:
     assert np.all(var_zero.numpy() == var_non_zero.numpy())
     # predictive mean changes after modifying mean function
     assert not np.all(mu_zero.numpy() == mu_non_zero.numpy())
+
+
+class test_bug_2091_ensureParameterTypeForLinearMeanFunction(unittest.TestCase):
+    """
+    See github issue #2091. This are tests to ensure that the prior is keeped if A is already given as Parameter.
+    """
+
+    # test_parameter_with_correct_shape
+    def test_parameter_with_correct_shape(self) -> None:
+        A = gpflow.Parameter(np.ones((1, 1)), dtype=np.float64)
+        linear_function = Linear(A, 1)
+        assert not linear_function.A.prior == None
+
+    # test_parameter_with_incorrect_shape
+    def test_parameter_with_incorrect_shape(self) -> None:
+        A = gpflow.Parameter(np.zeros(20), dtype=np.float64)
+        self.assertRaises(TypeError, Linear(A, 1))
--- a/tests/gpflow/test_functions.py
+++ b/tests/gpflow/test_functions.py
@@ -12,11 +12,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import unittest
 from typing import Any, Sequence, Tuple, Type
 
 import numpy as np
 import pytest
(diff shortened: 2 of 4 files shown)


JSchmiegel:
Sorry, but I am unsure, if I understand correctly, this merge is blocked because format checker tests fail. However, if I actually apply the `make format-check` command to the project, then I would change 67 files (that I have not changed in my original pull request). That seems a bit much to me. Is that correct? Should I commit it like this?

tests/gpflow/test_functions.py
```diff
@@ -371,3 +372,20 @@ def test_models_with_mean_functions_changes(model_class: Type[Any]) -> None:
     assert np.all(var_zero.numpy() == var_non_zero.numpy())
     # predictive mean changes after modifying mean function
     assert not np.all(mu_zero.numpy() == mu_non_zero.numpy())
+
+
+class test_bug_2091_ensureParameterTypeForLinearMeanFunction(unittest.TestCase):
```
uri-granta, line 377: In GPflow we use pytest for testing rather than unittest. You can easily convert this by changing the name of the test class, and changing the self.assertRaises to pytest's equivalent (see comment below).

```suggestion
class TestIssue2091EnsureParameterTypeForLinearMeanFunction:
```

(Also, remember to remove the `unittest` import above.)

tests/gpflow/test_functions.py
```diff
@@ -371,3 +372,20 @@ def test_models_with_mean_functions_changes(model_class: Type[Any]) -> None:
...
+    """
+    See github issue #2091. This are tests to ensure that the prior is keeped if A is already given as Parameter.
```
uri-granta, line 379:
```suggestion
    See github issue #2091. These are tests to ensure that the prior is kept if A is already given as Parameter.
```

tests/gpflow/test_functions.py
```diff
@@ -371,3 +372,20 @@ def test_models_with_mean_functions_changes(model_class: Type[Any]) -> None:
...
+    """
+
+    # test_parameter_with_correct_shape
+    def test_parameter_with_correct_shape(self) -> None:
+        A = gpflow.Parameter(np.ones((1, 1)), dtype=np.float64)
```
uri-granta, line 384: You forgot to set a prior on A!

```suggestion
        A = gpflow.Parameter(
            np.ones((1, 1)), dtype=np.float64, prior=tfp.distributions.Normal(0.0, 1.0)
        )
```
 
Also It's useful to include a docstring for each test too, something like
```python
        "Check that Linear copies the prior from a correctly shaped A Parameter"
```

tests/gpflow/test_functions.py
```diff
@@ -371,3 +372,20 @@ def test_models_with_mean_functions_changes(model_class: Type[Any]) -> None:
...
+        linear_function = Linear(A, 1)
+        assert not linear_function.A.prior == None
```
uri-granta, line 386: (it's more normal in Python to compare against None with `is`)

```suggestion
        assert linear_function.A.prior is not None
```

tests/gpflow/test_functions.py
```diff
@@ -371,3 +372,20 @@ def test_models_with_mean_functions_changes(model_class: Type[Any]) -> None:
...
+
+    # test_parameter_with_incorrect_shape
+    def test_parameter_with_incorrect_shape(self) -> None:
+        A = gpflow.Parameter(np.zeros(20), dtype=np.float64)
+        self.assertRaises(TypeError, Linear(A, 1))
```
uri-granta, line 391: (pytest way of doing this)

```suggestion
        with pytest.raises(TypeError):
            Linear(A, 1)
```

tests/gpflow/test_functions.py
```diff
@@ -234,8 +235,8 @@ def test_polynomial__1d() -> None:
     Y = p(X)
     assert_allclose(
         [
-            [1.0 + 2.0 * 1.0 + 3.0 * (1.0 ** 2)],
-            [1.0 + 2.0 * 2.0 + 3.0 * (2.0 ** 2)],
+            [1.0 + 2.0 * 1.0 + 3.0 * (1.0**2)],
```
uri-granta, line 238: This change suggests that you have the wrong version of black installed! To ensure `make format` works you should first run `make dev-install` which should install it. (I'm not sure it actually says this anywhere in CONTRIBUTING.md, sorry! I'll make sure to update the instructions.)

To check which version of black you have, run `pip show black`. The current version used by the tests is 20.8b1.

gpflow/functions.py
```diff
@@ -110,7 +110,15 @@ def __init__(self, A: TensorType = None, b: TensorType = None) -> None:
         MeanFunction.__init__(self)
         A = np.ones((1, 1), dtype=default_float()) if A is None else A
         b = np.zeros(1, dtype=default_float()) if b is None else b
-        self.A = Parameter(np.atleast_2d(A))
+        if isinstance(A, Parameter):
+            if len(A._shape) >= 2:
+                self.A = A
+            else:
+                raise TypeError(
```
uri-granta, line 117: This should probably be a ValueError not a TypeError:

```suggestion
                raise ValueError(
```

uri-granta (commented):
Thank you very much for writing this (and sorry for taking so long to get to look at it). I've added a few comments and suggestions. Feel free to ask if anything is unclear!

uri-granta (approved):
Looks good, thanks!

Issue #2106, opened 2024-03-24
Failure to install `GPflow` alongside `Tensorflow 2.16.1`


RiccardoBuscicchio:
# Build issue
Failure to install `GPflow` dependencies alongside `Tensorflow 2.16.1`

## To reproduce

**Minimal, reproducible example**
 I created a minimal conda environment with `python=3.10` only 
 ```bash 
conda create --name gpflowenv python=3.10 --no-default-packages
```

Then activated it and installed gpflow, assuming all the dependencies are dealt with. 
```bash 
conda activate gpflowenv
pip install gpflow
```
Afterwards, I check the relevant packages version
```python
import tensorflow as tf 
import keras 
print(tf.__version__) 
print(keras.__version__) 
``` 
which gives 

```python
 '2.16.1' '3.1.1' 
```
However when I `import gpflow` I can a missing package (sympy) error.
After installing it via pip, I get  another missing package error `tf_keras` (see the stack trace below).
After `pip install tf-keras` (version 2.16.0), importing `gpflow` doesn't return any error, only a warning about missing `TensorRT` (see second slack trace below).

I'd like to know whether `GPflow` is correctly installed as it is.

**Stack trace, or error message**

***Missing tf_keras***
```python
>> import gpflow
2024-03-24 16:58:02.638116: I tensorflow/core/util/port.cc:113] oneDNN custom operations are on. You may see slightly different numerical results due to floating-point round-off errors from different computation orders. To turn them off, set the environment variable `TF_ENABLE_ONEDNN_OPTS=0`.
2024-03-24 16:58:02.638473: I external/local_tsl/tsl/cuda/cudart_stub.cc:32] Could not find cuda drivers on your machine, GPU will not be used.
2024-03-24 16:58:02.640577: I external/local_tsl/tsl/cuda/cudart_stub.cc:32] Could not find cuda drivers on your machine, GPU will not be used.
2024-03-24 16:58:02.669390: I tensorflow/core/platform/cpu_feature_guard.cc:210] This TensorFlow binary is optimized to use available CPU instructions in performance-critical operations.
To enable the following instructions: AVX2 AVX512F AVX512_VNNI FMA, in other operations, rebuild TensorFlow with the appropriate compiler flags.
2024-03-24 16:58:03.120969: W tensorflow/compiler/tf2tensorrt/utils/py_utils.cc:38] TF-TRT Warning: Could not find TensorRT

[...]

PR #2107, opened 2024-03-27
Bump max_tf_ver to 2.16.1

uri-granta:
**PR type:** enhancement

**Related issue(s)/PRs:** #2106 

## Summary

Add support for Tensorflow 2.16.1. Since this uses Keras 3 as the default keras, I've had to:

- explicitly depend on `tf_keras` (Keras 2) which is no longer installed by default; fortunately this can done by specifying the `[tf]` extra with `tensorflow-probability` (as TFP also requires Keras 2)
- switch all uses of `tf.keras` to use `tf_keras` instead if it's available

A simpler approach would be to make `tf.keras` automatically use Keras 2 by setting the `TF_USE_LEGACY_KERAS` environment variable in `__init__.py`. However, this would prevent gpflow from being used alongside packages that use Keras 3.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -18,7 +18,7 @@ parameters:
     default: "2.8.4"
   max_tf_ver:
     type: string
-    default: "2.15.0"
+    default: "2.16.1"
   min_tfp_ver:
     type: string
     default: "0.12.0"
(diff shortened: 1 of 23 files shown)


khurram-ghani (approved):
LGTM.

PR #2146, opened 2026-05-14
Max/hierarchical kernel

mawbray:
**PR type:** new feature

**Related issue(s)/PRs:** No issue raised but discussed with @uri-granta beforehand.

## Summary

**Proposed changes**

Integration of arc and wedge hierarchical kernels for GPflow. 

* introduction of base hierarchical kernel
* arc and wedge kernel subclasses that simply modify the embedding used by the base (some kernel specific parameters are also defined). 
* notebook to show what problem setting these are useful in, how the design works etc.
* pinned setuptools to ensure pkg_resources is available.

* alternatives considered include multi-task approaches and tree kernels. 

### Minimal working example

```python
# Imports
import numpy as np
import tensorflow as tf
import gpflow
from gpflow.kernels import (
    ActivityCondition,
    ArcHierarchical,
    HierarchyNode,
)

np.random.seed(1793)
tf.random.set_seed(1793)

# Problem set up

def objective(X: np.ndarray) -> np.ndarray:
    x1, y1, x2, x3 = X[:, 0], X[:, 1], X[:, 2], X[:, 3]
    return (
        np.sin(2.0 * np.pi * x1)
        + (y1 > 0.5).astype(float) * 0.5 * np.cos(np.pi * x2 / 5.0)
        + (y1 < 0.5).astype(float) * 0.5 * x3
    ).reshape(-1, 1)

def sample_inputs(n: int) -> np.ndarray:
    x1 = np.random.uniform(0.0, 1.0, size=n)
    y1 = np.random.randint(0, 2, size=n).astype(float)
    x2 = np.random.uniform(0.0, 5.0, size=n)
    x3 = np.random.uniform(-1.0, 1.0, size=n)
    return np.stack([x1, y1, x2, x3], axis=-1)

X_train = sample_inputs(40)
Y_train = objective(X_train) + 0.05 * np.random.randn(40, 1)
X_test = np.array(
    [
        [0.3, 1.0, 2.0, 0.0],  # y1 = 1 branch
        [0.3, 0.0, 0.0, 0.5],  # y1 = 0 branch
        [0.7, 1.0, 4.0, 0.0],
        [0.7, 0.0, 0.0, -0.4],
    ]
)

indicator_dims = [1]

# Define disjunctive structure

hierarchy = [
    HierarchyNode("shared", feature_dims=[0], feature_bounds=[[0.0, 1.0]]),
    HierarchyNode(
        "branch_A",
        feature_dims=[2],
        feature_bounds=[[0.0, 5.0]],
        activity_condition=ActivityCondition({0: 1}),
    ),
    HierarchyNode(
        "branch_B",
        feature_dims=[3],
        feature_bounds=[[-1.0, 1.0]],
        activity_condition=ActivityCondition({0: 0}),
    ),
]

# Construct kernel
arc = ArcHierarchical(hierarchy=hierarchy, indicator_dims=indicator_dims)
print("conditional columns:", arc._n_cond)
print("unconditional columns:", arc._n_uncond)
print("angle init:", arc.angle.numpy())
print("radius init:", arc.radius.numpy())

# Build and train model
# Wrap in a Constant() factor so the GP can learn an overall variance.
arc_for_fit = ArcHierarchical(
    hierarchy=hierarchy, indicator_dims=indicator_dims
)
kernel = gpflow.kernels.Constant() * arc_for_fit
gpr = gpflow.models.GPR(
    data=(X_train, Y_train), kernel=kernel, noise_variance=0.05
)

print(f"LML before fit: {gpr.log_marginal_likelihood().numpy():+.3f}")
gpflow.optimizers.Scipy().minimize(
    gpr.training_loss, gpr.trainable_variables, options={"maxiter": 100}
)
print(f"LML after fit:  {gpr.log_marginal_likelihood().numpy():+.3f}")
print("learnt angle :", arc_for_fit.angle.numpy())
print("learnt radius:", arc_for_fit.radius.numpy())

# Prediction
mean, var = gpr.predict_f(X_test)
truth = objective(X_test).ravel()
for x, m, v, t in zip(X_test, mean.numpy().ravel(), var.numpy().ravel(), truth):
    print(
        f"  x = {x.tolist()}  ->  mean = {m:+.3f}  var = {v:.3f}  truth = {t:+.3f}"
    )
```

### Release notes

## PR checklist

- [ X] New features: code is well-documented
  - [ X] detailed docstrings (API documentation)
  - [ X] notebook examples (usage demonstration)
- [ X] Build checks
  - [ X] I locally tested that the tests pass (`make check-all`) (to the best of my knowledge any errors are not due to my additions).
- [ ] Release management


gpflow/kernels/hierarchical.py
```diff
@@ -0,0 +1,366 @@
...
+        self._required_is_ignore = tf.equal(self._required, _IGNORE)
+
+        if base_kernel is None:
+            base_kernel = Matern52()
+        if not isinstance(base_kernel, Stationary):
+            raise ValueError(
+                f"`base_kernel` must be a gpflow.kernels.Stationary instance; "
+                f"got {type(base_kernel).__name__}."
+            )
+        base_kernel.lengthscales.assign(tf.ones_like(base_kernel.lengthscales))
```
uri-granta, line 292: We should either clearly document that the base kernel will be modified, or make a copy here.

We should probably also validate the lengthscales of the base_kernel at the start (otherwise it might broadcast incorrectly).

gpflow/kernels/hierarchical.py
```diff
@@ -0,0 +1,366 @@
...
+            )
+
+        bounds = tf.convert_to_tensor(self.feature_bounds, dtype=tf.float64)
+        if bounds.shape.rank != 2 or bounds.shape[0] != len(feature_dims) or bounds.shape[1] != 2:
+            raise ValueError(
+                f"`feature_bounds` of node {self.name!r} must have shape "
+                f"[len(feature_dims), 2] = [{len(feature_dims)}, 2]; got "
+                f"{tuple(bounds.shape)}."
+            )
+        if bool(tf.reduce_any(bounds[:, 1] < bounds[:, 0]).numpy()):
```
uri-granta, line 137: Calling numpy() might cause issues with non-eager evaluation? (not sure) Might be safer to just use numpy arrays instead of tensors in this section.

gpflow/kernels/hierarchical.py
```diff
@@ -0,0 +1,366 @@
...
+            r_c \\sin(\\pi a_c v_c)\\, m_c,\\;
+            r_c \\cos(\\pi a_c v_c)\\, m_c
+        \\big),
+
+    so that inactive points sit at the origin and active points sit on a
+    circle whose phase depends on ``v_c``. A stationary base kernel then
+    evaluates covariance in the joint embedded space.
+    """
+
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
```
uri-granta: (this should repeat the signature of the parent method to help the type checker)

gpflow/kernels/hierarchical.py
```diff
@@ -0,0 +1,366 @@
...
+        if self._n_cond > 0:
+            v_c = tf.gather(v, self._cond_local_idx, axis=-1)
+            m_c = tf.gather(m_float, self._cond_local_idx, axis=-1)
+            parts.append(self._embed_conditional(v_c, m_c))
+        if not parts:  # pragma: no cover - unreachable: enforced by HierarchyNode/__init__
+            shape = tf.concat([tf.shape(X)[:-1], [0]], axis=0)
+            return tf.zeros(shape, dtype=v.dtype)
+        return tf.concat(parts, axis=-1)
+
+    def K(self, X: TensorType, X2: Optional[TensorType] = None) -> tf.Tensor:
```
uri-granta, line 379: K and K_diag should be marked with `@inherit_check_shapes`.

gpflow/kernels/hierarchical.py
```diff
@@ -0,0 +1,366 @@
...
+    def _normalise(self, X: TensorType) -> tf.Tensor:
+        X_cast = tf.cast(X, self._bounds.dtype)
+        v = tf.gather(X_cast, self._feature_dims, axis=-1)
+        lo, hi = self._bounds[:, 0], self._bounds[:, 1]
+        rng = hi - lo
+        safe_rng = tf.where(tf.abs(rng) < 1e-12, tf.ones_like(rng), rng)
+        return (v - lo) / safe_rng
+
+    @abc.abstractmethod
+    def _embed_conditional(self, v_c: tf.Tensor, m_c: tf.Tensor) -> tf.Tensor:
```
uri-granta, line 356: This should be marked with a `@check_shapes` decorator. Some of the other helper functions might benefit from it too.

gpflow/kernels/hierarchical.py
```diff
@@ -0,0 +1,366 @@
...
+    so that inactive points sit at the origin and active points sit on a
+    circle whose phase depends on ``v_c``. A stationary base kernel then
+    evaluates covariance in the joint embedded space.
+    """
+
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        super().__init__(*args, **kwargs)
+        if self._n_cond > 0:
+            self.angle = Parameter(
+                0.5 * tf.ones(self._n_cond, dtype=tf.float64),
```
uri-granta: (we should probably use `default_float()` here and elsewhere)

gpflow/kernels/hierarchical.py
```diff
@@ -0,0 +1,366 @@
...
+
+        bounds_tensor = tf.stack(flat_bounds_rows, axis=0)
+
+        self._n_feat = len(flat_feature_dims)
+        self._n_ind = len(indicator_dims)
+
+        self._feature_dims = tf.constant(flat_feature_dims, dtype=tf.int32)
+        self._indicator_dims = tf.constant(indicator_dims, dtype=tf.int32)
+        self._bounds = bounds_tensor
+        self._hierarchy = hierarchy
```
uri-granta, line 256: (we don't currently use this AFAICT, but I guess it might be useful to keep around; maybe even expose in a property?)

gpflow/kernels/hierarchical.py
```diff
@@ -0,0 +1,366 @@
...
+import numpy as np
+import tensorflow as tf
+import tensorflow_probability as tfp
+
+from ..base import Parameter, TensorType
+from ..utilities import positive, set_trainable, to_default_float
+from .base import ActiveDims, Kernel
+from .stationaries import Matern52, Stationary
+
+_IGNORE = -1
```
uri-granta, line 41: (could use a docstring or comment)

tests/gpflow/kernels/reference.py
```diff
@@ -106,3 +110,90 @@ def ref_periodic_kernel(
...
+    if uncond_idx:
+        parts.append(v[:, uncond_idx])
+    if cond_idx:
+        v_c = v[:, cond_idx]
+        m_c = mask[:, cond_idx].astype(float)
+        parts.append(embed_conditional(v_c, m_c))
+    return np.concatenate(parts, axis=-1) if parts else np.zeros((X.shape[0], 0))
+
+
+def ref_arc_hierarchical_kernel(
```
uri-granta, line 145: (missing docstrings for the public functions)

uri-granta (commented):
A few initial comments. Not a proper review yet but at a high level this looks good! Happy to discuss tomorrow.

gpflow/kernels/hierarchical.py
```diff
@@ -0,0 +1,366 @@
...
+        interpreted as positions within this sequence.
+    :param base_kernel: stationary base kernel applied in the joint embedded
+        space. Defaults to :class:`Matern52` if omitted.
+    :param active_dims: inherited from :class:`Kernel`. If supplied, both
+        feature dims and ``indicator_dims`` are interpreted in the sliced
+        coordinate system.
+    :param name: optional kernel name.
+    """
+
+    def __init__(
```
uri-granta, line 183: Consider marking this as experimental in the first release?

```suggestion
    @experimental
    def __init__(
```

gpflow/kernels/hierarchical.py
```diff
@@ -0,0 +1,366 @@
...
+        self,
+        hierarchy: Sequence[HierarchyNode],
+        indicator_dims: Sequence[int] = (),
+        base_kernel: Optional[Stationary] = None,
+        *,
+        active_dims: Optional[ActiveDims] = None,
```
uri-granta: We should add some tests for active_dims usage

tests/gpflow/kernels/test_hierarchical.py
```diff
@@ -0,0 +1,769 @@
...
+        np.testing.assert_allclose(Z[0, 3], 0.0)  # cos_x2 row 0
+        np.testing.assert_allclose(Z[1, 2], 0.0)  # sin_x3 row 1 (x3 inactive)
+        np.testing.assert_allclose(Z[1, 4], 0.0)  # cos_x3 row 1
+        # Unconditional column passes through normalised.
+        np.testing.assert_allclose(Z[:, 0], [0.5, 0.5])
+
+
+class TestBaseKernelPlumbing:
+    def test_default_base_kernel_is_matern52(self) -> None:
+        import gpflow
```
uri-granta: (move to top: local imports inconsistent with our style)

tests/gpflow/kernels/test_hierarchical.py
```diff
@@ -0,0 +1,769 @@
...
+        for g in grads:
+            assert g is not None
+            g_np = g.numpy()
+            assert np.all(np.isfinite(g_np))
+            assert np.any(np.abs(g_np) > 1e-12)
+
+
+# Silence unused-import warnings — HierarchicalEmbeddingKernel is part of the
+# public surface and re-exported here for downstream test ergonomics.
+_ = HierarchicalEmbeddingKernel
```
uri-granta: (no idea what this means, but I expect you can just remove HierarchicalEmbeddingKernel from the imports)

doc/sphinx/notebooks/advanced/hierarchical_kernels.pct.py
```diff
@@ -0,0 +1,195 @@
...
+#       format_version: '1.3'
+#       jupytext_version: 1.3.3
+#   kernelspec:
+#     display_name: Python 3
+#     language: python
+#     name: python3
+# ---
+
+# %% [markdown]
+# # Hierarchical Arc kernels
```
uri-granta: (this covers Wedge too, right?)

```suggestion
# # Hierarchical kernels
```

doc/sphinx/notebooks/advanced/hierarchical_kernels.pct.py
```diff
@@ -0,0 +1,195 @@
...
+# %%
+
+from gpflow.kernels import ArcHierarchical, Constant
+from gpflow.models import GPR
+from gpflow.optimizers import Scipy
+
+indicator_dims = [1]
+
+arc = ArcHierarchical(hierarchy=hierarchy, indicator_dims=indicator_dims)
+print("conditional columns:", arc._n_cond)
```
uri-granta: (if these are interesting enough to print, we should probably expose them via a property, rather than accessing the private attributes)

uri-granta (commented):
A few more first pass comments. Overall this looks very nice.

gpflow/kernels/hierarchical.py
mawbray: So l guess if someone tries to construct a hierarchynode inside @tf.function it would fail. Are GPFlow objects built so that construction can happen within tf.function calls generally? If so, I will revise.

tests_requirements.txt
```diff
@@ -8,7 +8,7 @@ pytest-random-order
 pytest-xdist
 pytest>=3.5.0
 types-Deprecated
-types-setuptools
+types-setuptools<80  # match setuptools<80 pin; newer stubs use _DictLike which conflicts with plain dict
 types-tabulate
```
uri-granta, line 13: I suspect you'll need to add the following line too, to fix the notebook error in notebook-test-max. Looks like this package is no longer pulled in automatically by tensorflow.

```
importlib_resources  # required by tensorflow-datasets when downloading datasets (gp_nn notebook)
```

setup.py
```diff
@@ -14,7 +14,7 @@
     "numpy<2",
     "packaging",
     "scipy",
-    "setuptools>=41.0.0",  # to satisfy dependency constraints
+    "setuptools>=41.0.0,<80.0.0",  # to satisfy dependency constraints, and dropping from >=82
```
uri-granta, line 20: We should discuss this, or at least raise an issue for it, as pinning like this always causes issue down the line.

gpflow/kernels/hierarchical.py
```diff
@@ -0,0 +1,366 @@
...
+        _check_non_negative_unique(indicator_dims, "indicator_dims")
+
+        for node in hierarchy:
+            for k in node.activity_condition.requirements:
+                if k >= len(indicator_dims):
+                    raise ValueError(
+                        f"`hierarchy` node {node.name!r} references indicator "
+                        f"index {k}, but only {len(indicator_dims)} indicator "
+                        f"dims were declared."
+                    )
```
pakhaykwok: Q: Should we also prevent redundant index in `indicator_dims`?

gpflow/kernels/hierarchical.py
pakhaykwok, line 292: Q: Kind of related to my question about `active_dims` in the meeting. I believe `Matern52()` will instantiate a kernel with `lengthscales` of size 1, i.e. a single value shared across all feature dimensions. To allow each feature to have its own lengthscale, one must specify `base_kernel` with `lengthscales` of the correct length here, so we should also check that it's length is correct.

gpflow/kernels/hierarchical.py
mawbray: Good Q! Not sure - might be useful to have a global decision vector shared between multiple models and one indicator may not actually enter into the activity conditions required for a given model?

gpflow/kernels/hierarchical.py
mawbray: I think this should be resolved by recent commits but let me know - also related to Uri's comment on silent mutation.

gpflow/kernels/hierarchical.py
```diff
@@ -0,0 +1,453 @@
...
+    """
+
+    @experimental
+    def __init__(
+        self,
+        hierarchy: Sequence[HierarchyNode],
+        indicator_dims: Sequence[int] = (),
+        base_kernel: Optional[Stationary] = None,
+        *,
+        active_dims: Optional[ActiveDims] = None,
```
pakhaykwok: C: If we are to allow the data to contain redundant columns, I suggest that we make `active_dims` compulsory, and that we check the set of required columns matches exactly to `active_dims`. This should be less error prone, I think.

gpflow/kernels/hierarchical.py
```diff
@@ -0,0 +1,453 @@
...
+        slice defined by ``active_dims`` is applied first, and the resulting
+        coordinates are then processed according to the hierarchy and indicators.
+    :param name: optional kernel name.
+    """
+
+    @experimental
+    def __init__(
+        self,
+        hierarchy: Sequence[HierarchyNode],
+        indicator_dims: Sequence[int] = (),
```
pakhaykwok: C: I feel that the reason that the different indices are hard to follow is due to `ActivityCondition` referencing `indicator_dims`, hence one needs to go from `ActivityCondition` -> `indicator_dims` -> `active_dims` to work out the relevant boolean columns. I wonder if it would be easier to follow if `ActivityCondition` reference directly `active_dims`, i.e. removing one layer of indexing.

setup.py
mawbray: agree

gpflow/kernels/hierarchical.py
mawbray: I guess there could be some intersection. Say if the indicator_dim is not specified in active_dims then we allow it as a redundant index. Otherwise we flag to the user or error?

gpflow/kernels/hierarchical.py
mawbray: Ah nice, I had the same thought on the tube yesterday ahha.

mawbray: See above^

gpflow/kernels/hierarchical.py
mawbray: Let me think about this.

pakhaykwok:
This might be a useful reference:
https://github.com/Prowler-io/automotive/blob/d13276a3c12b7db1b09b35051bcefc4fa17275a2/engine-tuning/sm_calibration/configs/yaml/model_config/regression_models/ev_sgpr.yaml#L16

This is how the EV custom kernel is defined in config, the kernel also requires different indices to be specified. It's worth thinking how this kernel will look like in config too.

gpflow/kernels/hierarchical.py
mawbray: Okay reconsidered. So I think indicator dims should refer to the sliced coordinate system (ie the vector yielded after active dims have been sliced). So in principle we should not allow redundant indicator_dims indices.

gpflow/kernels/hierarchical.py
mawbray: So indicator dims refers to dimensions in the sliced coordinate system, yielded after active_dims has been applied. Does this affect the above?

gpflow/kernels/hierarchical.py
pakhaykwok: Agree, if there're redundant columns which the kernel doesn't need, `active_dims` must be specified without those columns.

gpflow/kernels/hierarchical.py
pakhaykwok: Let's imagine the following under the current implementation:
`active_dims = [0, 1, 2, 3, 4, 5]`
`indicator_dims = [2, 4]`
`activity_condition = ActivityCondition({1:0})`
The actual boolean column with which `activity_condition` is referencing is `4`.

Alternatively, let's consider an implementation where `indicator_dims` is implicit:
`active_dims = [0, 1, 2, 3, 4, 5]`
`activity_condition = ActivityCondition({4:0})`
The actual boolean column with which `activity_condition` is referencing is `4`.

Here in the alternative implementation, you can see that I can get to `4` with one fewer step, also one fewer thing to specify. I guess I still struggle to see why `indicator_dims` needs to be provided explicitly.

mawbray: Okay yes this is a very good point. I have now revised accordingly :)

gpflow/kernels/hierarchical.py
mawbray: Okay so active_dims is no compulsory and is validated so that the feature_dims + indicator_dims have to equal the number of active_dims specified.

gpflow/kernels/hierarchical.py
```diff
@@ -0,0 +1,488 @@
...
+            )
+
+        key_to_pos: Dict[int, int] = {k: p for p, k in enumerate(indicator_dims)}
+
+        bounds_tensor = tf.stack(flat_bounds_rows, axis=0)
+
+        self._n_feat = len(flat_feature_dims)
+        self._n_ind = len(indicator_dims)
+
+        n_expected = self._n_feat + self._n_ind
```
pakhaykwok, line 228: SH: Now that we have made `active_dims` compulsory, we expect the tensor after slicing using `active_dims` to contain no redundant columns, we can therefore add a stronger check to ensure the sorted combined list of `flat_feature_dims` and `indicator_dims` to equal `list(range(len(width)))`.

mawbray: Resolved?

mawbray:
Agreed design is ready to go ! Will open a collection of smaller stacked PRs for integration to develop.

PR #2151, opened 2026-05-22
Add HierarchicalEmbeddingKernel abstract base and base-kernel plumbing

mawbray:
Introduces the abstract embedding kernel that all hierarchical kernels inherit from: the construction-time validation pipeline (active_dims width check, feature/indicator overlap check), the indicator-derived `indicator_dims` property, the `_normalise` / `_build_activity_mask` / `_embed` machinery, and the `K` / `K_diag` dispatch. Also handles base-kernel deep-copy with lengthscale forcing and trainability disabling — concrete subclasses carry the per-conditional-column scale parameters, so the base kernel's lengthscales must stay at 1.

Tests exercise the base via a private `_FakeEmbedKernel` subclass (stacks each masked conditional column with itself); the first real subclass arrives in the next PR of the stack.

This is PR2 of a six-PR stack to integrate the hierarchical kernels feature.

**PR type:** new feature 

**Related issue(s)/PRs:** 

## Summary

**Proposed changes**

See #2146. 

* Introducing a base hierarchical kernel class
* validates the logic from the hierarchy and makes sure it is consistent with active_dims.
* tested

### Minimal working example

```python
import numpy as np
import tensorflow as tf

from gpflow.kernels import ActivityCondition, HierarchyNode, HierarchicalEmbeddingKernel

np.random.seed(1793)
tf.random.set_seed(1793)


def objective(X: np.ndarray) -> np.ndarray:
    x1, y1, x2, x3 = X[:, 0], X[:, 1], X[:, 2], X[:, 3]
    return (
        np.sin(2.0 * np.pi * x1)
        + (y1 > 0.5).astype(float) * 0.5 * np.cos(np.pi * x2 / 5.0)
        + (y1 < 0.5).astype(float) * 0.5 * x3
    ).reshape(-1, 1)


def sample_inputs(n: int) -> np.ndarray:
    x1 = np.random.uniform(0.0, 1.0, size=n)
    y1 = np.random.randint(0, 2, size=n).astype(float)
    x2 = np.random.uniform(0.0, 5.0, size=n)
    x3 = np.random.uniform(-1.0, 1.0, size=n)
    return np.stack([x1, y1, x2, x3], axis=-1)


X_train = sample_inputs(40)
Y_train = objective(X_train) + 0.05 * np.random.randn(40, 1)
X_test = np.array(
    [
        [0.3, 1.0, 2.0, 0.0],  # y1 = 1 branch
        [0.3, 0.0, 0.0, 0.5],  # y1 = 0 branch
        [0.7, 1.0, 4.0, 0.0],
        [0.7, 0.0, 0.0, -0.4],
    ]
)

hierarchy = [
    HierarchyNode("shared", feature_dims=[0], feature_bounds=[[0.0, 1.0]]),
    HierarchyNode(
        "branch_A",
        feature_dims=[2],
        feature_bounds=[[0.0, 5.0]],
        activity_condition=ActivityCondition({1: 1}),
    ),
    HierarchyNode(
        "branch_B",
        feature_dims=[3],
        feature_bounds=[[-1.0, 1.0]],
        activity_condition=ActivityCondition({1: 0}),
    ),
]

# Here we show how the base class is used in subsequent PRs.
class ArcHierarchical(HierarchicalEmbeddingKernel):
    """The Arc kernel of Swersky et al. (2014).

    Each conditional column ``c`` (normalised value ``v_c`` in ``[0, 1]``,
    activity mask ``m_c``) is mapped into the plane via

    .. math::

        \\phi_c(v_c, m_c) = \\big(
            r_c \\sin(\\pi a_c v_c)\\, m_c,\\;
            r_c \\cos(\\pi a_c v_c)\\, m_c
        \\big),

    so that inactive points sit at the origin and active points sit on a
    circle whose phase depends on ``v_c``. A stationary base kernel then
    evaluates covariance in the joint embedded space.
    """

    def __init__(
        self,
        hierarchy: Sequence[HierarchyNode],
        base_kernel: Optional[Stationary] = None,
        *,
        active_dims: ActiveDims,
        name: Optional[str] = None,
    ) -> None:
        super().__init__(hierarchy, base_kernel, active_dims=active_dims, name=name)
        if self._n_cond > 0:
            self.angle = Parameter(
                0.5 * tf.ones(self._n_cond, dtype=default_float()),
                transform=tfp.bijectors.Sigmoid(to_default_float(0.1), to_default_float(0.9)),
                name="angle",
            )
            self.radius = Parameter(
                tf.ones(self._n_cond, dtype=default_float()),
                transform=positive(),
                name="radius",
            )

    @inherit_check_shapes
    def _embed_conditional(self, v_c: tf.Tensor, m_c: tf.Tensor) -> tf.Tensor:
        theta = np.pi * self.angle * v_c
        sin_part = self.radius * tf.sin(theta) * m_c
        cos_part = self.radius * tf.cos(theta) * m_c
        return tf.concat([sin_part, cos_part], axis=-1)

arc = ArcHierarchical(hierarchy, active_dims=[0,1,2,3])

print("conditional columns:", arc.n_cond_dims)
print("unconditional columns:", arc.n_uncond_dims)
print("indicator columns (derived):", arc.indicator_dims)
print("angle init:", arc.angle.numpy())
print("radius init:", arc.radius.numpy())


```

### Release notes

## PR checklist

- [ X] New features: code is well-documented
  - [ X] detailed docstrings (API documentation)
- [ X] Build checks
  - [ X] I locally tested that the tests pass (`make check-all`)
- [ ] Release management


gpflow/kernels/hierarchical.py
```diff
@@ -111,3 +137,258 @@ def __post_init__(self) -> None:
...
+        name: Optional[str] = None,
+    ) -> None:
+        super().__init__(active_dims=active_dims, name=name)
+
+        hierarchy = tuple(hierarchy)
+        if not hierarchy:
+            raise ValueError("`hierarchy` must contain at least one node.")
+        names = [node.name for node in hierarchy]
+        if len(set(names)) != len(names):
+            raise ValueError(f"`hierarchy` contains duplicate node names: {names}.")
```
pakhaykwok: Consider a stronger uniqueness check.

gpflow/kernels/hierarchical.py
```diff
@@ -111,3 +137,258 @@ def __post_init__(self) -> None:
...
+
+        required = [[_IGNORE] * self._n_ind for _ in range(self._n_feat)]
+        for j, ac in enumerate(flat_activity_conditions):
+            for k, v in ac.requirements.items():
+                required[j][key_to_pos[k]] = v
+        self._required = tf.constant(required, dtype=tf.int32)
+        self._required_is_ignore = tf.equal(self._required, _IGNORE)
+
+        if base_kernel is None:
+            base_kernel = Matern52()
```
pakhaykwok, line 277: Should we default lengthscales to a scalar or a tensor of length `d_embed`?

gpflow/kernels/hierarchical.py
```diff
@@ -111,3 +137,258 @@ def __post_init__(self) -> None:
...
+        ls_shape = tuple(base_kernel.lengthscales.shape)
+        if ls_shape not in ((), (1,), (d_embed,)):
+            raise ValueError(
+                f"`base_kernel.lengthscales` must be scalar or have shape "
+                f"({d_embed},) to match the embedded dimension "
+                f"`2 * n_cond + n_uncond`; got shape {ls_shape}."
+            )
+
+        base_kernel.lengthscales.assign(tf.ones_like(base_kernel.lengthscales))
+        set_trainable(base_kernel.lengthscales, False)
```
pakhaykwok, line 309: Why should it be untrainable?

gpflow/kernels/hierarchical.py
```diff
@@ -111,3 +137,258 @@ def __post_init__(self) -> None:
...
+            ignore_expanded,
+            tf.equal(ind_expanded, required_expanded),
+        )
+        return tf.reduce_all(match, axis=-1)
+
+    @check_shapes(
+        "X: [batch..., N, D]", "return: [batch..., N, D_f]"
+    )  # D_f is feature_dims (i.e. D_cond + D_uncond)
+    def _normalise(self, X: TensorType) -> tf.Tensor:
+        X_cast = tf.cast(X, self._bounds.dtype)
```
pakhaykwok: Should we follow `X`'s dtype instead?

tests/gpflow/kernels/test_hierarchical.py
```diff
@@ -115,3 +179,330 @@ def test_non_string_name_rejected(self) -> None:
...
+                [0.3, 1.0, 4.0, 0.0],
+            ],
+            dtype=tf.float64,
+        )
+        K = kernel.K(X).numpy()
+        assert K.shape == (3, 3)
+        assert np.all(np.isfinite(K))
+        np.testing.assert_allclose(K, K.T, atol=1e-12)
+        eigs = np.linalg.eigvalsh(K + 1e-10 * np.eye(3))
+        assert eigs.min() > -1e-8
```
pakhaykwok, line 573: May I also suggest adding tests where `active_dims` is not trivial, i.e. not starting from zero, or having gap, together with `X` with redundant columns? They are to ensure all the indexing and slicing are done correctly.

pakhaykwok, line 573: Also consider adding tests where the kernel is being call as part of a `tf.function`? They are to ensure that the kernel can be compiled.

pakhaykwok (commented):
Some suggestions.

gpflow/kernels/hierarchical.py
mawbray: The embedding itself has parameters which provide an equivalent function to kernel lengthscales. Combing the two makes the two parameters unidentifiable, which can cause issues for optimisers based on second order information (e.g. the LML can be flat on a connected subset of the domain and have no curvature which means the Hessian becomes singular when evaluated in that part of the domain).

mawbray: We should allow the lengthscales of unconditional dimensions to be learned however.

gpflow/kernels/hierarchical.py
mawbray: See discussion in the next comment.

gpflow/kernels/hierarchical.py
mawbray: This is now added.

mawbray:
I think this review should now be largely remedied thanks @pakhaykwok

gpflow/kernels/hierarchical.py
```diff
@@ -111,3 +138,280 @@ def __post_init__(self) -> None:
...
+    ) -> None:
+        super().__init__(active_dims=active_dims, name=name)
+
+        hierarchy = tuple(hierarchy)
+        if not hierarchy:
+            raise ValueError("`hierarchy` must contain at least one node.")
+        names = [node.name for node in hierarchy]
+        duplicate_names = [name for name, count in Counter(names).items() if count > 1]
+        if duplicate_names:
+            raise ValueError(f"`hierarchy` contains duplicate node names: {duplicate_names}.")
```
pakhaykwok, line 208: Q: what purpose does `node.name` serve? If it's simply for duplication checking, it's probably better to compare the nodes directly, as `HierarchyNode` supports the `==` operator.

gpflow/kernels/hierarchical.py
```diff
@@ -111,3 +138,280 @@ def __post_init__(self) -> None:
...
+                "`slice` whose `stop` is concrete; "
+                f"got {self._active_dims!r}."
+            )
+        if width != n_expected:
+            raise ValueError(
+                f"`active_dims` selects {width} column(s), but the hierarchy "
+                f"defines {self._n_feat} feature dimension(s) and "
+                f"{self._n_ind} indicator dimension(s) (total {n_expected}); "
+                f"`active_dims` must select exactly that many columns."
+            )
```
pakhaykwok: This is now redundant because of the check below.

gpflow/kernels/hierarchical.py
```diff
@@ -111,3 +138,280 @@ def __post_init__(self) -> None:
...
+            ignore_expanded,
+            tf.equal(ind_expanded, required_expanded),
+        )
+        return tf.reduce_all(match, axis=-1)
+
+    @check_shapes(
+        "X: [batch..., N, D]", "return: [batch..., N, D_f]"
+    )  # D_f is feature_dims (i.e. D_cond + D_uncond)
+    def _normalise(self, X: TensorType) -> tf.Tensor:
+        X_cast = tf.cast(X, X.dtype)
```
pakhaykwok: This is now redundant, as `X_cast` is simply `X`

pakhaykwok (approved):
LGTM, just a few minor comments.

gpflow/kernels/hierarchical.py
mawbray: Fair point -  `node.name` also provides a way of tying a disjunction back to some property of the underlying system / as a tag so I am inclined to keep it. This check is to ensure there are no duplications of nodes.

